PR #6, opened 2019-12-13
Fix temp-user cleanup job to cleanup queue/sync data

yufengwng:
Noticed an alert from `test` channel:
```
Error while executing job 'cleanup.temp-user'  Cannot delete or update a parent row: a foreign key constraint fails (peppertest.queued_event, CONSTRAINT queued_event_operator_user_fk FOREIGN KEY (operator_user_id) REFERENCES user (user_id))
```

It looks like we're queuing up events for temp users, so we need to clean those up when removing expired temp users. Also cleaning up data-sync requests as well.


yufengwng:
please run tests

ssettipalli (approved):
looks good to me

yufengwng:
please run tests

raiks2 (approved):
LGTM

PR #12, opened 2019-12-16
DDP 4232 Add EMAIL type, confirmEntry & related properties to TextQuestion

ssettipalli:
Look at https://broadinstitute.atlassian.net/browse/DDP-4232


ssettipalli:
Please run tests

ssettipalli:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/MiscUtil.java
```diff
@@ -62,6 +62,16 @@ public static void checkRegexPattern(String pattern) {
         }
     }
 
+    public static boolean isEmailFormatValid(String email) {
+        String emailRegex = "^[a-zA-Z0-9_+&*-]+(?:\\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,7}$";
```
bskinner: This regex does not properly handle TLDs longer than 7 characters. Do we really need to perform strict checking here? Unless we're going to be a lot more stringent with the regexp (a full RFC 5322-compliant regex is an absolute beast), we might want to just do a check for a `@` followed by at least one `.`- it'll let through a lot of invalid addresses, but we can rely on the SMTP server to catch the bad ones when it goes to send.

bskinner (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/util/MiscUtil.java
ssettipalli: How about this regex: "^[A-Za-z0-9+_.-]+@(.+)$"

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1263,6 +1263,14 @@ Question.Text:
           - TEXT
           - ESSAY
           - SIGNATURE
+          - EMAIL
+      confirmEntry:
+        type: boolean
+      confirmText:
```
yufengwng: Maybe `confirmPrompt` sounds better?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/TextQuestionDef.java
```diff
@@ -29,6 +29,16 @@
...
 
+    @SerializedName("confirmEntry")
+    private boolean confirmEntry;
+
+    @Valid
+    @SerializedName("confirmPromptTemplate")
+    private Template confirmPromptTemplate;
+
+    @SerializedName("mismatchMessage")
+    private String mismatchMessage;
```
yufengwng: I thought we said this will be a `Template` as well?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/TextQuestion.java
```diff
@@ -39,13 +39,24 @@
     @SerializedName("suggestions")
     private List<String> suggestions;
 
+    @SerializedName("confirmEntry")
+    private boolean confirmEntry;
+
+    @SerializedName("confirmEntryText")
```
yufengwng: `confirmPrompt`?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/MiscUtil.java
bskinner: Per RFC 6532, email addresses were extended to support full UTF-8 characters in both the local part and the domain part so just checking for the english & numeric digits will kick out valid addresses as well. That regex also does not require at least one period in the domain-part- `test@domain` is accepted.

If we want to go dead simple, `\S+\@\S+\.\S+` is adequate.

ssettipalli:
please run tests

ssettipalli:
please run tests

yufengwng (approved):
LGTM

PR #17, opened 2019-12-17
DDP-4053 spec: permanent announcement messages

yufengwng:
DDP-4053 adds feature for "permanent" announcement messages. To help clients identify permanent messages, we're adding a `guid` to message objects returned from the API and a `permanent` flag.

Corresponding API/Angular changes will be made separately. This PR is only the OpenAPI spec changes.

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1522,6 +1522,7 @@ AnnouncementModel:
   properties:
     guid:
       type: string
+      format: uuid
       description: an identifier for this message
       example: c8716864-3cd8-4a4b-8736-d2eec9756ad9
     permanent:
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1516,8 +1516,17 @@ Validation.IntRange:
 AnnouncementModel:
   type: object
   required:
+    - guid
+    - permanent
     - message
   properties:
+    guid:
+      type: string
+      description: an identifier for this message
+      example: c8716864-3cd8-4a4b-8736-d2eec9756ad9
+    permanent:
+      type: boolean
+      description: if true, then message is permanent and should not be able to be dismissed by user
     message:
       type: string
 Drug:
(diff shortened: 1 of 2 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1516,8 +1516,17 @@ Validation.IntRange:
 AnnouncementModel:
   type: object
   required:
+    - guid
```
fat23cat, line 1519: Could you please explain why do we need `guid` here?

yufengwng: I believe @bskinner suggested it. It might come in handy if frontend clients are caching the permanent messages. If so, then the guid is there so you can identify which message is which, and helps prevent displaying duplicate messages. Don't think clients need to use it if it's not needed.

fat23cat: Okay, spec looks good to me. Do we have a frontend ticket for these changes? I could jump on it tomorrow.

yufengwng: I don't know what's required on the frontend yet. If we're okay that permanent msg can be "dismissed" then we won't have to do anything. Otherwise, we'll need changes.

bskinner: Yep- that's it. It gives the clients a way to track individual messages (whether it's client-side only behavior, or with API interactions as well)

bskinner:
@yufengwng Is this ready for review?

yufengwng:
This is intentionally not ready for review yet. I'll mark it ready when the API changes are done, so we don't "publish" the documentation change ahead of actual implementation change :)

yufengwng:
Server-side changes in #27 is ready to go, so this is ready for review.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1516,8 +1516,17 @@ Validation.IntRange:
...
+    - permanent
     - message
   properties:
+    guid:
+      type: string
```
bskinner, line 1524: If we're using a real UUID for this, we can define this using the `format` property. It's not explicitly defined by the OpenAPI spec, but we can use arbitrary strings here (tooling should just fall back to the `type` if the `format` is undefined)
```suggestion
      type: string
      format: uuid
```

yufengwng: Oh nice

yufengwng:
please run tests

bskinner (approved):
👍

PR #21, opened 2019-12-18
DDP-4053 osteo: add announcement msg for former proxies

yufengwng:
After child participant registers for the study, we create a new announcement message for their proxy. When former proxy logs in, they will see it and it cannot be dismissed.


zyme (dismissed):

yufengwng:
This one is ready to go! _please run tests_

yufengwng:
please run tests

ssettipalli (approved):
looks good to me

PR #23, opened 2019-12-19
End-of-year cleanup

yufengwng:
Cleaning up a few things to close out the year.


pepper-apis/api-build.sh
```diff
@@ -220,6 +220,64 @@ function docker_nginx_push {
     docker push "broadinstitute/pepper-nginx:$tag"
 }
 
+function print_usage() {
+    echo 'usage: api-build.sh <version> <environment> <dir> <option>...'
```
bskinner: Considering using `$0`, `$(basename $0)`, or `` `basename $0` `` here instead of hardcoding the script name.

pepper-apis/api-build.sh
```diff
@@ -220,6 +220,64 @@ function docker_nginx_push {
...
+  --nginx           build and push docker image for nginx
+  --test            build and test docker images locally
+EOM
+}
+
+###
+#
+# Main
+
+if (( $# == 1 )); then
```
bskinner, line 271: Why is the `((` construct being used here instead of `[[`?

bskinner (commented):
A question & suggestion, but looks good otherwise

pepper-apis/api-build.sh
yufengwng: sure

pepper-apis/api-build.sh
yufengwng: I think people recommend using `((` instead of `[[` when doing numeric comparisons in bash?

zyme (approved):

yufengwng:
please run tests

PR #25, opened 2020-01-02
builder: allow activity validation list to be optional

yufengwng:
Make the `validations` key optional so we don't have to backfill it.


yufengwng:
please run tests

ssettipalli (approved):

PR #40, opened 2020-01-08
DDP-4276 Copying answers from one activity instance to another

zyme:
DDP-4276 is about copying answers from one instance to another to help participants avoid doing double entry.  This PR is _just_ the code around copying the answers from one place to another.  This might appear a bit over-engineered, but while working through composites, I thought it was worth doing this now so that down the road we have some flexibility with how we copy answers of various types.

There's a new interface called `AnswerCopier`, with implementations for each kind of question.  Most implementations are very similar until you get to `CompositeAnswerCopier`, where all the fun is.  Each copier class has wiggle room to return `AnswerCopyResult` derivatives, which in earlier drafts where quite different, but are now all the same.  After reviewing the implementation for copying data from questions to other places (such as profile) with Marco, I felt we needed a different approach for copying data between activities that was a bit closer to the database.

`CompositeAnswerCopyConfiguration` is basically a from/to for child answers.  I expect these values to be baked into some TBD schema changes and hocon, which I hadn't added to DDP-4276.  I welcome suggestions on this.

I added `equals()` and `hashCode()` in a few places where doing so struck me as low-risk to aid in tests.  In other places, such as AnswerRow, I resorted to `instanceof` and friends in tests.

At the moment, lots of classes sit inside `ActivityAnswerCopierSql`.  They could be moved out to better locations--what do folks think?
 
  - [x] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  
 ## Risk Assessment
    - [x] There is an elevated risk to security and privacy.  We're copying around data here, so let's be double careful.
 
 ### FUD Score 
  - [x] :relaxed: All good, business as usual!  Once I had some time to plan this out, things mostly fit into place.
 
 in jira.
 - [x] Demoing this requires some `main()` to create a new activity instance, copy stuff, and then look at either DB data or load things up in the browser.  We could safely merge this up and then start messing with things on dev as a next step.
 
 ## UI/UX
  - [ ] There ain't no UI/UX in these changes
 
###  Backend
   - [x] I have considered error handling and notification to slack alert rooms via `LOG.error()`.
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  - [x] Some analysis has been done on performance, but more is needed on full dev.
  
## Testing
 
## Release
     - [ ] Participant UX
     - [ ] Study staff


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
```diff
@@ -0,0 +1,405 @@
...
+                                    long destinationInstanceId,
+                                    long createdAtEpochMillis,
+                                    long lastUpdatedAtEpochMillis,
+                     CompositeAnswerCopyConfiguration compositeCopyConfiguration);
+    }
+
+    /**
+     * Copies top-level answer and composite_answer_item (when directed by compositeCopyConfig)
+     */
+    abstract class BaseAnswerCopier<T extends AnswerCopyResult> implements AnswerCopier {
```
MocanaAtBroad: Looks like for all subclasses `T` is `AnswerCopyResult`. Maybe this doesn't need to be generic?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
```diff
@@ -0,0 +1,405 @@
...
+
+
+        public T copyAnswer(Handle handle, String sourceQuestionStableId, String destinationQuestionStableId,
+                                           long sourceInstanceId, long destinationInstanceId, long createdAtEpochMillis,
+                            long lastUpdatedAtEpochMillis, CompositeAnswerCopyConfiguration compositeCopyConfig) {
```
MocanaAtBroad: Doesn't look like we use argument `compositeCopyConfig`

MocanaAtBroad: A subjective style preference: the argument pattern for a bunch of these methods appears to be:
- sourceStableId
- destinationStableId
- sourceInstanceId 
- destinationInstanceId
Since in my mind it is the combination of actitivty instance id and question stable id that define what a source and destination are, I would find more comforting to see the source specific args next to each other and dest specific args next to each other.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.sql.stg
```diff
@@ -0,0 +1,107 @@
...
+	from
+    answer a2
+    where
+    a2.question_id = src_a.question_id
+    and
+    a2.operator_user_id = src_a.operator_user_id
+    and
+    a2.activity_instance_id = src_ai.activity_instance_id)
+>>
+
```
MocanaAtBroad, line 20: Would it simplify query if instead of `last_update_at` subquery, you  `SORT` by that column and then use `LIMIT 1` to get the first item ?

MocanaAtBroad: Might be an equivalent query, but it would be more orthodox to have the query arguments be used in the `WHERE` clause rather than as part of the `JOIN` clauses.

MocanaAtBroad: Is `operator_user_id` needed?

MocanaAtBroad (commented):
Need to split, but here are some initial observations.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
zyme: An earlier iteration had composite-specific data in a `CompositeCopyResult` which was interesting for a time.  I don't have a concrete scenario for when we might want to introduce a new subclass, just more of a vague "might be useful someday".  I'll de-genericize.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
zyme: Will do!  I actually went back and forth on this one: "should I make each src/dest pair closer or put all sources and all destinations closer"?

zyme: @MocanaAtBroad it's true that `BaseAnswerCopier` doesn't use `compositeCopyConfig`, but it's here as per the signature of `AnswerCopier`.  We could set things up so that `BaseAnswerCopier` doesn't implement `AnswerCopier` so it's more of a helper that implementing classes can use directly instead of via `super`.  What's your preference?

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.sql.stg
zyme: Oy, my prehistoric oracle days are coming back to haunt me--order by and limit definitely make this simpler!  I'm not sure about performance, but I suspect that it's a wash.  Gotta do the sort somewhere, right?

zyme: I don't think we need to restrict this by operator.  We're just copying whatever the latest answer is, regardless of who put it in.  Although now I'm second guessing myself.  Is there a scenario where this would get us into trouble?

zyme: Will definitely move the args into a `where`.

spookymotion (dismissed):
LGTM

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @UseStringTemplateSqlLocator
+    @SqlQuery("getLatestAnswer")
+    @RegisterConstructorMapper(AnswerToCopy.class)
+    AnswerToCopy getLatestAnswer(@Bind("questionStableId") String sourceQuestionStableId,
+                         @Bind("activityInstanceId") long sourceInstanceId);
+
+    @UseStringTemplateSqlLocator
+    @SqlUpdate("copyTextAnswer")
+    void copyTextAnswer(@Bind("sourceAnswerId") long answerId, @Bind("destinationAnswerId") long destinationAnswerId);
```
yufengwng, line 41: Should return type be `int` for number of rows inserted?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/copyanswer/AgreementAnswerCopier.java
```diff
@@ -0,0 +1,20 @@
...
+public class AgreementAnswerCopier extends BaseAnswerCopier {
+
+    @Override
+    public AnswerCopyResult copyAnswer(Handle handle, long sourceInstanceId, String sourceQuestionStableId, long destinationInstanceId,
+                                       String destinationQuestionStableId, long createdAtEpochMillis, long lastUpdatedAtEpochMillis,
+                                       CompositeAnswerCopyConfiguration compositeCopyConfiguration) {
+        AnswerCopyResult baseCopyResult = super.copyAnswer(handle, sourceInstanceId, sourceQuestionStableId, destinationInstanceId,
+                destinationQuestionStableId, createdAtEpochMillis, lastUpdatedAtEpochMillis, compositeCopyConfiguration);
+
+        handle.attach(ActivityAnswerCopierSql.class).copyAgreementAnswer(baseCopyResult.getSourceAnswerId(),
```
yufengwng, line 15: How about we pass in an instance of `ActivityAnswerCopierSql` into the constructor instead of calling `handle.attach()`? We should already have one when constructing the copiers in the `switch` statement, e.g. `new TextAnswerCopier(this)`. Once you have that, you can also get rid of the `Handle` argument since we can call `copierSql.getHandle()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/copyanswer/BooleanAnswerCopier.java
```diff
@@ -0,0 +1,20 @@
...
+public class BooleanAnswerCopier extends BaseAnswerCopier {
+
+    @Override
+    public AnswerCopyResult copyAnswer(Handle handle, long sourceInstanceId, String sourceQuestionStableId, long destinationInstanceId,
+                                       String destinationQuestionStableId, long createdAtEpochMillis, long lastUpdatedAtEpochMillis,
+                                       CompositeAnswerCopyConfiguration compositeCopyConfiguration) {
+        AnswerCopyResult baseCopyResult = super.copyAnswer(handle, sourceInstanceId, sourceQuestionStableId, destinationInstanceId,
+                destinationQuestionStableId, createdAtEpochMillis, lastUpdatedAtEpochMillis, compositeCopyConfiguration);
+
+        handle.attach(ActivityAnswerCopierSql.class).copyBooleanAnswer(baseCopyResult.getSourceAnswerId(),
```
yufengwng, line 15: For error handling, should we check the result of `copyBooleanAnswer()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.java
yufengwng: Is there any more utility to `BaseAnswerCopier`? Since it only calls one method `copyBaseAnswer()` I feel like we should just get rid of the inheritance and have the various copiers just do the base copy.

zyme:
Making this PR official, now that I've extracted the bajillion classes from `ActivityAnswerCopierSql` into better places.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/copyanswer/AgreementAnswerCopier.java
zyme: Interesting idea!  I'll give this a whirl.  One concern with that approach is that it might be easier for clients to loose track of state and potentially use two different connections when they should only be using one.  Will address this with some javadoc.

yufengwng: But if we only pass in `ActivityAnserCopierSql` then the only way to get a connection is to call `copierSql.getHandle()` so I don't see how we might end up with two different connections?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/copyanswer/BooleanAnswerCopier.java
zyme: I'll go through each subclass and apply a similar check.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityAnswerCopierSql.sql.stg
```diff
@@ -0,0 +1,101 @@
...
+dest_q.question_id,
+src_a.operator_user_id,
+dest_ai.activity_instance_id,
+:createdAt,
+:lastUpdatedAt
+from
+answer src_a
+join question_stable_code as dest_q_code on dest_q_code.stable_id = :destinationStableId
+join question as dest_q on dest_q.question_stable_code_id = dest_q_code.question_stable_code_id
+join activity_instance as dest_ai on dest_ai.activity_instance_id = :destinationActivityInstanceId and dest_q.study_activity_id = dest_ai.study_activity_id
```
yufengwng, line 98: I'm a bit worried about this piece of sql. Question stableId is not unique enough -- we'll need to restrict by study id as well. We can probably get that by joining over some tables using `sourceInstanceId`.

Also, probably not an issue now but what if there's different revisions of the question? We probably need to restrict it down to the question_id that matches up with the `destinationInstanceId`.

zyme: Nice catch--I'll add a few more criteria and for both uniqueness and revisioning.

yufengwng:
Is this ready to go? Maybe we can merge this in first, and fix things up as we work off of the `ddp-4276-feature` feature branch?

yufengwng:
Going to move this PR forward by merging into the feature branch `ddp-4276-feature`. Further improvements will be done separately.

PR #45, opened 2020-01-10
osteo: allow save for prequal country questions

yufengwng:
Setting `allowSave=true` on the country questions for Osteo's prequal so we can clear out those answers for DDP-4278.


yufengwng:
please run tests

MocanaAtBroad (approved):

PR #48, opened 2020-01-14
DDP 4202 post consent flow

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4202
Implemented using first Impl sketch (reuse existing  RELEASE_SELF activity)
Once reconsent (CONSENT) is completed RELEASE_SELF is created.
Emails are triggered as per scenario#1 [CONSENT and RELEASE_SELF] event configurations.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/user/EnrollmentStatusType.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/user/EnrollmentStatusType.java
@@ -11,7 +11,7 @@ public enum EnrollmentStatusType {
     ENROLLED(true, false, false,  true),
     EXITED_BEFORE_ENROLLMENT(false, true, true, false),
     EXITED_AFTER_ENROLLMENT(false, true, true, false),
-    CONSENT_SUSPENDED(true, false, true, true);
+    CONSENT_SUSPENDED(true, false, false, true);
 
     private final boolean shouldReceiveCommunications;
 
--- a/study-builder/studies/osteo/study.conf
+++ b/study-builder/studies/osteo/study.conf
@@ -2400,7 +2400,8 @@
         ! (user.studies["CMI-OSTEO"].forms["PARENTAL_CONSENT"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["PARENTAL_CONSENT"].isStatus("COMPLETE")
-        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE"))
+        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE")
+        && !user.studies["CMI-OSTEO"].forms["RELEASE_SELF"].hasInstance())
       """,
       "dispatchToHousekeeping": true,
       "order": 1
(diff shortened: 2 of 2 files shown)


yufengwng (commented):
Looks good so far

ssettipalli:
please run tests

study-builder/studies/osteo/study.conf
```diff
@@ -2400,7 +2400,8 @@
         ! (user.studies["CMI-OSTEO"].forms["PARENTAL_CONSENT"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["PARENTAL_CONSENT"].isStatus("COMPLETE")
-        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE"))
+        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE")
+        && !user.studies["CMI-OSTEO"].forms["RELEASE_SELF"].hasInstance())
```
ssettipalli: needed to check if minor is aged-up and if so cancel pdf generation with minor template

study-builder/studies/osteo/study.conf
```diff
@@ -2417,7 +2418,8 @@
         ! (user.studies["CMI-OSTEO"].forms["CONSENT_ASSENT"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].hasInstance()
         && user.studies["CMI-OSTEO"].forms["CONSENT_ASSENT"].isStatus("COMPLETE")
-        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE"))
+        && user.studies["CMI-OSTEO"].forms["RELEASE_MINOR"].isStatus("COMPLETE")
+        && !user.studies["CMI-OSTEO"].forms["RELEASE_SELF"].hasInstance())
```
ssettipalli: needed to check if minor is aged-up and if so cancel pdf generation with minor template

ssettipalli:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/model/user/EnrollmentStatusType.java
```diff
@@ -11,7 +11,7 @@
     ENROLLED(true, false, false,  true),
     EXITED_BEFORE_ENROLLMENT(false, true, true, false),
     EXITED_AFTER_ENROLLMENT(false, true, true, false),
-    CONSENT_SUSPENDED(true, false, true, true);
+    CONSENT_SUSPENDED(true, false, false, true);
```
yufengwng, line 14: I think the initial idea with `consent_suspended(readonly: true)` was so that it always results in their existing data being read-only until they re-consent. One drawback with changing it to `false` is that read-only is no longer guaranteed and that studies will need to setup configuration to appropriately mark activities read-only.

study-builder/studies/osteo/study.conf
yufengwng: I think I get what we're trying to say here, maybe it's clearer to write expression as below?
```
! ( has_parental()...
&& is_complete())
|| has_release_self()
```

pepper-apis/src/main/java/org/broadinstitute/ddp/model/user/EnrollmentStatusType.java
yufengwng: I think we can make this change, but we need to keep in mind to remind studies to add configuration to mark stuff read-only

yufengwng (approved):

ssettipalli:
please run tests

PR #52, opened 2020-01-16
Ddp 3989 create activity in workflow if missing

raiks2:

The origin of this PR is https://broadinstitute.atlassian.net/browse/DDP-3989, which, however, contains only a general idea. The actual functionality this PR implements is more specific: it generalizes the idea of creating the "next" activity instance in the workflow if it doesn't exist. We were doing it for the start activities before, but now this is done for in `WorkflowService` for _any_ activity.
 
  - [x] New feature (non-breaking change which adds functionality)
  
 ## Risk Assessment
 
 ### FUD Score 
  - [x] :scream:I'm sweaty and nervous
 
 - [] They are user-visible in dev as a regular user journey and require no additional instructions.
 in jira.
 - [x] I have no idea how to demo this.  Please help me!
 
 ## UI/UX
 
###  Backend
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  - [x] I'm unsure about stability, performance, fault tolerance, and graceful degradation.  Please help!
  
## Testing
 
## Release
     - [ ] Participant UX
     - [ ] Study staff


pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetWorkflowRoute.java
```diff
@@ -72,11 +70,7 @@ public WorkflowResponse handle(Request request, Response response) {
             LOG.info("Finding suggestions using 'from' state {}", fromState);
 
             return workflowService
-                    .suggestNextState(handle, userGuid, studyGuid, fromState)
-                    .map(nextState -> {
-                        checkAndCreateStartActivityInstance(handle, fromState, nextState, operatorGuid, userGuid);
-                        return nextState;
-                    })
+                    .suggestNextState(handle, operatorGuid, userGuid, studyGuid, fromState)
```
raiks2, line 73: Since the functionality is now incorporated into the `WorkflowService`, there's no need for special handling anymore.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetWorkflowRoute.java
```diff
@@ -88,25 +82,6 @@ public WorkflowResponse handle(Request request, Response response) {
...
-                ActivityInstanceDto instanceDto = handle.attach(ActivityInstanceDao.class)
-                        .insertInstance(activityState.getActivityId(), operatorGuid, userGuid, InstanceStatusType.CREATED, false);
-                LOG.info("Created start activity instance with guid '{}' for user guid '{}' using operator guid '{}' and activity id {}",
-                        instanceDto.getGuid(), userGuid, operatorGuid, activityState.getActivityId());
-            } else {
-                LOG.info("User guid '{}' already has start activity instance with guid '{}'", userGuid, instanceGuid);
-            }
-        }
-    }
-
```
raiks2, line 109: This method is not needed anymore - `WorkflowService` will do the job.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/SendEmailRoute.java
```diff
@@ -49,10 +51,10 @@ public SendEmailRoute(WorkflowService workflowService) {
 
     @Override
     public Object handle(Request request, Response response, SendEmailPayload payload) throws Exception {
+        DDPAuth ddpAuth = RouteUtil.getDDPAuth(request);
         String email = payload.getEmail();
-
         String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
-
+        String operatorGuid = ddpAuth.getOperator();
```
raiks2: For inserting a new activity instance, we need an `operatorGuid`. `SendEmailRoute` is a bit special because it's called in the context of a study rather than a user. However, since it requires a token, I expect it to be able to extract the `operatorGuid`. Does it make sense?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -28,7 +32,13 @@ public WorkflowService(PexInterpreter interpreter) {
         this.interpreter = interpreter;
     }
 
-    public Optional<WorkflowState> suggestNextState(Handle handle, String userGuid, String studyGuid, WorkflowState fromState) {
+    public Optional<WorkflowState> suggestNextState(
+            Handle handle,
+            String operatorGuid,
```
raiks2, line 42: `operatorGuid` parameter was added to allow the activity instance insertion.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetWorkflowRouteTest.java
```diff
@@ -152,28 +150,6 @@ public void testGet_fromActivityState_acceptsActivityCode() {
...
-        given().auth().oauth2(token)
-                .queryParam(QueryParam.FROM, StateType.ACTIVITY)
-                .queryParam(QueryParam.INSTANCE_GUID, fromInstance.getGuid())
-                .when().get(url).then().assertThat()
-                .statusCode(200).contentType(ContentType.JSON)
-                .body("next", equalTo(StateType.ACTIVITY.name()))
-                .body("activityCode", equalTo(nextActivity.getActivityCode()))
-                .body("instanceGuid", is(nullValue()));
-    }
-
```
raiks2: This test checks if the route returns NULL `instanceGuid` for the "next" activity with a missing instance, it doesn't work this way now that we create a missing activity instance.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PutFormAnswersRouteTest.java
```diff
@@ -460,7 +459,7 @@ public void testWorkflowResponse_unknownSuggestion_whenNoSatisfactoryTransitions
     }
 
     @Test
-    public void testWorkflowResponse_activityStateSuggestion_providesActivityCode() {
+    public void testWorkflowResponse_activityStateSuggestion_providesActivityCodeAndInstanceGuid() {
```
raiks2, line 462: The workflow response now contains a code of an automatically created activity.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/WorkflowServiceTest.java
```diff
@@ -86,6 +87,7 @@ public static void setup() {
                     mgmtSecret,
                     sendgridApiKey);
             userGuid = data.getUserGuid();
+            operatorGuid = data.getUserGuid();
```
raiks2, line 100: For tests, we consider `operatorGuid` = `userGuid`.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/SendEmailRoute.java
yufengwng: Actually, I don't think `SendEmailRoute` is an authenticated route. This means we're not going to have a `DDPAuth` token nor an operatorGuid. In these cases, I think we should just use `participantGuid` as operator guid.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -59,7 +69,12 @@ public WorkflowService(PexInterpreter interpreter) {
             }
         }
 
-        return Optional.ofNullable(next);
+        Optional<WorkflowState> nextState = Optional.ofNullable(next);
+        nextState.ifPresent(nextWfState -> {
+            createActivityInstanceIfMissing(handle, fromState, nextWfState, operatorGuid, userGuid);
```
yufengwng: A bit of nitpicking, but have we considered just calling this method directly right before the return, instead of having to nest it inside the `ifPresent()`? It's just a plain old `if` and we can also pull up the type check as well:
```
if (next != null && next.type() == ACTIVITY) {
    create((ActivityState) next);
}
return Optional()
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -75,4 +90,29 @@ public WorkflowResponse buildStateResponse(Handle handle, String userGuid, Workf
...
+        }
+        ActivityState activityState = (ActivityState) nextState;
+        String instanceGuid = handle.attach(JdbiActivityInstance.class)
+                .findLatestInstanceGuidByUserGuidAndActivityId(userGuid, activityState.getActivityId())
+                .orElse(null);
+        boolean activityInstanceIsMissing = instanceGuid == null;
+        if (activityInstanceIsMissing) {
+            ActivityInstanceDto instanceDto = handle.attach(ActivityInstanceDao.class)
+                    .insertInstance(activityState.getActivityId(), operatorGuid, userGuid, InstanceStatusType.CREATED, false);
+            LOG.info("Created start activity instance with guid '{}' for user guid '{}' using operator guid '{}' and activity id {}",
```
yufengwng:
```suggestion
            LOG.info("Created activity instance with guid '{}' for user guid '{}' using operator guid '{}' and activity id {}",
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -75,4 +90,29 @@ public WorkflowResponse buildStateResponse(Handle handle, String userGuid, Workf
...
+    ) {
+        if (nextState.getType() != StateType.ACTIVITY) {
+            return;
+        }
+        ActivityState activityState = (ActivityState) nextState;
+        String instanceGuid = handle.attach(JdbiActivityInstance.class)
+                .findLatestInstanceGuidByUserGuidAndActivityId(userGuid, activityState.getActivityId())
+                .orElse(null);
+        boolean activityInstanceIsMissing = instanceGuid == null;
+        if (activityInstanceIsMissing) {
```
yufengwng, line 119: Should we also check that the activity definition (in `study_activity` table) has `max_instances_per_user` is `== null || > 0`? If it's zero it means we can't have any instances. I doubt we'll see that in practice, but just in case?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -59,7 +69,12 @@ public WorkflowService(PexInterpreter interpreter) {
...
-        return Optional.ofNullable(next);
+        Optional<WorkflowState> nextState = Optional.ofNullable(next);
+        nextState.ifPresent(nextWfState -> {
+            createActivityInstanceIfMissing(handle, fromState, nextWfState, operatorGuid, userGuid);
+        });
+
+        return nextState;
     }
 
     public WorkflowResponse buildStateResponse(Handle handle, String userGuid, WorkflowState state) {
```
yufengwng, line 85: Now that we're always returning an `instanceGuid`, do you think it makes sense to include the instance guid directly in the `ActivityState` object? So we don't have to do another query to find it in `buildStateResponse()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
raiks2: Yes, thanks for suggesting it, sounds like a viable alternative. However, it seems to me that the gain in speed is infinitesimal while readability can become slightly worse. Also, since we already use `Optional`, makes sense to follow the same style. Given all that, I'd stick to the original code 😃

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
raiks2: Thanks, that's a really important point!

raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/SendEmailRoute.java
raiks2: Yeah, makes sense, changed this.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
raiks2: Yes, this optimization seemed natural after the changes were made to `WorkflowService` and avoiding one more DB trip is always nice. However, I tinkered with it a bit and got a feeling that it doesn't worth the trouble - too many breaking changes for a small gain. If you insist, let's do that, but I'd leave it as-is. @zyme, what do you think?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -75,4 +95,72 @@ public WorkflowResponse buildStateResponse(Handle handle, String userGuid, Workf
...
+                participantId,
+                participantGuid,
+                instanceId,
+                activityId,
+                studyId,
+                InstanceStatusType.CREATED
+        );
+        EventService.getInstance().processAllActionsForEventSignal(handle, activityCreationSignal);
+    }
 }
```
raiks2: Ok, so that's how the post-instance-creation event processing should look like according to my understanding: we publish an activity instance creation signal and then ask the `EventService` to process all the event configurations associated with it. @yufengwng please point out if I'm missing anything here.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/WorkflowServiceTest.java
```diff
@@ -360,6 +372,136 @@ public void testBuildStateResponse_activityState_providesActivityMetadata() {
...
+                    )
+            );
+
+            Optional<String> latestInstanceGuid = handle.attach(JdbiActivityInstance.class)
+                    .findLatestInstanceGuidByUserGuidAndActivityId(userGuid, triggeredFormActivity.getActivityId());
+            assertTrue(latestInstanceGuid.isPresent());
+            handle.rollback();
+        });
+    }
+
```
raiks2: Making sure that when a missing instance is created, all dependent events get processed.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/WorkflowServiceTest.java
```diff
@@ -360,6 +372,136 @@ public void testBuildStateResponse_activityState_providesActivityMetadata() {
...
+                            actParticipatingInTransition.get("activityTransitionedFromId")
+                    )
+            );
+
+            Optional<String> latestInstanceGuid = handle.attach(JdbiActivityInstance.class)
+                    .findLatestInstanceGuidByUserGuidAndActivityId(userGuid, actParticipatingInTransition.get("activityTransitionedToId"));
+            assertTrue(latestInstanceGuid.isPresent());
+            handle.rollback();
+        });
+    }
```
raiks2: This test and a bunch of similar tests below make sure that a new instance is created/not created, depending on the `maxInstancePerUser` values.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -75,4 +95,70 @@ public WorkflowResponse buildStateResponse(Handle handle, String userGuid, Workf
...
+    private void processActionsForActivityCreationSignal(
+            Handle handle,
+            String operatorGuid,
+            long participantId,
+            String participantGuid,
+            long instanceId,
+            long activityId,
+            long studyId
+    ) {
+        long operatorId = handle.attach(JdbiUser.class).getUserIdByGuid(operatorGuid);
```
raiks2, line 152: Observation: the DAO method returns `Optional` while the column in `study_activity.study_id` is `NOT NULL`. This means that `Optional<Long>` can be replaced with just `long`. Let's clean it up later in another PR.

raiks2:
OK, folks, so after adding code for processing events dependent on an instance creation and adding more tests I believe it's ready for another round of code review.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetWorkflowRouteTest.java
yufengwng: We should keep this test. The point of the test is to check that it accepts instanceGuid in query params (as the test name suggests) not that the route returns null for instanceGuid. We can change it to check `is(notNullValue())`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
yufengwng: Since it's no longer just the "start" activity, we should remove `start` from this log.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
yufengwng: I'm okay with refactoring it separately.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
yufengwng: Looks right according to my understanding as well.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetWorkflowRouteTest.java
raiks2: Ok, thanks, didn't realize that. Put it back.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
raiks2: Done.

yufengwng (approved):
This looks good to me

raiks2:
Please run tests

raiks2:
Please run tests

PR #53, opened 2020-01-16
DDP 4202 suspend consent refactor

ssettipalli:
Changes to update All existing activity instances as read-only in database when consent is suspended. 
Changes will ensure that we are not dependent on study config to identify and mark activity instances as read-only with consent suspend.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUserStudyEnrollment.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUserStudyEnrollment.java
@@ -310,17 +310,15 @@ public interface JdbiUserStudyEnrollment extends SqlObject {
         return changeUserStudyEnrollmentStatus(userId, studyId, newEnrollmentStatus, null);
     }
 
-    default long suspendUserStudyConsent(String userGuid, String studyGuid) {
+    default long suspendUserStudyConsent(long userId, long studyId) {
         //update EnrollmentStatus to EnrollmentStatusType.CONSENT_SUSPENDED
         //Update All existing activity instances as read-only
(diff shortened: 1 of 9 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUserStudyEnrollment.java
```diff
@@ -307,6 +310,18 @@ default long changeUserStudyEnrollmentStatus(long userId,
         return changeUserStudyEnrollmentStatus(userId, studyId, newEnrollmentStatus, null);
     }
 
+    default long suspendUserStudyConsent(String userGuid, String studyGuid) {
+        //update EnrollmentStatus to EnrollmentStatusType.CONSENT_SUSPENDED
+        //Update All existing activity instances as read-only
+        long id = changeUserStudyEnrollmentStatus(
+                userGuid, studyGuid, EnrollmentStatusType.CONSENT_SUSPENDED, null);
+        Set<Long> instanceIds = getActivityInstanceDao().findActivityInstanceIdsByStudyAndUser(studyGuid, userGuid);
+        long userId = getJdbiUser().getUserIdByGuid(userGuid);
```
yufengwng: We have `policy.getStudyId()` and `candidate.getParticipantUserId()`. I think using ids instead of guids will simplify these changes.

ssettipalli: Its easy to use Ids but felt guid's is more user / client friendly way.
Will change to Ids

ssettipalli:
please run tests

yufengwng (approved):
LGTM

ssettipalli:
please run tests

PR #55, opened 2020-01-21
DDP-4276 osteo: adding copying for about_child -> about_you

yufengwng:
This PR adds copying configuration we'll need for Osteo.

### DIFF
--- a/study-builder/studies/osteo/study.conf
+++ b/study-builder/studies/osteo/study.conf
@@ -2366,7 +2366,7 @@
       "order": 1
     },
 
-    ## reconsent events
+    # reconsent events
 
     ## collecting child's contact email
     {
@@ -2486,7 +2486,6 @@
     },
 
     ## events at age-up
-
     {
       "trigger": {
         "type": "REACHED_AOM"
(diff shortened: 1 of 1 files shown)


study-builder/studies/osteo/study.conf
```diff
@@ -2606,43 +2647,174 @@
...
+              "questionStableId": "CHILD_HOW_HEAR"
+            },
+            "target": {
+              "type": "ANSWER",
+              "questionStableId": "HOW_HEAR"
+            }
+          },
+        ]
       },
+      "cancelExpr": """!user.studies["CMI-OSTEO"].forms["ABOUTCHILD"].hasInstance()""",
```
ssettipalli, line 2814: wondering if we should add ["ABOUTCHILD"] status = "Completed" for additional safety !!

ssettipalli (approved):

study-builder/studies/osteo/study.conf
yufengwng: I think that might result in the wrong behavior. User is not required to submit the `about_child` activity, so it is likely that when we get there during age-up, the `about_child` is not complete yet. Regardless of whether it's complete or not, it's still safe to do copying -- if there's no answer, we don't copy anything over.

PR #54, opened 2020-01-16
DDP-4276 implement copy handlers

yufengwng:
## Context and the big picture

This PR implements the actual copying functionality, which processes through the copy configuration and copy answers from source location to the appropriate target. Both the copy-answer and create-instance event actions now hook into this "executor" to do the copying. This implementation is feature-parity with old copying functionality, so existing studies should continue to work in the same way.
 
 ## What type of change is this?
  
  - [ ] Bugfix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Experimental proof-of-concept
  - [ ] Infosec/Compliance remediation
  
 ## Risk Assessment
  - [x] These changes are **HIGH RISK**.
    - [x] There is an elevated risk that may impact a large number of features
    - [x] There is an elevated risk to security and privacy (since we're copying answers)
    - [ ] These changes introduce new 3rd party dependencies, alter the network perimeter, or contain
    major upgrades to persistent storage or SaaS dependencies such as auth0, sendgrid, or easypost    
  - [ ] These changes are low risk, do not impact security/privacy, do not impact pepper shared
  components/modules, and have no major upgrades to 3rd party services or libraries
 
 ### FUD Score 
 Overall, how are you feeling about these changes?
  - [ ] :relaxed: All good, business as usual!
  - [x] :sweat_smile: There might be some issues here (the composite answers are tough...)
  - [ ] :scream:I'm sweaty and nervous
 
 ## How to we demo these changes?
 How does one observe these changes in a deployed system?  Note that _user visible_ encompasses
 many personas--not just patients and study staff, but ops your fellow devs, compliance, etc.
 
 - [x] They are user-visible in dev as a regular user journey and require no additional instructions.
 - [ ] Getting dev into a state where this is user-visible requires some tech fiddling.  I have
 documented these steps in the related ticket.
 - [ ] Requires other features before it's human visible.  I have documented the blocking issues
 in jira.
 - [ ] I have no idea how to demo this.  Please help me!
 
 ## UI/UX
  - [x] There ain't no UI/UX in these changes
  - [ ] My changes have passed muster with Erin and Jen via an over-the-shoulder review, screenshots, etc.
  - [ ] Erin and Jen have not had an opportunity to provide feedback yet
 
###  Backend
   - [x] I have considered error handling and notification to slack alert rooms via `LOG.error()`.
     #### Route Changes
[...]

PR #49, opened 2020-01-14
DDP-4276 database changes for copy configuration

yufengwng:
## Context and the big picture

This PR encompasses the database schema changes, data migration/backfilling, model classes, and DAO queries. Once we have these, we can then work on updating copy-answer action handler and study-builder.

Tips on reading this PR:
1. Start with liquibase XML file and spend your time here -- there's some critical things going on here.
2. Skim through the new `ddp.model.copy` package and the classes in there.
3. Read `CopyConfigurationDao`, its test, and its related sql files.
4. Skim through the event configuration / event action changes -- these are just plumbing to get thing setup for processing.

_NOTE: tests are mostly passing but things don't really work until we implement the other changes, so we're merging into a feature branch for now._

 ## What type of change is this?
  
  - [ ] Bugfix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)
  - [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Experimental proof-of-concept
  - [ ] Infosec/Compliance remediation
  
 ## Risk Assessment
  - [x] These changes are **HIGH RISK**.
    - [x] There is an elevated risk that may impact a large number of features
    - [ ] There is an elevated risk to security and privacy
    - [ ] These changes introduce new 3rd party dependencies, alter the network perimeter, or contain
    major upgrades to persistent storage or SaaS dependencies such as auth0, sendgrid, or easypost    
  - [ ] These changes are low risk, do not impact security/privacy, do not impact pepper shared
  components/modules, and have no major upgrades to 3rd party services or libraries
 
 ### FUD Score 

  - [x] :relaxed: All good, business as usual!
  - [ ] :sweat_smile: There might be some issues here
  - [ ] :scream:I'm sweaty and nervous
 
 ## How to we demo these changes?

 - [ ] They are user-visible in dev as a regular user journey and require no additional instructions.
 - [ ] Getting dev into a state where this is user-visible requires some tech fiddling.  I have
 documented these steps in the related ticket.
 - [x] Requires other features before it's human visible.
 - [ ] I have no idea how to demo this.  Please help me!
 
 ## UI/UX
  - [x] There ain't no UI/UX in these changes
  - [ ] My changes have passed muster with Erin and Jen via an over-the-shoulder review, screenshots, etc.
  - [ ] Erin and Jen have not had an opportunity to provide feedback yet
 
###  Backend
   - [x] I have considered error handling and notification to slack alert rooms via `LOG.error()`.
[...]

PR #59, opened 2020-01-22
Ddp 4276 answer-copying feature

yufengwng:
Final PR to get answer-copying feature into mainline. Some of the pieces had been reviewed in separate smaller PRs:

* #49 for database/model changes
* #50 for study-builder changes
* #54 for implementation of copying
* #55 for osteo specific configuration for copying


yufengwng:
please run tests

yufengwng:
please run tests

yufengwng:
please run tests

study-builder/studies/osteo/study.conf
```diff
@@ -1499,8 +1350,53 @@
       },
       "action": {
         "type": "COPY_ANSWER",
-        "copySourceQuestionStableId": "PARENTAL_CONSENT_CHILD_DOB",
-        "copyTarget": "PARTICIPANT_PROFILE_BIRTH_DATE",
+        "copyConfigPairs": [
```
zyme, line 1353: So cool to see this as "just configuration" :)

pepper-apis/src/test/java/org/broadinstitute/ddp/util/TestFormActivity.java
```diff
@@ -206,12 +221,20 @@ public TestFormActivity build(Handle handle, long userId, String studyGuid) {
...
             if (picklistSingleListOptions != null) {
                 var question = PicklistQuestionDef
-                        .buildSingleSelect(PicklistRenderMode.LIST, "PL" + Instant.now().toEpochMilli(), Template.text("picklist prompt"))
+                        .buildSingleSelect(PicklistRenderMode.LIST, "PL" + Instant.now().toEpochMilli(), Template.text("pl single prompt"))
                         .addOptions(List.copyOf(picklistSingleListOptions))
                         .build();
                 result.picklistSingleListQuestion = question;
                 picklistBlocks.add(new QuestionBlockDef(question));
             }
+            if (picklistMultiListOptions != null) {
```
zyme, line 230: How did I miss `TestFormActivity` last week?  Nice bit of test util code.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,176 @@
...
+            targetInstance.putAnswer(result);
+        }
+        CompositeAnswer targetParentAnswer = (CompositeAnswer) result;
+
+        // Line up the rows between source and target composite answers
+        List<AnswerRow> sourceRows = parentAnswer.getValue();
+        List<AnswerRow> targetRows = targetParentAnswer.getValue();
+        while (targetRows.size() < sourceRows.size()) {
+            targetRows.add(new AnswerRow());
+        }
```
zyme, line 103: Do we need to be concerned about a case where `targetRows.size() > sourceRows.size()` ?

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,176 @@
...
+
+            for (int colIdx = 0; colIdx < sourceRow.size(); colIdx++) {
+                Answer sourceChild = sourceRow.get(colIdx);
+                if (sourceChild == null || !sourceChild.getQuestionStableId().equals(sourceQuestion.getStableId())) {
+                    continue;
+                }
+                Answer targetChild = targetRow.get(colIdx);
+                if (targetChild == null) {
+                    targetChild = createTargetAnswer(sourceChild, targetInstance.getId(), targetQuestion);
+                    jdbiCompositeAnswer.insertChildAnswerItems(
```
zyme, line 121: Kudos to whoever made this `@SqlBatch`'d already--the performance numbers @yufengwng shared look good, so making this sort of optimization way back in the early days of composites was smart :)

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,176 @@
...
+        } else if (type == QuestionType.TEXT) {
+            String value = ((TextAnswer) sourceAnswer).getValue();
+            targetAnswer = new TextAnswer(null, targetQuestion.getStableId(), null, value);
+        } else {
+            throw new DDPException("Unhandled copying for answer type " + type);
+        }
+        return answerDao.createAnswer(operatorId, targetInstanceId, targetQuestion.getId(), targetAnswer);
+    }
+
+    private Answer updateTargetAnswer(Answer sourAnswer, long targetInstanceId, Answer targetAnswer) {
```
zyme: `sourAnswer` indeed!  I would prefer this exception be thrown directly by client code that thinks it needs to issue an update.  It's possible that someone could see the method name, _not_ look at the implementation, and think "Oh cool, I'll just call `updateTargetAnswer` and then...surprise!

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
yufengwng: I pondered it a bit but not sure how we want to handle it. For now, we're only copying from the size range of the source since that's what's available in the source.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
yufengwng: Ahh typo! And sure, we can have it just throw an exception for now.

yufengwng:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToProfileCopier.java
```diff
@@ -0,0 +1,134 @@
...
+    public void copy(FormResponse sourceInstance, QuestionDto sourceQuestion, CopyLocationType target) {
+        Answer sourceAnswer = sourceInstance.getAnswer(sourceQuestion.getStableId());
+        if (sourceAnswer != null) {
+            copySourceAnswer(sourceInstance, sourceAnswer, target);
+        } else {
+            sourceAnswer = extractSourceChildAnswer(sourceInstance, sourceQuestion);
+            if (sourceAnswer != null) {
+                copySourceAnswer(sourceInstance, sourceAnswer, target);
+            } else {
+                LOG.info("No source answer to copy from activity instance {} and question {} to target {}",
```
zyme, line 49: I think this is going to generate a lot of noise in our logs.  I suggest we remove this.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToProfileCopier.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public Answer extractSourceChildAnswer(FormResponse sourceInstance, QuestionDto sourceQuestion) {
+        CompositeQuestionDto parentDto = parentDtosByChildId.get(sourceQuestion.getId());
+        if (parentDto == null) {
+            parentDto = jdbiCompositeQuestion
+                    .findParentDtoByChildQuestionId(sourceQuestion.getId())
+                    .orElse(null);
+            if (parentDto == null) {
+                LOG.info("No parent composite question found for source question {}", sourceQuestion.getStableId());
+                return null;
```
zyme, line 90: `Optional`?

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToProfileCopier.java
```diff
@@ -0,0 +1,134 @@
...
+            return null;
+        }
+
+        List<Answer> answers = parentAnswer.getValue().stream()
+                .flatMap(row -> row.getValues().stream())
+                .filter(child -> sourceQuestion.getStableId().equals(child.getQuestionStableId()))
+                .collect(Collectors.toList());
+        if (answers.size() > 1) {
+            throw new DDPException("Composite child question " + sourceQuestion.getStableId() + " has multiple answers");
+        } else if (answers.size() == 1) {
```
zyme, line 109: My suggestion here should be taken with some salt in terms of modern VM performance and silly micro-optimizations, but consider `!answers.isEmpty()` here and `answers.iterator().next()` on L110.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/copy/CopyConfigurationPair.java
```diff
@@ -0,0 +1,41 @@
+package org.broadinstitute.ddp.model.copy;
+
+public class CopyConfigurationPair {
+
+    private long id;
```
zyme, line 5: Other than `order`, the fields are all immutable, so these could be `final`.  I find this handy not so much for performance as for avoiding oopsies when writing constructors.

pepper-apis/src/main/resources/db-changes/schema/DDP-4276-restructure-copy-answer.xml
```diff
@@ -0,0 +1,165 @@
...
+
+                    INSERT INTO copy_configuration_pair (copy_configuration_id, source_location_id, target_location_id, execution_order)
+                    VALUES (cfg_id, source_loc_id, target_loc_id, 1);
+
+                    UPDATE copy_answer_event_action SET copy_configuration_id = cfg_id WHERE event_action_id = action_id;
+                END LOOP;
+                CLOSE actions;
+            END
+        </createProcedure>
+        <sql>CALL backfill_copy_action_configurations()</sql>
```
zyme, line 143: Should we issue some logging in the stored procedure to get a sense of progress on the backfill?  The number of rows is quite small, and not very different (at all) between prod and non-prod so maybe this isn't helpful.

pepper-apis/src/main/resources/db-changes/schema/DDP-4276-restructure-copy-answer.xml
```diff
@@ -0,0 +1,165 @@
...
+
+                    SELECT umbrella_study_id INTO study_id FROM event_configuration
+                     WHERE event_action_id = action_id;
+
+                    INSERT INTO copy_configuration (study_id) VALUES (study_id);
+                    SET cfg_id = LAST_INSERT_ID();
+
+                    INSERT INTO copy_location (copy_location_type_id)
+                    SELECT copy_location_type_id FROM copy_location_type WHERE copy_location_type_code = 'ANSWER';
+                    SET source_loc_id = LAST_INSERT_ID();
```
zyme, line 125: If we're being really paranoid, checking `LAST_INSERT_ID` for nonzero is the way to go.

pepper-apis/src/main/resources/db-changes/schema/DDP-4276-restructure-copy-answer.xml
```diff
@@ -0,0 +1,165 @@
...
+
+        <dropColumn tableName="copy_answer_event_action" columnName="source_question_stable_id"/>
+        <dropColumn tableName="copy_answer_event_action" columnName="copy_answer_target_id"/>
+
+        <dropTable tableName="copy_answer_target"/>
+    </changeSet>
+
+    <changeSet author="yufeng" id="20200110-expand-answer-timestamps-to-millis">
+        <update tableName="answer">
+            <column name="created_at" valueComputed="created_at * 1000"/>
```
zyme, line 161: Yay milliseconds!

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.sql.stg
```diff
@@ -0,0 +1,117 @@
+group ActivityInstanceSql;
+
+all_form_instance_responses_with_latest_status_and_answers_select() ::= <<
+select ai.activity_instance_id   as a_instance_id,
```
zyme, line 4: Now _this_ is what I call SQL!

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/CopyConfigurationDao.sql.stg
```diff
@@ -0,0 +1,40 @@
...
+       src_ans.question_stable_code_id as source_question_stable_code_id,
+       (select stable_id from question_stable_code
+         where question_stable_code_id = src_ans.question_stable_code_id
+       ) as source_question_stable_id,
+       tgt.copy_location_id as target_copy_location_id,
+       (select copy_location_type_code from copy_location_type
+         where copy_location_type_id = tgt.copy_location_type_id
+       ) as target_copy_location_type,
+       tgt_ans.question_stable_code_id as target_question_stable_code_id,
+       (select stable_id from question_stable_code
```
zyme, line 21: I find the select-in-the-select pattern handy for this kind of operation when compared to a big ol' pile of outer joins. 👍

pepper-apis/src/test/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopierTest.java
```diff
@@ -0,0 +1,402 @@
...
+            assertEquals(QuestionType.TEXT, row3.get(0).getQuestionType());
+            assertEquals("row 3 col 1", ((TextAnswer) row3.get(0)).getValue());
+
+            var row4 = rows.get(3).getValues();
+            assertEquals(2, row4.size());
+            assertNull(row4.get(0));
+            assertEquals(QuestionType.DATE, row4.get(1).getQuestionType());
+            assertEquals(new DateValue(2020, 9, null), ((DateAnswer) row4.get(1)).getValue());
+
+            handle.rollback();
```
zyme, line 328: Most of the tests here are rolling back, but some aren't.  Would it muddy the waters to have a `handle` field and `@Before` to get a new one, and `@After` to rollback?

zyme: Could be a fun sidebar on creating a `@JdbiRollback` test method annotation for general use.

zyme (commented):
Overall this is great.  There are some places where immutable fields aren't `final`--not a big deal, just something to consider going forward as a good practice to set off alarms in IDEs about missing assignments/initialization.

CC @Rinchan20 let's make sure to do some extra testing around this, using sparse data for data copying and verifying that instance creation that doesn't use instance copying continues to not copy answers.

Still waiting for tests to complete for me locally...

zyme:
please run tests

zyme (approved):

pepper-apis/src/test/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopierTest.java
yufengwng: Ohh that sounds nice. Sorta like JUnit's `@Rule` and `ExpectedException`. I'll investigate separately.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToProfileCopier.java
yufengwng: I can remove in separate cleanup PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.java
```diff
@@ -0,0 +1,73 @@
...
+            @BindList(value = "userGuids", onEmpty = BindList.EmptyHandling.NULL) Set<String> userGuids,
+            @Define("limitActivities") boolean limitActivities,
+            @BindList(value = "activityCodes", onEmpty = BindList.EmptyHandling.NULL) Set<String> activityCodes);
+
+    @UseStringTemplateSqlLocator
+    @SqlQuery("bulkQueryFormResponsesWithAnswersByUsersAndActivityIds")
+    @RegisterConstructorMapper(value = FormResponse.class, prefix = "a")
+    @UseRowReducer(FormResponsesWithAnswersForUsersReducer.class)
+    Stream<FormResponse> findFormResponsesWithAnswersByUsersAndActivityIds(
+            @Define("byId") boolean byId,
```
MocanaAtBroad, line 51: Maybe ought to be `byUserId`?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.java
```diff
@@ -0,0 +1,73 @@
...
+            @BindList(value = "userIds", onEmpty = BindList.EmptyHandling.NULL) Set<Long> userIds,
+            @BindList(value = "userGuids", onEmpty = BindList.EmptyHandling.NULL) Set<String> userGuids,
+            @BindList(value = "activityIds", onEmpty = BindList.EmptyHandling.NULL) Set<Long> activityIds);
```
MocanaAtBroad, line 54: Reading query, the other options for `onEmpty` (THROW or VOID) seem like better choices (at least for `activityIds`)

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,173 @@
...
+import org.broadinstitute.ddp.model.activity.types.QuestionType;
+import org.jdbi.v3.core.Handle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AnswerToAnswerCopier {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AnswerToAnswerCopier.class);
+
+    private final Handle handle;
```
MocanaAtBroad, line 35: `handle` appears to go unused.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,173 @@
...
+            List<SelectedPicklistOption> value = ((PicklistAnswer) sourceAnswer).getValue();
+            targetAnswer = new PicklistAnswer(null, targetQuestion.getStableId(), null, value);
+        } else if (type == QuestionType.TEXT) {
+            String value = ((TextAnswer) sourceAnswer).getValue();
+            targetAnswer = new TextAnswer(null, targetQuestion.getStableId(), null, value);
+        } else {
+            throw new DDPException("Unhandled copying for answer type " + type);
+        }
+        return answerDao.createAnswer(operatorId, targetInstanceId, targetQuestion.getId(), targetAnswer);
+    }
```
MocanaAtBroad, line 172: More of meta comment: I think we have this sort of if(type == QuestionType)/case statement logic in multiple places in our code where we want to exhaustively deal with all the different kinds of questions.

I think this makes it tough to add new question types and keep them up to date; need to figure out all the places in code base that deal with question types and make sure your question is dealt with.

Maybe a possible future refactoring would be to invert the this functionality and ask: if we have a new question type, what interfaces do we have to implement so that this new question type can plug into all the other question functionality (querying, copying, updating, etc)?

In this specific case it appears we are creating question specific answers. Maybe we can have a AnswerBuilder<BooleanQuestion> or something of the sort that the question type implementor would have to create.

Then code operating on question and answers can strictly deal with the object at a generic API level and not have to worry too much (or at all) with specific question/answer types.

At any rate, not asking for changes, just that I have seen and written code of this sort that needs to deal with every question type before and think we can do this in  cleaner/simpler fashion.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
```diff
@@ -0,0 +1,173 @@
...
+            copySourceAnswer(sourceAnswer, targetInstance, targetQuestion);
+        } else {
+            Answer parentAnswer = retrieveParentQuestion(sourceQuestion)
+                    .map(parentDto -> (CompositeAnswer) sourceInstance.getAnswer(parentDto.getStableId()))
+                    .orElse(null);
+            if (parentAnswer != null) {
+                copyFromParentAnswer((CompositeAnswer) parentAnswer, sourceQuestion, targetInstance, targetQuestion);
+            } else {
+                LOG.info("No source answer to copy from activity instance {} and question {}",
+                        sourceInstance.getGuid(), sourceQuestion.getStableId());
```
MocanaAtBroad, line 63: IMO this method is doing a little too much, specifically there is some logic and operation here trying to determine if this is a composite or not and if yes try to fetch that answer. Would seem clearer to have this method expect the caller to have retrieved the answer already. I think OK to treat composite as separate/special case.

MocanaAtBroad: Are we precluding copying to/from a simple question/composite question?
Very realistic case is composite question with allowMultiple=false. E.g., a composite question/multiple=false with children "firstName" and "lastName". In this case each child ought to be treated a just a "simple" question in the activity instance and copied to/from any other.

MocanaAtBroad (commented):
Some initial stuff...

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.java
yufengwng: I think `byId` is clear enough. There's only `userGuid` in here, and it's mostly an "internal" sql method that shouldn't be used that widely but rather be exposed more nicely via DAO methods.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.java
yufengwng: Curious why the other two variants will be better? Seems better to always return something than `THROW`-ing. We could use `VOID` (which results in sql like `where .. in ()`) which pretty much results in the same behavior as `NULL` (i.e. `where .. in (null)`).

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
yufengwng: Right. I left it here in case we need to do other operations.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
yufengwng: I get the same feeling too. The alternative is basically having the answer objects insert themselves, which as you mentioned will boil down to having an interface that answer objects implement that takes a `Handle` as argument and inserts themselves into the database. But up to this point, we have consolidated all the database code in `ddp.db` package, so spreading out the DB code to the answer objects seems like a significant paradigm shift.

pepper-apis/src/main/java/org/broadinstitute/ddp/copy/AnswerToAnswerCopier.java
yufengwng: _Re: method doing too much._ Out-sourcing the answer lookup to the caller is definitely an alternative. The caller (currently only `CopyExecutor`) has all the question DTOs too, so might be able to leverage some sharing of parent question DTOS between the copiers.

_Re: precluding._ Let's see if I understand the question. Are you asking if it allows copying from a child question to a non-composite question? This is not supported right now because I have not implemented it. The use-case is definitely reasonable, but the ticket is already a big piece of work and it doesn't seem like we will leverage that feature any time soon, so I took a shortcut here. I should have listed it as another limitation in `studies/study-schema.conf`, sorry about that. It will be a nice exercise for folks who want to jump in on the copier code!

PR #60, opened 2020-01-22
Fix export of birth date

yufengwng:
Fixed an issue with exports in `dev`.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
@@ -21,6 +21,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -612,14 +620,11 @@ void enrichWithDSMEventDates(Handle handle,
...
                 if (birthDate != null && aomRule != null) {
-                    LocalDate localBirthDate = birthDate.asLocalDate().orElse(null);
-                    if (localBirthDate != null) {
-                        dateOfMajority = aomRule.getDateOfMajority(localBirthDate);
-                    }
+                    dateOfMajority = aomRule.getDateOfMajority(birthDate);
                 }
             }
 
-            participant.setBirthDate(birthDate.asLocalDate().orElse(null));
```
yufengwng, line 622: Alerts in Slack are pointing to this line. If there's no birth date, then it looks like this will throw an NPE.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -601,8 +603,14 @@ void enrichWithDSMEventDates(Handle handle,
         dataset.forEach(participant -> {
             DateValue diagnosisDate = medicalRecordService
                     .getDateOfDiagnosis(handle, participant.getUser().getId(), studyId).orElse(null);
-            DateValue birthDate = medicalRecordService
-                    .getDateOfBirth(handle, participant.getUser().getId(), studyId).orElse(null);
+
+            // First grab from the profile, if it's not there then look in activity data using medical-record service
```
yufengwng, line 606: Some studies, like Osteo, will have the birth date copied into the profile, so I think we should grab from there as a more "canonical" source. And then fallback to `MedicalRecordService` for studies that doesn't have copying (like Angio/Brain).

yufengwng:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
spookymotion: Annoying. Thanks for fixing this.

spookymotion (dismissed):

yufengwng:
please run tests

spookymotion (approved):

PR #65, opened 2020-01-27
osteo: show CHILD_CONTACT when logging in

yufengwng:
Fixes DDP-4357.

If user has not completed the `CHILD_CONTACT` activity yet, then workflow should show it when they login. When a user logs back in, frontend client calls Workflow API with `from` state `RETURN_USER`, so we add a transition rule there.


yufengwng:
please run tests

raiks2 (approved):

PR #72, opened 2020-01-30
Make age-up check process more fault-tolerant

yufengwng:

Updated the age-up job/service so that there's better isolation between studies and participants. Also changed how we run the `REACHED_AOM` events, i.e. we'll only run these events after participant has aged-up AND they have an invitation ready.

Some nice properties after these changes:
* Errors in one study will not affect another study
* Errors for one participant in a study will not affect other participants
* Status will always be changed to `consent_suspended` even if we run into errors processing events
* We'll run the `REACHED_AOM_PREP` events first, so that age-up process can be kicked off
* Since we added restriction of having an invitation, once they do have invitation, we kick off the `REACHED_AOM` events. This means that age-up process will be somewhat automated and continues to work even if parent fills out Child Contact _after_ someone reached their date-of-majority.
* And since we changed `consent_suspended` status _before_ the `REACHED_AOM_PREP` events, it means we marked existing activities read-only before creating the Child Contact activity, so that even in the situation where we found out about age-up before we created this activity, we still get the desired read-only behavior while Child Contact is still editable to continue the process.
 
  - [x] Bugfix (non-breaking change which fixes an issue)
  
 ## Risk Assessment
    - [x] There is an elevated risk to security and privacy
 
 ### FUD Score 
 
 in jira.
 
 ## UI/UX
 
###  Backend
   - [x] I have considered error handling and notification to slack alert rooms via `LOG.error()`.
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  
## Testing
 
## Release
     - [ ] Participant UX
     - [ ] Study staff

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/schedule/CheckAgeUpJob.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/schedule/CheckAgeUpJob.java
@@ -17,7 +17,6 @@ import org.broadinstitute.ddp.pex.PexInterpreter;
 import org.broadinstitute.ddp.pex.TreeWalkInterpreter;
 import org.broadinstitute.ddp.service.AgeUpService;
 import org.broadinstitute.ddp.util.ConfigUtil;
-import org.jdbi.v3.core.Handle;
 import org.quartz.CronScheduleBuilder;
 import org.quartz.DisallowConcurrentExecution;
 import org.quartz.Job;
@@ -75,7 +74,7 @@ public class CheckAgeUpJob implements Job {
         try {
             LOG.info("Running job '{}'", getKey());
             long start = Instant.now().toEpochMilli();
-            TransactionWrapper.useTxn(TransactionWrapper.DB.APIS, this::run);
+            run();
             long elapsed = Instant.now().toEpochMilli() - start;
             LOG.info("Job '{}' completed in {} ms", getKey(), elapsed);
         } catch (Exception e) {
@@ -84,12 +83,13 @@ public class CheckAgeUpJob implements Job {
         }
     }
 
-    private void run(Handle handle) {
+    private void run() {
         AgeUpService service = new AgeUpService();
         PexInterpreter interpreter = new TreeWalkInterpreter();
-        List<GovernancePolicy> policies = handle.attach(StudyGovernanceDao.class)
-                .findAllPolicies()
-                .collect(Collectors.toList());
+        List<GovernancePolicy> policies = TransactionWrapper.withTxn(TransactionWrapper.DB.APIS, handle ->
+                handle.attach(StudyGovernanceDao.class)
+                        .findAllPolicies()
+                        .collect(Collectors.toList()));
         Collections.shuffle(policies);
         for (var policy : policies) {
             if (policy.getAgeOfMajorityRules().isEmpty()) {
@@ -98,7 +98,8 @@ public class CheckAgeUpJob implements Job {
             }
             try {
                 LOG.info("Running age-up check for study {}", policy.getStudyGuid());
-                service.runAgeUpCheck(handle, interpreter, policy);
+                // Perform check for each study within a transaction so any unhandled errors for study will not affect other studies.
+                TransactionWrapper.useTxn(TransactionWrapper.DB.APIS, handle -> service.runAgeUpCheck(handle, interpreter, policy));
                 LOG.info("Finished age-up check for study {}", policy.getStudyGuid());
             } catch (DDPException e) {
                 LOG.error("Error while checking age-up for study {}, continuing", policy.getStudyGuid(), e);
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/service/AgeUpService.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/service/AgeUpService.java
@@ -10,6 +10,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.broadinstitute.ddp.db.TransactionWrapper;
+import org.broadinstitute.ddp.db.dao.InvitationDao;
 import org.broadinstitute.ddp.db.dao.JdbiUserStudyEnrollment;
 import org.broadinstitute.ddp.db.dao.StudyGovernanceDao;
 import org.broadinstitute.ddp.model.activity.types.EventTriggerType;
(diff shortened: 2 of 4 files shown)


yufengwng:
please run tests

raiks2 (commented):
Looks good in general, just double-check that the interplay of savepoints doesn't cause any unpredicted effects. Looks like you are doing some checks in the test already, so probably nothing else is required.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/schedule/CheckAgeUpJob.java
```diff
@@ -98,7 +98,8 @@ private void run(Handle handle) {
             }
             try {
                 LOG.info("Running age-up check for study {}", policy.getStudyGuid());
-                service.runAgeUpCheck(handle, interpreter, policy);
+                // Perform check for each study within a transaction so any unhandled errors for study will not affect other studies.
```
zyme, line 101: Ah, so this was a "one bad apple ruins the bunch" kind of scenario.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/AgeUpServiceTest.java
```diff
@@ -174,6 +189,58 @@ public void testRunAgeUpCheck_reachedAgeOfMajorityPrep() {
...
+                    Instant.now().toEpochMilli(), null, 0, null, null, true, 1);
+
+            GovernancePolicy policy = new GovernancePolicy(1L, testData.getStudyId(), testData.getStudyGuid(), new Expression("true"));
+            policy.addAgeOfMajorityRule(new AgeOfMajorityRule("true", 20, 4));
+            assertEquals("should have only completed 1 candidate",
+                    1, service.runAgeUpCheck(handle, interpreter, policy));
+
+            try {
+                // Wait for events to be ready
+                TimeUnit.SECONDS.sleep(1);
```
zyme, line 215: Is there a way to dip into internal state of event dispatch and query it directly to ensure the things you expect to happen have happened?  Sleeping will slow things down and we might get a false failure with a `sleep()`.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/AgeUpServiceTest.java
```diff
@@ -174,6 +189,58 @@ public void testRunAgeUpCheck_reachedAgeOfMajorityPrep() {
...
+            } catch (InterruptedException ignored) {
+            }
```
zyme, line 218: You might want to log this,  as an interruption here might lead to a test failure.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/AgeUpService.java
```diff
@@ -88,39 +92,85 @@ public int runAgeUpCheck(Handle handle, PexInterpreter interpreter, GovernancePo
...
+                            userGuid, studyGuid, e);
+                    handle.rollbackToSavepoint(checkpoint("status", studyGuid, userGuid));
+                    continue;
+                }
+            }
+
+            if (shouldPrepForAgeUp && !candidate.hasInitiatedPrep()) {
+                LOG.info("Candidate {} in study {} has reached preparation for age-of-majority, processing age-up preparation events",
+                        userGuid, studyGuid);
+                handle.savepoint(checkpoint("prep", studyGuid, userGuid));
```
zyme: Since the checkpoint strings are such a critical part of this, they should be factored out to constants.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/AgeUpServiceTest.java
yufengwng: That's what I wanted to do as well. But given the state of the current DAOs, I worry that it will introduce refactors/unrelated code that I'll need to add to this PR. I'm positive we can remove this sleep after we done some minimal DAO/tests cleanup.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/AgeUpService.java
```diff
@@ -88,39 +92,85 @@ public int runAgeUpCheck(Handle handle, PexInterpreter interpreter, GovernancePo
...
+                try {
+                    EventSignal signal = new EventSignal(
+                            candidate.getParticipantUserId(),
+                            candidate.getParticipantUserId(),
+                            candidate.getParticipantUserGuid(),
+                            policy.getStudyId(),
+                            EventTriggerType.REACHED_AOM_PREP);
+                    EventService.getInstance().processAllActionsForEventSignal(handle, signal);
+                    preppedCandidateIds.add(candidate.getId());
+                    handle.release(checkpoint("prep", studyGuid, userGuid));
```
zyme: Might be more robust to have a `finally` do the `release` here and above.  I'm not sure if there would need to be additional error handling if the savepoint is gone after rolling back.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/schedule/CheckAgeUpJob.java
```diff
@@ -98,7 +98,8 @@ private void run(Handle handle) {
...
+                TransactionWrapper.useTxn(TransactionWrapper.DB.APIS, handle -> service.runAgeUpCheck(handle, interpreter, policy));
```
zyme, line 102: Since `runAgeUpCheck` runs in its own transaction now, I think it might be cleaner to have `runAgeUpCheck` use three distinct transactions, rather than savepoint/rollback, since explicit savepoints and rollbacks are a bit more complicated.

yufengwng: Not sure I fully understand the suggestion. `runAgeUpCheck` is running within a transaction already. Are we suggesting to open additional transactions within this to do the different pieces of change? I was thinking having transactions-within-transactions is discouraged?

If the concern is around managing the savepoints, which I'm also concerned about, I think we can provide methods to help manage the lifecycle for it. Idea here is very similar to what we do for opening a transaction. E.g. let's say we have this method signature in `TransactionWrapper`:
```java
public static <X extends Exception> void useSavepoint(String name, Handle handle, HandleConsumer<X> callback) throws X {}
```
Then we can use it to wrap the "critical region" of the code that mutates data:
```java
try {
    EventSignal signal = new EventSignal(..., REACHED_AOM_PREP);
    TransactionWrapper.useSavepoint(checkpoint("prep", study, user), handle,
            h -> EventService.getInstance().processAllActionsForEventSignal(h, signal));
} catch (Exception e) {
    LOG.error("Error ... rolling back and skipping",);
    continue;
}
```
What do folks think?

yufengwng:
please run tests

yufengwng:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/schedule/CheckAgeUpJob.java
zyme: I like the `useSavepoint()` approach!

zyme (approved):

PR #79, opened 2020-02-05
Ddp 4378 normalize activity name

yufengwng:
## Context and the big picture

Renamed
* `name` -> `title`
* `dashboardName` -> `name`

Will also fixes DDP-4376.

 ## What type of change is this?
  
  - [x] Bugfix (non-breaking change which fixes an issue)
  - [ ] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Experimental proof-of-concept
  - [ ] Infosec/Compliance remediation
  - [x] Refactor
  
 ## Risk Assessment
  - [ ] These changes are **HIGH RISK**.
    - [ ] There is an elevated risk that may impact a large number of features
    - [ ] There is an elevated risk to security and privacy
    - [ ] These changes introduce new 3rd party dependencies, alter the network perimeter, or contain
    major upgrades to persistent storage or SaaS dependencies such as auth0, sendgrid, or easypost    
  - [x] These changes are low risk, do not impact security/privacy, do not impact pepper shared
  components/modules, and have no major upgrades to 3rd party services or libraries
 
 ### FUD Score 
 Overall, how are you feeling about these changes?
  - [x] :relaxed: All good, business as usual!
  - [ ] :sweat_smile: There might be some issues here
  - [ ] :scream:I'm sweaty and nervous
 
 ## How to we demo these changes?
 How does one observe these changes in a deployed system?  Note that _user visible_ encompasses
 many personas--not just patients and study staff, but ops your fellow devs, compliance, etc.
 
 - [x] They are user-visible in dev as a regular user journey and require no additional instructions.
 - [ ] Getting dev into a state where this is user-visible requires some tech fiddling.  I have
 documented these steps in the related ticket.
 - [ ] Requires other features before it's human visible.  I have documented the blocking issues
 in jira.
 - [ ] I have no idea how to demo this.  Please help me!
 
 ## UI/UX
  - [x] There ain't no UI/UX in these changes
  - [ ] My changes have passed muster with Erin and Jen via an over-the-shoulder review, screenshots, etc.
  - [ ] Erin and Jen have not had an opportunity to provide feedback yet
 
###  Backend
   - [ ] I have considered error handling and notification to slack alert rooms via `LOG.error()`.
     #### Route Changes
     - [ ] My changes involve a new route changes to an existing route, so the who/what/when are logged using the usual logback mechanism
at the _start_ of the route, prior to any validation, as well as at the end of the route when work completes.
[...]

PR #80, opened 2020-02-05
DDP-4378 api: rename activity name and title

yufengwng:
Renamed activity attributes:
* `name` -> `title`
* `dashboardName` -> `name`

Backend changes are in #79, Angular SDK changes are in https://github.com/broadinstitute/ddp-angular/pull/45

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -10,7 +10,6 @@ ActivityInstanceStatusType:
     - COMPLETE
     - CREATED
     - IN_PROGRESS
-    - TERMINATED
 FormType:
   title: FormType
   type: string
@@ -114,25 +113,49 @@ Activity:
   required:
     - activityType
     - formType
+    - activityCode
     - guid
+    - title
+    - subtitle
     - readonly
     - status
     - listStyleHint
     - sections
     - lastUpdatedText
     - lastUpdated
+    - isFollowup
   properties:
     activityType:
       $ref: '../pepper.yml#/components/schemas/ActivityType'
     formType:
       $ref: '../pepper.yml#/components/schemas/FormType'
+    activityCode:
+      type: string
+      description: unique identifier of the activity
     guid:
       type: string
-      description: the unique identifier for this instance of an activity
+      description: unique identifier for this instance of the activity
     status:
       type: string
+    title:
+      type: string
+      description: title of the activity to be displayed
+    subtitle:
+      type: string
+      nullable: true
+      description: |
+        A string to display to the user as a subtitle for the activity. If the
+        `ddp-Content-Style` header is set to `BASIC`, this field will contain
+        unstructured text.
     readonly:
       type: boolean
+    readonlyHint:
+      type: string
+      nullable: true
+      description: |
+        A string to display when an activity is in a read-only state (the
+        `readonly` flag is `true`). If the `ddp-Content-Style` header is set to
+        `BASIC`, this field will contain unstructured text.
     listStyleHint:
       $ref: '../pepper.yml#/components/schemas/ListStyle'
     introduction:
@@ -143,20 +166,6 @@ Activity:
         $ref: '../pepper.yml#/components/schemas/Section'
     closing:
       $ref: '../pepper.yml#/components/schemas/Section'
-    subtitle:
-      type: string
-      nullable: true
-      description: |
-        A string to display to the user as a subtitle for the activity. If the
-        `ddp-Content-Style` header is set to `BASIC`, this field will contain
-        unstructured text.
-    readonlyHint:
-      type: string
-      nullable: true
-      description: |
-        A string to display when an activity is in a read-only state (the
-        `readonly` flag is `true`). If the `ddp-Content-Style` header is set to
-        `BASIC`, this field will contain unstructured text.
     lastUpdatedText:
       type: string
       nullable: true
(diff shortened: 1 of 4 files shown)


bskinner:
Is there a critical reason why this needs to be in now, or is this just a cosmetic change?

yufengwng:
It's not critical. I have fixed the underlying issue on the server-side while keeping the API the same. When's a better time to submit this API change, @bskinner?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -10,7 +10,6 @@ ActivityInstanceStatusType:
     - COMPLETE
     - CREATED
     - IN_PROGRESS
-    - TERMINATED
```
bskinner, line 13: This change is not noted in the PR's description- why is this needed at this time?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -114,49 +113,59 @@ Activity:
   required:
     - activityType
     - formType
+    - activityCode
     - guid
+    - title
```
bskinner, line 118: Are these values guaranteed to be present in the response, or are we considering them nullable?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -169,18 +178,21 @@ Activity:
...
   type: object
   title: Activity Summary
   required:
     - activityCode
-    - activityDashboardName
-    - activityDescription
+    - activityType
+    - activitySubtype
     - activityName
+    - activityTitle
```
bskinner, line 192: What is the requirement for this change? This is not cosmetic, and results in a breaking change to the API.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1023,15 +1040,15 @@ PrequalifierModel:
   type: object
   required:
     - guid
-    - name
+    - title
```
bskinner, line 1052: What is the requirement behind this change? This is a breaking change to the API and will require client modifications.

bskinner (changes requested):
There are non-cosmetic changes to the API which break compatibility with existing clients.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: This status is no longer in the API (it has been removed some time ago). Just cleaning things up.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: `title` is guaranteed and non-null.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: I'm not aware of a product/stakeholder requirement for this change. This was done as part of refactor for DDP-4378, which helps resolve bug DDP-4376.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: We can postpone API change. When is a better time?

yufengwng: Also, this `PrequalifierModel` is used for this "prequalifer" API which I don't think any studies uses. I'm inclined to remove this API completely.

bskinner: Let me take a look at both the apps and verify that. If neither one is using it, I'm game for adding a ticket to remove the API. One less vestigial endpoint will be nice.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Let's try and keep unrelated changes to their own PR in the future. Even if it's not being used, there may be client logic & dependencies on this and it does constitute a breaking change.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Looking at [DDP-4378](https://broadinstitute.atlassian.net/browse/DDP-4378), both the `title` and `subtitle` are mentioned to be `optional`, not required, which doesn't seem to mesh with this. Are these the properties being referred to in the ticket?

yufengwng: They're optional as in study does not have to define it. If `title` is not defined, it will be blank. `subtitle` is nullable.

zyme:
Ping!  This one is on hold...is now a better time to pop it off the queue?

bskinner:
If we can batch up a bunch of API changes together, I'd be great to get this out.

yufengwng:
Recent API changes so far are additive. I think this is the only change that's breaking. I can pull together some doc describing all the changes if that helps?

bskinner:
That would be helpful. I want to get as many of these in at one time as possible since each release requires me to create a new tag for both the apps & SDKs, update the dependencies in each app, and resubmit to the App Stores for review. In the future, expect at least a 2-3 week lead time for any breaking changes.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Good point. However, I truly believe that this was added in here haphazardly and wasn't cleaned up. The chances of clients depending on this value is UNLIKELY since the sever will not return such a value. I'll like to remove this. This has been noted in DDP-4471. Resolving comment.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: I believe this is addressed. Resolving comment.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Tracked in DDP-4471. Resolving comment.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Took note of this in DDP-4515. Resolving comment.

yufengwng:
FYI I have revived this PR.

bskinner (approved):
Let's do it! The matching ticket in Jira is labeled as being a breaking change as well, correct?

yufengwng:
please run tests

fat23cat (approved):

yufengwng:
please run tests

PR #81, opened 2020-02-06
Ddp 4366 update routes to respect language header

raiks2:
 This PR is about using a unified approach for figuring out a preferred user's language. The premise for it is https://broadinstitute.atlassian.net/browse/DDP-4366. In order to be able to figure out a list of languages an entity is translated into (a question or an activity instance, for example), a new "languages supported by the study" concept was introduced.

For this new stuff to work, we will need some backfilling: a list of supported languages must be specified for each study, or else a fallback language will be chosen. I created a ticket for that: https://broadinstitute.atlassian.net/browse/DDP-4394.
 
 ## Risk Assessment
 
 ### FUD Score 
  - [] :scream:I'm sweaty and nervous
 
 - [x] Requires other features before it's human visible.  I have documented the blocking issues
 in jira.
 
 ## UI/UX
  - [x There ain't no UI/UX in these changes
 
###  Backend
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  
## Testing
 
## Release
 - [x] Releasing these changes requires special handling
     - [ ] Participant UX
     - [ ] Study staff


pepper-apis/src/main/resources/db-changes/schema/DDP-4366-study-language-table.xml
```diff
@@ -0,0 +1,18 @@
...
+    <changeSet author="sretsia" id="20200502-create-study-language-table">
+        <createTable tableName="study_language">
+            <column name="umbrella_study_id" type="bigint">
+                <constraints nullable="false" foreignKeyName="sl_study_id_us_study_id_fk" references="umbrella_study(umbrella_study_id)"/>
+            </column>
+            <column name="language_code_id" type="bigint">
+                <constraints nullable="false" foreignKeyName="sl_lang_code_id_lc_lang_code_id_fk" references="language_code(language_code_id)"/>
+            </column>
+        </createTable>
+        <addPrimaryKey tableName="study_language" columnNames="umbrella_study_id, language_code_id" constraintName="sl_pk"/>
```
yufengwng: I thought the pattern we're following is to prefer having an auto-increment primary key column? Should we add one and make this a unique key instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
```diff
@@ -149,4 +159,18 @@ public static ActivityInstanceDto findAccessibleInstanceOrHalt(Response response
...
+     * from Accept-Language header, information from the user profile and languages
+     * supported by the study.
+     */
+    public static Locale resolvePreferredUserLanguage(
+            Handle handle, String acceptLanguageHeader, Locale preferredLocale, String studyGuid
+    ) {
+        List<LanguageRange> acceptLanguages = StringUtils.isNotEmpty(acceptLanguageHeader)
+                ? LanguageRange.parse(acceptLanguageHeader) : new ArrayList<>();
+        Set<Locale> localesSupportedByStudy = handle.attach(StudyDao.class).getSupportedLocalesByGuid(studyGuid);
+        return I18nUtil.resolvePreferredLanguage(preferredLocale, acceptLanguages, localesSupportedByStudy);
```
yufengwng: I don't see changes in `I18nUtil` in this PR. Currently, `I18nUtil` won't behave the way we want. It will prefer the `preferredLocale` over `acceptLanguages` locales, whereas we want `acceptLanguages` to have higher priority than `preferredLocale`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/StudyDao.java
```diff
@@ -24,4 +28,16 @@
     @SqlQuery("select * from study_exit_request where user_id = :userId")
     @RegisterConstructorMapper(StudyExitRequest.class)
     Optional<StudyExitRequest> findExitRequestForUser(@Bind("userId") long userId);
+
+    @SqlQuery(
+            "select lc.iso_language_code from study_language sl inner join language_code lc on sl.language_code_id = lc.language_code_id"
+            + " inner join umbrella_study us on sl.umbrella_study_id = us.umbrella_study_id where us.guid = :studyGuid"
+    )
+    Set<String> getSupportedLanguageCodesByGuid(@Bind("studyGuid") String studyGuid);
```
yufengwng: I recommend naming this `findSupportedLanguageCodesbyGuid`. Likewise for the method below. I believe naming things using the `find*` pattern is better than `get*` since those have connotations with getters.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
```diff
@@ -73,38 +78,37 @@ public Object handle(Request request, Response response) {
...
+                    handle, acceptLanguageHeader, ddpAuth.getPreferredLocale(), studyGuid
+            );
+            String isoLangCode = preferredUserLanguage.getLanguage();
+
+            LOG.info("Attempting to find a translation for the following language: {}", isoLangCode);
+            Optional<ActivityInstance> inst = actInstService.getTranslatedActivity(
+                    handle, userGuid, instanceDto.getActivityType(), instanceGuid, isoLangCode, style
+            );
+            if (!inst.isPresent()) {
+                LOG.info("Failed to find a translation to the '{}' language code for the activity instance "
```
yufengwng: Is this a truly unexpected event and qualifies for a `log.error()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
```diff
@@ -149,4 +159,18 @@ public static ActivityInstanceDto findAccessibleInstanceOrHalt(Response response
...
+     * Figures out a preferred user language taking into account the language weights
+     * from Accept-Language header, information from the user profile and languages
+     * supported by the study.
+     */
+    public static Locale resolvePreferredUserLanguage(
+            Handle handle, String acceptLanguageHeader, Locale preferredLocale, String studyGuid
+    ) {
+        List<LanguageRange> acceptLanguages = StringUtils.isNotEmpty(acceptLanguageHeader)
+                ? LanguageRange.parse(acceptLanguageHeader) : new ArrayList<>();
+        Set<Locale> localesSupportedByStudy = handle.attach(StudyDao.class).getSupportedLocalesByGuid(studyGuid);
```
yufengwng: Hmmm, is there a way to have `StudyDao` return a set of `LanguageCode` objects which has _both_ `iso_language_code` and `language_code_id`? So that we don't have to do yet another database query in the routes to `getLanguageCodeId`? Maybe something like
```java
class LangCode {
    private long langId;
    private String isoCode;
    ...
    public asLocale() {
      return Locale.forLanguageTag(isoCode);
    }
}
```

yufengwng (changes requested):
Do we need to update OpenAPI docs?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
```diff
@@ -149,4 +159,18 @@ public static ActivityInstanceDto findAccessibleInstanceOrHalt(Response response
...
+    /**
+     * Figures out a preferred user language taking into account the language weights
+     * from Accept-Language header, information from the user profile and languages
+     * supported by the study.
+     */
+    public static Locale resolvePreferredUserLanguage(
+            Handle handle, String acceptLanguageHeader, Locale preferredLocale, String studyGuid
+    ) {
+        List<LanguageRange> acceptLanguages = StringUtils.isNotEmpty(acceptLanguageHeader)
+                ? LanguageRange.parse(acceptLanguageHeader) : new ArrayList<>();
```
zyme: How about `Collections.emptyList()` instead of `new ArrayList`?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
zyme: I'd like us to do this whenever possible--those repeated calls for uk -> pk eventually will add up.

pepper-apis/src/main/resources/db-changes/schema/DDP-4366-study-language-table.xml
zyme: Agreed, this should be an autogenerated primary key, and a separate unique key for `umbrella_study_id` + `language_code_id`.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
```diff
@@ -73,38 +78,37 @@ public Object handle(Request request, Response response) {
...
+                        + "with GUID {}", isoLangCode, instanceGuid);
+                throw new DDPException("Unable to find activity instance " + instanceGuid + " of type " + instanceDto.getActivityType());
```
zyme:
```suggestion
                throw new DDPException("Unable to find activity instance " + instanceGuid + " of type " + instanceDto.getActivityType()
                        + " in " + isoLangCode);
```

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
zyme: Since this is just before a `throws`, let's just `throw`, since the logging will pick it up.

zyme (commented):
Could we isolate all the header checking using the pattern of a `Filter` that reads the language header and puts it into attribute storage?  Something like [TokenConverterFilter](https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/TokenConverterFilter.java).

We could then isolate `RouteUtil.resolvePreferredUserLanguage()` to that filter and set an attribute that is the language to use, which may or may not be what's in the profile.  And that filter could be robust to non-auth routes as well.

As part of this, to @yufengwng's point, we should update the API docs, adding note in a central place that describes that precedence for translations.

pepper-apis/src/main/resources/db-changes/schema/DDP-4366-study-language-table.xml
raiks2: Ok, I will add an autogenerated PK although I can't imagine a situation where we gonna reference such a relation table by a PK.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/StudyDao.java
raiks2: Agree, the `find` prefix reflects the intention better. We still have `get` methods in `DAOs` throughout the code, it would be nice to rename them to make things consistent.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
raiks2: That's what I was also thinking about when writing this code. Will change.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/LanguageResolutionFilter.java
```diff
@@ -0,0 +1,73 @@
...
+    public void handle(Request request, Response response) {
+        try {
+            String acceptLanguageHeader = request.headers(RouteConstants.ACCEPT_LANGUAGE);
+            Matcher matcher = Pattern.compile("/studies/(\\w+)").matcher(request.url());
+            String studyGuid = matcher.find() ? matcher.group(0) : null;
+            // The "supported languages" notion is an attribute of a study, thus is doesn't
+            // make any sense outside of the study context
+            boolean supportedLanguagesCanBeDetected = studyGuid != null;
+            if (!supportedLanguagesCanBeDetected) {
+                LOG.warn(
```
zyme: Let's remove the `warn()` call--there are all kinds of routes outside of the study and this will cause a lot of uninteresting log noise.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/LanguageResolutionFilter.java
```diff
@@ -0,0 +1,73 @@
...
+                    handle -> handle.attach(StudyDao.class).getSupportedLocalesByGuid(studyGuid)
+            );
+            Locale preferredLanguage =  I18nUtil.resolvePreferredLanguage(
+                    RouteUtil.getDDPAuth(request).getPreferredLocale(), acceptLanguages, localesSupportedByStudy
+            );
+            request.attribute(USER_LANGUAGE, preferredLanguage);
+            LOG.info("Added the preferred user language {} to the attribute store", preferredLanguage.getLanguage());
+        } catch (Exception e) {
+            LOG.error("Error while figuring out the user language", e);
+            halt(401);
```
zyme: Let's not `halt()` here, but just log an error.  If something goes wrong with this filter, since it's a `/*` filter, the whole API will be borked.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/LanguageResolutionFilter.java
```diff
@@ -0,0 +1,73 @@
...
+public class LanguageResolutionFilter implements Filter {
+
+    public static final String USER_LANGUAGE = "USER_LANGUAGE";
+    private static final Logger LOG = LoggerFactory.getLogger(LanguageResolutionFilter.class);
+
+    @Override
+    public void handle(Request request, Response response) {
+        try {
+            String acceptLanguageHeader = request.headers(RouteConstants.ACCEPT_LANGUAGE);
+            Matcher matcher = Pattern.compile("/studies/(\\w+)").matcher(request.url());
```
zyme: `RouteUtil.STUDIES_PATH_MARKER` is the place to look for `studies`--can you change to refer to that?

zyme: Actually, a nice refactor would be to clean up `RouteUtil.STUDIES_PATH_MARKER` and `RouteConstants.API.STUDY_ALL` and make `"studies"` more D-R-Y.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/I18nUtil.java
```diff
@@ -133,9 +133,9 @@ public static Locale resolvePreferredLanguage(@Nullable Locale preferredLocale,
         if (null != preferredLocale) {
             LanguageRange preferred = new LanguageRange(preferredLocale.toLanguageTag(), LanguageRange.MAX_WEIGHT);
 
-            // This is the locale the user has specifically chosen, so insert it
-            // at the top of the list so it's the first MAX_WEIGHT option.
-            priorityList.add(0, preferred);
+            // This is the locale the user has specifically chosen, insert it
+            // at the bottom of the list so it has the lowest priority
+            priorityList.add(preferred);
```
raiks2, line 138: This is the change that gives languages in `Accept-Language` header a priority over the language from the user profile.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -274,6 +275,8 @@ public void handle(Request request, Response response) throws Exception {
             }
         });
 
+        before(API.BASE + "/*", new LanguageResolutionFilter());
```
raiks2: Should we really mount the filter for all routes or limit it to ones under `/user/{userGuid}/studies/{studyGuid}`? I mean, does it make sense outside the study context given that we use the notion "languages supported by the study"?

yufengwng: That's a good point. I feel making it a filter just for `/user/{}/studies{}` routes and renaming it to something like `StudyLanguageResolveFilter` seems reasonable.

zyme: Agreed!  If/when we move  other  translated assets (thinking of `en.json`-type things from angular), we can adjust later on.

zyme (commented):
Looking  good so far.  I suspect that to really vet this, we'll  want to step through deployed frontends on mobile and web.  CC @bskinner .

I  do think it's important to include API  documentation with this PR to convey how language  selection precedent and defaults work.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
raiks2: Done, renamed the filter and made it applicable to `/user/*studies/*` only. Additionally, put back the warning and adjusted it to suggest remounting the filter under `/user/*studies/*` if someone inadvertently enables it for other routes.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/LanguageResolutionFilter.java
raiks2: Makes sense, done.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/LanguageResolutionFilter.java
raiks2: Now that we invoke the filter only in the study context, makes sense to put it back to prevent someone from inadvertently applying the filter to other routes.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
raiks2: We do it now.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
raiks2: Done.

bskinner:
At least on the iOS client, it should be automatically including an `Accept-Language` header based off which locale the phone is currently configured for. Once this hits `dev`, we can get verify the behavior and make sure the Android app is doing the same.

> Looking good so far. I suspect that to really vet this, we'll want to step through deployed frontends on mobile and web. CC @bskinner .
> 
> I do think it's important to include API documentation with this PR to convey how language selection precedent and defaults work.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,98 @@
...
+        // will have to look into the db to get the id of the fallback language
+        LanguageDto preferredLanguage = Optional.ofNullable(supportedStudyLocaleToLang.get(preferredLocale))
+                .orElseGet(
+                        () -> new LanguageDto(
+                                handle.attach(JdbiLanguageCode.class).getLanguageCodeId(preferredLocale.getLanguage()),
+                                preferredLocale.getLanguage()
+                        )
+                );
+        return preferredLanguage;
+    }
```
raiks2: I created this `public` method to separate plumbing from logic and make `StudyLanguageResolutionFilter` testable.

raiks2:
Please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/I18nUtil.java
```diff
@@ -133,9 +133,9 @@ public static Locale resolvePreferredLanguage(@Nullable Locale preferredLocale,
         if (null != preferredLocale) {
             LanguageRange preferred = new LanguageRange(preferredLocale.toLanguageTag(), LanguageRange.MAX_WEIGHT);
```
yufengwng, line 134: Should this be `MIN_WEIGHT` instead to lower its priority?

raiks2:
Ok, I believe all of the important and not-so-important stuff is addressed. Please approve if we are good to go or let me know if there's anything else.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,98 @@
...
+                        request.url()
+                );
+                return;
+            }
+            LOG.info("The supported languages can be detected for the study {}", studyGuid);
+            Locale ddpAuthPreferredLocale = RouteUtil.getDDPAuth(request).getPreferredLocale();
+            LanguageDto preferredLanguage = TransactionWrapper.withTxn(
+                    handle -> getPreferredLanguage(handle, acceptLanguageHeader, ddpAuthPreferredLocale, studyGuid)
+            );
+            request.attribute(USER_LANGUAGE, preferredLanguage);
```
yufengwng: Should we also set the `Content-Language` header on the `response` object as well? [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)
> the server then selects one of the proposals, uses it and informs the client of its choice with the Content-Language response header

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,98 @@
...
+                ? LanguageRange.parse(acceptLanguageHeader) : Collections.emptyList();
+        Set<LanguageDto> languagesSupportedByStudy = handle.attach(StudyDao.class).findSupportedLanguagesByGuid(studyGuid);
+        Map<Locale, LanguageDto> supportedStudyLocaleToLang = languagesSupportedByStudy.stream()
+                .collect(Collectors.toMap(LanguageDto::toLocale, lang -> lang));
+        Locale preferredLocale = I18nUtil.resolvePreferredLanguage(
+                ddpAuthPreferredLocale, acceptLanguages, supportedStudyLocaleToLang.keySet()
+        );
+        // If we are lucky, the preferred language will be among the list
+        // of supported languages and we'll spare a db call. Otherwise, we
+        // will have to look into the db to get the id of the fallback language
```
yufengwng: One optimization we can do is to cache the fallback lang's id (maybe just a `static long` in this class?) since it isn't going to change. It's nice-to-have, so up to you to add it or not.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,98 @@
...
+        LanguageDto preferredLanguage = Optional.ofNullable(supportedStudyLocaleToLang.get(preferredLocale))
+                .orElseGet(
+                        () -> new LanguageDto(
+                                handle.attach(JdbiLanguageCode.class).getLanguageCodeId(preferredLocale.getLanguage()),
```
yufengwng: Now that we have `LanguageDto`, I'll recommend adding a `findLanguage() -> LanguageDto` method instead of `new`-ing one up.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiStudyLanguage.java
```diff
@@ -0,0 +1,13 @@
...
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+public interface JdbiStudyLanguage {
+    @SqlUpdate(
+            "insert into study_language(umbrella_study_id, language_code_id) values ("
+            + " (select umbrella_study_id from umbrella_study where guid = :studyGuid),"
+            + " (select language_code_id from language_code where iso_language_code = :isoCode))"
+    )
+    int insert(@Bind("studyGuid") String umbrellaStudyGuid, @Bind("isoCode") String isoLanguageCode);
```
yufengwng: Can we put this in `StudyDao` instead of creating yet another JDBI file? Something like `long insertSupportedLanguage()` maybe?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiStudyLanguage.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.ddp.db.dao;
+
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+public interface JdbiStudyLanguage {
+    @SqlUpdate(
```
yufengwng: I'll recommend changing return type to `long` and adding `@GetGeneratedKeys` since the table now have an auto-increment PK.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
yufengwng: How about we start with package private? That should be enough for testing. If it truly needs to be `public` we can change it then.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,120 @@
...
+                    }
+            );
+        } catch (Exception e) {
+            LOG.error("Error while figuring out the user language", e);
+        }
+    }
+
+    static String createContentLanguageHeaderFromLocale(Locale locale) {
+        // E.g. "en-US"
+        return String.format("%s-%s", locale.getLanguage(), locale.getCountry());
```
yufengwng: The languages we have in database are not actual locales (e.g. `en` and not `en-US`), so if the `Locale` is from their user profile, then I think the Locale is just `en`, so won't the content-language header become `en-`?

I think it's simpler and less error-prone to just call [`Locale.toLanguageTag()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#toLanguageTag()).

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -0,0 +1,120 @@
...
+                        Locale preferredLocale = StudyLanguageResolutionFilter.getPreferredLocale(
+                                handle, acceptLanguageHeader, ddpAuthPreferredLocale, studyGuid
+                        );
+                        LanguageDto preferredLanguage = StudyLanguageResolutionFilter.convertLocaleToLanguageDto(
+                                handle, preferredLocale
+                        );
+                        request.attribute(USER_LANGUAGE, preferredLanguage);
+                        LOG.info("Added the preferred user language {} to the attribute store", preferredLanguage.getIsoCode());
+                        String contentLanguageHeader = StudyLanguageResolutionFilter.createContentLanguageHeaderFromLocale(preferredLocale);
+                        response.header(RouteConstants.CONTENT_LANGUAGE, contentLanguageHeader);
```
yufengwng: Will this mean that we'll always add the `content-language` header? What happens if the route itself throws an error or returns an error response? Does it make sense to include `content-langauage` header in that case?

If not, maybe one idea is to have an `after()` filter to handle this? I.e. `StudyContentLanguageFilter` will just need to look if `response.statusCode()` is `2xx` and `request.attribute()` has `user_language`, then we tag on a `content-language`. Or maybe leave it up to the individual routes to add this header?

yufengwng (commented):
Looking good! Just a few more comments

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
raiks2: I chose the last option: checking for the status code == 200 and the presence of `USER_LANGUAGE` attribute.

raiks2:
Please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/I18nUtil.java
raiks2: Changing it this way apparently makes the Locale.lookup to prefer the fallback language over the preferred one from the user profile - breaking tests down. I would leave it as-is unless you are adamant to change it.

raiks2:
Please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/I18nUtil.java
yufengwng: Interesting. It should have the same weight as the fallback, but comes before the fallback in the priority list. Let's leave it for now then.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
yufengwng: Hmm, saw latest change but not sure how it solves issue... This filter runs _before_ the actual route handlers, so this means we have set the `content-language` header even before the route code executes. Did you mean to pull this code out to a separate filter class that's mounted in a `after()`?

raiks2:
Split the `StudyLanguageResolutionFilter` into `StudyLanguageResolutionFilter` and `StudyLanguageContentTypeSettingFilter`. The latter is the `after(...)` filter that sets the `Content-Language` header only for routes that have the `USER_LANGUAGE` attribute and whose HttpStatus is 200.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
raiks2: Yes, you're right, I overlooked the fact that unlike classic servlet filters, Spark `before` filters are executed only **before** the route. Split the filter into two and mounted the new filter in an `after()`.

raiks2:
Please run tests

yufengwng (approved):
LGTM

Do we need to backfill "supported languages" for angio/brain/osteo studies? Is that captured in a different ticket and is that being worked on?

raiks2:
Yes, there is a ticket for backfilling languages supported by studies: https://broadinstitute.atlassian.net/browse/DDP-4394, it's not started. Until then, we will return entities (activities etc.) translated to the fallback language (English).

raiks2:
Please run tests

raiks2:
Please run tests

raiks2:
Please run tests

PR #98, opened 2020-02-20
Ddp 4366 switch routes to using lang from attr store

raiks2:
Switches one more route to the new scheme of things described in https://broadinstitute.atlassian.net/browse/DDP-4366. `GetStudyDetailRoute` is in the study context but its path doesn't match `*/user/*studies/*` and it was not switched over because of that.

The route was not originally covered by tests so those were added along the way to make sure the change doesn't break anything.
  
 ## Risk Assessment
  - [x] These changes are low risk, do not impact security/privacy, do not impact pepper shared
 
 ### FUD Score 
 
 in jira.
 
 ## UI/UX
 
###  Backend
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.
 - [ ] There are no analytics requirements for these changes

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  
## Testing
 - [] I have written automated negative tests
 
## Release
     - [ ] Participant UX
     - [ ] Study staff

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
@@ -278,6 +278,8 @@ public class DataDonationPlatform {
         // - StudyLanguageContentLanguageSettingFilter sets the "Content-Language" header later on
         before(API.BASE + "/user/*/studies/*", new StudyLanguageResolutionFilter());
         after(API.BASE + "/user/*/studies/*", new StudyLanguageContentLanguageSettingFilter());
+        before(API.BASE + "/studies/*", new StudyLanguageResolutionFilter());
+        after(API.BASE + "/studies/*", new StudyLanguageContentLanguageSettingFilter());
 
         enableCORS("*", String.join(",", CORS_HTTP_METHODS), String.join(",", CORS_HTTP_HEADERS));
         setupCatchAllErrorHandling();
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
@@ -3,6 +3,7 @@ package org.broadinstitute.ddp.db.dao;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.Optional;
 
 import org.broadinstitute.ddp.constants.SqlConstants;
 import org.broadinstitute.ddp.db.dto.StudyI18nDto;
@@ -22,6 +23,17 @@ public interface JdbiUmbrellaStudyI18n extends SqlObject {
                 + "AND i18n.language_code_id = lang.language_code_id")
     @RegisterRowMapper(UmbrellaStudyI18nDtoMapper.class)
     List<StudyI18nDto> findTranslationsByStudyId(@Bind("id") long id);
+
+    @SqlQuery("SELECT lc.iso_language_code, i18n.name, i18n.summary "
+            + "FROM i18n_umbrella_study AS i18n JOIN language_code AS lc "
+            + " ON i18n.language_code_id = lc.language_code_id"
+            + " WHERE i18n.umbrella_study_id = :id AND lc.language_code_id = :langCodeId"
+    )
+    @RegisterRowMapper(UmbrellaStudyI18nDtoMapper.class)
+    Optional<StudyI18nDto> findTranslationByStudyIdAndLanguageCodeId(
+            @Bind("id") long id,
+            @Bind("langCodeId") long languageCodeId
+    );
 
     @SqlUpdate("insert into i18n_umbrella_study "
             + "(umbrella_study_id,language_code_id,name,summary) values "
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
@@ -44,6 +44,9 @@ public interface JdbiUmbrellaStudyI18n extends SqlObject {
                 @Bind("name") String name,
                 @Bind("summary") String summary);
 
+    @SqlUpdate("DELETE FROM i18n_umbrella_study WHERE id = :id")
+    int deleteById(@Bind("id") long id);
+
     class UmbrellaStudyI18nDtoMapper implements RowMapper<StudyI18nDto> {
         @Override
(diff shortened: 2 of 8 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
```diff
@@ -23,6 +24,17 @@
     @RegisterRowMapper(UmbrellaStudyI18nDtoMapper.class)
     List<StudyI18nDto> findTranslationsByStudyId(@Bind("id") long id);
 
+    @SqlQuery("SELECT lc.iso_language_code, i18n.name, i18n.summary "
+            + "FROM i18n_umbrella_study AS i18n JOIN language_code AS lc "
+            + " ON i18n.language_code_id = lc.language_code_id"
+            + " WHERE i18n.umbrella_study_id = :id AND lc.language_code_id = :langCodeId"
+    )
```
raiks2: This query is needed to fetch the translation for the language taken from the attribute store.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -278,6 +278,8 @@ public void handle(Request request, Response response) throws Exception {
         // - StudyLanguageContentLanguageSettingFilter sets the "Content-Language" header later on
         before(API.BASE + "/user/*/studies/*", new StudyLanguageResolutionFilter());
         after(API.BASE + "/user/*/studies/*", new StudyLanguageContentLanguageSettingFilter());
+        before(API.BASE + "/studies/*", new StudyLanguageResolutionFilter());
+        after(API.BASE + "/studies/*", new StudyLanguageContentLanguageSettingFilter());
```
raiks2, line 282: We have to mount filters 2 times because `*/studies/*` doesn't work. `*studies/`, on the other hand, seems to be too permissive (and doesn't work anyway).

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyI18n.java
```diff
@@ -23,6 +24,17 @@
...
+    @RegisterRowMapper(UmbrellaStudyI18nDtoMapper.class)
+    Optional<StudyI18nDto> findTranslationByStudyIdAndLanguageCodeId(
+            @Bind("id") long id,
+            @Bind("langCodeId") long languageCodeId
+    );
```
raiks2, line 36: This method is needed to fetch the translation for the language taken from the attribute store. Yes, `RegisterConstructorMapper` is better, but I used `RegisterRowMapper` to be consistent with other methods in this DAO.

yufengwng (commented):
Looks good so far. There's also `GetStudiesRoute`, does that need to be updated as well? I believe these routes are used exclusively on the mobile app, so CC @bskinner just in case.

raiks2:
@yufengwng, for `GetStudiesRoute` we will need a list of all languages supported by **all** studies. To do that, `StudyLanguageResolutionFilter` should be able to fetch this data (not just languages for a single study) and put it to the attribute store (say, inside a `Map`). It's not a big deal, but this will, however, force it to store excessive information on each call because in most cases it's not needed.

Or we can add some logic to the filter making it aware of the route it precedes and making a decision (don't like this idea). 

Or we can consider the `GetStudiesRoute` a special case for now. I'm leaning towards this option. What do you think?

yufengwng:
Ah you make a good point! `GetStudiesRoute` is for all studies under a umbrella. Since it's already doing translation lookup for each study, maybe we can leave it as-is? Or maybe we can update the implementation so that it's closer to the new language lookup mechanism (i.e. looking at "supported languages" instead of `availableTranslations`)?

raiks2:
Yeah, I think let's leave it as-is but, as you suggest, alter the language lookup mechanism by using the "supported languages" concept.

bskinner:
Thanks for the ping! as long as the changes are not changing the format of the response, I think we should be good.

> Looks good so far. There's also `GetStudiesRoute`, does that need to be updated as well? I believe these routes are used exclusively on the mobile app, so CC @bskinner just in case.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetStudiesRouteTest.java
```diff
@@ -82,7 +93,29 @@ private String buildGetStudiesUrl(String umbrellaGuid) {
     }
 
     @Test
-    public void testHappyPath() throws IOException {
+    public void test_givenStudyExists_andIsTranslated_whenRouteIsCalled_thenItReturns200_andNameIsTranslated() {
```
raiks2: Gave a more meaningful name to the test along the way.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudiesRoute.java
```diff
@@ -68,20 +72,16 @@ public Object handle(Request request, Response response) throws Exception {
...
-                });
-
-                Locale userLocale = I18nUtil.resolvePreferredLanguage(authInfo.getPreferredLocale(), acceptLanguages,
-                        availableTranslations.keySet());
-                StudyI18nDto preferredTranslation = availableTranslations.get(userLocale);
+                Set<LanguageDto> supportedLanguages = handle.attach(StudyDao.class).findSupportedLanguagesByGuid(study.getGuid());
+                Set<Locale> supportedLocales = supportedLanguages.stream()
+                        .map(LanguageDto::toLocale)
+                        .collect(Collectors.toSet());
+                Locale userLocale = I18nUtil.resolvePreferredLanguage(authInfo.getPreferredLocale(), acceptLanguages, supportedLocales);
```
yufengwng, line 79: I think there's the possibility that we return a list of studies where one study uses `en` and another study uses another language, since we're doing this resolution for each study individually. Not sure how that will play out, but we can figure that out later.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudiesRoute.java
```diff
@@ -68,20 +72,16 @@ public Object handle(Request request, Response response) throws Exception {
...
+                Long langCodeId = handle.attach(JdbiLanguageCode.class).getLanguageCodeId(userLocale.toLanguageTag());
```
yufengwng, line 80: You might want to cache this id lookup, since we're doing this in a for-loop, and that the studies will very likely all want `en`.

yufengwng (approved):

raiks2:
Please run tests

PR #104, opened 2020-02-26
docs: add api spec for participant-info endpoint

yufengwng:
Came across an API that was not documented in our OpenAPI docs, so added some notes for it.

Rendered:
Screen Shot 2020-02-26 at 1 20 39 PM


yufengwng:
Not going to add this since we might just deprecate this API.

PR #106, opened 2020-02-27
Backfill auth0 tenant password policy

yufengwng:
## Context

Okay, this thing is a little funny. The password requirements, or policy, is actually configured for a _tenant_ in Auth0's dashboard. In our database, the table is called `study_password_requirements` but the reference is to an `auth0_tenant` and not a study. Also, our study-builder is mainly "study centric" so I wasn't sure where to put this piece of "tenant configuration"...

Anyways, did the following:

* Added support in study-builder to insert tenant password policy/requirements
* Added code we can run in production to insert this for CMI tenant via Angio
* Made some tweaks so that `maxIdenticalConsecutiveCharacters` is optional

## Checklist

## FUD Score

- [x] Getting dev into a state where this is user-visible requires some tech fiddling. I will add steps to release notes.

## Testing

## Release

- [x] Releasing these changes requires special handling and I will document the special procedures in the release plan document

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1872,6 +1872,7 @@ StudyPasswordRequirements:
     - isLowercaseLetterRequired
     - isSpecialCharacterRequired
     - isNumberRequired
+    - maxIdenticalConsecutiveCharacters
   properties:
     minLength:
       type: integer
@@ -1885,8 +1886,8 @@ StudyPasswordRequirements:
       type: boolean
     maxIdenticalConsecutiveCharacters:
       type: integer
-      default: 2
-      minimum: 0
+      nullable: true
+      example: 2
 #####
 #
 # Workflows
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1876,6 +1876,7 @@ StudyPasswordRequirements:
   properties:
     minLength:
       type: integer
+      minimum: 1
     isUppercaseLetterRequired:
       type: boolean
     isLowercaseLetterRequired:
@@ -1886,6 +1887,7 @@ StudyPasswordRequirements:
       type: boolean
     maxIdenticalConsecutiveCharacters:
       type: integer
+      minimum: 0
       nullable: true
       example: 2
 #####
--- a/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilder.java
+++ b/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilder.java
@@ -14,6 +14,7 @@ import org.broadinstitute.ddp.db.dao.JdbiClientUmbrellaStudy;
 import org.broadinstitute.ddp.db.dao.JdbiLanguageCode;
 import org.broadinstitute.ddp.db.dao.JdbiOLCPrecision;
 import org.broadinstitute.ddp.db.dao.JdbiSendgridConfiguration;
+import org.broadinstitute.ddp.db.dao.JdbiStudyPasswordRequirements;
 import org.broadinstitute.ddp.db.dao.JdbiUmbrella;
 import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
 import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudyI18n;
@@ -26,6 +27,7 @@ import org.broadinstitute.ddp.db.dao.StudyGovernanceDao;
 import org.broadinstitute.ddp.db.dto.Auth0TenantDto;
 import org.broadinstitute.ddp.db.dto.ClientDto;
 import org.broadinstitute.ddp.db.dto.StudyDto;
+import org.broadinstitute.ddp.db.dto.StudyPasswordRequirementsDto;
 import org.broadinstitute.ddp.db.dto.UmbrellaDto;
 import org.broadinstitute.ddp.db.dto.UserDto;
 import org.broadinstitute.ddp.exception.DDPException;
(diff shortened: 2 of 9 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1885,8 +1886,8 @@ StudyPasswordRequirements:
       type: boolean
     maxIdenticalConsecutiveCharacters:
       type: integer
-      default: 2
-      minimum: 0
+      nullable: true
```
yufengwng, line 1891: @bskinner this property should be nullable.

study-builder/studies/study-example.conf
```diff
@@ -4,7 +4,16 @@
     # The domain url, with a trailing slash at the end.
     "domain": "string",
     "mgmtClientId": "string",
-    "mgmtSecret": "string"
+    "mgmtSecret": "string",
+    # Optional. The password policy configured for the Auth0 database connection.
+    "passwordPolicy": {
```
yufengwng, line 9: @charlottedye I don't think we need this for Prion since it's not on mobile, but just letting you know of this.

study-builder/studies/angio/study.conf
```diff
@@ -2,7 +2,15 @@
     "tenant": {
         "domain": ${auth0.domain},
         "mgmtClientId": ${auth0.mgmtClientId},
-        "mgmtSecret": ${auth0.mgmtSecret}
+        "mgmtSecret": ${auth0.mgmtSecret},
+        "passwordPolicy": {
```
yufengwng, line 6: Not really too sure where to put this. I only have this in Angio, which should be enough. Should I also add to other `study.conf` files? Will that be too redundant?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/TenantPasswordPolicy.java
```diff
@@ -0,0 +1,34 @@
...
+import com.typesafe.config.Config;
+import org.broadinstitute.ddp.db.dao.JdbiAuth0Tenant;
+import org.broadinstitute.ddp.db.dto.Auth0TenantDto;
+import org.broadinstitute.ddp.studybuilder.StudyBuilder;
+import org.jdbi.v3.core.Handle;
+
+/**
+ * A general task to help insert password policy for a tenant.
+ */
+public class TenantPasswordPolicy implements CustomTask {
```
yufengwng, line 14: We'll run this in production to backfill.

study-builder/studies/angio/study.conf
```diff
@@ -2,7 +2,15 @@
...
+            "minLength": 8,
+            "requireUppercase": true,
+            "requireLowercase": true,
+            "requireNumber": true,
+            "requireSpecialChar": false,
+            "maxConsecutive": null
```
yufengwng, line 12: I believe these are the settings in Auth0 dashboard. Please double-check.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: How can it be null?  If there's a default, the value of the field is always defined.

yufengwng: I'm not sure why it says `default: 2` in the first place. I don't see it being defaulted to 2 any where, whereas I do see that it can be `null`.

bskinner: How are we expecting clients to handle a `null` value here, and how would the server be responding? Given the context, even if the server is returning `null`, I don't think it fits the API here- `null` does not make sense for an integer in this case, and feel it should be regarded as a bug in the server's implementation as opposed to an issue with the API docs.

zyme: A bit of reverse engineering here...I think `null` means "use as many identical consecutive chars as you like".

Screen Shot 2020-02-27 at 4 02 48 PM

study-builder/studies/angio/study.conf
zyme: I think we want it in all CMI studies--they'll all be on the mobile app some time :)

yufengwng: Okay the funny thing here is that this is a _tenant configuration_. So once we do this for Angio, all the other CMI studies will have this because they're with the same tenant. I just wasn't sure if I should put this in all the other `study.conf` files too...

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: I think Zim's right -- if it's `null` then there's no max limit on consecutive chars. Maybe @raiks2 can provide more tips here.

bskinner: I withdraw my objection- you're both correct on this. I was hung up on it being a `minimum` value field, not a max. That said, I think we should still have the `minimum` attribute on the type- bounding this to all positive values >= 0.

study-builder/studies/angio/study.conf
zyme: It looks like it's at the per-db-in-tenant level in auth0.  We currently only have a single database connection, but in theory we could have > 1 per tenant.  I wonder if we shouldn't copy-paste this into the schema but instead use the auth0 [management API for the connection](https://community.auth0.com/t/getting-password-rules-for-a-given-connection/15111) and hardcode a mapping of `low`, `medium`, etc. to the structured data that the app needs.

This would probably require the [password complexity route](https://pepper-dev.datadonationplatform.org/docs/api/#operation/getStudyPasswordRequirements) to include something we can map to the connection--probably the client id would suffice.  That would punt the problem of "What if we have a client that can use two or more connections?" down the road, which seems okay.  A use case for one client with multiple db connections (that is, multiple non-oauth/social auths) doesn't leap out at me.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Okay, this one is a bit funny. I think Auth0 only allows it to be `2` so I think it's either `null` or `2`. I don't know why it's an `integer` -- it probably should have just been a `boolean`. Anways, added in `minimum`.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilder.java
```diff
@@ -203,6 +205,40 @@ private Auth0TenantDto getTenantOrInsert(Handle handle) {
         return dto;
     }
 
+    public void insertTenantPasswordPolicy(Handle handle, Auth0TenantDto tenantDto) {
+        if (cfg.hasPath("tenant.passwordPolicy")) {
+            JdbiStudyPasswordRequirements jdbiPasswordPolicy = handle.attach(JdbiStudyPasswordRequirements.class);
+            StudyPasswordRequirementsDto policyDto = jdbiPasswordPolicy.getById(tenantDto.getId()).orElse(null);
+            if (policyDto == null) {
```
raiks2, line 212: The same here, what about an early return with `if (policyDto != null) { ... return }`?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilder.java
```diff
@@ -203,6 +205,40 @@ private Auth0TenantDto getTenantOrInsert(Handle handle) {
         return dto;
     }
 
+    public void insertTenantPasswordPolicy(Handle handle, Auth0TenantDto tenantDto) {
+        if (cfg.hasPath("tenant.passwordPolicy")) {
```
raiks2, line 209: My be just return early to reduce nesting levels? I.e. if (cfg.hasPath("tenant.passwordPolicy" == null)) { ... return }

raiks2 (commented):
LGTM in general, just some nitpicks.

yufengwng:
I think I'm going to abandon this approach in light of recent discussions. Please see latest comments in Slack and ticket DDP-4502.

PR #111, opened 2020-03-02
study-builder: manage angio/brain/mbc emails

yufengwng:
Switched Angio/Brain/MBC studies to use email configurations so we can manage it using study-builder. Most of this is updating the email templates. But most important changes are in the `sendgrid_emails.conf` files.

### DIFF
--- a/study-builder/studies/brain/sendgrid_emails.conf
+++ b/study-builder/studies/brain/sendgrid_emails.conf
@@ -1,0 +1,116 @@
+{
+  "sendgridEmails": [
+    {
+      "key": "bloodNotReceived4Weeks",
+      "name": "BRAIN_BLOOD_SENT_REMINDER",
+      "subject": "Brain Cancer Project Check-In",
+      "filepath": "emails/blood_sent_reminder.html",
+      "render": true
+    },
+    {
+      "key": "bloodReceived",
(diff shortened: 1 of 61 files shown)


study-builder/studies/brain/sendgrid_emails.conf
```diff
@@ -0,0 +1,116 @@
+{
```
ssettipalli, line 1: Quick count of PEPPER_DEV_BRAIN_* templates in sendgrid shows 18 templates where as here we seem to have only 16 templates !!

ssettipalli:
Tried update of MBC template and looked good. 
This is very handy and avoids manual work during prod deploy. 
Thank you

study-builder/studies/brain/sendgrid_emails.conf
yufengwng: A few of those are old or not being used so they're not added to this config.

yufengwng: This config should match up to all the keys we have in `vars.conf` for brain.

ssettipalli: ok. May be we should delete the two unused old templates.

ssettipalli (approved):
looks good to me

yufengwng:
please run tests

PR #120, opened 2020-03-12
DDP-4502 refactor password policy route

yufengwng:
## Context

See old PR #106. This PR refactors the password policy API:

* We now require a `clientId` query param.
* Instead of storing policy in database, we make request out to Auth0's Management API and map policy name to policy requirements.
* New password policy response schema. See api docs.

## Checklist

- [x] I have labeled the type of changes involved using the `C-*` labels.
- [x] I have assessed potential risks and labeled using the `R-*` labels.
- [x] I have considered error handling and alerts, and added `L-*` labels as needed.
- [x] I have considered security and privacy, and added `I-*` labels as needed
- [x] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

_Overall, how are you feeling about these changes?_

- [x] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [x] They are user-visible in dev as a regular user journey and require no additional instructions.
- [ ] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [x] I have written automated positive tests
- [x] I have written automated negative tests
- [x] I have poked around locally to verify proper functionality
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [x] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #121, opened 2020-03-13
Cache auth0 management token in api client

yufengwng:
## Context

While working on PR #120, I noticed that we're not caching the access token for Auth0's Management API. The token is valid for a day or so. `Auth0MgmtTokenHelper` does help cache the token across invocations, however, we're always discarding and `new`-ing up new instances of the token helper, effectively throwing away the cached token.

Following recent patterns, I have added a `Auth0ManagementClient` as an interface in front of Auth0's Management API. I believe the API client should be the one to manage its own token, so this PR merges `Auth0MgmtTokenHelper` into `Auth0ManagementClient` and adds more "global" token caching. I also believe that our interactions with Auth0 should be more than a "util", so my eventual goal is to move us from `Auth0Util` to `Auth0ManagementClient`.

## Checklist

## FUD Score

## Testing

## Release


zyme (commented):
/pullapprove-test

```yml
version: 2

notifications:
  - when: pull_request.opened
    if: "base.ref == 'test' or base.ref == 'staging' or base.ref == 'production'"
    comment: |
      Hey @{{ author }}, looks like we're doing a deploy to @{{base}}.  Cool!
      The reviewers will let you know when it's approved.
  - when: pullapprove.approved
    if: "base.ref == 'test' or base.ref == 'staging' or base.ref == 'production'"
    comment: |
      Thanks @{{ author }}, your @{{base}} deployment has been approved.  Please follow the merge instructions.

groups:
  test-deployment:
    required: 2
    users:
      - zyme
      - yufengwng

    conditions:
    - "base.ref == 'test'"
```

yufengwng:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/client/Auth0ManagementClient.java
```diff
@@ -1,37 +1,152 @@
...
+            return domain;
+        }
+    }
+
+    /**
+     * Get access token for Auth0's Management API, ensuring that it has not expired yet.
+     *
+     * @return token
+     */
+    public String getToken() {
```
ssettipalli, line 99: getToken is just invoking --> getValidToken which is a synchronized method
That meant every call to getToken() is synchronized
More than 90% of the time we probably return cached Token and we don't need to synch/lock the entire call/method.
We only need to synchronize while generating new Token.
something like below should be more efficient in terms of performance. 
public String getToken() {
        if (shouldRequestToken()) {
            getValidToken() //generateNewToken() //synchronized method call
       }
        return TOKEN_CACHE.get(tokenLookupKey).getToken();
}
May be we can consider using ConcurrentHashMap for TOKEN_CACHE for any additional concurrency issues!

yufengwng: I took that straight from `Auth0MgmtTokenHelper`. But good points, let me try your way.

yufengwng:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/client/Auth0ManagementClient.java
yufengwng: done

ssettipalli: nice

ssettipalli (approved):
looks good to me

PR #123, opened 2020-03-13
DDP-4141 add dob validation for age-up

yufengwng:
This PR fixes a bug where the DOB validations are not showing up in age-up scenario. The expectation is that in this age-up scenario, we want to check that the DOB entered by user matches what we got previously from their parent, which is stored in their profile.

HOW TO APPLY FIX:
* Since this is only on `dev`, I will manually delete the old validations and rerun the patch on `dev`.


yufengwng:
please run tests

ssettipalli (approved):
looks good

yufengwng:
please run tests

PR #139, opened 2020-03-16
DDP-3636-take-umbrella-guid-instead-of-name: updated GetStudiesRoute …

raiks2:

Switches `GetStudiesRoute` from using an umbrella name to umbrella GUID. We need to create a separate ticket for front-end changes.
 
 ## Risk Assessment
 
 ### FUD Score 
 
 - [x] These changes are internal and not visible to a regular user
 in jira.
 
 ## UI/UX
 
###  Backend
     #### Route Changes
     - [x] My changes involve a new route changes to an existing route, so the who/what/when are logged using the usual logback mechanism
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
 ## Performance and Stability
  
## Testing
 - [x] The changes doesn't require writing new tests. Existing ones are not failing.
 
## Release
     - [ ] Participant UX
     - [ ] Study staff


pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudiesRoute.java
```diff
@@ -40,14 +40,14 @@
 
     @Override
     public Object handle(Request request, Response response) throws Exception {
-        final String umbrellaIdentifier = request.queryParams(RouteConstants.QueryParam.UMBRELLA);
+        String umbrellaGuid = request.queryParams(RouteConstants.QueryParam.UMBRELLA_GUID);
```
raiks2: We can leave the old query parameter name (`name`) and prevent breaking the front-end. However, `name` containing a GUID sounds wrong. What's more important here - backward compatibility or correctness?

pepper-apis/docs/specification/src/endpoints/studies.yml
```diff
@@ -6,9 +6,9 @@ get:
   description: returns a list of summary information for each of the studies under a given umbrella.
   parameters:
     - in: query
-      name: umbrella
+      umbrellaGuid: umbrella GUID
```
raiks2: We can leave the old query parameter (`name`) and prevent breaking the front-end. However, a name containing a GUID sounds wrong. What's more important here - backward compatibility or correctness?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudiesRoute.java
```diff
@@ -40,14 +40,14 @@
...
 
-        if (StringUtils.isBlank(umbrellaIdentifier)) {
+        if (StringUtils.isBlank(umbrellaGuid)) {
             ApiError error = new ApiError(ErrorCodes.REQUIRED_PARAMETER_MISSING, "Missing umbrella parameter");
             ResponseUtil.haltError(response, HttpStatus.SC_BAD_REQUEST, error);
         }
 
-        LOG.debug("Received request for summary of studies under {}", umbrellaIdentifier);
+        LOG.debug("Received request for summary of studies under {}", umbrellaGuid);
```
yufengwng, line 50: Can we add
```suggestion
        LOG.debug("Received request for summary of studies under umbrella {}", umbrellaGuid);
```

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
```diff
@@ -174,6 +174,7 @@
         public static final String EMAIL  = "email";
         public static final String SUCCESS  = "success";
         public static final String UMBRELLA = "umbrella";
+        public static final String UMBRELLA_GUID = "umbrellaGuid";
```
yufengwng: If we're renaming it, maybe remove the `UMBRELLA` above since that's the old one?

pepper-apis/docs/specification/src/endpoints/studies.yml
yufengwng: Since this whole thing is effectively a breaking change, I'm inclined to say to just go ahead and rename the query parameter as well so we avoid potential confusion down the line.

By the way, the query parameter _name_ is not `name` but `umbrella`. Aka it's `name: <the_name>` and not `<the_name>: description`.

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
bskinner: Why do we need to rename this at all? The parameter is already called `umbrella`- the description can make it clear we're looking for the GUID, although I don't think it's as big a deal since we're using GUIDs as UIDs everywhere else.

pepper-apis/docs/specification/src/endpoints/studies.yml
bskinner: I'd prefer we not rename this- I don't feel that adding `GUID` to the name contributes enough value to justify a breaking change. The query parameter is already named `umbrella` and that should be descriptive enough.

raiks2: Ok, so we have different opinions when it comes to renaming the query string param. On the one hand, renaming it will break the front-end. On the other, calling what is actually a GUID a "name" is quite confusing. Adding to this, the content of `umbrella.name` and `umbrella.guid` in the DB is the same now.

We need a consensus here.

bskinner: But it's not. The name of the query parameter was `umbrella`- nowhere was the parameter referred to as `name`. Regardless, a change in the backend data structure does not necessitate a change to the API.

This change needs to be reverted because it is no longer compliant with the OpenAPI spec.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudy.java
```diff
@@ -52,9 +52,9 @@
             + " from umbrella_study as us"
             + " join umbrella as u on us.umbrella_id = u.umbrella_id"
             + " left join olc_precision op on op.olc_precision_id = us.olc_precision_id"
-            + " where u.umbrella_name = :umbrella")
```
bskinner, line 55: This is the only spot that needed a change- looking up the umbrella using the GUID instead of it's human readable name.

pepper-apis/docs/specification/src/endpoints/studies.yml
yufengwng: Yup, just to reiterate what @bskinner said -- the query param name is `umbrella`. And either way is fine with me. If we're fine with keeping it as-is, then seems like less hassle to me. As the author of PR, what's your thoughts @raiks2?

zyme: My vote would be to keep this as `umbrella` but change the logic and docs to clarify that this is the guid.

raiks2:
Ok, so a compromise: the query string parameter's name is unchanged while everything from the route down to the persistence layer uses the term `umbrellaGuid`. The change is thus invisible to the clients but we benefit from the proper logic and naming.

yufengwng (approved):
👍

raiks2:
Please run tests

PR #141, opened 2020-03-16
Tcell instrumentation

zyme:
When we go to appengine, we'll lose our nginx WAF.  So we'll run [tcell RASP](https://earlyaccess.rapid7.com/tcell/) instead.  Still needs some work in terms of how the configs get rendered/bundled in circle.

As per @adrazhi, no need to redirect `tcell.log` off-container since what we rely on for tcell is API-based pushes to tcell.

We could update develop docs to run tcell locally.  I welcome opinions on that.


pepper-apis/src/appengine/app.yaml
```diff
@@ -3,16 +3,14 @@ runtime: java11
...
 manual_scaling:
   instances: 2
 
 network:
   instance_tag: study-server # necessary for firewall rules so that DSM will accept https requests from pepper
 
-# get size from env var, try this without mx, ms
-entrypoint: java -Dddp.firecloudKeysDir=fc_keys -Dconfig.file=application.conf DataDonationPlatform.jar
+# will need to move all contents of tcell into deploy, and copy tcell_agent.config after rending
+entrypoint: java -javaagent:tcell/tcellagent.jar-Dddp.firecloudKeysDir=fc_keys -Dconfig.file=application.conf DataDonationPlatform.jar
```
yufengwng: Is it just me, or does checking in jars and dynamic libs into source seems a bit odd? Is there no way to manage this, say, through maven? Can we push this into docker image builds?

zyme: It is unusual, but I haven't found a supported mvn repo.   I like the idea to push it into docker image for builds and keep it out of "source".

zyme: However...I'm not sure if this plays nice with GAE.  Maybe we could put the jars into buckets, then grab those as part of deploy?   CC @MocanaAtBroad

zyme: After a bit of fiddling, I'm going to try to put the tcell tarball into a bucket (sans secrets) and then fetch that in circle.  Should be reproducible as long as we version that tarball accordingly and don't update it.

yufengwng: 🎉 woohoo!

.circleci/config.yml
```diff
@@ -17,6 +17,25 @@ executors:
...
+            TAR_FILE_URL=`gsutil ls $TCELL_PATH  | sort -r | head -1`
+
+            if [ -z $TAR_FILE_URL ]
+            then
+              echo "Could not find tcell archive ${TCELL_PATH}"
+              exit 1
+            fi
+            TAR_FILE_PATH="/tmp/$(basename -- $TAR_FILE_URL)"
+            gsutil -u broad-ddp-prod cp  $TAR_FILE_URL $TAR_FILE_PATH
+            cd ~/repo/ddp-workspace
```
MocanaAtBroad: Going to have to run this, but maybe this is meant for ddp-angular? I don't think we have a ddp-workspace directory in `ddp-study-server` repo.

pepper-apis/src/appengine/app.yaml
MocanaAtBroad: What I have done in the past with libraries that are not in Maven is to install them into local repo and then refer to them in Maven build as any other library.
Trick was to setup a module under main pom that when it got built it would grab the library and executed a maven install on it.
The solution here appears to work and this is just a runtime dependency, so maybe we leave it alone?

zyme:
@MocanaAtBroad @yufengwng after some review of various other options (put in docker, host in bucket), I think we should keep this as-is.  It's simple and easy to manage, and source-code-ish.  Okay if I merge up?

.circleci/config.yml
MocanaAtBroad: This path works?

zyme:
@SimoneMaiwald just a heads up on some tcell backend stuff in the context of appengine.

.circleci/config.yml
zyme: Oops, let me  clean this up.

.circleci/config.yml
```diff
@@ -21,9 +21,26 @@ executors:
...
     docker: # todo: create new image endpoint, add this in admin tab for faster starts
     - image: *ci_support_image
-    working_directory: ~/repo/pepper-apis  
+    working_directory: ~/repo/pepper-apis
 
 commands:
+  install-tcell:
+    steps:
+      - run:
+          name: Download tcell files from bucket
```
yufengwng, line 59: Is the plan to simply check tcell files into source? So we don't need this CircleCI command anymore?

zyme: In the plan-of-the-moment, we'd just grab them from a bucket.  In which case we can delete the files here :)

.circleci/config.yml
```diff
@@ -21,9 +21,26 @@ executors:
...
+          command: |
+            vault read -format=json  secret/pepper/prod/v1/conf | jq -r .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
```
yufengwng, line 61: We're reading the `prod` secrets in vault. Does that mean CircleCI have access to our prod secrets? That's a bit concerning...

_Edit: I think this was how things were setup originally. But now that I'm taking a second look at this, I'm a bit concerned, especially because most of our configs/secrets are in that same vault path..._

yufengwng: Is there a way to restrict a vault token to reading certain paths? Maybe we should put the service account in a separate path so we isolate build/deploy secrets from actual application secrets? Also, I understand that this is a bit tangential to tcell, so I'm okay pushing this up and discussing this point separately.

SimoneMaiwald (approved):

PR #144, opened 2020-03-17
DDP-4508 . Populate prequal activity during migration

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4508

StudyDataLoader.createPrequal() is the main change

rest of the changes are cleanup to remove unused passed parameter Handle.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
@@ -257,6 +257,47 @@ public class StudyDataLoader {
         return pepperUser.getUserGuid();
     }
 
+    public ActivityInstanceDto createPrequal(Handle handle, String participantGuid, long studyId, String ddpCreated,
+                                             JdbiActivity jdbiActivity,
+                                             ActivityInstanceDao activityInstanceDao,
+                                             ActivityInstanceStatusDao activityInstanceStatusDao,
+                                             AnswerDao answerDao) throws Exception {
+
+        Long studyActivityId = jdbiActivity.findIdByStudyIdAndCode(studyId, "PREQUAL").get();
+        Instant instant;
+        try {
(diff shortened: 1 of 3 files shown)


ssettipalli:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
```diff
@@ -257,6 +257,47 @@ public String loadParticipantData(Handle handle, JsonElement datstatData, JsonEl
...
+
+        answerTextQuestion("PREQUAL_LAST_NAME", participantGuid, dto.getGuid(),
+                profileDto.getLastName(), answerDao);
+
+        List<SelectedPicklistOption> options = new ArrayList<SelectedPicklistOption>();
+        options.add(new SelectedPicklistOption("DIAGNOSED"));
+        answerPickListQuestion("PREQUAL_SELF_DESCRIBE", participantGuid, dto.getGuid(),
+                options, answerDao);
+
+        //add 30seconds to created_date and populate complete status
```
yufengwng, line 294: Is 30 seconds safe? Should it be some time after user is created but before the About You is created?

yufengwng: We took a closer look at this -- it doesn't seem to have any know side effects so this way should be fine.

ssettipalli: Basically I am using participant created date so that the prequal activity creation date is close to reality rather than current system date. 
To avoid DB constraint violation while inserting second status row for complete status, I had to change the timestamp, hence added 30seconds. 
These dates are not used any where, we should be fine.

yufengwng (approved):
👍

ssettipalli:
please run tests

PR #150, opened 2020-03-23
Ddp 4134 auth0 client id to be scoped to tenant

raiks2:
 Implements https://broadinstitute.atlassian.net/browse/DDP-4134 which is about dropping the assumption that `auth0_client_id` and `auth0_name` are unique and using a combination of `auth0_client_id`, `auth0_tenant_id` instead.
 
 ## Risk Assessment
 
 ### FUD Score 
 
- [x] These changes should not alter the behavior of the system in its current state
 in jira.
 
 ## UI/UX
 
###  Backend
     #### Route Changes
     ### Housekeeping Changes
  ### Client-side

## Analytics
 accordingly.

## Security and Privacy

  ###  Backend
    
  ### Browser
  
  ### Mobile
  - [ ] TBD 
 
 ## Performance and Stability
  
## Testing
 
## Release
     - [ ] Participant UX
     - [ ] Study staff


pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
```diff
@@ -104,6 +105,7 @@ public static void main(String[] args) throws Exception {
         options.addOption("gb", true, "google bucket file");
         options.addOption("p", false, "preprocess");
         options.addOption("pf", true, "preprocess file name");
+        options.addOption("t", true, "Auth0 tenant id");
```
raiks2, line 108: Not sure if a tenant id or a auth0 domain should be used here.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
```diff
@@ -57,6 +57,21 @@ long insertClient(@Bind("clientName")String clientName,
...
+            + "WHERE "
+            + "     c.auth0_client_id = :auth0ClientId and c.auth0_tenant_id = t.auth0_tenant_id"
+            + "     AND t.auth0_domain = :auth0Domain"
+    )
+    @RegisterConstructorMapper(ClientDto.class)
+    Optional<ClientDto> getClientByAuth0ClientAndDomain(
+            @Bind("auth0ClientId") String auth0ClientId,
+            @Bind("auth0Domain") String auth0Domain
+    );
+
```
raiks2: From the https://broadinstitute.atlassian.net/browse/DDP-4134 implementation sketch, I inferred that a combination of `auth0_client_id` and `auth0_domain` is unique. This was reinforced by examining the `auth0_tenant` table definition that defines a UNIQUE KEY for `auth0_domain` column.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
```diff
@@ -42,14 +42,16 @@
     public static final String EMAIL_OPTION = "e";
     public static final String PASSWORD_OPTION = "p";
     public static final String CLIENT_OPTION = "c";
+    public static final String TENANT_OPTION = "t";
```
raiks2: Should we take a `auth0Domain` instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
```diff
@@ -42,14 +42,16 @@
...
 
     public static void main(String[] args) {
         Options options = new Options();
 
         options.addRequiredOption(STUDY_OPTION, "study", true, "study guid (note caps sensitivity");
         options.addRequiredOption(EMAIL_OPTION, "email", true, "email for the account");
         options.addRequiredOption(PASSWORD_OPTION, "password", true, "password for the account");
-        options.addRequiredOption(CLIENT_OPTION, "client", true, "auth client id");
+        options.addRequiredOption(CLIENT_OPTION, "client", true, "auth0 client id");
+        options.addRequiredOption(TENANT_OPTION, "tenant", true, "auth0 tenant id");
```
raiks2, line 54: Should we take a `auth0Domain` instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/security/AesUtil.java
```diff
@@ -32,6 +32,17 @@
...
+            "    SELECT "
+            + "      CONCAT(auth0_tenant_id, '-', auth0_client_id) AS auth0_client_id_tenant_id,"
+            + "      auth0_signing_secret"
+            + "  FROM"
+            + "      client";
+    private static final String UPDATE_SIGNING_SECRET_QUERY =
+            "  UPDATE client "
+            + "SET auth0_signing_secret = :auth0_signing_secret "
+            + "WHERE auth0_client_id = :auth0_client_id "
+            + "AND auth0_tenant_id = :auth0_tenant_id";
```
raiks2: I discovered that these queries are used 2 times, so I put them into constants.

pepper-apis/src/main/java/org/broadinstitute/ddp/security/AesUtil.java
```diff
@@ -160,23 +171,34 @@ public static void main(String[] args) throws Exception {
...
-                        "select auth0_client_id, auth0_signing_secret from client").mapToMap().list();
+                List<Map<String, Object>> clientList = handle.createQuery(SELECT_CLIENT_TENANT_AND_SIGNING_SECRET_QUERY)
+                        .mapToMap()
+                        .list();
                 for (Map<String, Object> client : clientList) {
                     String auth0Secret = (String) client.get("auth0_signing_secret");
-                    String auth0ClientId = (String) client.get("auth0_client_id");
-                    System.out.println("auth0clientId = " + auth0ClientId);
+                    String auth0ClientIdTenantIdStr = (String) client.get("auth0_client_id_tenant_id");
+                    ClientIdTenantId clientIdTenantId = new ClientIdTenantId(auth0ClientIdTenantIdStr);
```
raiks2: Since we need `auth0ClientId` and `auth0TenantId` now, we glue them together, use as a map key and consequently split afterward.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/UserDao.java
```diff
@@ -392,8 +392,12 @@ private User insertNewUser(Handle handle, String auth0ClientId, String auth0User
     /**
      * Returns user permissions for the given operator user guid.
      */
-    public UserPermissions queryUserPermissionsByOperatorGuidAndClientId(Handle handle, String operatorGuid,
-                                                                         String auth0ClientId) {
+    public UserPermissions queryUserPermissionsByOperatorGuidClientIdAndAuth0Domain(
```
zyme: At some point, the `AndFooAndBar` starts to get a bit unwieldy.  This might be that point.  How about just `queryUserPermissions()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/security/JWTConverter.java
```diff
@@ -131,9 +131,12 @@ private DDPAuth convertJWT(String jwt) {
                     try {
                         DecodedJWT validToken = verifyDDPToken(jwt, jwkProvider);
                         String ddpUserGuid = validToken.getClaim(Auth0Constants.DDP_USER_ID_CLAIM).asString();
-                        UserPermissions userPermissions = userDao.queryUserPermissionsByOperatorGuidAndClientId(handle,
+                        UserPermissions userPermissions = userDao.queryUserPermissionsByOperatorGuidClientIdAndAuth0Domain(
+                                handle,
                                 ddpUserGuid,
-                                auth0ClientId);
+                                auth0ClientId,
+                                auth0Domain
```
zyme, line 138: Oh goodie, we already had `auth0Domain` in the token :)

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
zyme: I vote `tenant`

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
zyme: I get a bit confused by `domain`, especially with [auth0 custom domains](https://auth0.com/docs/custom-domains).  Tenant is less confusing to me.

pepper-apis/src/main/resources/db-changes/schema/DDP-4134-change-unique-keys-in-client-table.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="sretsia" id="20200323-change-unique-keys-in-client-table">
+        <dropUniqueConstraint tableName="client" uniqueColumns="client_name" constraintName="client_name"/>
```
zyme: Maybe I'm opening things up too much, but I wonder if we should just remove `client_name` from the database.  It's entirely redundant, since the client already has a name stored in auth0.  Doesn't seem like it's used very much.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
```diff
@@ -10,11 +10,20 @@
...
     private String auth0ClientId;
 
-    public CreateTemporaryUserPayload(String auth0ClientId) {
+    @NotBlank
+    @SerializedName("auth0Domain")
+    private String auth0Domain;
+
+    public CreateTemporaryUserPayload(String auth0ClientId, String auth0Domain) {
         this.auth0ClientId = auth0ClientId;
+        this.auth0Domain = auth0Domain;
```
zyme, line 18: In light of the recent discussion around  breaking API changes, I think we can make things easier on clients by looking up the `auth0Domain` based on `auth0ClientId` and erroring out if there are 0 or > 1 rows returned.  The change is still breaking, but there would be a grace period where clients are unlikely to break until we see duplicate client ids appear.

I believe we already include this value as `auth0Audience` on the angular side, so this should be a small change for angular.  Not sure about mobile app.

CC @bskinner @fat23cat @MocanaAtBroad

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
```diff
@@ -10,11 +10,20 @@
     @SerializedName("auth0ClientId")
     private String auth0ClientId;
 
-    public CreateTemporaryUserPayload(String auth0ClientId) {
+    @NotBlank
+    @SerializedName("auth0Domain")
```
zyme, line 13: Needs API doc changes in this PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRoute.java
```diff
@@ -33,14 +33,21 @@
     public PasswordPolicy handle(Request request, Response response) {
         String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
         String clientId = request.queryParams(RouteConstants.QueryParam.AUTH0_CLIENT_ID);
+        String domain = request.queryParams(RouteConstants.QueryParam.AUTH0_DOMAIN);
```
zyme: Also needs API doc change.  We could make the server more forgiving here as well by looking for a single client with the given `clientid` if `domain` is missing, and if 0 or > 1 are  found, error out.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
```diff
@@ -34,16 +34,17 @@
     @Override
     public Object handle(Request request, Response response) throws Exception {
         String auth0ClientId = request.queryParams(QueryParam.AUTH0_CLIENT_ID);
+        String auth0Domain = request.queryParams(QueryParam.AUTH0_DOMAIN);
```
zyme, line 37: API doc changes here as well please.  Maybe a utility query in `JdbiClient` to do the query-by-client-and-see-if-you-get-just-one thing?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
```diff
@@ -94,7 +95,7 @@ public static void main(String[] args) throws Exception {
         options.addOption("m", false, "Also load mailing address");
         options.addOption("o", true, "Output/Report csv file");
         options.addOption("u", true, "Users to migrate (comma separated list of altpids)");
-        options.addOption("c", true, "Auth0 client name");
+        options.addOption("c", true, "Auth0 client id");
```
zyme, line 98: CC @ssettipalli since he's the data  loader expert ;)

ssettipalli: Debated to user client_id or client name and ended up using client name because Its more human readable and consistent across regions for a study. 
Looks like I need to use tenantId as well in addition to client Id , not a problem.
StudyDataLoader changes look good.  :+1:

raiks2: Thank you, @ssettipalli. The problem here is that if eradicate `clientName` completely as @zyme suggests above (and which makes sense because we are not using it in any meaningful way), we won't be able to use it anymore and `auth0ClientId` will remain our only option.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ClientDao.java
```diff
@@ -24,30 +24,30 @@
...
-     *                           keep it consistent with the auth0 client name
      * @param auth0ClientId      auth0's internal client id
      * @param auth0ClientSecret  auth0's secret for the client
      * @param studyGuidsToAccess list of guids that the client should
      *                           have access to
      * @return client.client_id
      */
-    default long registerClient(String clientName,
-                                String auth0ClientId,
+    default long registerClient(String auth0ClientId,
```
yufengwng, line 33: We need to update `study-builder` as well since it calls one of these DAO methods for inserting client for study.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ClientDao.java
```diff
@@ -24,30 +24,30 @@
...
                                 String encryptionKey,
                                 Long auth0TenantId) {
 
         String encryptedClientSecret = AesUtil.encrypt(auth0ClientSecret, encryptionKey);
 
-        long clientId = getClientDao().insertClient(clientName, auth0ClientId, encryptedClientSecret, auth0TenantId,
+        long clientId = getClientDao().insertClient(auth0ClientId, encryptedClientSecret, auth0TenantId,
                                                     null);
-        LOG.info("Inserted client {} for {}", clientId, clientName);
+        LOG.info("Inserted client {} for {}", clientId);
```
yufengwng:
```suggestion
        LOG.info("Inserted client {}", clientId);
```

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ClientDao.java
```diff
@@ -80,18 +80,17 @@ default Long getClientIdByAuth0ClientAndDomain(String auth0ClientId, String auth
...
         // First remove all client__umbrella_study entries for this client Id
-        Long clientId = getClientDao().getClientIdByName(clientName).orElse(null);
+        Long clientId = getClientDao().getClientIdByAuth0ClientIdAndAuth0TenantId(auth0ClientId, auth0TenantId).orElse(null);
         if (clientId != null) {
             getClientUmbrellaStudyDao().deleteByInternalClientId(clientId);
         }
 
         // now remove the client itself
-        return getClientDao().deleteClientByName(clientName);
+        return getClientDao().deleteClientByAuth0ClientIdAndAuth0TenantId(auth0ClientId, auth0TenantId);
```
yufengwng: Why not `deleteByClientId()` since we have already queried the `clientId` above?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
```diff
@@ -57,6 +55,20 @@ long insertClient(@Bind("clientName")String clientName,
     Optional<Long> getClientIdByAuth0ClientAndDomain(@Bind("auth0ClientId") String auth0ClientId,
                                                      @Bind("auth0Domain") String auth0Domain);
 
+    @SqlQuery("SELECT "
+            + "     c.client_id, c.auth0_client_id, c.auth0_signing_secret, "
+            + "     c.web_password_redirect_url, c.is_revoked, c.auth0_tenant_id "
+            + "FROM "
+            + "     client c, auth0_tenant t "
```
yufengwng, line 63: Can you try using `from ... join` syntax for new code?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
```diff
@@ -72,38 +84,50 @@ long insertClient(@Bind("clientName")String clientName,
...
+    Optional<ClientDto> findByAuth0ClientIdAndAuth0TenantId(
+            @Bind("auth0ClientId") String auth0ClientId,
+            @Bind("auth0TenantId") long auth0TenantId
+    );
 
-    @SqlUpdate("UPDATE client SET web_password_redirect_url = :url WHERE auth0_client_id = :auth0ClientId")
-    int updateWebPasswordRedirectUrlByAuth0ClientId(
+    @SqlUpdate(
+            "UPDATE client SET web_password_redirect_url = :url WHERE auth0_client_id = :auth0ClientId "
+            + "AND EXISTS (SELECT 1 FROM auth0_tenant WHERE auth0_domain = :auth0Domain)"
```
yufengwng: This checks that the auth0 domain exists but I don't see how it guarantees that it's the one we're looking for. Shouldn't we use `update ... join ... set` syntax here? Likewise for the SQL below.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClientUmbrellaStudy.java
```diff
@@ -18,9 +18,12 @@
             + "join client as c on c.client_id = cus.client_id "
             + "join umbrella_study as us on us.umbrella_study_id = cus.umbrella_study_id "
             + "join auth0_tenant t on t.auth0_tenant_id = us.auth0_tenant_id "
-            + "where c.auth0_client_id = :auth0ClientId and not c.is_revoked "
-            + "and c.auth0_tenant_id = t.auth0_tenant_id")
-    List<String> findPermittedStudyGuidsByAuth0ClientId(@Bind("auth0ClientId") String auth0ClientId);
+            + "where c.auth0_client_id = :auth0ClientId and t.auth0_domain = :auth0Domain "
+            + "and not c.is_revoked and c.auth0_tenant_id = t.auth0_tenant_id")
```
yufengwng: Can you move the `c.auth0_tenant_id = t.auth0_tenant_id` up to the `join` clause for `auth0_tenant`?

pepper-apis/src/test/java/org/broadinstitute/ddp/filter/DsmAuthFilterTest.java
```diff
@@ -56,13 +57,17 @@ public void testInvalidToken() {
 
     @Test
     public void testValidTokenWithInvalidClient() {
+        String updateIsRevokedQueryTemplate = "UPDATE client SET is_revoked = %d WHERE auth0_client_id = '%s' "
+                + " AND EXISTS (SELECT 1 FROM auth0_tenant where auth0_domain = '%s')";
```
yufengwng: Again, shouldn't we do a `join` here?

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateTemporaryUserRouteTest.java
```diff
@@ -34,13 +36,15 @@
...
     private static String url;
+    private static String auth0Domain;
 
     @BeforeClass
     public static void setup() {
         TransactionWrapper.useTxn(handle -> testData = TestDataSetupUtil.generateBasicUserTestData(handle));
         url = RouteTestUtil.getTestingBaseUrl() + RouteConstants.API.TEMP_USERS;
         usersToDelete = new HashSet<>();
         gson = new Gson();
+        auth0Domain = ConfigManager.getInstance().getConfig().getConfig(ConfigFile.AUTH0).getString(ConfigFile.DOMAIN);
```
yufengwng: Integration tests have access to the config object already, you can use `RoutTestUtils.getConfig()`.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRouteTest.java
```diff
@@ -53,6 +56,7 @@
     public static void setup() {
         testData = TransactionWrapper.withTxn(TestDataSetupUtil::generateBasicUserTestData);
         auth0ClientId = testData.getTestingClient().getAuth0ClientId();
+        auth0Domain = ConfigManager.getInstance().getConfig().getConfig(ConfigFile.AUTH0).getString(ConfigFile.DOMAIN);
```
yufengwng: Again, `RouteTestUtils` is your friend!

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PostPasswordResetRouteTest.java
```diff
@@ -40,8 +43,10 @@ public static void setup() {
                 handle -> {
                     GeneratedTestData testData = TestDataSetupUtil.generateBasicUserTestData(handle);
                     auth0ClientId = testData.getTestingClient().getAuth0ClientId();
-                    handle.attach(JdbiClient.class).updateWebPasswordRedirectUrlByAuth0ClientId(testRedirectUrl, auth0ClientId);
+                    handle.attach(JdbiClient.class)
+                            .updateWebPasswordRedirectUrlByAuth0ClientIdAndAuth0Domain(testRedirectUrl, auth0ClientId, auth0Domain);
                     token = testData.getTestingUser().getToken();
+                    auth0Domain = ConfigManager.getInstance().getConfig().getConfig(ConfigFile.AUTH0).getString(ConfigFile.DOMAIN);
```
yufengwng: `RouteTestUtils`

pepper-apis/src/test/resources/db-testscripts/baseline-seed-test-data.xml
```diff
@@ -51,18 +51,17 @@
...
         <insert tableName="client">
-            <column name="client_name" value="demo-web-client"/>
             <column name="is_revoked" valueBoolean="false"/>
             <column name="auth0_domain" value="https://ddp-dev.auth0.com/"/>
             <column name="auth0_client_id" value="5ps6MnYXmzpuo4OePvfcW0Dfv6cMPNIC"/>
             <column name="auth0_signing_secret" value="20JhCH3DTn905Vm90Rcy/DGNqQw9EJ/XOGvPG8yyQV7+kkL4riGVGKEuTzZ6xKOQMFL9wKA8ZBC8h1WqKI4xeh9dX0BqBfRrAuubSyHPrOP2qUX4DcO4SqXqyO5CEnr4"/>
         </insert>
         <insert tableName="client__umbrella_study">
-            <column name="client_id" valueComputed="(select client_id from client where client_name = 'demo-web-client')"/>
+            <column name="client_id" valueComputed="(select client_id from client where auth0_client_id = '5ps6MnYXmzpuo4OePvfcW0Dfv6cMPNIC' and auth0_domain = 'https://ddp-dev.auth0.com/')"/>
```
yufengwng, line 60: Hmmm, are we sure we want to put this id in here?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
yufengwng: Can we use the auth0 domain/tenant **string**? I don't think it's developer-friendly to have to give the **numeric** id and that id might change across databases and environments.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
yufengwng: Again, might be better to take the tenant/domain string instead of numeric id...

pepper-apis/src/main/java/org/broadinstitute/ddp/security/AesUtil.java
yufengwng: Instead of concat-then-split by hand, you can have JDBI do it, just like how we do it for most other DAOs stuff. I don't remember the method names off-the-top of my head, but you can use their fluent API something like `handle.createQuery(..).registerConstructorMapper(..).mapTo(ClientIdTenantId.class).list()`. Note, I think you have to make this class `public` so JDBI can access it via reflection.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
yufengwng: This sounds like a nice idea for backwards-compatibility!

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRoute.java
yufengwng: In light of backwards-compatibility, perhaps we can try NOT require the `domain` query param and use the one associated with the study itself, since they should both be in the same domain?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
yufengwng: Question: who is passing in this `domain` query param? It looks like this might be Auth0 calling back to the server? Maybe you need to update auth0 email configs? (look in `study-builder/tenants/cmi/emails/change_password.html`).

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
bskinner: That's an excellent idea- I'm behind it!

yufengwng: FYI I'm tracking API breaking changes here: https://broadinstitute.atlassian.net/browse/DDP-4471.

yufengwng:
@raiks2 since there are some (potential) API breaking changes here, are there changes we can make on the Angular SDK side? And if so, should we do that concurrently along with this ticket?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ClientDao.java
raiks2: Yeah, but this method didn't exist. Added it as well as made the method logic more straightforward.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateTemporaryUserRouteTest.java
raiks2: Done.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRouteTest.java
raiks2: Done!

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PostPasswordResetRouteTest.java
raiks2: Done.

pepper-apis/src/test/resources/db-testscripts/baseline-seed-test-data.xml
raiks2: Not sure what you mean actually... Since we are not using `client_name` anymore, the only reliable way to find a client is the `auth0_client_id` + `auth0_tenant_id` combination.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
raiks2: Done.

pepper-apis/src/test/java/org/broadinstitute/ddp/filter/DsmAuthFilterTest.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClientUmbrellaStudy.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ClientDao.java
raiks2: Done.

pepper-apis/src/main/resources/db-changes/schema/DDP-4134-change-unique-keys-in-client-table.xml
raiks2: Done, removed `client.client_name` and changed the code using it.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/UserDao.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/security/AesUtil.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRoute.java
raiks2: I implemented the approach suggested by @zyme: if a `domain` is missing, we check if the `clientId` is unique. If it is, we proceed or else fail. If the `domain` is present, we use it. From the client's point of view, there is no difference until we start getting `clientId` clashes.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
raiks2: I implemented the approach suggested by @zyme: if a domain is missing, we check if the clientId is unique. If it is, we proceed or else fail. If the domain is present, we use it. From the client's point of view, there is no difference until we start getting clientId clashes.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
raiks2: I implemented the approach suggested by @zyme: if a domain is missing, we check if the clientId is unique. If it is, we proceed or else fail. If the domain is present, we use it. From the client's point of view, there is no difference until we start getting clientId clashes.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
raiks2: Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
raiks2: @yufengwng, not quite sure what the tenant **string** is. Could you please give an example?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
raiks2: @yufengwng, not quite sure what the tenant string is. Could you please give an example?

raiks2:
Please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/UserAdminCLI.java
yufengwng: Yup, something like `https://ddp-dev.auth0.com/`.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
```diff
@@ -36,11 +39,36 @@ protected int getValidationErrorStatus() {
...
+                // Left for backward compatibility. It's expected that for some time
+                // there will be no clashes between clients with the same Auth0 client id
+                // When they start to occur, change the check accordingly
+                LOG.info("Domain query parameter is missing, checking if the auth0 client id '{}' is unique", auth0ClientId);
+                int numClients = handle.attach(JdbiClient.class).countClientsWithSameAuth0ClientId(auth0ClientId);
+                if (numClients > 1) {
+                    String msg = String.format(
+                            "Auth0 client id '%s' is not unique, please provide a auth0Domain value for disambiguation", auth0ClientId
+                    );
+                    LOG.warn(msg);
```
yufengwng: I think we should make this a `log.error()` so we get alerted about this.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
```diff
@@ -36,11 +39,36 @@ protected int getValidationErrorStatus() {
...
+            ClientDto clientDto = null;
+            boolean isDomainSpecified = auth0Domain != null && !auth0Domain.isBlank();
+            if (isDomainSpecified) {
+                clientDto = handle.attach(JdbiClient.class).getClientByAuth0ClientAndDomain(auth0ClientId, auth0Domain).orElse(null);
+            } else {
+                // Left for backward compatibility. It's expected that for some time
+                // there will be no clashes between clients with the same Auth0 client id
+                // When they start to occur, change the check accordingly
+                LOG.info("Domain query parameter is missing, checking if the auth0 client id '{}' is unique", auth0ClientId);
+                int numClients = handle.attach(JdbiClient.class).countClientsWithSameAuth0ClientId(auth0ClientId);
```
yufengwng: Might be nicer to just do a query for all clients with this auth0_client_id (i.e. return a `List<ClientDto>`) so we avoid doing an extra query just for the count.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRoute.java
yufengwng: It will work, but I don't think that approach is necessary for this API. That's because both the study and client should ideally belong to the same tenant/domain (if they're not, what does that mean?). So we can just use the domain from the study. We can even grab the `auth0TenantId` from the `StudyDto` object. If we can avoid changing the API (even if it's non-breaking), I think that will be better way to go so we don't have to support something that's no needed. Welcome other suggestions here.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoaderMain.java
ssettipalli: 👍

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetStudyPasswordPolicyRoute.java
raiks2: Well, I couldn't figure out what you mean by the domain "associated with the study itself" (the `umbrella_study` table has nothing resembling a domain). However, I used `auth0TenantId` from `StudyDto` as you suggest so there's no breaking change in this route anymore.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
raiks2: Sounds good to me! Done.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
raiks2: Yeah, makes sense. Done.

raiks2:
Ok, apparently the most important changes are done. We need to decide whether to use `tenantId` or `domain` in command-line parameters of `UserAdminCLI` and `StudyLoaderMain` (there are 2 opinions by @zyme and @yufengwng) and then we should be good to go.

raiks2:
Please run tests

yufengwng:
We don't really use the other CLIs, so the main concern will be the DataLoader ones. If Sampath is okay with just using the id, then either way is fine with me. 

I am fine with StudyDataLoader using Id. :+1:

ssettipalli:
please run tests

ssettipalli:
please run tests

pepper-apis/docs/specification/src/endpoints/temporary-users.yml
```diff
@@ -9,5 +9,8 @@ post:
   responses:
     201:
       $ref: '../pepper.yml#/components/responses/CreateTempUserResponse'
+    400:
```
bskinner, line 12: Anything in the `responses` map must be a `Response` object, or a `$ref`- not both. An object reference may not also have a `description`.

Relevant docs:
* [Responses object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#responsesObject)
* [Response object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#responseObject)
* [Reference (`$ref`) object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#referenceObject)

ssettipalli:
please run tests

ssettipalli:
@bskinner @yufengwng Can you please take one final look at this PR.
Thanks.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
```diff
@@ -43,7 +44,27 @@ public Object handle(Request request, Response response) throws Exception {
...
+            // Left for backward compatibility. It's expected that for some time
+            // there will be no clashes between clients with the same Auth0 client id
+            // When they start to occur, change the check accordingly
+            LOG.info("Domain query parameter is missing, checking if the auth0 client id '{}' is unique", auth0ClientId);
+            int numClients = TransactionWrapper.withTxn(
+                    handle -> handle.attach(JdbiClient.class).countClientsWithSameAuth0ClientId(auth0ClientId)
+            );
+            if (numClients > 1) {
+                String msg = "Auth0 client id '{}' is not unique, please provide a auth0Domain value for disambiguation";
+                LOG.error(msg);
```
yufengwng, line 61: Need to pass argument to the `log`. Also, looks like this `msg` string is what we pass back to client. I'll recommend cleaning up the string a little. e.g. try
```
Auth0 client id '{}' is not unique, please also provide domain query parameter`
```
Likewise for `CreateTemporaryUserRoute.java` (note this one takes body payload and not query param).

pepper-apis/docs/specification/src/endpoints/temporary-users.yml
yufengwng: Can you expand this a little bit to better reflect what the API is returning? Maybe something like the following:
```
    400:
      description: if domain is not provided and client id is not unique
      content:
        application/json:
          schema:
            $ref: '../pepper.yml#/components/schemas/Error.BadPayload'
```

pepper-apis/docs/specification/src/endpoints/studies.password-policy.yml
```diff
@@ -28,7 +28,7 @@ get:
           schema:
             $ref: '../pepper.yml#/components/schemas/StudyPasswordPolicy'
     400:
-      description: if client id is missing
+      description: if client id is missing or if domain is not provided and client id is not unique
```
yufengwng: I don't think we need this change... this route does not need `domain` since we're using the same domain as the study. I recommend reverting this so clients won't get confused.

pepper-apis/docs/specification/src/endpoints/temporary-users.yml
yufengwng: _Edit: oops! I missed it, I see it's there!_

And I just realized that we didn't update the payload... In `requestBodies/CreateTempUserRequestBody'`, we'll need to add a new `properties` object:
```
auth0Domain:
  type: string
  description: ...
```

yufengwng (commented):
Found a few more things... almost there!

ssettipalli:
please run tests

ssettipalli:
please run tests

yufengwng (approved):
I think this is ready to go

ssettipalli:
please run tests

PR #152, opened 2020-03-24
DDP-4512 . thank-you email fix for MBC about-you survey

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4512

Generate a thank you email once next activity, consent is created rather than about-you complete.
consent is created only when supported country is selected.  

requirement:
thank-you email should be generated when about-you is submitted with supported country.
thank-you email should NOT be generated when about-you is submitted with NOT supported country.
This is same as brain study.


ssettipalli:
please run tests

yufengwng (approved):
👍

PR #157, opened 2020-04-02
MBC user enrollment fix

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4523
Migrated user should be updated as "enrolled" if blood consent and blood survey is completed.
Handle followup instance only creation

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
@@ -700,6 +700,12 @@ public class StudyDataLoader {
             return;
         }
 
+        String status = getStringValueFromElement(surveyData, "survey_status");
+        if (status.equalsIgnoreCase("CREATED")) {
+            LOG.warn("Created followup survey instance but no data ");
+            return;
+        }
+
         processSurveyData(handle, "followupsurvey", surveyData, mappingData,
                 studyDto, userDto, instanceDto, answerDao);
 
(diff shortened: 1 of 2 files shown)


ssettipalli:
please run tests

ssettipalli:
please run tests

pepper-apis/src/main/java/org/broadinstitute/ddp/util/StudyDataLoader.java
```diff
@@ -570,6 +570,8 @@ public void loadBloodReleaseSurveyData(Handle handle,
         //MBC might have dup physicians because physician list comes in both release and bdrelease surveys
         processPhysicianList(handle, surveyData, userDto, studyDto,
                 "physician_list", InstitutionType.PHYSICIAN, "bdreleasesurvey", instanceDto);
+
+        updateUserStudyEnrollment(handle, surveyData, userDto.getUserGuid(), studyDto.getGuid());
```
yufengwng, line 574: Do we need an equivalent event configuration in `study.conf`? What happens to migrated users who _just_ finished blood-consent, _and then_ finishes blood-release in new Pepper? Do they also get `enrolled`?

ssettipalli: thanks for reminding.  good find. Thought about it but slipped my mind during implementation.
added it. thank you.

ssettipalli: yes they need to be enrolled too

ssettipalli:
please run tests

study-builder/studies/mbc/study.conf
```diff
@@ -1636,6 +1636,19 @@
...
+        "activityCode": "BLOODRELEASE",
+        "statusType": "COMPLETE"
+      },
+      "action": {
+        "type": "USER_ENROLLED"
+      },
+      "maxOccurrencesPerUser": 1,
+      "dispatchToHousekeeping": false,
+      "order": 1
+    },
```
yufengwng, line 1651: I'm still not 100% sure how this will affect the overall study, since MBC's workflows are a bit more complicated... but I guess we'll find out from testing.

yufengwng (approved):

study-builder/studies/mbc/study.conf
ssettipalli: tested the scenario in local. looked good.

PR #159, opened 2020-04-07
data exporter cli: add flag to do all elastic indices

yufengwng:
Instead of running 4 different commands separately, we should just run a single command. Also, running separate commands is not optimal since each command will have to warm up cache and query data separately.


ssettipalli (approved):
looks good to me. . 
This should help cut downtime during releases if we need to reload all elastic indices of a study

yufengwng:
please run tests

PR #163, opened 2020-04-09
[rc_v1.8] data exporter cli: add flag to do all elastic indices

yufengwng:
Cherry-picking this to the RC branch. Again, this is just change to a CLI tool we use during deployment. Will not affect production features.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/util/DataExporterCli.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/util/DataExporterCli.java
@@ -32,6 +32,7 @@ import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
 import org.broadinstitute.ddp.db.dto.StudyDto;
 import org.broadinstitute.ddp.export.ActivityExtract;
 import org.broadinstitute.ddp.export.DataExporter;
+import org.broadinstitute.ddp.model.study.Participant;
 
 /**
  * Utility to run data export via command-line. Note that a config file is required, and the main thing we look for in the configuration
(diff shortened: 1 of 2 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/util/DataExporterCli.java
```diff
@@ -149,13 +155,56 @@ private void initExporter() {
...
-            // Warm up the activity cache
-            exporter.extractActivities(handle, studyDto);
+            System.out.println("[export] starting activity definitions export...");
+            start = System.currentTimeMillis();
+            List<ActivityExtract> activities = exporter.exportActivityDefinitionsToElasticsearch(handle, studyDto, cfg);
+            elapsed = System.currentTimeMillis() - start;
+            System.out.println(String.format("[export] took %d ms (%.2f s)", elapsed, elapsed / 1000.0));
+
+            System.out.println("[export] starting structured json export...");
+            start = System.currentTimeMillis();
```
ssettipalli, line 183: Looks like we can extract a private method  and avoid some code duplication. Same statements are invoked for 3 indices and probably same from other method too. 
Not lot of statements though.

yufengwng: I think it's fine -- it's just timing statements. I'll like to keep this the same as what's merged into `develop`. We can refactor later if we'll like.

ssettipalli: ok. If you want to revisit it later, thats fine.

ssettipalli (approved):

PR #63, opened 2020-01-24
Study server to appengine

MocanaAtBroad:
## Context and the big picture
Paying for docker is expensive.  App engine is cheap.  Jenkins is hard.  Circle is easy.  So we're moving study-server from jenkins to circle, and using app engine instead of docker-compose.

Along the way, we're changing connection pooling from DBCP to Hikari.  It's faster, more robust, and the GAE examples use it.  We had experimented with changing this in the long ago future, so it's not an unknown.  There were a bunch of changes needed in our tests to accommodate some unusual testing scenarios around password rotation, and some fiddling to make sure we got all the pool settings correct.

Let's bring in the circle config learnings from the frontend appengine PR.

@MocanaAtBroad I accidentally overwrote your initial notes here.  Totally accidental.  Sorry 😢 
 
 ## What type of change is this?
  
  - [ ] Bugfix (non-breaking change which fixes an issue)
  - [X] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Experimental proof-of-concept
  - [ ] Infosec/Compliance remediation
  
 ## Risk Assessment
  - [x] These changes are **HIGH RISK**.
    - [X] There is an elevated risk that may impact a large number of features (Just about everything depends on database connections!)
    - [ ] There is an elevated risk to security and privacy
    - [ ] These changes introduce new 3rd party dependencies, alter the network perimeter, or contain
    major upgrades to persistent storage or SaaS dependencies such as auth0, sendgrid, or easypost    
  - [ ] These changes are low risk, do not impact security/privacy, do not impact pepper shared
  components/modules, and have no major upgrades to 3rd party services or libraries
 
 ### FUD Score 
 Overall, how are you feeling about these changes?
  - [X] :relaxed: All good, business as usual!
  - [ ] :sweat_smile: There might be some issues here
  - [ ] :scream:I'm sweaty and nervous
 
 ## How to we demo these changes?
 All behinds the scenes. No demoing!
 - [ ] They are user-visible in dev as a regular user journey and require no additional instructions.
 - [ ] Getting dev into a state where this is user-visible requires some tech fiddling.  I have
 documented these steps in the related ticket.
 - [ ] Requires other features before it's human visible.  I have documented the blocking issues
 in jira.
 - [ ] I have no idea how to demo this.  Please help me!
 
 ## UI/UX
  - [ x] There ain't no UI/UX in these changes
[...]

PR #168, opened 2020-04-14
Back-end build/test/deploy with parallel tests

MocanaAtBroad:
## Context
**Updates from April 21 2020**
At this point full functionality should be in place, including new database dev URLs in Vault. 
If you want to see parallel tests in action, will need to checkout this branch and execute the new script:
```
./run_ci_tests.sh
```
The OpenAPI docs deployment to GAE functionality is in separate PR branched off this PR:
https://github.com/broadinstitute/ddp-study-server/pull/172

**Previous comments**

This PR contains enhancement over PR link below and at this point it supersedes it.
https://github.com/broadinstitute/ddp-study-server/pull/63

Changes are mainly meant to speed up running tests by running them in parallel.
From comments in original PR looks like we need to be able to deploy open api docs. That has not happened yet.

Also need to adjust URLs in Vault to deal with new consolidated databases.

## Checklist

- [ X ] I have labeled the type of changes involved using the `C-*` labels.
- [  X ] I have considered security and privacy, and added `I-*` labels as needed

## FUD Score

- [X ] :sweat_smile: There might be some issues here

Need to change code and push it to different branches.

## Testing

- [X ] I have written zero automated tests but have poked around locally to verify proper functionality

## Release

- [ X ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document


.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
+version: 2.1
+
+references:
+  ci_support_image: &ci_support_image
+                       broadinstitute/study-server-build:2020-04-10G
```
MocanaAtBroad, line 5: The image that has everything in it.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+      # start pubsub emulator
+      - image: broadinstitute/study-server-build:pubsub-1
+      - image: circleci/mysql:5.7-ram
+        environment: &DB_VARS
+          MYSQL_ROOT_PASSWORD: rootpw
+          MYSQL_DATABASE: test_db
+          MYSQL_USER: user
+          MYSQL_PASSWORD: passw0rd
+    resource_class: medium+ #default medium not big enough for tests
+
```
MocanaAtBroad, line 50: Test executor starts a MySQL and a pubsub emulator. Emulator does not have to be started up from test.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+          name: Run maven
+          command: |
+            # Some images the $HOME directory is not writable. Let's be specific for .m2/repository location
+            mvn compile package -DskipTests -file pom.xml -Dmaven.repo.local=<< parameters.m2_repo >>
+            mvn package -DskipTests -f housekeeping-pom.xml -Dcheckstyle.skip -Dmaven.repo.local=<< parameters.m2_repo >>
+      - save_cache:
+          key: *m2_cache_key
+          paths:
+            - << parameters.root_path >>/.m2
+
```
MocanaAtBroad: save and restore cache allows us to save local copies of Java dependencies without downloading the Internet during each build.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+  setup-shared-env:
+    description: "Setup shared ENV vars used by multiple scripts"
+    steps:
+      - run:
+          name: Set environment to be used by build
+          command: |
+            echo "export VAULT_TOKEN=$(vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_ROLE_SECRET_ID)" >> $BASH_ENV
+            echo "export SHORT_GIT_SHA=$(echo '<< pipeline.git.revision >>' | cut -c 1-7)" >> $BASH_ENV
+            source $BASH_ENV
```
MocanaAtBroad, line 78: Where we generate the vault token. Uses CircleCi configuration. VAULT_ROLE_ID is injected by CircleCi

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            ;;
+            *)
+              echo "Cannot map << pipeline.git.branch >> to a deployment environment"
+              exit 17
+            ;;
+            esac
+            echo "Setting deployment ENVIRONMENT to $DEPLOY_ENV"
+            echo "export ENVIRONMENT=$DEPLOY_ENV" >> $BASH_ENV
+            source $BASH_ENV          
+
```
MocanaAtBroad, line 114: Figure out deployment environment based on branch. Store it in environment variable.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+    steps:
+      - run:
+          name: show environment variables
+          command: env
+      - run:
+          name: api-build.sh
+          command: ./api-build.sh v1 ${ENVIRONMENT:-dev} . --gae-config
+          environment:
+            USE_DOCKER: false
+
```
MocanaAtBroad, line 126: And here we use that ENVIRONMENT variable.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+              sleep 1
+            done
+            echo Failed waiting for MySQL && exit 1
+      - run:
+          name: Setup test databases
+          command: |
+            mysql -h 127.0.0.1 -u root --password=$MYSQL_ROOT_PASSWORD test_db --execute="CREATE DATABASE << parameters.studyserverdb >> CHARACTER SET utf8 COLLATE utf8_general_ci"
+            mysql -h 127.0.0.1 -u root --password=$MYSQL_ROOT_PASSWORD test_db --execute="CREATE DATABASE << parameters.housekeepingdb >>b CHARACTER SET utf8 COLLATE utf8_general_ci"
+          environment: *DB_VARS
+
```
MocanaAtBroad, line 146: When running the test executor, we prep the database to run the tests.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+          command: |
+            echo "pipeline.id << pipeline.id >>"
+            echo "pipeline.number  << pipeline.number  >>"
+            echo "pipeline.project.git_url << pipeline.project.git_url >>"
+            echo "pipeline.project.type << pipeline.project.type >>"
+            echo "pipeline.git.tag << pipeline.git.tag >>"
+            echo "pipeline.git.branch << pipeline.git.branch >>"
+            echo "pipeline.git.revision << pipeline.git.revision >>"
+            echo "pipeline.git.base_revision << pipeline.git.base_revision >>"
+
```
MocanaAtBroad, line 161: Some debug.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            DATE=`date +%F`
+            BUILDS_BUCKET='<<parameters.builds_bucket>>'
+            JAR_BASE_NAME='<<parameters.jar_base_name>>'
+            JAR_BUILD_BUCKET_DIR='<<parameters.jar_build_bucket_dir>>'
+            JAR_FILE_NAME="${JAR_BASE_NAME}_${DATE}_${SHORT_GIT_SHA}.jar"
+            LOCAL_JAR_FILE_PATH="./target/${JAR_BASE_NAME}.jar"
+            JAR_BUCKET_URL="gs://${BUILDS_BUCKET}${JAR_BUILD_BUCKET_DIR}/${JAR_FILE_NAME}"
+            gsutil cp "${LOCAL_JAR_FILE_PATH}" "${JAR_BUCKET_URL}"
+            echo "${JAR_FILE_NAME} successfully uploaded to ${JAR_BUCKET_URL}"
+
```
MocanaAtBroad, line 189: Store the given jar into our build repo.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            JAR_FILE_URL=`gsutil ls $JAR_FILE_URL_PATTERN  | sort -r | head -1`
+            if [ -z $JAR_FILE_URL ]
+              then
+                echo "Could not find archive using pattern ${JAR_FILE_URL_PATTERN}"
+                exit 1
+            fi
+            JAR_FILE_PATH="target/${JAR_BASE_NAME}.jar"
+            mkdir -p $(dirname $JAR_FILE_PATH)
+            gsutil cp  $JAR_FILE_URL $JAR_FILE_PATH
+            echo ""
```
MocanaAtBroad, line 224: Get the jar out of repo for deployment.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            cp ../target/<<parameters.jar_base_name>>.jar .
+            cp ../src/appengine/<<parameters.jar_base_name>>.yaml .
+            cp ../output-config/application.conf .
+            cp -r ../output-config/fc_keys .
+      - auth-gcp-service-account
+      - run:
+          name: Deploy <<parameters.jar_base_name>> to GAE
+          command: |
+            set -u
+            gcloud app deploy --version="${SHORT_GIT_SHA}" --stop-previous-version --project "broad-ddp-${ENVIRONMENT}" <<parameters.deploy_dir>>/<<parameters.jar_base_name>>.yaml
```
MocanaAtBroad, line 250: With jar in the target directory (there because we just built or because we pulled it out of our repo), we deployment to GAE.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            # Sometime deleting versions fails. If there are less than 3 versions then we are not going to worry about it
+            # and set flag so this job step does not fail when exit error value is generated
+            if [[ ! -z $VERSIONS_TO_DELETE ]] && [[ "${#VERSIONS_TO_DELETE[@]}" -lt 3 ]]; then
+              set +e
+            fi
+
+            for CURRENT_VERSION in $VERSIONS_TO_DELETE; do
+              echo "Deleting version $CURRENT_VERSION of service $SERVICE"
+              gcloud app versions delete --quiet --service "$SERVICE" "$CURRENT_VERSION" --project "broad-ddp-${ENVIRONMENT}"
+            done;
```
MocanaAtBroad, line 276: Everytime we deploy, we create a new version of the deployment. There is a limit to how many you can have. And if we use automatic scaling, older versions keep running. We choose to remove older versions each time we deploy.
Notice special logic here to allow for failure removing versions. Sometimes the `versions delete` command fails for unknown reason.
We are going to let it go as long as we don't have more than 3 versions accumulated.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+        type: string
+    steps:
+      - run:
+          name: Update deployments sheet
+          command: |
+            set -u
+            set +x
+            vault read -format=json secret/pepper/${ENVIRONMENT}/v1/conf  | jq -r .data.gcp.serviceKey |
+            /opt/build-utils/reportdeploy.sh "$RELEASE_SHEET_ID" '<<parameters.jar_base_name>> server' "$ENVIRONMENT" "$CIRCLE_BUILD_NUM" "$CIRCLE_BUILD_URL"
+
```
MocanaAtBroad, line 291: We append info about to our deployment Google sheet.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+              echo "Target environment was not set"
+              exit 1
+            else
+              echo "Target vault env: $TARGET_ENV"
+              echo "From $PWD"
+            fi
+            vault read --format=json secret/pepper/${TARGET_ENV}/v1/conf |
+              jq -r .data.gcp.serviceKey |
+              gcloud auth activate-service-account --key-file=-
+
```
MocanaAtBroad, line 313: When we want to call a gcloud command, we need to do an auth first. We keep credentials in vault.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+          path: *repo_path
+      - setup-shared-env
+      - compile
+      # will be needed for tests later on
+      - render-configs
+      - persist_to_workspace:
+          root: *root_path
+          paths:
+            - repo
+
```
MocanaAtBroad: The compile job. Notice `persist_to_workspace`. This saves files so they can be retrieved by another job in same workflow. Here we choose to store everything: source, generated classes and jar, etc.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+            mvn --batch-mode -Dcheckstyle.skip test -f pom.xml \
+                    -Ditext.license=output-config/itextkey.xml \
+                    -Dddp.firecloudKeysDir=output-config/fc_keys \
+                    -Dconfig.file=output-config/testing-circleci.conf  \
+                    -Dmaven.repo.local=<< parameters.m2_repo >> \
+                    -Dsurefire.excludesFile=$IGNORE_CLASS_LIST_PATH   # Key line to do tests in parallel
+
+      - store_test_results:
+          path: target/surefire-reports
+
```
MocanaAtBroad, line 435: A little refactoring of shared maven execution.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+    working_directory: *pepper_apis_path
+    steps:
+      - attach_workspace:
+          at: *root_path
+      - setup-shared-env
+      - store-jar:
+          jar_base_name: DataDonationPlatform
+      - store-jar:
+          jar_base_name: Housekeeping
+
```
MocanaAtBroad, line 535: Self explanatory. Notice that we reattach the workspace we saved in the compile job.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+    working_directory: *pepper_apis_path
+    steps:
+      - attach_workspace:
+          at: *root_path
+      - setup-shared-env
+      - deploy-jar:
+          jar_base_name: DataDonationPlatform
+      - deploy-jar:
+          jar_base_name: Housekeeping
+
```
MocanaAtBroad, line 552: This one is used in the branch develop workflow. We are not interested in storing jar builds in repo, so we deploy directly from build contents. Once again, we attach the saved compilation workspace.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+      - setup-shared-env
+      - retrieve-jar:
+          jar_base_name: DataDonationPlatform
+      - deploy-jar:
+          jar_base_name: DataDonationPlatform
+      - retrieve-jar:
+          jar_base_name: Housekeeping
+      - deploy-jar:
+          jar_base_name: Housekeeping    
+
```
MocanaAtBroad, line 569: This is the deployment from jars we have stored before.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+      - store-jar-job:
+          filters:
+            branches:
+              only:
+                - *rc
+                - *hotfix
+                - study-server-parallel-tests
+          requires:
+            - parallel-tests-job
+
```
MocanaAtBroad, line 618: Workflow defined by the `requires:` property. Each job waits for the job that it `requires`.

.circleci/config.yml
```diff
@@ -0,0 +1,515 @@
...
+          filters:
+            branches:
+              only:
+                - test
+                - staging
+                - production
+                # @todo remove when done
+                - test_deleteme1
+
+
```
MocanaAtBroad, line 631: This is the the other workflow, just waiting for the branches to be pushed to deploy.

pepper-apis/Dockerfile-build
```diff
@@ -0,0 +1,52 @@
+# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
+FROM broadinstitute/dsde-toolbox:latest AS dsdetoolbox
+
+FROM maven:3.6.3-jdk-11 AS maven
+
+FROM google/cloud-sdk:alpine
+
```
MocanaAtBroad, line 7: Tried to create as compact an image as possible as waiting for these images to spin up in CircleCi can be a big time waster.
The two big changes from the previous image were:
1) Using an Alpine base image
2) Leaving out the gcloud beta tools
Was able to bring the image down from over 800MB to below 300MB and on average much faster start up times in CircleCi (from, on average 90 seconds to on average 20 something seconds).
CIrcleCi image start times do tend to fluctuate a lot though, but this has made a big difference.

pepper-apis/api-build.sh
```diff
@@ -52,15 +52,18 @@ function push_docs_image {
...
+  INPUT_DIR=config DOCS_PROXIED_HOST=$DOCS_PROXIED_HOST NGINX_PROXIED_HOST=$NGINX_PROXIED_HOST OUTPUT_DIR=$BUILD_OUTDIR ENVIRONMENT=$ENVIRONMENT VERSION=$VERSION DIR=$DIR MANIFEST=build-manifest.rb ruby configure.rb -y
+
+  if [[ "$1" == "local" ]]; then
+      echo "rendering runtime configs locally"
+      INPUT_DIR=config NO_SYSLOG=true DOCS_PROXIED_HOST=$DOCS_PROXIED_HOST NGINX_PROXIED_HOST=$NGINX_PROXIED_HOST OUTPUT_DIR=$OUTDIR ENVIRONMENT=$ENVIRONMENT DIR=$DIR/$OUTDIR VERSION=$VERSION ruby configure.rb -y
+      cp $OUTDIR/docker-compose.yaml $DIR
+  elif [[ "$1" == "gae" ]]; then
+      echo "gae configs"
+      INPUT_DIR=config NO_SYSLOG=true DOCS_PROXIED_HOST=$DOCS_PROXIED_HOST NGINX_PROXIED_HOST=$NGINX_PROXIED_HOST OUTPUT_DIR=$OUTDIR ENVIRONMENT=$ENVIRONMENT DIR=$DIR/$OUTDIR VERSION=$VERSION GAE=true ruby configure.rb -y
+  fi
```
MocanaAtBroad, line 67: Needed our own specialized GAE config settings.

pepper-apis/build-utils/gcloud-pubsub-emulator/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY --from=builder /usr/bin/wait-for /usr/bin
+COPY --from=builder /go/bin/pubsubc   /usr/bin
+COPY                run.sh            /run.sh
+
+ENV PORT 8442
+
+EXPOSE $PORT
+
+CMD /run.sh $PORT
```
MocanaAtBroad, line 29: Leaving out the gcloud beta tools (which include the emulator) saved a lot of time in start up by keeping base build image thin. But we still need a pubsub emulator to run tests.
Turns out starting pubsub emulator in separate image is much faster.

pepper-apis/config/application.conf.ctmpl
```diff
@@ -2,18 +2,24 @@
...
 {
-    "dbUrl": "{{$conf.Data.dbUrl}}",
+    {{if eq $gae "true"}}
+        "dbUrl": "{{$conf.Data.gaeDbUrl}}",
+        "housekeepingDbUrl": "{{$conf.Data.gaeHousekeepingDbUrl}}",
+    {{else}}
+        "dbUrl": "{{$conf.Data.dbUrl}}",
+        "housekeepingDbUrl": "{{$conf.Data.housekeepingDbUrl}}",
+    {{end}}
+    "defaultTimezone": "{{$conf.Data.defaultTimezone}}",
```
MocanaAtBroad, line 21: This is the main difference in our generated GAE-specific config file: the database URLs are different. Do not use sqlproxy. Permissions need to be set properly.

pepper-apis/config/manifest.rb
```diff
@@ -10,6 +10,7 @@
 copy_secret_from_path "secret/pepper/#{$env}/pepper-cert.crt"
 render "ddp.conf.ctmpl"
 render "application.conf.ctmpl"
+render "testing-circleci.conf.ctmpl"
 copy_file "post_deploy_smoketest.sh"
```
MocanaAtBroad, line 13: We have an alternative to inmemory-test.conf to run our tests in CircleCi.

pepper-apis/config/testing-circleci.conf.ctmpl
```diff
@@ -0,0 +1,183 @@
...
+{{with $auth0 := index $testingConf.Data "auth0"}}
+{{with $testingAuth := index $testingConf.Data "auth0"}}
+{{with $elasticsearchConf := vault (printf "secret/pepper/%s/%s/elasticsearch" $environment $version)}}
+
+{
+    # If set to true, tests will use MySQL from Docker container and dbUrl and housekeepingDbUrl will be ignored
+    "useDisposableTestDbs": false,
+    "dbUrl": "jdbc:mysql://localhost:3306/studyservicedb?user=root&password=rootpw&characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC&useSSL=false&sessionVariables=innodb_strict_mode=on,tx_isolation='READ-COMMITTED',sql_mode='TRADITIONAL'",
+    "housekeepingDbUrl": "jdbc:mysql://localhost:3306/housekeepingdb?user=root&password=rootpw&characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC&useSSL=false&sessionVariables=innodb_strict_mode=on,tx_isolation='READ-COMMITTED',sql_mode='TRADITIONAL'",
+    "maxConnections": 50,
```
MocanaAtBroad: Maybe we should look at consolidating configuration? Here are the main differences to run tests in CircleCi.: we start mysql server with a RAM disk (TestContainer was difficult to run in CircleCi) and needed to tweak the URLs to use this MySQL instance.

pepper-apis/configure.rb
```diff
@@ -1,15 +1,25 @@
...
   puts "VERSION not set"
   exit 1
 }
 
-$env = ENV.fetch("ENV") { |_|
-  puts "ENV not set"
+$env = ENV.fetch("ENVIRONMENT") { |_|
+  puts "ENVIRONMENT var not set"
   exit 1
 }
```
MocanaAtBroad, line 19: The name ENV was problematic in either consult-template (I don't recall what the problem was at this point)

pepper-apis/configure.rb
```diff
@@ -171,24 +181,31 @@ def render_from_path(path, output_file_name = nil)
...
+                  $dsde_toolbox_image_name,
+                  "consul-template", "-config=/etc/consul-template/config/config.json",
+                  "-template=#{file_name}:#{output_file_name}",
+                  "-once"
+    ]
+  else
+    vault_cmd = ["consul-template", "-config=/etc/consul-template/config/config.json",
+                              "-template=#{file_name}:#{output_file_name}", "-once"]
+  end
+
```
MocanaAtBroad, line 207: Allow for possibility of having consul-template and vault installed locally. This is what happens in CircleCi docker image (calling out to from from Docker in CircleCi was complicated).

pepper-apis/parent-pom.xml
```diff
@@ -547,12 +558,12 @@
...
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.22.0</version>
+                <version>2.22.1</version>
                 <dependencies>
                     <dependency>
                         <groupId>org.apache.maven.surefire</groupId>
                         <artifactId>surefire-junit47</artifactId>
-                        <version>2.22.0</version>
+                        <version>2.22.1</version>
                     </dependency>
```
MocanaAtBroad, line 582: Ran into a bug with surefire that was fixed in point release.

pepper-apis/pom.xml
```diff
@@ -19,16 +19,17 @@
...
 			  <groupId>org.apache.maven.plugins</groupId>
 			  <artifactId>maven-jar-plugin</artifactId>
+			  <version>3.2.0</version>
 			  <configuration>
 				<archive>
 				  <manifest>
-					<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                      <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                      <mainClass>org.broadinstitute.ddp.DataDonationPlatform</mainClass>
 				  </manifest>
```
MocanaAtBroad: Simplifies and speeds up start up in GAE.

pepper-apis/src/appengine/DataDonationPlatform.yaml
```diff
@@ -0,0 +1,25 @@
...
+network:
+  instance_tag: study-server # necessary for firewall rules so that DSM will accept https requests from pepper
+
+entrypoint: java -Xmx820m -Xmx820m -jar DataDonationPlatform.jar
+
+# GAE does not like the "=" character in entrypoint and does not support "." in environment variables
+env_variables:
+  config_file: application.conf
+  ddp_firecloudKeysDir: fc_keys
+
```
MocanaAtBroad, line 24: Tried F1 instances with half the memory. Performance was obviously pathetic. I expect that this arrangement (matchine type, scaling policy, and configuration) will have to be adjusted with actual usage.

pepper-apis/src/appengine/Housekeeping.yaml
```diff
@@ -0,0 +1,14 @@
...
+manual_scaling:
+  instances: 1
+
+entrypoint: java -Xmx820m -Xmx820m -jar Housekeeping.jar
+
+# GAE does not like the "=" character in entrypoint and does not support "." in environment variables
+env_variables:
+  config_file: application.conf
+  ddp_firecloudKeysDir: fc_keys
+
```
MocanaAtBroad, line 13: B4 allows for manual scaling and we can just make sure it is running all the time.

pepper-apis/src/appengine/DataDonationPlatform.yaml
MocanaAtBroad: Note that with this setting no instance is started until someone tries to access it. And it goes dormant down to 0 instances after a period of no usage.

pepper-apis/src/appengine/Housekeeping.yaml
MocanaAtBroad, line 13: Note that we specify runtime config params here. GAE generates warning if you expand the command line to include params with the "-", e.g. `-DconfigFile=`

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -229,8 +234,14 @@ private static void start() {
...
-        int port = cfg.getInt(ConfigFile.PORT);
-        port(port);
+        // app engine's port env var wins
+        int configFilePort = cfg.getInt(ConfigFile.PORT);
+        String appEnginePort = System.getenv(APP_ENGINE_PORT);
+        if (appEnginePort != null) {
+            port(Integer.parseInt(appEnginePort));
+        } else {
+            port(configFilePort);
+        }
```
MocanaAtBroad, line 225: Changes above allow for config parameters that we would have set via properties can be picked up via environment variables. GAE dictates the PORT number to use.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -33,6 +30,14 @@
...
  */
 public class TransactionWrapper {
+    // We are setting the VMs default timezone expectation that it will be same on server
+    // or at least specified in DB connection URL
+    // if timezones don't match, times and dates might be miscalculated/improperly converted by JDBC connector
+    // https://bugs.mysql.com/bug.php?id=91112
+    // Note 03/10/2020: Release of fixed Java connector is imminent and this might no longer be necessary
+    static {
+        TimeZone.setDefault(TimeZone.getTimeZone(ConfigManager.getInstance().getConfig().getString(ConfigFile.DEFAULT_TIMEZONE)));
+    }
```
MocanaAtBroad, line 41: As comment indicates, this is likely to be a temporary fix. When new driver class is available these lines should be able to be safely removed.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -337,10 +361,31 @@ private static Handle openJdbiWithAuthRetry(DB db) {
...
+        config.setConnectionTimeout(1000);
+        config.setMaxLifetime((28800 * 1000) - (3 * 60 * 1000));  // as per hikari docs, a few minutes less than mysql wait_timeout
+        config.setPoolName(db.name());
+
+        // todo arz leverage allowPoolSuspension and mxbeans to fully  automate password rotation
+
+        return new HikariDataSource(config);
+    }
+
+
```
MocanaAtBroad, line 388: A lot of changes here were merged from Zim's PR to fix issues found when running tests.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiActivityInstance.java
```diff
@@ -45,7 +45,7 @@ default String generateUniqueGuid() {
 
     @SqlQuery("select activity_instance_id from activity_instance where "
             + "activity_instance_guid = :activityInstanceGuid")
-    long getActivityInstanceId(String activityInstanceGuid);
+    long getActivityInstanceId(@Bind("activityInstanceGuid") String activityInstanceGuid);
```
MocanaAtBroad, line 49: Trying to get past problems in my IDE. Once in a while my IntelliJ compiler simply will not apply it properly. Look forward to day when we can get rid of -parameter flag for real. A small contribution in that direction.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiBooleanQuestion.java
```diff
@@ -14,7 +14,8 @@
 
     @SqlUpdate("insert into boolean_question (question_id,true_template_id,false_template_id)"
             + " values(:questionId,:trueTemplateId,:falseTemplateId)")
-    int insert(long questionId, long trueTemplateId, long falseTemplateId);
+    int insert(@Bind("questionId")long questionId, @Bind("trueTemplateId")long trueTemplateId,
+               @Bind("falseTemplateId") long falseTemplateId);
```
MocanaAtBroad, line 19: @bind

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiDateFieldType.java
```diff
@@ -2,10 +2,11 @@
...
 import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 
 public interface JdbiDateFieldType extends SqlObject {
 
     @SqlQuery("select date_field_type_id from date_field_type where date_field_type_code = :dateFieldCode")
-    long getTypeId(DateFieldType dateFieldCode);
+    long getTypeId(@Bind("dateFieldCode") DateFieldType dateFieldCode);
 }
```
MocanaAtBroad, line 12: @bind

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiDateQuestionFieldOrder.java
```diff
@@ -13,9 +14,9 @@
...
 
     @SqlQuery("queryOrderedFieldsByQuestionId")
-    List<DateFieldType> getOrderedFieldsByQuestionId(long questionId);
+    List<DateFieldType> getOrderedFieldsByQuestionId(@Bind("questionId") long questionId);
 
     @SqlUpdate("insert")
     @GetGeneratedKeys
-    long insert(long questionId, long dateFieldTypeId, int displayOrder);
+    long insert(@Bind("questionId") long questionId, @Bind("dateFieldTypeId") long dateFieldTypeId, @Bind("displayOrder") int displayOrder);
 }
```
MocanaAtBroad, line 22: @bind

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiDateRenderMode.java
```diff
@@ -2,10 +2,11 @@
...
 import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 
 public interface JdbiDateRenderMode extends SqlObject {
 
     @SqlQuery("select date_render_mode_id from date_render_mode where date_render_mode_code = :renderMode")
-    long getModeId(DateRenderMode renderMode);
+    long getModeId(@Bind("renderMode")DateRenderMode renderMode);
 }
```
MocanaAtBroad, line 12: @bind

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiExpression.java
```diff
@@ -20,19 +20,19 @@ default String generateUniqueGuid() {
...
+    long insert(@Bind("expressionGuid") String expressionGuid, @Bind("expressionText") String expressionText);
 
     default Expression insertExpression(String text) {
         String guid = generateUniqueGuid();
         return new Expression(insert(guid, text), guid, text);
     }
 
     @SqlUpdate("update expression set expression_text = :text where expression_guid = :guid")
-    int updateByGuid(String guid, String text);
+    int updateByGuid(@Bind("guid") String guid, @Bind("text") String text);
```
MocanaAtBroad, line 36: @Bind

pepper-apis/src/main/java/org/broadinstitute/ddp/exception/InvalidConfigurationException.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Exception thrown when there is an unrecoverable issue with
+ * configuration file values.
+ */
+public class InvalidConfigurationException extends DDPException {
+
+    public InvalidConfigurationException(String message) {
+        super(message);
+    }
```
MocanaAtBroad, line 11: Part of Hikari Pool changes.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
```diff
@@ -28,7 +28,16 @@
...
+    public static final File TYPESAFE_CONFIG_FILE;
+
+    static {
+        // For benefit of GAE. Does not like command line options with "=" characters and env variables with "."
+        String configFileName = System.getenv(TYPESAFE_CONFIG_SYSTEM_VAR.replace('.', '_'));
+        if (configFileName == null) {
+            configFileName = System.getProperty(TYPESAFE_CONFIG_SYSTEM_VAR);
+        }
+        TYPESAFE_CONFIG_FILE = new File(configFileName);
+    }
```
MocanaAtBroad, line 40: Getting config file from environment variable as discussed higher up.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/LiquibaseUtil.java
```diff
@@ -31,36 +30,44 @@
...
+        HikariConfig config = new HikariConfig();
+        config.setJdbcUrl(dbUrl);
+        dataSource = new HikariDataSource(config);
+    }
+
+    @Override
+    public void close()  {
+        if (dataSource != null) {
+            dataSource.close();
+        }
```
MocanaAtBroad, line 55: Reading config Liquibase config files from jar instead of file system.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/LiquibaseUtil.java
```diff
@@ -104,18 +106,18 @@ private void runPepperAPIsGlobalMigrations() throws LiquibaseException {
...
      * Runs the specific changelog file. When migration fails, this will attempt to rollback the changes, as applicable.
      */
-    private void runMigrations(String changelogFile) throws LiquibaseException {
+    private void runMigrations(String changelogFile) throws LiquibaseException, SQLException {
         Liquibase liquibase = null;
         String tag = null;
         try {
-            liquibase = new Liquibase(changelogFile, new FileSystemResourceAccessor(), new JdbcConnection(conn));
+            liquibase = new Liquibase(changelogFile, new ClassLoaderResourceAccessor(), new JdbcConnection(dataSource.getConnection()));
             logLocks(liquibase.listLocks());
```
MocanaAtBroad, line 121: Drop FileSystemResourceAccessor and use ClassLoaderResourceAccessor: the key to read config files from jar file.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PubSubEmulator.java
```diff
@@ -149,6 +153,27 @@ public static void shutdown() throws IOException {
...
+                CloseableHttpResponse response =
+                        httpClient.execute(new HttpGet("http://" + EMULATOR_HOST));
+        ) {
+            return response.getStatusLine().getStatusCode() == 200;
+        } catch (IOException e) {
+            String msg = "There was problem initializing CloseableHttpClient";
+            LOG.error(msg, e);
+            throw new DDPException(msg, e);
+        }
+
```
MocanaAtBroad: If emulator is already running, we don't start a new one. We setup our own in CircleCi test environment.

pepper-apis/src/test/java/org/broadinstitute/ddp/db/PooledPasswordRotationTest.java
```diff
@@ -101,23 +97,26 @@ public void testPasswordRotation() throws Exception {
...
+
+        // re-apply overrides
+        ConfigManager configManager = ConfigManager.getInstance();
+        for (Map.Entry<String, String> override : originalConfigOverrides.entrySet()) {
+            configManager.overrideValue(override.getKey(), override.getValue());
+        }
+        TransactionWrapper.reloadDbPoolConfiguration(true);
+        LOG.info("Reset test database config file to its original state.");
     }
 }
```
MocanaAtBroad, line 122: Zim's changes to get pooled password rotation working on Hikari.

pepper-apis/src/test/java/org/broadinstitute/ddp/db/TransactionWrapperTest.java
```diff
@@ -373,7 +384,7 @@ public void testJdbi_useTxn_outOfConnections() {
                 });
             });
         } catch (DDPException e) {
-            assertTrue(e.getCause().getMessage().matches(".*Pool exhausted"));
+            assertTrue(e.getCause().getMessage().matches(".*Connection is not available.*"));
         }
     }
 }
```
MocanaAtBroad, line 390: Some tweaks to work with Hikari. Mostly different error messages.

pepper-apis/src/test/java/org/broadinstitute/ddp/model/event/NotificationEventActionTest.java
```diff
@@ -120,6 +120,11 @@ public void test_invitationEmail_usesLatestNonVoidedInvitation() {
                     testData.getStudyId(), testData.getUserId(), "test111@datadonationplatform.org");
             InvitationDto invitation2 = invitationFactory.createInvitation(InvitationType.AGE_UP,
                     testData.getStudyId(), testData.getUserId(), "test222@datadonationplatform.org");
+            try {
+                Thread.sleep(2);
+            } catch (InterruptedException e) {
+                fail("Got interrupted while sleeping");
+            }
             InvitationDto invitation3 = invitationFactory.createInvitation(InvitationType.AGE_UP,
```
MocanaAtBroad, line 128: This is a workaround to actual bug in code, otherwise test suite fails regularly:
https://broadinstitute.atlassian.net/browse/DDP-4527

pepper-apis/src/test/java/org/broadinstitute/ddp/util/LiquibaseUtilTest.java
```diff
@@ -85,10 +82,10 @@ public void testSuccessfulRollbackWhenChangeSetFails() {
...
     public void testFailedRollbackWhenChangeSetFails() {
         String script = DB_SCRIPTS_DIR + "/liquibase-migrate-success.xml";
-        LiquibaseUtil.runChangeLog(driver, testDbUrl, script);
+        LiquibaseUtil.runChangeLog(testDbUrl, script);
         try {
             script = DB_SCRIPTS_DIR + "/liquibase-migrate-rollback-fail.xml";
-            LiquibaseUtil.runChangeLog(driver, testDbUrl, script);
+            LiquibaseUtil.runChangeLog(testDbUrl, script);
             fail("Expected migrations to fail and trigger rollback");
         } catch (DDPException e) {
```
MocanaAtBroad, line 90: Liquibase changes: from files to contents of jar.

.circleci/config.yml
```diff
@@ -0,0 +1,552 @@
...
+      - retrieve-jar:
+          jar_base_name: Housekeeping
+      - deploy-jar:
+          jar_base_name: Housekeeping    
+
+parameters:
+  on_demand:
+    type: boolean
+    default: false
+
```
MocanaAtBroad, line 574: This parameter is used to distinguish between a build initiated by a git push and a build initiate by a remote API call. We created a script to do the on demand builds. In this PR somewhere.

pepper-apis/scripts/runcitests.sh
```diff
@@ -0,0 +1,20 @@
...
+
+BRANCH=${1-$DEFAULT_BRANCH}
+
+echo "Will try to run build/test on branch: $BRANCH"
+curl -u "${CI_TOKEN}:" -X POST --header "Content-Type: application/json" -d "{
+                                  \"branch\": \"$BRANCH\",
+                                  \"parameters\": {
+                                      \"on_demand\": true
+                                  }
+}" "https://circleci.com/api/v2/project/${CI_PROJECT_SLUG}/pipeline"
```
MocanaAtBroad: Script to initiate a build/test from command line. Pass in the branch you want to build or `develop` by default.
This might be the alternative to "Please run tests" as I have not found a way to drive builds from PR comments in CircleCi.

pepper-apis/parent-pom.xml
```diff
@@ -366,7 +377,7 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>mysql</artifactId>
-            <version>1.10.6</version>
+            <version>1.12.3</version>
             <scope>test</scope>
         </dependency>
```
MocanaAtBroad, line 398: In trying to TestContainers to work in CircleCi, I upgraded the library. As it turns out we don't use them at all in CircleCi, but tests run with newer library and newer feature *MIGHT* allow to run tests faster, so leave upgrade in.

pepper-apis/parent-pom.xml
```diff
@@ -491,7 +502,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.8.4</version>
+                <version>0.8.5</version>
                 <executions>
```
MocanaAtBroad, line 521: Supports newer JDKs which I experimented with (tried another thing that need work). Does not hurt anything, so might as well use newer one.

pepper-apis/src/test/java/org/broadinstitute/ddp/util/TestDataSetupUtil.java
```diff
@@ -503,12 +501,12 @@ public static synchronized void insertStaticTestData() {
         scripts.add(BASELINE_SEED_TEST_DATA);
         scripts.add(MIGRATE_LEGACY_STUDY_CLIENT_TENANT_AND_ENABLE_CONSTRAINTS);
         try {
-            Driver driver = new com.mysql.jdbc.Driver();
+            Driver driver = new com.mysql.cj.jdbc.Driver();
```
yufengwng: Driver is not used?

pepper-apis/src/test/java/org/broadinstitute/ddp/db/TransactionWrapperTest.java
```diff
@@ -77,22 +78,28 @@ public void testPoolExhaustionBehavior() {
...
-                    gotPoolExhaustedError.set(e.getCause().getMessage().toLowerCase().contains("pool"));
+                    ResultSet rs = queryTestData(handle.getConnection());
+                    int numRows = 0;
+                    while (rs.next()) {
+                        numRows++;
+                    }
+                    assertEquals(1, numRows);
+                    try {
+                        TransactionWrapper.useTxn((handle2) -> {
+                            LOG.info("Got connection when we shouldn't have");
```
yufengwng: Should this be a `fail()` instead of just a log?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PubSubEmulator.java
```diff
@@ -149,6 +153,27 @@ public static void shutdown() throws IOException {
...
+        } else {
+            return pubsubProcess != null;
+        }
+    }
+
+    private static boolean isExternalEmulatorRunning() {
+        try (
+                CloseableHttpClient httpClient = HttpClients.createDefault();
+                CloseableHttpResponse response =
+                        httpClient.execute(new HttpGet("http://" + EMULATOR_HOST));
```
yufengwng: Can also be `PUBSUB_BASE_URL`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -337,10 +361,31 @@ private static Handle openJdbiWithAuthRetry(DB db) {
...
     }
 
+    private HikariDataSource createDataSource(int maxConnections, String dbUrl, DB db) {
+        HikariConfig config = new HikariConfig();
+        config.setJdbcUrl(dbUrl);
+        config.addDataSourceProperty("cachePrepStmts", "true");
+        config.addDataSourceProperty("prepStmtCacheSize", "250");
+        config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
+        config.addDataSourceProperty("cacheServerConfiguration", "true");
+        config.setInitializationFailTimeout(0); // don't fail to  create the pool if the db is not available
```
yufengwng, line 375: May I ask why this is desirable?

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -214,7 +215,11 @@ private static void start() {
         boolean doLiquibase = cfg.getBoolean(ConfigFile.DO_LIQUIBASE);
         int maxConnections = cfg.getInt(ConfigFile.NUM_POOLED_CONNECTIONS);
 
-        String firecloudKeysLocation = System.getProperty(ConfigFile.FIRECLOUD_KEYS_DIR_ENV_VAR);
+        // For benefit of GAE. Does not like command line options with "=" characters and env variables with "."
+        String firecloudKeysLocation = System.getenv(ConfigFile.FIRECLOUD_KEYS_DIR_ENV_VAR.replace('.', '_'));
```
yufengwng: One of these days we should remove the firecloud stuff...

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -229,8 +234,14 @@ private static void start() {
         int requestThreadTimeout = cfg.getInt(ConfigFile.THREAD_TIMEOUT);
         String healthcheckPassword = cfg.getString(ConfigFile.HEALTHCHECK_PASSWORD);
 
-        int port = cfg.getInt(ConfigFile.PORT);
-        port(port);
+        // app engine's port env var wins
+        int configFilePort = cfg.getInt(ConfigFile.PORT);
+        String appEnginePort = System.getenv(APP_ENGINE_PORT);
```
yufengwng: Can we just call it `ENV_PORT`? No need to be specific to appengine.

pepper-apis/scripts/runcitests.sh
yufengwng: Can you rename the script to `run_ci_tests.sh`?

pepper-apis/pom.xml
yufengwng: Oh wow, I didn't realize the indentation is so off here... can you help fix the indent?

pepper-apis/parent-pom.xml
```diff
@@ -22,6 +22,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
```
yufengwng, line 47: Let's remove the `commons-dbcp2` dependency below?

pepper-apis/config/testing-circleci.conf.ctmpl
yufengwng: Is `dbUrl` and `housekeepingDbUrl` the only differences? Maybe we should add an env variable that is read in `MySqlTestContainerUtil`. Something like `DDP_NO_TEST_CONTAINERS` or `DDP_FORCE_LOCAL_DB`?

pepper-apis/config/application.conf.ctmpl
yufengwng: At which point can we just use one set of urls? And how do we transition to that?

pepper-apis/build-utils/gcloud-pubsub-emulator/make.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+docker build -t broadinstitute/study-server-build:pubsub-1 .
+docker push broadinstitute/study-server-build:pubsub-1
```
yufengwng, line 4: Is this name set-in-stone? Maybe it's better to namespace our stuff with `ddp-`?

pepper-apis/Dockerfile-java-build
```diff
@@ -0,0 +1,54 @@
+# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
```
yufengwng, line 1: So this `Dockerfile-java-build` is for `broadinstitute/study-server-build:java-2020-04-15A`, and the other `Dockerfile-build` is for `broadinstitute/study-server-build:2020-04-10G`?

.circleci/config.yml
```diff
@@ -0,0 +1,596 @@
...
+              /home/circleci/repo
+  m2_cache_key: &m2_cache_key
+                  pom.xml1-{{ checksum "pom.xml" }}
+  m2_cache_restore_key: &m2_cache_keys
+    keys:
+      - *m2_cache_key
+      - pom.xml1-
+  m2_repo_path: &m2_repo_path
+             /home/circleci/.m2/repository
+  pepper_apis-path: &pepper_apis_path
```
yufengwng: `pepper_apis_path`?

.circleci/config.yml
```diff
@@ -0,0 +1,596 @@
...
+              DEPLOY_ENV=test
+            ;;
+            test_deleteme1)
+              DEPLOY_ENV=dev
+            ;;
+            staging)
+              DEPLOY_ENV=staging
+            ;;
+            staging)
+              DEPLOY_ENV=prod
```
yufengwng: Hmmm 🤔

.circleci/config.yml
```diff
@@ -0,0 +1,596 @@
...
+            # and set flag so this job step does not fail when exit error value is generated
+            if [[ ! -z $VERSIONS_TO_DELETE ]] && [[ "${#VERSIONS_TO_DELETE[@]}" -lt 3 ]]; then
+              set +e
+            fi
+
+            for CURRENT_VERSION in $VERSIONS_TO_DELETE; do
+              echo "Deleting version $CURRENT_VERSION of service $SERVICE"
+              gcloud app versions delete --quiet --service "$SERVICE" "$CURRENT_VERSION" --project "broad-ddp-${ENVIRONMENT}"
+            done;
+            :
```
yufengwng: Looks like an empty line with a colon? Or are we using the bash `:` builtin?

.circleci/config.yml
```diff
@@ -0,0 +1,596 @@
...
+              circleci tests split --split-by=timings --timings-type=classname > $THIS_NODE_TESTS_PATH
+
+
+            IGNORE_CLASS_LIST_PATH="${TEST_DATA_PATH}/surefire_classnames_ignore_list"
+
+            # Filter out the ones that don't run in this node
+            cat $SUREFIRE_CLASS_NAMES_PATH | \
+              grep -xvf $THIS_NODE_TESTS_PATH > $IGNORE_CLASS_LIST_PATH
+
+            // save path for next steps
```
yufengwng:
```suggestion
            # save path for next steps
```

yufengwng (commented):
Looking good as far as I can tell. I recommend squashing your commits to make the PR more approachable (otherwise, I fear the commits will drown out the comments...)

.circleci/config.yml
MocanaAtBroad: Mistake!

.circleci/config.yml
```diff
@@ -0,0 +1,653 @@
...
+            # And ignore the Suite itself
+            echo 'org.broadinstitute.ddp.route.IntegrationTestSuite'  >> $IGNORE_CLASS_LIST_PATH
+
+            cp $THIS_NODE_TESTS_PATH "${TEST_DATA_PATH}" /this_node_circleci_tests
+
+            # save path for next steps
+            echo "export IGNORE_CLASS_LIST_PATH=${IGNORE_CLASS_LIST_PATH}" >> $BASH_ENV
+            source $BASH_ENV
+            echo "Total number of test classes that will be skipped in this node: $(wc -l $IGNORE_CLASS_LIST_PATH)"
+      - store_artifacts:
```
MocanaAtBroad, line 370: Where we figure out how to split tests over parallel jobs. The ignore list of class names is what `maven test` users to skip tests (instead of passing the list of tests to execute)

.circleci/config.yml
```diff
@@ -0,0 +1,653 @@
...
+    description: Run test server that applies liquibase and load test data
+    steps:
+      - run-maven-with-pepper-options:
+          phase: exec:java
+          additional_options: '-Dexec.mainClass=org.broadinstitute.ddp.route.IntegrationTestSuite -Dexec.classpathScope=test'
+          background: true
+      - wait_for_server:
+          port: 5556
+          timeout: 60
+
```
MocanaAtBroad, line 469: How we start a DataDonationPlatform server that will be used by (most) of our tests.

.circleci/config.yml
```diff
@@ -0,0 +1,653 @@
...
+          path: *repo_path
+      - setup-shared-env
+      - render-configs
+      - create-test-dbs
+      - run-maven-with-pepper-options:
+          phase: test
+          skip_tests: false
+          additional_options: '-Dtest=ServerInSameProcessIntegrationTestSuite'
+          restore_m2_cache: true
+
```
MocanaAtBroad, line 487: This is the job where we run the tests that cannot run in parallel. Thing in common is that these tests modify data on the server side to operate, so server needs to be in same Java process as the test. Luckily, it is just a handful.

.circleci/config.yml
```diff
@@ -0,0 +1,653 @@
...
+          skip_checkstyle: false
+      - render-configs
+      - create-test-dbs
+      - start-ddp-test-server # Forcing Liquibase to execute ahead of tests so test timings are not eschewed by order
+      - compute-parallel-test-split
+      - run-maven-with-pepper-options:
+          phase: test
+          skip_tests: false
+          additional_options: '-PCircleciParallelBuild -Dsurefire.excludesFile=$IGNORE_CLASS_LIST_PATH'
+          save_m2_cache: true
```
MocanaAtBroad, line 513: Note we using Maven profile. More info in pom.

pepper-apis/Dockerfile-java-build
MocanaAtBroad: Yes. Turns out that in CircleCi the time to load the docker images is an important chunk of our build time. And there seems to be relationship between the size of the image and the time it takes to load.
So for the java compile/build I tried to keep the images as small as possible. Leaving out the gcloud stuff (and in particular the gcloud pub sub emulator) saved a ton of MBs.
Now one thing that is a little frustrating with CircleCi is the high variability of Docker image startup times: Sometimes these Java images are loading in a few seconds, and some other times 2 + minutes. Most of the time they are now taking 30 seconds or less.
Not super satisfied with our current management of images and their mapping to Docker images. I am kind of punting on that problem for now, but it is real (though not a huge one).

pepper-apis/Dockerfile-java-build
```diff
@@ -0,0 +1,42 @@
...
+    # to support configure.rb
+    ruby \
+    ruby-json  \
+    && rm -rf /var/cache/apk/*
+
+# keep track of git hash
+ARG GIT_SHA
+RUN echo $GIT_SHA > /git_sha.txt
+
+CMD "/bin/bash"
```
MocanaAtBroad, line 42: The Java docker image. Trying to keep minimum needed to run the compile/tests job in CircleCi.

pepper-apis/build-utils/gcloud-pubsub-emulator/make.sh
MocanaAtBroad: Not set in stone. Its a utility image. And to create docker rep we need to go through BITS. Maybe we need a ddp-utility repo. Open to suggestions.

pepper-apis/config/application.conf.ctmpl
MocanaAtBroad: Not sure:
The GAE URLs look like so:
`jdbc:mysql:///pepperdevclone?cloudSqlInstance=broad-ddp-dev:us-central1:ddp-dev&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=XXXXX&password=XXXXXX&characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC&useSSL=false&sessionVariables=innodb_strict_mode=on,tx_isolation='READ-COMMITTED',sql_mode='TRADITIONAL'`
So no reference to `sqlproxy`. I expect we will still need to use sqlproxy when trying to get to databases locally. Perhaps we can figure out a way to update the generated config files to create the sqlproxy urls from the "real" ones using a flag?

pepper-apis/config/testing-circleci.conf.ctmpl
MocanaAtBroad: I'll take a look at that. Just found an old error on the config file and had to correct in both. So this setup is not really acceptable.

pepper-apis/scripts/runcitests.sh
MocanaAtBroad: Not a fan of underscores in script names (harder to type), but will rename for the sake of consistency; I see we already have some scripts with underscores.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
MocanaAtBroad: @zyme ?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PubSubEmulator.java
```diff
@@ -149,6 +154,29 @@ public static void shutdown() throws IOException {
...
+                    httpClient.execute(new HttpGet("http://" + EMULATOR_HOST));
+            return response.getStatusLine().getStatusCode() == 200;
+        } catch (HttpHostConnectException e) {
+            LOG.debug("Could not connect to pubsub server. Must not be running");
+            return false;
+        } catch (IOException e) {
+            String msg = "There was problem initializing CloseableHttpClient";
+            LOG.error(msg, e);
+            throw new DDPException(msg, e);
+        }
```
MocanaAtBroad, line 180: In CircleCi we are starting a pubsub emulator from a Docker image. If the port is being used we are going to assume a pubsub emulator already running and no need to start our own from Java.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -336,4 +361,30 @@ public static void doSuiteTearDown() {
...
+            throw new DDPException(msg, e);
+        }
+
+    }
+
+    public static void main(String[] args) {
+        initializeDatabase();
+        startupTestServer();
+    }
+
```
MocanaAtBroad, line 388: Refactored IntegrationSuite here so we can start a test server that has applied to the common database changes. Tried to preserve existing functionality. Definitely room for improvement here.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/ServerInSameProcessIntegrationTestSuite.java
```diff
@@ -0,0 +1,17 @@
...
+ * differently then other route tests that just need access to the server via http and and the database
+ */
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+        GetParticipantInfoRouteTest.class,
+        ListCancersRouteTest.class,
+        GetDsmDrugSuggestionsRouteTest.class
+})
+public class ServerInSameProcessIntegrationTestSuite extends IntegrationTestSuite {
+}
```
MocanaAtBroad, line 17: These are the special test that are run separate from all the others in CircleCi as the comment indicates.
Kind of hacky. Tried hard to find a way with JUnit to tag just these tests or add them dynamically but could not find a good solution. Open to suggestions on how to achieve same thing more cleanly.

pepper-apis/src/test/java/org/broadinstitute/ddp/db/housekeeping/dao/KitCheckDaoTest.java
```diff
@@ -49,6 +49,9 @@ public static void setUp() {
             testData = TestDataSetupUtil.generateBasicUserTestData(handle);
             userGuid = testData.getUserGuid();
             studyGuid = testData.getStudyGuid();
+            JdbiMailAddress jdbiAddress = handle.attach(JdbiMailAddress.class);
+            jdbiAddress.findDefaultAddressForParticipant(userGuid)
+                    .ifPresent(address -> jdbiAddress.deleteAddress(address.getId()));
         });
```
MocanaAtBroad, line 55: The test testCheckForKits_userEnrolled_missingMailAddress started failing consistently. I believe that the test was expecting that the user it was using did not have a mailing address. But because now the split test functionality changes the order of tests, it is likely that another test added an address. This change ensures test sets things up as expected.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
yufengwng: To clarify what I mean in case there's confusion: we don't have any actual production integration with firecloud. So I feel these `firecloud_keys_dir` things are somewhat a relic of past prototyping, and will be nice to remove them so we don't have to add stuff we don't need on the command line.

pepper-apis/src/test/java/org/broadinstitute/ddp/util/TestDataSetupUtil.java
MocanaAtBroad: Guess not. 520427c

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PubSubEmulator.java
MocanaAtBroad: That is nicer. cee618e

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
MocanaAtBroad: I went a little bit beyond command line arguments and environment variables:
https://github.com/broadinstitute/ddp-study-server/pull/174

pepper-apis/src/test/java/org/broadinstitute/ddp/db/TransactionWrapperTest.java
MocanaAtBroad: https://github.com/broadinstitute/ddp-study-server/blob/f1c2bf0c25787c26b921bed84d5a44d6ef8eae82/pepper-apis/src/test/java/org/broadinstitute/ddp/db/TransactionWrapperTest.java#L95

.circleci/config.yml
MocanaAtBroad: https://github.com/broadinstitute/ddp-study-server/blob/f1c2bf0c25787c26b921bed84d5a44d6ef8eae82/.circleci/config.yml#L100

.circleci/config.yml
MocanaAtBroad: The builtin. Intended to be a no-op with exit code 0 . Last command in this script needs to exit with a 0. I have added a comment to clarify.
https://github.com/broadinstitute/ddp-study-server/blob/f1c2bf0c25787c26b921bed84d5a44d6ef8eae82/.circleci/config.yml#L278

.circleci/config.yml
MocanaAtBroad: I tripped over this one before I saw your comment.

.circleci/config.yml
yufengwng: Can we just use `exit 0`?

MocanaAtBroad:
It ought to do the same thing 44eebc8 .

pepper-apis/config/testing-circleci.conf.ctmpl
MocanaAtBroad: Consolidated testing config files: a670d35

pepper-apis/config/application.conf.ctmpl
yufengwng: Hmm, I'm not sure that is necessary? Because for sqlproxy, we just need the sql instance id (which is `broad-ddp-dev:us-central1:ddp-dev` here). Then we just need the user credentials. So the GAE URL has everything we need to connect locally.

pepper-apis/src/appengine/DataDonationPlatform.yaml
yufengwng: Is this scaling behavior desired? How long does it take to spin up a new instance after a period of no usage?

pepper-apis/src/appengine/Housekeeping.yaml
yufengwng: Wait a minute.. is it only a warning and not an error?

pepper-apis/src/appengine/DataDonationPlatform.yaml
MocanaAtBroad: I was observing about 20 seconds after that first GET/PUT/Whatever. Not terrible for dev/test or even staging. But we will need to figure out something different for prod.

pepper-apis/src/appengine/Housekeeping.yaml
MocanaAtBroad: A warning to the effect of since you included these characters we will have to spawn a shell before we can launch your java process which slows down start up time (I forget the specific wording but got fed up with seeing that warning each time).

pepper-apis/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -153,20 +159,35 @@ public static void setup() {
 
         LogbackConfigurationPrinter.printLoggingConfiguration();
 
-        RouteTestUtil.loadConfig();
+        initializeDatabase();
+
+        if (!isTestServerRunning()) {
+            startupTestServer();
+            insertTestData();
```
yufengwng, line 165: `insertTestData()` is already called within `startupTestServer()`?

.circleci/config.yml
```diff
@@ -0,0 +1,633 @@
...
+            test)
+              DEPLOY_ENV=test
+            ;;
+            test_deleteme1)
+              DEPLOY_ENV=dev
+            ;;
+            staging)
+              DEPLOY_ENV=staging
+            ;;
+            prod)
```
yufengwng: I think this deploy branch should be named `production`.

yufengwng (commented):
Looks good. Don't see anything else that jumps out at me.

pepper-apis/config/application.conf.ctmpl
MocanaAtBroad: Right. The GAE URL is the "real" url that sqlproxy is connecting to for us.

If starting from scratch, I would break out the individual fields out of the jdbc url (particularly the credentials) and set those from within the connection creation code. Something distasteful of a URL that includes a password.

But I am trying preserve existing functionality and GAE deployments and our current deployments to live at the same time.

And I am also wary on going on another yak shaving expedition. I wholly agree this can be better but I hope you would agree that we can take on cleaning up the connection URLs as a separate undertaking.

.circleci/config.yml
MocanaAtBroad: 66b170b

yufengwng: Wait... can you swap this? The branch should be `production` and the env should be `prod`.

MocanaAtBroad: Note to self: when making changes at the end of the day be triply careful as you are more likely to make stupid mistakes.
ff3dd42c

yufengwng (approved):
looks good as far as I can tell

MocanaAtBroad:
Please run tests.

PR #173, opened 2020-04-17
Event-driven management of housekeeping

yufengwng:
## Context

Currently, there's no easy way for "management"-style operations on Pepper or Housekeeping. For example, operations like data export need to be ran manually and is often very slow because you're doing it locally. Another operation is clearing any caches. This is done by rebooting Pepper or Housekeeping, which is not ideal and prevents us from reaching zero-downtime.

## Proposal

We hook both Pepper and Housekeeping into pubsub, and drive them via pubsub messages. We will assign each of them a pubsub topic, which we can publish messages into to tell Pepper/Housekeeping to do some task (clear caches, export data, etc). Pepper and Housekeeping is capable of slightly different things, so we use different pubsub topics. Current thinking is to support common operational tasks, but we might expand this design to handle other things.

#### Messages

Each message has a `type` attribute that tells us what task it is. Some tasks are simply and will not need any other pieces of data. Other tasks will need additional data, which will be provided in the message body as a serialized JSON payload. A message can easily be published either via cmdline or gcloud console.

#### Broadcast and Unicast

Each "instance" of Pepper/Housekeeping will create a subscription for itself upon startup. This is done because GAE instances are designed to be "interchangeable" and ephemeral, so there isn't really a way to uniquely identity each instance, thus we have each instance automatically create a subscription.

Since each instance will have it's own subscription to the event topic, sending a message to the topic will effectively "broadcast" the message to each instance. This is good -- some tasks, like clear cache, we will want all instances to do. For "unicast" and targeting just one instance, add a `sub` attribute to the message with a specific subscription name.

#### Security

You need to be authenticated in order to publish messages to the topics, so we won't need to worry about random messages. Also, the task that we'll have Pepper/Housekeeping perform are well-specified, so if we get an event we don't know about then that's an error.

#### Issues

We shouldn't have issues setting this up for Housekeeping, because it already deals with pubsub. However, for Pepper, this adds a new dependency.

Automatic creation of subscriptions will work, but there's a bit of worry that we'll end up with a lot of "abandoned" subscriptions. This is because everything you reboot Pepper or Housekeeping, it'll create a new subscription. This is probably fine for now. We can periodically remove old subscriptions. To find which subscriptions are "active", send a `PING` message. Also, we can set a expiration on a subscription, so that pubsub will automatically delete them if there's no activity. Right now we set it to 10 days, so that should be good enough.


ssettipalli:
Event driven management feature looks good. We will have good use in Housekeeping for sure.
However in pepper backend as of now we might not need it other than refreshing cache when new study is deployed. In prod for most cases we do restart after new study deploy. May be we can avoid additional overhead on pepper backend for now and have it only in Housekeeping.

ssettipalli:
Overall looks good to me and will be useful feature to be able to trigger these Housekeeping events.

ssettipalli (approved):

PR #176, opened 2020-04-25
DDP-4551: Cross-study registration bug: Auth0 rule

MocanaAtBroad:
## Context
Fix involves changes Auth0 rule and front-end . This PR contains modification to Auth0.

**Question for reviewers**: I branched off `test`. Is this the right place to branch off?

Note: I modified the Auth0 rule code in code repo, but what is in repo is slightly different than the rule I found in Auth0 dev and in test tenants. @bskinner The part that is different appears to relate to SALT stuff. Do take a look.

The delta I am proposing would look the same with code in Auth0 but we should resolve disparity.

I have applied my rule delta to the dev environment. Might impact other app deployments!

## Checklist

## FUD Score

- [X ] :sweat_smile: There might be some issues here

Walk through steps in bug report.

## Testing

- [ X] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release
Release of this feature requires updating the Auth0 rule, most likely through Auth0 management page.
- [X ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document


pepper-apis/auth0-config/rules/register-user-in-pepper.js
```diff
@@ -103,6 +103,12 @@ function (user, context, callback) {
 
         console.log(context);
 
+        // This is the token renewal case. Let's avoid going through pepper registration
+        if (context.request.query.renew_token_only) {
+            context.idToken[pepperUserGuidClaim] = user.app_metadata.user_guid;
+            return callback(null, user, context);
+        }
+
```
MocanaAtBroad: The basic problem was that token renewals ("Click Continue to extend your session") have been all along generating calls to to the pepper '/register' route.
While unnecessary, under normal circumstances it is not a problem to call /register twice.
Problem comes in when a second study is openned in the same browser. Original Auth0 session cookie gets overwritten, most importantly potentially changing the Auth0 user id.
So when renewal for token goes through with the correct study but with the wrong user id, you end up with user incorrectly registered in another study.
These 3 lines of code just make sure that token renewals never go on to call /register route. Note that the renewal token returned by Auth0 can still be incorrect! 
This situation is detected and handled in the client. See related ddp-angurlar PR:
(https://github.com/broadinstitute/ddp-angular/pull/97)

yufengwng (changes requested):
LGTM.

Re: disparity between checked-in code and what's on `dev`/`test`. I took a look and I see code is same as what's on `test`. `dev` has some new code that seems to relate to GAE. @MocanaAtBroad does that ring a bell?

Re: branching. `test` branch is a deployment branch (so is `staging` and `production` branches). We should not PR off of a deployment branch. Can you point your branch against `develop` so we can get this into the next release? Will approve after this is done.

fat23cat (approved):

yufengwng (approved):

yufengwng:
If this gets merged in for `v1.8.2` release, then please add instructions for updating Auth0 to `v1.8.2` release notes.

PR #182, opened 2020-04-29
Add CI checks for study-builder

yufengwng:
## Context

We occasionally breaks study-builder because it's not part of pepper-apis in the source directory and we do not have compile/checkstyle checks on it. No more. This PR adds a step in CircleCI to ensure that study-builder compiles and passes checkstyle. Also moved pepper-apis and housekeeping compilation/checkstyle checks earlier in the process.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -519,12 +519,13 @@ commands:
           condition: << parameters.restore_m2_cache >>
           steps:
             - restore_cache: *m2_cache_keys
-      - run: |
-          pwd
-          mvn install --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
-          cd << parameters.repo_path >>/study-builder
-          ls -l
-          mvn package --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+      - run:
+          name: Compiling pepper-apis then study-builder
+          command: |
+            mvn install -f parent-pom.xml --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+            mvn install --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+            cd << parameters.repo_path >>/study-builder
+            mvn package --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
 
 jobs:
   compile-and-run-test-suite-job:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -520,7 +520,7 @@ commands:
           steps:
             - restore_cache: *m2_cache_keys
       - run:
-          name: Compiling and installing Pepper to m2 repository
+          name: Compiling and installing pepper-apis to m2 repository
           command: mvn install --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
       - run: cd << parameters.repo_path >>/study-builder
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -568,8 +568,10 @@ jobs:
       - checkout:
           path: *repo_path
       - setup-shared-env
-      - compile-all-jars:
+      - run-maven-with-pepper-options:
+          phase: process-test-classes
           restore_m2_cache: true
+          skip_checkstyle: true
       - render-configs
       - create-test-dbs
       - start-ddp-test-server # starts DDP server. And execute Liquibase DB population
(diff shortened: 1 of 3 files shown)


pepper-apis/parent-pom.xml
```diff
@@ -406,6 +405,12 @@
             <groupId>org.quartz-scheduler</groupId>
             <artifactId>quartz</artifactId>
             <version>2.3.0</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.zaxxer</groupId>
+                    <artifactId>HikariCP-java6</artifactId>
+                </exclusion>
+            </exclusions>
```
yufengwng, line 413: Quartz comes with an old version of HikariCP (for java 6 !) which conflicts with what we declared. Excluding it so we get the right version into the jar. We're not using database persistence of jobs in Quartz so this should matter to Quartz functionality.

.circleci/config.yml
```diff
@@ -502,6 +502,32 @@ commands:
...
+            - restore_cache: *m2_cache_keys
+      - run:
+          name: Compiling pepper-apis, housekeeping, and study-builder
+          command: |
+            mvn install --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+            mvn install -f parent-pom.xml --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+            mvn package -f housekeeping-pom.xml --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+            cd << parameters.repo_path >>/study-builder
+            mvn package --batch-mode -DskipTests -Dmaven.repo.local=<< parameters.m2_repo >>
+
```
MocanaAtBroad, line 537: Does running `install` catch problems? If not we should just run `test-compile` as creating the mega jars is slow (was trying to keep the many-times-a-day/be-done-already run tests as fast as possible). Had been deferring building the jars on purpose.
Also was shooting for keeping the number `mvn` invocations to a minimum. take time.
And checkstyle execution was only happening on the non-parallel compile-run-test-suite-job . That job has less to do/finishes faster and checkstyle was touching everything there.

.circleci/config.yml
```diff
@@ -533,6 +561,8 @@ jobs:
       - checkout:
           path: *repo_path
       - setup-shared-env
+      - compile-all-jars:
+          restore_m2_cache: true
       - run-maven-with-pepper-options:
           phase: process-test-classes
           restore_m2_cache: true
```
MocanaAtBroad: We only need to restore m2_cache once: it reads it back from wherever circle archives .m2/repository .

MocanaAtBroad:
Will share that I have found it helpful to look at the CircleCi job executions in the console, particularly the order of the commands and the output of the maven execution (depending on when  m2 repo is stored to cache you might end up downloading libraries from repos repeatedly).

.circleci/config.yml
yufengwng: `install` doesn't catch additional problems, but I think we have to do it given the fact that study-builder is a separate thing and needs the pepper-apis dependency in .m2. I can probably turn the other `package` ones into `test-compile`.

About the parallel job: initially I didn't want to put this in that job at all. We don't need two jobs doing the same compile task. But I was thinking if there was compile/checkstyle error then the parallel job will catch it too late in the game. Maybe that's a non-concern? Should I just remove it from that job?

.circleci/config.yml
yufengwng: Ah ha! The more you know! Will change it.

.circleci/config.yml
MocanaAtBroad: I think we can take it out of the parallel-*-job. Does not add value to have it there too and is the slower job. Doing compile-all-jars only in the compile-and-run-test-suite-jobs likely to keep the overall build time the same. Hopefully the caching will be OK.

yufengwng: On second thought, I think `compile-all-jars` does add value in the parallel job. We need to do a compile anyways for the parallel job to run the tests, but now we're moving it up earlier in the process. This way, if there was any compile/checkstyle issues, we don't even need to do the parallel test thingy, so we fail faster.

MocanaAtBroad: Don't think this is deal breaker, but probably does not make it faster. Since the `parallel-compile-and-test-job`  job runs in parallel to the `compile-and-run-test-suite-job` that would include the `compile-all-jars step` ,  both jobs would find the problem around the same time, but then `parallel-compile-and-test-job` would take longer to complete.
We probably splitting hairs anyways. Probably just an addition 20 or 30 secs. But on the other hand sometimes we can get impatient.

yufengwng: I reverted parallel job. PR is ready to go.

MocanaAtBroad (approved):

PR #195, opened 2020-05-09
Invitation APIs and recruitment invitation type

yufengwng:
This PR should cover almost all the APIs and other changes we need in terms of invitations.

* Refactored invitations so they're a bit more general
* Added `RECRUITMENT` invitation types
* Added a `notes` field to invitations for storing freeform notes
* Added handling of new invitations in `/register` route
* Added `CheckInvitationStatusRoute` route that clients can call to check invite code
* Added `GetInvitationRoute` for looking up an invite
* Added `UpdateInvitationRoute` for saving invitation notes

NOTE: have not implemented study admin API restrictions yet, will handle in separate PRs.


pepper-apis/src/main/java/org/broadinstitute/ddp/route/CheckInvitationRoute.java
```diff
@@ -0,0 +1,71 @@
...
+                return HttpStatus.SC_BAD_REQUEST;
+            } else if (invitation.isVoid()) {
+                LOG.error("Invitation {} is voided", invitationGuid);
+                return HttpStatus.SC_BAD_REQUEST;
+            } else if (invitation.isAccepted()) {
+                LOG.error("Invitation {} has already been accepted", invitationGuid);
+                return HttpStatus.SC_BAD_REQUEST;
+            } else {
+                LOG.info("Invitation {} is valid", invitationGuid);
+                return HttpStatus.SC_OK;
```
zyme: Intent here is to *not* divulge detailed status, but just have a UI say something like "We're sorry, that invitation is invalid" more generally, right?  Looks good.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -225,6 +166,95 @@ public Object handle(Request request, Response response, UserRegistrationPayload
...
+                    ResponseUtil.halt422ErrorResponse(response, ErrorCodes.GOVERNANCE_POLICY_VIOLATION);
+                }
+            } else {
+                LOG.error("No governance policy for study {}.  Why is a client registering user {} with invitation {} ?",
+                        studyGuid, auth0UserId, invitationGuid);
+                ResponseUtil.halt422ErrorResponse(response, ErrorCodes.GOVERNANCE_POLICY_VIOLATION);
+            }
+
+            return user;
+        } else if (invitation.getInvitationType() == InvitationType.RECRUITMENT) {
```
zyme, line 214: I like the introduction of a new type.  Maybe introduce an interface like `InvitationValidator` instead of a type-based if-then?

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
```diff
@@ -0,0 +1,25 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="yufeng" id="20200508-refactor-invitations">
+        <dropNotNullConstraint tableName="invitation" columnName="user_id" columnDataType="bigint"/>
+        <dropNotNullConstraint tableName="invitation" columnName="contact_email" columnDataType="varchar(255)"/>
+        <dropUniqueConstraint tableName="invitation" constraintName="invitation_guid_uk"/>
+
+        <addUniqueConstraint tableName="invitation" columnNames="study_id, invitation_guid" constraintName="invitation_study_uk"/>
```
zyme, line 12: Cool, so we can still have multiple invitations per user.  So we could recruit someone with an invite, and then age do an age up, and everything works :)

pepper-apis/src/test/java/org/broadinstitute/ddp/db/dao/InvitationDaoTest.java
```diff
@@ -26,25 +25,25 @@ public void testInsertUpdateAndRead() {
...
-            assertTrue(invitation.getCreatedAt().before(new Timestamp(Instant.now().toEpochMilli())));
+            assertTrue(invitation.getCreatedAt().isBefore(Instant.now()));
 
             // set various dates
-            Timestamp createdAt = invitation.getCreatedAt();
-            Timestamp acceptedAt = new Timestamp(createdAt.getTime() + (2 * 10000));
-            Timestamp verifiedAt = new Timestamp(createdAt.getTime() + (4 * 10000));
-            Timestamp voidedAt = new Timestamp(createdAt.getTime() + (6 * 10000));
+            Instant createdAt = invitation.getCreatedAt();
+            Instant acceptedAt = createdAt.plus(2 * 10000, ChronoUnit.MILLIS);
```
zyme, line 48: Yay units!

zyme (commented):
Overall makes sense to me.  Thanks for the early check in.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
```diff
@@ -1,49 +1,59 @@
...
-public class VerifyInvitationRoute extends ValidatedJsonInputRoute<InvitationPayload> {
+public class VerifyInvitationRoute extends ValidatedJsonInputRoute<VerifyInvitationPayload> {
 
     private static final Logger LOG = LoggerFactory.getLogger(VerifyInvitationRoute.class);
 
     @Override
-    public Object handle(Request request, Response response, InvitationPayload invitation) throws Exception {
+    public Object handle(Request request, Response response, VerifyInvitationPayload invitation) throws Exception {
+        String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
         String invitationGuid = invitation.getInvitationId();
```
MocanaAtBroad: In Pepper when I see the `Id` suffix on a Java POJO I expect it to be the database row id. Consider sticking with `Guid`

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
```diff
@@ -1,49 +1,59 @@
...
+
+        LOG.info("Attempting to verify invitation {} in study {}", invitationGuid, studyGuid);
 
         TransactionWrapper.useTxn(handle -> {
+            StudyDto studyDto = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
+            if (studyDto == null) {
+                LOG.warn("Could not find study with guid {}", studyGuid);
+                return;
+            }
+
```
MocanaAtBroad: Meta comment: high time that we expanded our own enhanced Route class (maybe ValidatedJsonRoute maybe a new one) that auto extracts the the user object and study object and injects into the body of the "handle" method like we do here with payload.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
```diff
@@ -1,49 +1,59 @@
...
             InvitationDao invitationDao = handle.attach(InvitationDao.class);
-            Optional<InvitationDto> invitationDto = invitationDao.findByInvitationGuid(invitationGuid);
+            Optional<InvitationDto> invitationDto = invitationDao.findByInvitationGuid(studyDto.getId(), invitationGuid);
 
             invitationDto.ifPresent(invite -> {
                 if (invite.canBeVerified()) {
-                    int numRows = invitationDao.updateVerifiedAt(verifiedAt, invitationGuid);
-
+                    int numRows = invitationDao.updateVerifiedAt(invite.getInvitationId(), verifiedAt);
                     if (numRows > 1) {
```
MocanaAtBroad: IMO this is overly-paranoid given we control schema.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
MocanaAtBroad: Never mind. I see that the id suffix is "legacy".

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
```diff
@@ -0,0 +1,25 @@
...
+        <modifyDataType tableName="invitation" columnName="created_at" newDataType="datetime(6)"/>
+        <modifyDataType tableName="invitation" columnName="verified_at" newDataType="datetime(6)"/>
+        <modifyDataType tableName="invitation" columnName="voided_at" newDataType="datetime(6)"/>
+        <modifyDataType tableName="invitation" columnName="accepted_at" newDataType="datetime(6)"/>
```
MocanaAtBroad, line 16: good call on datetime(6). We should consider doing the refactoring on the whole database.

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
MocanaAtBroad: I see now why we have the check for multiple updated rows in VerifyInvitationCheckRoute. Don't see how it helps us to loosen the guid uniqueness constraint to just within a study, though.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CheckInvitationRoute.java
```diff
@@ -0,0 +1,71 @@
...
+                LOG.info("Invitation check by client clientId={}, tenant={}",
+                        payload.getAuth0ClientId(), payload.getAuth0Domain());
+            } else {
+                LOG.error("Invitation check by client which does not have access to study: clientId={}, tenant={}",
+                        payload.getAuth0ClientId(), payload.getAuth0Domain());
+                return HttpStatus.SC_BAD_REQUEST;
+            }
+
+            // todo: check recaptcha
+
```
MocanaAtBroad: We'll see how it goes with the recaptcha: planning on doing a POC really soon. I am thinking that the best way to handle it would be as part of a filter and that we would ideally turning on the recaptcha option from study builder for anything that we can configure there (at this point would be any activity instance). Not asking you to do anything at this point (you are appropriately leaving it out) but just sharing preliminary thoughts.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
```diff
@@ -1,49 +1,59 @@
...
 import org.broadinstitute.ddp.util.ResponseUtil;
-import org.broadinstitute.ddp.util.TimestampUtil;
 import org.broadinstitute.ddp.util.ValidatedJsonInputRoute;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
 
-public class VerifyInvitationRoute extends ValidatedJsonInputRoute<InvitationPayload> {
+public class VerifyInvitationRoute extends ValidatedJsonInputRoute<VerifyInvitationPayload> {
```
MocanaAtBroad: Was confused about difference between **Verify**InvitationRoute and **Check**InvitationRoute as the words are generally synonyms. A short comment might be in order.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -225,6 +166,95 @@ public Object handle(Request request, Response response, UserRegistrationPayload
...
+
+            int numRows = handle.attach(InvitationDao.class)
+                    .assignAcceptingUser(invitation.getInvitationId(), user.getId(), Instant.now());
+            if (numRows != 1) {
+                throw new DDPException("Updated " + numRows + " for invitation " + invitationGuid);
+            }
+
+            return user;
+        } else {
+            throw new DDPException("Unhandled invitation type " + invitation.getInvitationType());
```
MocanaAtBroad, line 221: This feels more like a BAD REQUEST. (Actually, what is our default handler returning on uncaught Runtime exceptions? 500?)

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/BeforeWithExclusionFilter.java
```diff
@@ -33,4 +34,11 @@ public static void beforeWithExclusion(String filterPath, String pathToExclude,
         beforeWithExclusion(filterPath, List.of(pathToExclude), filter);
     }
 
+    public static void afterWithExclusion(String filterPath, List<String> pathsToExclude, Filter filter) {
+        after(filterPath, new ExcludePathFilterWrapper(filter, pathsToExclude));
+    }
+
+    public static void afterWithExclusion(String filterPath, String pathToExclude, Filter filter) {
+        afterWithExclusion(filterPath, List.of(pathToExclude), filter);
+    }
```
MocanaAtBroad: Since we are doing `after` and `before`, maybe we should rename this class now? `FilterWithExclusion`?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
MocanaAtBroad: Found why you doing it later. Still not sure we should not be enforcing global uniqueness of invitation code.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: We have already done some sanity checks and validation, so this part of the code is for assigning the invitation.

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
yufengwng: Actually, the `numRows` checks are just paranoid and not related to guid uniqueness.

I loosened the uniqueness constraints to allow studies to generate their own invitation codes instead of random UUIDs the system generates. For example `testboston` wants to have their own nice invitation codes. If other studies want to generate their own invitation codes, we might run into conflicts. Also, it kind of makes sense that invitations are study-specific things, thus scoped to a study.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
yufengwng: I moved these `numRows` checks into the DAO itself, so we don't have to do this everywhere we use the DAO.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
yufengwng: done

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: If we got here, it's a programmer error since we didn't add new code to handle new type. Our default handler returns 500.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/BeforeWithExclusionFilter.java
yufengwng: I simply removed the whole thing and used the wrapping filter directly.

pepper-apis/src/test/java/org/broadinstitute/ddp/model/event/NotificationEventActionTest.java
```diff
@@ -116,20 +114,15 @@ public void test_invitationEmail_providesInvitationIdSubstitution() {
...
+            InvitationDto invitation1 = invitationFactory.createAgeUpInvitation(
                     testData.getStudyId(), testData.getUserId(), "test111@datadonationplatform.org");
-            InvitationDto invitation2 = invitationFactory.createInvitation(InvitationType.AGE_UP,
+            InvitationDto invitation2 = invitationFactory.createAgeUpInvitation(
                     testData.getStudyId(), testData.getUserId(), "test222@datadonationplatform.org");
-            try {
-                Thread.sleep(2);
-            } catch (InterruptedException e) {
-                fail("Got interrupted while sleeping");
-            }
```
yufengwng, line 127: @MocanaAtBroad I kind of think that we got into that timing issue on CircleCI because we were only using millisecond precision. I have expanded precision of these timestamps, so hopefully we don't run into that timing issue again and give us 2 seconds back :)

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -54,6 +54,12 @@ paths:
     $ref: 'endpoints/join.mailing.list.yml'
   /studies/{study}:
     $ref: 'endpoints/studies.study.yml'
+  /studies/{study}/invitation-check:
+    $ref: 'endpoints/studies.invitation-check.yml'
+  /studies/{study}/invitation-verify:
+    $ref: 'endpoints/studies.invitation-verify.yml'
+  /studies/{study}/invitations/{invite}:
+    $ref: 'endpoints/studies.invitations.yml'
```
yufengwng: Screen Shot 2020-05-13 at 2 37 25 PM

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -54,6 +54,12 @@ paths:
     $ref: 'endpoints/join.mailing.list.yml'
   /studies/{study}:
     $ref: 'endpoints/studies.study.yml'
+  /studies/{study}/invitation-check:
+    $ref: 'endpoints/studies.invitation-check.yml'
+  /studies/{study}/invitation-verify:
```
yufengwng, line 63: @fat23cat I had to change the URL of this API to this. I think we need to update Angular SDK here: https://github.com/broadinstitute/ddp-angular/blob/develop/ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/invitationsServiceAgent.service.ts.

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
```diff
@@ -0,0 +1,35 @@
+post:
+  operationId: checkInvitationStatus
```
yufengwng: Screen Shot 2020-05-13 at 2 38 07 PM

pepper-apis/docs/specification/src/endpoints/studies.invitation-verify.yml
```diff
@@ -0,0 +1,24 @@
+post:
+  operationId: verifyInvitation
```
yufengwng: Screen Shot 2020-05-13 at 2 38 28 PM

pepper-apis/docs/specification/src/endpoints/studies.invitations.yml
```diff
@@ -0,0 +1,101 @@
...
+      type: string
+  - in: path
+    name: invite
+    required: true
+    description: the invitation id
+    schema:
+      type: string
+
+get:
+  operationId: getInvitation
```
yufengwng: Screenshot_2020-05-13 Pepper

pepper-apis/docs/specification/src/endpoints/studies.invitations.yml
```diff
@@ -0,0 +1,101 @@
...
+                nullable: true
+              userLoginEmail:
+                description: login email of associated user, if invitation is accepted and user has a login account
+                type: string
+                nullable: true
+    404:
+      description: invitation not found
+
+patch:
+  operationId: patchInvitation
```
yufengwng: Screen Shot 2020-05-13 at 2 40 49 PM

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
```diff
@@ -0,0 +1,29 @@
...
+
+        <addColumn tableName="invitation">
+            <column name="notes" type="varchar(16000)"/>
```
yufengwng, line 19: Should be plenty of room for writing notes?

study-builder/studies/testboston/study.conf
```diff
@@ -158,28 +164,68 @@
...
+      },
+      "to": [
+        {
+          "type": "DASHBOARD",
+          "expression": "true"
+        }
+      ]
+    },
+
+    # prescreen workflow
```
yufengwng, line 191: These are outdated. Will fix them up in separate PRs.

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
```diff
@@ -0,0 +1,35 @@
...
+              type: string
+            auth0Domain:
+              description: the domain of the client
+              type: string
+            invitationId:
+              description: the id of the invitation
+              type: string
+            recaptchaToken:
+              description: the user's reCAPTCHA token
+              type: string
```
MocanaAtBroad, line 32: Zip code?

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
```diff
@@ -0,0 +1,35 @@
...
+            - auth0ClientId
+            - auth0Domain
+            - invitationId
+            - recaptchaToken
+          properties:
+            auth0ClientId:
+              description: the id of the client
+              type: string
+            auth0Domain:
+              description: the domain of the client
```
MocanaAtBroad, line 25: Why not just the study guid?

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
```diff
@@ -0,0 +1,35 @@
...
+              description: the id of the invitation
+              type: string
+            recaptchaToken:
+              description: the user's reCAPTCHA token
+              type: string
+  responses:
+    200:
+      description: invitation exists and is valid and can be used by user
+    400:
+      description: invitation not valid
```
MocanaAtBroad, line 37: Error bodies?

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
yufengwng: Don't know concretely what we're doing with zip codes yet. That will be in different PR.

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
yufengwng: Study guid is already in the URL path. Requiring auth0 client/domain here so we know which client is calling in, and hopefully acts as a bit more deterrent for whackos.

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
yufengwng: No error bodies. Is there info you're expecting here? This is a public API, so I think the less stuff we return the better.

yufengwng: Oh, you mean those `ApiError` objects? I can put in a very generic one here.

MocanaAtBroad: Yeah. Since we concluded that we would do just one POST with the data, the error body needs describe the error for anything that is wrong with submission.

yufengwng: Sorry, I might be out-of-date here. What did we conclude and what are we POST-ing here?

MocanaAtBroad: Thought, for potential subjects: invitationCode , zip, and recaptcha user response.
I believe you were talking about reusing the route for CRs so in that case recaptcha would not be required (though they obviously would have to be authenticated already and have CR role).
Then question would be how you handle ZIP code check for CRs. The simplest approach would be to include the warning message in a response body. Alternative is to have client check it separately. I think the first approach is simpler, but if you feel strongly about leaving up to client to check, I am good with that too.

yufengwng: I was thinking zip code will be a separate API. Either way, I think that is outside scope of this PR, and I will like to address it separate from this PR. Is that okay?

pepper-apis/docs/specification/src/pepper.yml
fat23cat: PR is ready https://github.com/broadinstitute/ddp-angular/pull/110

pepper-apis/docs/specification/src/endpoints/studies.invitations.yml
```diff
@@ -0,0 +1,109 @@
...
+                type: string
+              createdAt:
+                description: UTC timestamp of when invitation was created, in ISO-8601 format
+                type: string
+                example: 2020-05-13T17:58:33.887Z
+              voidedAt:
+                description: UTC timestamp of when invitation is voided
+                type: string
+                nullable: true
+              verifiedAt:
```
bskinner: What is the type of this? A timestamp in UTC should be an integer, not a string, unless we're using an ISO 8601 datetime here in the UTC zone

yufengwng: fixed

pepper-apis/docs/specification/src/endpoints/studies.invitation-check.yml
MocanaAtBroad: I still think one remaining issue here is that this API is only for GUI that follows the rules; if I have a valid invitation code I could choose to bypass this whole interaction and go directly to register in Auth0. We might be missing something like a short-lived authorization code that would tell /register route that I was given a green light to register.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: Is there something we'll like changed here?

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
yufengwng: I feel I have addressed question about uniqueness. Resolving.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/VerifyInvitationRoute.java
yufengwng: Good idea, but outside of scope of this PR. Maybe we refactor later.

pepper-apis/src/main/resources/db-changes/schema/refactor-invitations.xml
yufengwng: Let's refactor later.

MocanaAtBroad (approved):

PR #202, opened 2020-05-14
Fix parameter path in docs

yufengwng:
The linter (speccy) doesn't like the hierarchy in `parameters.yml`.


MocanaAtBroad (approved):
Builds for me!

PR #205, opened 2020-05-15
DDP-4433 init profile

yufengwng:
## Context

We can now initialize user profile during registration with data passed in from Auth0. This includes user's selected language and first/last name. These can be passed to Auth0 by the client, and then passed on to backend from there.

Also made some changes to `testboston`'s universal login to show custom fields and how we handle that via `user_metadata`.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
@@ -5,6 +5,7 @@ import static spark.Spark.halt;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
 import com.auth0.exception.Auth0Exception;
@@ -19,18 +20,19 @@ import org.broadinstitute.ddp.db.dao.ClientDao;
 import org.broadinstitute.ddp.db.dao.DataExportDao;
 import org.broadinstitute.ddp.db.dao.EventDao;
 import org.broadinstitute.ddp.db.dao.InvitationDao;
-import org.broadinstitute.ddp.db.dao.JdbiCountry;
 import org.broadinstitute.ddp.db.dao.JdbiLanguageCode;
 import org.broadinstitute.ddp.db.dao.JdbiMailingList;
 import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
 import org.broadinstitute.ddp.db.dao.JdbiUserStudyEnrollment;
 import org.broadinstitute.ddp.db.dao.QueuedEventDao;
+import org.broadinstitute.ddp.db.dao.StudyDao;
 import org.broadinstitute.ddp.db.dao.StudyGovernanceDao;
 import org.broadinstitute.ddp.db.dao.UserDao;
 import org.broadinstitute.ddp.db.dao.UserGovernanceDao;
 import org.broadinstitute.ddp.db.dao.UserProfileDao;
 import org.broadinstitute.ddp.db.dto.EventConfigurationDto;
 import org.broadinstitute.ddp.db.dto.InvitationDto;
+import org.broadinstitute.ddp.db.dto.LanguageDto;
 import org.broadinstitute.ddp.db.dto.StudyDto;
 import org.broadinstitute.ddp.exception.DDPException;
 import org.broadinstitute.ddp.json.LocalRegistrationResponse;
@@ -318,7 +320,7 @@ public class UserRegistrationRoute extends ValidatedJsonInputRoute<UserRegistrat
             user = userDao.createUser(auth0Domain, auth0ClientId, auth0UserId);
             LOG.info("Registered user {} with client {}", auth0UserId, auth0ClientId);
         }
-        initProfilePreferredLanguage(handle, user, payload.getAuth0ClientCountryCode());
+        initializeProfile(handle, user, payload);
         return user;
     }
 
@@ -345,7 +347,7 @@ public class UserRegistrationRoute extends ValidatedJsonInputRoute<UserRegistrat
 
         User governedUser = handle.attach(UserDao.class).findUserById(gov.getGovernedUserId())
                 .orElseThrow(() -> new DDPException("Could not find governed user with id " + gov.getGovernedUserId()));
-        initProfilePreferredLanguage(handle, governedUser, payload.getAuth0ClientCountryCode());
+        initializeProfile(handle, governedUser, payload);
 
         int numInstancesReassigned = handle.attach(ActivityInstanceDao.class)
                 .reassignInstancesInStudy(policy.getStudyId(), operatorUser.getId(), gov.getGovernedUserId());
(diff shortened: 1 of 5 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -454,25 +456,62 @@ private User upgradeTemporaryUser(Response response, UserDao userDao, User tempU
...
+                break;
+            }
+        }
+
+        // If no language provided or language is not one of study's supported ones,
+        // then use study's default language. If all else fails, fallback to English.
         if (languageId == null) {
-            languageId = handle.attach(JdbiLanguageCode.class).getLanguageCodeId(EN_LANGUAGE_CODE);
+            languageId = supportedLanguages.stream()
+                    // .filter(lang -> lang.isDefault())     TODO: use the filter when ready
```
yufengwng, line 474: This is waiting on DDP-4475. CC @ssettipalli

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -97,8 +91,18 @@ function (user, context, callback) {
 
         if (context.request.query.invitation_id) {
             pepper_params.invitationId = context.request.query.invitation_id;
+            console.log('Invitation id passed in (via query) = ' + pepper_params.invitationId);
         } else if (context.request.body.invitation_id) {
             pepper_params.invitationId = context.request.body.invitation_id;
+            console.log('Invitation id passed in (via body) = ' + pepper_params.invitationId);
+        }
+
+        if (context.request.query.language) {
```
yufengwng, line 100: This will pick up the new `language` parameter that frontend is now passing in. CC @fat23cat @charlottedye.

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -126,6 +130,16 @@ function (user, context, callback) {
             context.idToken[pepperUserGuidClaim] = user.app_metadata.user_guid;
             return callback(null, user, context);
         } else {
+            user.user_metadata = user.user_metadata || {};
+            if (user.user_metadata.first_name) {
+                pepper_params.firstName = user.user_metadata.first_name;
+                console.log('User metadata has first name = ' + pepper_params.firstName);
+            }
+            if (user.user_metadata.last_name) {
```
yufengwng, line 138: First/last name can be collected via Auth0's [custom fields](https://auth0.com/docs/libraries/custom-signup), which will store user's input into their `user_metadata`. These will then be picked up here and passed to backend.

@MocanaAtBroad, if we're implementing a custom page for `testboston` instead of using the Lock, then I think we can add two new input fields and pass these along in the request to Auth0?

study-builder/tenants/testboston/pages/login.html
```diff
@@ -156,6 +156,31 @@
         showPasswordRecentText = true;
     }
 
+    if (mode === 'signup') {
+        realOptions.additionalSignUpFields = [
+            {
+                name: "first_name",
```
yufengwng, line 162: Name of custom fields should exactly match these, i.e. `first_name` and `last_name`, so they will get properly picked up in the `register-user-in-pepper.js` rule above. @charlottedye do these name look about right to you in terms of how you'll use it for Prion?

yufengwng (commented):
Still need to figure out how to deal with custom fields for _local registration_. If I figure something out, I'll do a PR in the `ddp-angular` repo.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -454,25 +456,62 @@ private User upgradeTemporaryUser(Response response, UserDao userDao, User tempU
...
+                    .findFirst()
+                    .orElseGet(() -> handle.attach(JdbiLanguageCode.class).getLanguageCodeId(EN_LANGUAGE_CODE));
         }
 
         var profileDao = handle.attach(UserProfileDao.class);
         UserProfile profile = profileDao.findProfileByUserId(user.getId()).orElse(null);
 
+        // If any name part is blank, then don't use it.
+        String firstName = StringUtils.defaultIfBlank(payload.getFirstName(), null);
+        String lastName = StringUtils.defaultIfBlank(payload.getLastName(), null);
```
MocanaAtBroad: Can we trim()?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -454,25 +456,62 @@ private User upgradeTemporaryUser(Response response, UserDao userDao, User tempU
...
+
+        // If no language provided or language is not one of study's supported ones,
+        // then use study's default language. If all else fails, fallback to English.
         if (languageId == null) {
-            languageId = handle.attach(JdbiLanguageCode.class).getLanguageCodeId(EN_LANGUAGE_CODE);
+            languageId = supportedLanguages.stream()
+                    // .filter(lang -> lang.isDefault())     TODO: use the filter when ready
+                    .map(LanguageDto::getId)
+                    .findFirst()
+                    .orElseGet(() -> handle.attach(JdbiLanguageCode.class).getLanguageCodeId(EN_LANGUAGE_CODE));
```
MocanaAtBroad: Consider refactoring the code in this method up to here to another called something like `determineLanguageId()`.

study-builder/tenants/testboston/pages/login.html
```diff
@@ -156,6 +156,31 @@
...
+                placeholder: "Your first name",
+                validator: function(input) {
+                  return {
+                     valid: input != null && input.length > 0,
+                     hint: "Can't be blank"
```
MocanaAtBroad, line 167: Let's check for trim()ed length.

study-builder/tenants/testboston/pages/login.html
```diff
@@ -156,6 +156,31 @@
...
+                  };
+                }
+            },
+            {
+                name: "last_name",
+                placeholder: "Your last name",
```
MocanaAtBroad, line 173: Does auth0 provide way to internationalize these (a question. Not asking you do it).

MocanaAtBroad (commented):
Very minor comments. Let's try to merge this in ASAP.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: trimmed!

study-builder/tenants/testboston/pages/login.html
yufengwng: ✂️ done!

study-builder/tenants/testboston/pages/login.html
yufengwng: On a cursory look, [here](https://auth0.com/docs/universal-login/i18n#internationalization-in-the-classic-universal-login-experience) is what Auth0 says about i18n. We're using their old/classic universal login, so sounds like they have built-in i18n for their Lock widget. However, I couldn't find mentions of what they do for custom fields. Since `validator` is just a function, I imagine we can roll our own -- just two fields for now -- using our own dictionary (pseudocode):
```js
var dict = {
  en: { blank_hint: 'can't be blank!' },
  es: { blank_hint: '...' }
};
...
validator: function(input) {
  return { ... hint: dict[language].blank_hint ... };
}
```

MocanaAtBroad (approved):

yufengwng:
Merging this in now. Please feel free to continue to add comments, and I'll address them in follow-up PRs.

study-builder/tenants/testboston/pages/login.html
charlottedye: first_name and last_name are the fields Prion is using.

study-builder/tenants/testboston/pages/login.html
charlottedye: I looked at this and this definitely seems like the approach that I was considering for the custom field internationalization.  One thing I haven't looked into much yet is the non-custom fields.  For some of those fields (and some elements we specifically add), Prion uses the language dictionary to specify values other than the default.  My sense is that Auth0 wouldn't be able to translate those, so we might do something similar for those.  For fields where we are using the Auth0 default, it might still be worth doing something like this so we know we have consistent translation.  I believe Auth0 specifies a set of available languages, so it's also possible that we could have a situation where we are using a language that Auth0 doesn't translate things into.  So certainly at least some of the time we will need to use an approach like this, and it is possible we might want to do this all of the time.

PR #210, opened 2020-05-19
DDP-4674 admin API to create user login account

yufengwng:
## Context

New admin API to create an auth0 account for user. Idea is to use Auth0's Management API to create new account and trigger password reset flow. After that, we will have the "ticket" URL. Lastly, we run study events, so studies can configure whatever they want to do at this point, e.g. grab the ticket URL and send an email to user.

## Checklist

## FUD Score

## Testing

- [x] I have tried it locally to verify proper functionality

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
@@ -57,6 +57,7 @@ import org.broadinstitute.ddp.pex.PexInterpreter;
 import org.broadinstitute.ddp.pex.TreeWalkInterpreter;
 import org.broadinstitute.ddp.route.AddProfileRoute;
 import org.broadinstitute.ddp.route.AdminCreateStudyParticipantRoute;
+import org.broadinstitute.ddp.route.AdminCreateUserLoginAccountRoute;
 import org.broadinstitute.ddp.route.AdminLookupInvitationRoute;
 import org.broadinstitute.ddp.route.AdminUpdateInvitationDetailsRoute;
 import org.broadinstitute.ddp.route.CheckIrbPasswordRoute;
@@ -166,14 +167,14 @@ public class DataDonationPlatform {
 
     public static final String MDC_STUDY = "Study";
     public static final String MDC_ROUTE_CLASS = "RouteClass";
+    public static final String PORT = "PORT";
+    public static final int DEFAULT_RATE_LIMIT_MAX_QUERIES_PER_SECOND = 10;
+    public static final int DEFAULT_RATE_LIMIT_BURST = 15;
     private static final Logger LOG = LoggerFactory.getLogger(DataDonationPlatform.class);
     private static final String[] CORS_HTTP_METHODS = new String[] {"GET", "PUT", "POST", "OPTIONS", "PATCH"};
     private static final String[] CORS_HTTP_HEADERS = new String[] {"Content-Type", "Authorization", "X-Requested-With",
             "Content-Length", "Accept", "Origin", ""};
     private static final Map<String, String> pathToClass = new HashMap<>();
-    public static final String PORT = "PORT";
-    public static final int DEFAULT_RATE_LIMIT_MAX_QUERIES_PER_SECOND = 10;
-    public static final int DEFAULT_RATE_LIMIT_BURST = 15;
     private static Scheduler scheduler = null;
 
     /**
@@ -267,7 +268,7 @@ public class DataDonationPlatform {
         var jsonSerializer = new NullableJsonTransformer();
         SimpleJsonTransformer responseSerializer = new SimpleJsonTransformer();
 
-        if (cfg.hasPath(ConfigFile.API_RATE_LIMIT.MAX_QUERIES_PER_SECOND)  && cfg.hasPath(ConfigFile.API_RATE_LIMIT.BURST)) {
+        if (cfg.hasPath(ConfigFile.API_RATE_LIMIT.MAX_QUERIES_PER_SECOND) && cfg.hasPath(ConfigFile.API_RATE_LIMIT.BURST)) {
             int maxQueriesPerSecond = cfg.getInt(ConfigFile.API_RATE_LIMIT.MAX_QUERIES_PER_SECOND);
             int burst = cfg.getInt(ConfigFile.API_RATE_LIMIT.BURST);
             LOG.info("Will use rate limit {} with burst {}", maxQueriesPerSecond, burst);
@@ -309,6 +310,7 @@ public class DataDonationPlatform {
         post(API.ADMIN_STUDY_PARTICIPANTS, new AdminCreateStudyParticipantRoute(), jsonSerializer);
         post(API.ADMIN_STUDY_INVITATION_LOOKUP, new AdminLookupInvitationRoute(), jsonSerializer);
         post(API.ADMIN_STUDY_INVITATION_DETAILS, new AdminUpdateInvitationDetailsRoute(), jsonSerializer);
+        post(API.ADMIN_STUDY_USER_LOGIN_ACCOUNT, new AdminCreateUserLoginAccountRoute(), jsonSerializer);
 
         // These filters work in a tandem:
         // - StudyLanguageResolutionFilter figures out and sets the user language in the attribute store
@@ -339,6 +341,8 @@ public class DataDonationPlatform {
                 .addTempUserWhitelist(HttpMethod.patch, API.USER_ACTIVITY_ANSWERS)
                 .addTempUserWhitelist(HttpMethod.put, API.USER_ACTIVITY_ANSWERS)
         );
+        patch(API.UPDATE_USER_PASSWORD, new UpdateUserPasswordRoute(), responseSerializer);
+        patch(API.UPDATE_USER_EMAIL, new UpdateUserEmailRoute(), responseSerializer);
 
         // Governed participant routes
         get(API.USER_STUDY_PARTICIPANTS, new GetGovernedStudyParticipantsRoute(), responseSerializer);
@@ -464,17 +468,6 @@ public class DataDonationPlatform {
 
         // Routes calling DSM
         get(API.PARTICIPANT_STATUS, new GetDsmParticipantStatusRoute(new DsmClient(cfg)), responseSerializer);
-
-        patch(
-                API.UPDATE_USER_PASSWORD,
-                new UpdateUserPasswordRoute(),
-                responseSerializer
-        );
-        patch(
-                API.UPDATE_USER_EMAIL,
-                new UpdateUserEmailRoute(),
-                responseSerializer
-        );
 
         boolean runScheduler = cfg.getBoolean(ConfigFile.RUN_SCHEDULER);
         if (runScheduler) {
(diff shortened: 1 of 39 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
```diff
@@ -0,0 +1,178 @@
...
+import spark.Response;
+
+public class AdminCreateUserLoginAccountRoute extends ValidatedJsonInputRoute<CreateUserLoginAccountPayload> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AdminCreateUserLoginAccountRoute.class);
+
+    @Override
+    protected Class<CreateUserLoginAccountPayload> getTargetClass(Request request) {
+        return CreateUserLoginAccountPayload.class;
+    }
```
MocanaAtBroad, line 46: I think this is the default already.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            var userDao = handle.attach(UserDao.class);
+            User user = userDao.findUserByGuid(userGuid).orElse(null);
+            if (user == null) {
+                String msg = "Could not find user with guid " + userGuid;
+                LOG.warn(msg);
+                throw ResponseUtil.haltError(HttpStatus.SC_NOT_FOUND, new ApiError(ErrorCodes.NOT_FOUND, msg));
+            }
+
```
MocanaAtBroad, line 78: We need some code so we don't have to write these 20+ lines every again.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
```diff
@@ -0,0 +1,178 @@
...
+                throw ResponseUtil.haltError(HttpStatus.SC_BAD_REQUEST,
+                        new ApiError(ErrorCodes.OPERATION_NOT_ALLOWED, "User already has login account"));
+            } else if (user.isTemporary()) {
+                LOG.error("Attempted to create login account for temporary user, userGuid={}, studyGuid={}", userGuid, studyGuid);
+                throw ResponseUtil.haltError(HttpStatus.SC_BAD_REQUEST,
+                        new ApiError(ErrorCodes.OPERATION_NOT_ALLOWED, "User is a temporary user"));
+            }
+
+            Auth0ManagementClient auth0Mgmt = createManagementClient(handle, studyGuid);
+            String auth0ClientId = payload.getAuth0ClientId();
```
MocanaAtBroad: Hmm. Should client be dictating what the Auth0 client id? This user is authorized already on this study and and server know what the client id. Maybe we don't need this coming from client?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
```diff
@@ -0,0 +1,178 @@
...
+            // Create auth0 user
+            String email = payload.getEmail();
+            String randomPassword = GuidUtils.randomPassword();
+            var userResult = auth0Mgmt.createAuth0User(dbConn.getName(), email, randomPassword);
+            if (userResult.hasThrown() || userResult.hasError()) {
+                Exception e = userResult.hasThrown() ? userResult.getThrown() : userResult.getError();
+                LOG.error("Error creating new auth0 account for user {}", user.getGuid(), e);
+                throw ResponseUtil.haltError(HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                        new ApiError(ErrorCodes.SERVER_ERROR, "Error setting up login account"));
+            } else {
```
MocanaAtBroad: Consider finding out what a bad password error looks like and including a a couple of retries. Gonna happen that Auth0 will not like one of the passwords that have been autogenerated.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: I had issue with the default not working when trying to use spy in test, so overridden it.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
```diff
@@ -0,0 +1,178 @@
...
+            var ticketResult = auth0Mgmt.createPasswordResetTicket(auth0User.getId(), payload.getRedirectUrl());
+            if (ticketResult.hasThrown() || ticketResult.hasError()) {
+                Exception e = ticketResult.hasThrown() ? ticketResult.getThrown() : ticketResult.getError();
+                LOG.error("Error creating password reset ticket for user {}", user.getGuid(), e);
+                throw ResponseUtil.haltError(HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                        new ApiError(ErrorCodes.SERVER_ERROR, "Error setting up login account"));
+            } else {
+                LOG.info("Created password reset ticket for user {}", user.getGuid());
+            }
+
```
MocanaAtBroad: Consider moving some of the Auth0 interactions to separate component. This is a lot to do from a request handler. Will be easier to test and/or mock too.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: > user is authorized already

By "user" do you mean the study admin? This `auth0ClientId` is different. This is the client to _sign up the user with_. Whenever a user signs up in Auth0, they are associated with an auth0 database connected to the client. If we're using separate clients in `testboston` (e.g. one for CRC and one for regular app), then I imagine we won't want the users to be in the CRC client's database connection, if it even has one. I'm not sure how we'll figure it out in the backend, so leaving it up to the caller.

MocanaAtBroad: Not understanding how client would have a better idea what auth0 client id to use to sign up a subject on a study than the server. I was expecting that there would only be one valid Auth0 client that could be used to signup subjects. Perhaps this API call is meant to extend to admin signups ? Would expect that in that case too there would only be one valid value. Not seeing it, but maybe I am missing an important detail here.

yufengwng: No, this is not meant for admin signups. Perhaps I'm misunderstanding things, but here's some of my notes:

* A study may have multiple clients. E.g. one for Angular, one for mobile, one for admin interface, etc.
* A client _does not_ have to have a database connection, e.g. only social.
* A database connection is required to sign up a user with email/password.
* The client used to call this API (aka the client that authenticated the operator who is calling this API) might not be the one we want to use to sign up users.

Given what I know, I'm not sure how the server will figure out what client to use. Can it be an arbitrary choice? Perhaps we need more study configuration to inform the server. Any help or suggestion here is appreciated.

Also, my intent here is that for `testboston`, the CRC client calling this will pass in the client id of the regular app client because that's the one where other invited users are signing up with. Maybe this is asking too much of frontend clients. I'll figure something else out then.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: I have tried the letters in the alphabet used here, so I know those work. I don't see any settings in Auth0 password policies that will restrict what letters you can use, so I think the alphabet is good. The only issues I see here is if a study had set a minimum length is that greater than what we use here, or we somehow generate password with 2+ same consecutive letters AND study has that setting turned on. In those cases, I see retries will definitely help (for min length, probably need to figure out a way to bump up the length).

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: Okay, added some code to infer client in the backend for now. We can make it more flexible by adding study configuration later. The request just takes `email` now.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: Added same checks and retries which should be sufficient for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AdminCreateUserLoginAccountRoute.java
yufengwng: Refactored a substantial potion, can refactor more later.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -339,6 +341,8 @@ private static void start() {
                 .addTempUserWhitelist(HttpMethod.patch, API.USER_ACTIVITY_ANSWERS)
                 .addTempUserWhitelist(HttpMethod.put, API.USER_ACTIVITY_ANSWERS)
         );
+        patch(API.UPDATE_USER_PASSWORD, new UpdateUserPasswordRoute(), responseSerializer);
```
MocanaAtBroad, line 344: Yikes! Why does this route exist? Cannot see any references from Angular.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/study/PasswordPolicy.java
```diff
@@ -91,6 +92,73 @@ public Integer getMaxRepeatedChars() {
...
+            }
+            int len = j - curr;
+            if (len > maxLen) {
+                maxLen = len;
+            }
+            curr = j; // Can skip over contiguous sequence since any further counts will be less.
+        }
+        return maxLen;
+    }
+
```
MocanaAtBroad, line 161: OK. I guess if you really really want to make sure you have a valid password.

MocanaAtBroad (approved):
Was looking to see if we record or log somewhere who created the account for the the user with email X . I see in once spot we log that that we are about to create an account with user with guid xx, but would rather not have to look up guid, lookup auth0 id, go to Auth0 and look at email. I think just a log entry would be fine at this point.
Could not totally follow the logic over to housekeeping. Looks like we are just have a new email type with a new attribute, but not sure?

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
yufengwng: I believe it's used in the mobile app.

yufengwng:
Re: logging. Yep, there's no place to store who created login account for user right now. We'll need some database changes if we want that down the line.

Re: housekeeping. There's not much going on there. We add a new event type here so study can hook into that. For example, it should configure an email to send out the password reset link. The email part is the same old housekeeping stuff we have been doing.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
bskinner: Yep- this a features used by the Android and iOS apps.

PR #212, opened 2020-05-20
DDP 4361 profile birthdate

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4361
opted for birth_date to be string rather than LocalDate just to be safe


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1067,12 +1067,18 @@ ProfileModel:
       birthYear:
         type: integer
         example: 1970
+        deprecated: true
```
yufengwng, line 1071: Nice!

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1067,12 +1067,18 @@ ProfileModel:
...
         type: integer
         example: 1
+        deprecated: true
       birthDayInMonth:
         type: integer
         example: 1
+        deprecated: true
+      birthDate:
+        type: string
+        example: 2019-10-19
```
yufengwng: We can also add `format: date` here. ([docs](https://swagger.io/docs/specification/data-models/data-types/#string))

pepper-apis/src/main/java/org/broadinstitute/ddp/route/AddProfileRoute.java
```diff
@@ -66,7 +68,7 @@ public Object handle(Request request, Response response, Profile profile) throws
                             .setFirstName(profile.getFirstName())
                             .setLastName(profile.getLastName())
                             .setSexType(sexType)
-                            .setBirthDate(profile.getBirthDate())
+                            .setBirthDate(profile.getBirthDate() != null ? LocalDate.parse(profile.getBirthDate()) : null)
```
yufengwng, line 72: Since this is a string, I think we should wrap `parse()` in try/catch and return `400` if it's an invalid date.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/ProfileRouteTest.java
```diff
@@ -35,6 +37,8 @@
     private static final Integer birthMonth = 3;
     private static final Integer birthDayInMonth = 15;
     private static final Integer birthYear = 1995;
+    private static final LocalDate birthDate = LocalDate.of(1995, Month.MARCH, 15);
+    private static final String invalidBirthDate = "1999-29-09"; //yyyy-mm-dd
```
bskinner: Is this format string the expected format of the date, or the format of the birth date string?

pepper-apis/src/main/java/org/broadinstitute/ddp/json/Profile.java
```diff
@@ -33,10 +36,13 @@
     private String firstName;
     @SerializedName(LAST_NAME)
     private String lastName;
+    @SerializedName(BIRTH_DATE)
+    private String birthDate;
 
     public Profile(UserProfile other) {
         LocalDate birthDate = other.getBirthDate();
         UserProfile.SexType sexType = other.getSexType();
+        this.birthDate = birthDate != null ? birthDate.toString() : null;
```
bskinner, line 45: Is there a reason why we're converting the date to a string instead of storing a reference to the source object and performing the conversion as needed?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchProfileRoute.java
```diff
@@ -56,10 +56,8 @@ public Object handle(Request request, Response response) throws Exception {
         boolean providedLastName = json.has(Profile.LAST_NAME);
         String lastName = providedLastName ? payload.getLastName() : null;
 
-        boolean providedBirthDate = json.has(Profile.BIRTH_YEAR)
-                && json.has(Profile.BIRTH_MONTH)
-                && json.has(Profile.BIRTH_DAY_IN_MONTH);
-        LocalDate birthDate = providedBirthDate ? parseBirthDate(payload) : null;
+        boolean providedBirthDate = json.has(Profile.BIRTH_DATE);
```
bskinner, line 60: What's going on here? The old properties were marked as deprecated, not removed. We should still be supporting the old method.

bskinner (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchProfileRoute.java
ssettipalli: hmm. 
I guess I need to accept both types (3 date elements or full date) and act accordingly. 
Profile.BIRTH_DATE trumps deprecated elements. 
will make needed changes

pepper-apis/src/main/java/org/broadinstitute/ddp/json/Profile.java
ssettipalli: I am doing that to Serialize and deserialize. 
@SerializedName(BIRTH_DATE)
private String birthDate;

pepper-apis/src/test/java/org/broadinstitute/ddp/route/ProfileRouteTest.java
ssettipalli: I just mentioned it there as format of the date; 29 invalid month

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchProfileRoute.java
ssettipalli: updated to be backward compatible to deprecated elements. 
added a test case too.

yufengwng (approved):
Looks reasonable.

bskinner (approved):
Looks good!

PR #219, opened 2020-05-26
DDP-4285. activity status in activity detail api update

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4285

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -110,7 +110,7 @@ Activity:
     - title
     - subtitle
     - readonly
-    - status
+    - statusCode
     - listStyleHint
     - sections
     - lastUpdatedText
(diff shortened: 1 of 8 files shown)


yufengwng (approved):

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -127,8 +127,9 @@ Activity:
     guid:
       type: string
       description: unique identifier for this instance of the activity
-    status:
-      type: string
+    statusCode:
```
ssettipalli, line 130: @fat23cat 
Pavel: This is change to user activities API/Route . 
If front-end is using "status" field, we need a front-end change to be compatible with this backend change. 
This is not summary, individual activity instance details.
"/user/%s/studies/%s/activities/%s" call that gets user activity details.

bskinner (approved):
Looks good 👍

pepper-apis/docs/specification/src/components/schemas.yml
fat23cat: Frontend doesn't use that field.

PR #223, opened 2020-05-26
testboston: create both covid-survey and address activity

yufengwng:
Update based on latest Slack conversations.

### DIFF
--- a/study-builder/studies/testboston/study-events.conf
+++ b/study-builder/studies/testboston/study-events.conf
@@ -2,6 +2,38 @@
   # TODO: more to come
   "events": [
 
+    # activity creation
+    {
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
+        "activityCode": ${id.act.consent},
+        "statusType": "COMPLETE"
+      },
+      "action": {
+        "type": "ACTIVITY_INSTANCE_CREATION",
+        "activityCode": ${id.act.covid_survey}
+      },
+      "cancelExpr": """user.studies["testboston"].forms["COVID_SURVEY"].hasInstance()""",
+      "maxOccurrencesPerUser": 1,
+      "dispatchToHousekeeping": false,
+      "order": 1
+    },
+    {
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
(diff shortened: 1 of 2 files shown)


study-builder/studies/testboston/study-events.conf
```diff
@@ -2,6 +2,38 @@
...
+    # activity creation
+    {
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
+        "activityCode": ${id.act.consent},
+        "statusType": "COMPLETE"
+      },
+      "action": {
+        "type": "ACTIVITY_INSTANCE_CREATION",
+        "activityCode": ${id.act.covid_survey}
```
ssettipalli, line 14: just curious: What's the benefit of making activityCode value substitution variable rather than having actual value like "COVID_SURVEY" here !!!
cancelExpr in next line has activityCode value rather than substitution variable any way !!

ssettipalli (approved):
looks good to me

study-builder/studies/testboston/study-events.conf
yufengwng: Short answer: benefit is we don't have to hardcode things and it's in one place.

Long answer:
* One benefit is that all the relevant identifiers for the study (e.g. study guid, activity codes, question stable ids, etc) are in one place. I placed them in `sub.conf` file so we can use them throughout the various config files. Since they are in one place, if we ever need to change these ids or lookup these ids, there's one place to look.
* Another benefit is catching typos. If we used a regular string here, it's just a string. But if we use variable substitution and you made a typo, the substitution will fail and we catch it before we try inserting stuff into database.
* In fact, I actually wanted to use them everywhere, including in PEX expressions. However, HOCON and Typesafe Config library does not support variable substitutions within strings themselves, so I ended up hardcoding the ids in the PEX. You can still do it, but you end up having to do a whole bunch of double-quote escaping. Take a look below. Imagine having to do that for multi-line string... it will be awful and unwieldy.
    ```
    "user.studies[\""${id.study}"\"].forms[\""${id.act.covid_survey}"\"].hasInstance()"
    ```

PR #229, opened 2020-05-28
DDP-4338. Fix UserAuthCheckFilter responses

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4338

Changes can be considered high impact.
Please let me know if you prefer error messages to be different.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
@@ -135,7 +135,7 @@ public class UserAuthCheckFilter implements Filter {
         }
 
         if (!canAccess) {
-            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "Invalid temporary user");
+            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "user is not authorized");
             throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, apiError);
         }
     }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
@@ -122,24 +122,20 @@ public class UserAuthCheckFilter implements Filter {
                 handle.attach(JdbiUser.class).findByUserGuid(tempUserGuid));
 
         boolean canAccess = true;
-        String message = null;
         if (tempUser == null) {
             LOG.warn("Could not find temporary user with guid '{}'", tempUserGuid);
-            message = "Could not find temporary user with guid: " + tempUserGuid;
             canAccess = false;
         } else if (!tempUser.isTemporary()) {
             LOG.error("User with guid '{}' is not a temporary user but is used to access"
                     + " temp-user whitelisted path without a token", tempUserGuid);
-            message = "User is not a temporary user";
             canAccess = false;
         } else if (tempUser.isExpired()) {
             LOG.warn("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
-            message = "Temporary user passed expiration time";
             canAccess = false;
         }
 
         if (!canAccess) {
-            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, message);
+            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED,"Invalid temporary user");
             throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, apiError);
         }
     }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
@@ -133,7 +133,7 @@ public class UserAuthCheckFilter implements Filter {
             message = "User is not a temporary user";
             canAccess = false;
         } else if (tempUser.isExpired()) {
-            LOG.error("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
+            LOG.warn("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
             message = "Temporary user passed expiration time";
             canAccess = false;
         }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
@@ -135,7 +135,7 @@ public class UserAuthCheckFilter implements Filter {
         }
 
         if (!canAccess) {
-            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED,"Invalid temporary user");
+            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "Invalid temporary user");
             throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, apiError);
         }
     }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
@@ -1,16 +1,16 @@
 package org.broadinstitute.ddp.filter;
 
-import static spark.Spark.halt;
-
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.http.HttpStatus;
 import org.broadinstitute.ddp.constants.ErrorCodes;
 import org.broadinstitute.ddp.constants.RouteConstants.PathParam;
 import org.broadinstitute.ddp.db.TransactionWrapper;
 import org.broadinstitute.ddp.db.dao.JdbiUser;
 import org.broadinstitute.ddp.db.dto.UserDto;
 import org.broadinstitute.ddp.exception.DDPException;
+import org.broadinstitute.ddp.json.errors.ApiError;
 import org.broadinstitute.ddp.security.DDPAuth;
 import org.broadinstitute.ddp.util.ResponseUtil;
 import org.broadinstitute.ddp.util.RouteUtil;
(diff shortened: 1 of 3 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -115,27 +114,33 @@ private void handleTemporaryUser(Request request) {
 
         if (!inWhitelist) {
             LOG.warn("Request '{} {}' is not in temp-user whitelist", requestedMethod, requestedPath);
-            throw halt(401);
+            throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED,
+                    new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "Request is not in temp-user whitelist"));
```
ssettipalli, line 118: Do we prefer error code to be "OPERATION_NOT_ALLOWED" rather than AUTH_CANNOT_BE_DETERMINED !!

bskinner: I'd lean towards `OPERATION_NOT_ALLOWED` (or some variant of that). It's not that the authentication can't be verified, the user is simply not permitted.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -115,27 +114,33 @@ private void handleTemporaryUser(Request request) {
...
             LOG.warn("Could not find temporary user with guid '{}'", tempUserGuid);
+            message = "Could not find temporary user with guid: " + tempUserGuid;
             canAccess = false;
         } else if (!tempUser.isTemporary()) {
             LOG.error("User with guid '{}' is not a temporary user but is used to access"
                     + " temp-user whitelisted path without a token", tempUserGuid);
+            message = "User is not a temporary user";
             canAccess = false;
         } else if (tempUser.isExpired()) {
             LOG.error("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
```
bskinner: A nice-to-have, but while you're working on this, let's think about downgrading this from `error` to `info` or something so we don't flood Slack. A client passing in an expired token shouldn't be an exceptional event.

bskinner (dismissed):
👍 with a note

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -115,27 +114,33 @@ private void handleTemporaryUser(Request request) {
...
+            message = "Temporary user passed expiration time";
             canAccess = false;
         }
 
         if (!canAccess) {
-            throw halt(401);
+            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, message);
+            throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, apiError);
```
yufengwng, line 139: I'm still not too sure about why we need to add a body to `401`. Correct me if I'm wrong, but I recall that the point of just returning a `401` is so that we don't leak any information. If we're really going to add a body, I suggest making it as general as possible, e.g.
```
{
  "code": "UNAUTHORIZED",
  "message": "Unauthorized"
}
```

yufengwng (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
yufengwng: Maybe I'm being too paranoid, but can we make all the scenarios return a general body like this one?
```
{
  "code": "UNAUTHORIZED",
  "message": "Unauthorized"
}
```

yufengwng: Happy to be convinced otherwise though.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
ssettipalli: Probably should have been discussed while sizing and approving this ticket. 

I have seen three types of patterns in API's:
1. halt(401) no body
2. Return Error (just code) object in Response 
3. Return ApiError (both code & message) object in Response 
From the API tickets in this sprint , looks like we are moving towards # 3

Should we try to keep things consistent OR ok to use all 3 ways depending on the API. (Probably process meeting topic !)
If we want to implement yufengs above suggestions, we probably don't even need to use ApiError object , just return code if we are keeping the message same as code OR again for consistency use ApiError !!
We need to make sure we are not including any info that can lead to information leak/security concern while returning ApiError with message.

bskinner: Have a comment regarding the overarching error structure coming, but in the meantime- should we be using a `401 UNAUTHORIZED` in the auth filter? A `403 FORBIDDEN` might be better since a `401 UNAUTHORIZED` implies that it can be fixed if the client performs the proper authorization.

bskinner: Returning a response body isn't necessarily leaking- in this case, it's just following the pattern for our platform of including error details. I like returning a more vague `ApiError`, and logging more verbose details to the back end, but we still need to include some details so the caller can figure out what's up without access to the server. What if we restrict all client-facing errors in this filter to:

_if the client ID is not stored, enabled, whitelisted, etc- just say it's not authorized._
```json
{
    "code": "UNAUTHORIZED",
    "message": "client $CLIENT_ID_FROM_REQUEST is not authorized"
}
```

_the client ID cleared whatever checks we implemented, but access to the study is restricted for one reason or another (it's disabled, doesn't exist, etc)_
```json
{
    "code": "UNAUTHORIZED",
    "message": "study $STUDY_ID_FROM_REQUEST is not authorized"
}
```

_this could easily be used for temporary users as well, since we have a GUID. No need to treat them any differently in the eyes of the client :) _
```json
{
    "code": "UNAUTHORIZED",
    "message": "user $USER_ID_FROM_REQUEST is not authorized"
}
```

yufengwng: @ssettipalli (2) is outdated and is in the process of being removed in favor to (3). 

Regarding the `ApiError` message, what will a client do with that message? Why not just return a generic message in all cases?

bskinner: At least in the sense of the iOS client, those errors tend to be more for the developer and application logging than the end user. The messaging isn't consistent or useful enough to present to the user in a non-confusing manner, so   the client tries its best to take action based on the result, and lets the user know when it has no idea what to do. In this case, presenting a user with a message reading simply "unauthorized" is frustrating and not helpful, and the developer doesn't have any information to figure out what they need to do in order to move to the next step.

ssettipalli: Probably returned message should be more meaningful & useful rather than same or same meaning in a sentence as code

yufengwng: I had been under impression that we don't want to return any additional information other than a `401` so we don't give attackers any chances. I recall that we mentioned that if we returned some message about user or study, then you can infer that this particular user/study exists.

Can someone check in with the rest of the team about this change?

bskinner: It should, but we still need to define who it's supposed to be useful & meaningful **to**. I feel we should be targeting the developers with most of our errors- popping up an alert to the user is a worst-case scenario when handling errors- it's extremely disruptive and removes the user from the context of what they're doing.

ssettipalli: I will flag this ticket/PR for now and we can discuss during next process meeting or post standup !!!

bskinner: The issues arises when there are different responses depending on the user's state. In this case, we also need to balance that with the utility of the message itself. I don't feel that these messages share an unreasonable amount of information- aren't saying whether the user (or client) exists or not, just that whatever GUID was passed in isn't authorized.

If the client sending request only gets part of the request correct, given the length of the identifiers, it's far more likely that it's going to be a developer attempting to get their client working than an attacker.

> I had been under impression that we don't want to return any additional information other than a `401` so we don't give attackers any chances. I recall that we mentioned that if we returned some message about user or study, then you can infer that this particular user/study exists.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -115,27 +114,33 @@ private void handleTemporaryUser(Request request) {
...
             canAccess = false;
         } else if (!tempUser.isTemporary()) {
             LOG.error("User with guid '{}' is not a temporary user but is used to access"
                     + " temp-user whitelisted path without a token", tempUserGuid);
+            message = "User is not a temporary user";
             canAccess = false;
         } else if (tempUser.isExpired()) {
-            LOG.error("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
+            LOG.warn("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
+            message = "Temporary user passed expiration time";
```
yufengwng: If I'm understanding, the proposal is to return a generic message about the _class_ of problem and not anything specific. Perhaps we should change these to return something like `...invalid temporary user..` instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -130,12 +130,13 @@ private void handleTemporaryUser(Request request) {
...
             canAccess = false;
         } else if (tempUser.isExpired()) {
-            LOG.error("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
+            LOG.warn("Temporary user with guid '{}' had already expired at time {}ms", tempUserGuid, tempUser.getExpiresAtMillis());
             canAccess = false;
         }
 
         if (!canAccess) {
-            throw halt(401);
+            ApiError apiError = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "Invalid temporary user");
```
yufengwng: I'm still a little hesitant about this. Note that this auth filter is used for BOTH auth users and temp users. This temp user check is a different code path and you do not need an auth token to access. Client ID and study guid is effectively public information. And our API docs are public as well. It won't be hard to get client id, study guid, and API path correct. So now, only thing you need to guess is the temp user guid. And it just so happens that we tell callers that "invalid temporary user". Whereas previously we don't say anything at all.

How about we return a payload in the regular code path but nothing in the temp user code path?

ssettipalli: @bskinner Is that ok with you ?

yufengwng: Or return a generic message like "Unauthenticated".

bskinner: Why are we telling anyone it's a temp user? It's divulging privileged information- the user GUID is public knowledge, the type of the user is not. In this case, using a different status code for temporary users is **also** sharing privileged information.

if there's something wrong with the user GUID, i'd just say return "user $GUID is not authorized" or "user is not authorized". No need to treat it any differently. It's got a user GUID, it's a type of user.

yufengwng: That works too. @ssettipalli can you make that change?

yufengwng (approved):

PR #231, opened 2020-05-28
DDP-4703 support special vars in activity templating

yufengwng:
## Context

Add support for substituting "special vars" in templates, so we can pull in participant's name into activity content.

THIS DOES NOT SUPPORT ALL FORMS OF TEMPLATING. Only templates within an activity are supported, and only if they hook into the bulk rendering. This should cover major items like content blocks, question prompts and embedded components.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nTemplateConstants.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nTemplateConstants.java
@@ -2,4 +2,7 @@ package org.broadinstitute.ddp.content;
 
 public class I18nTemplateConstants {
     public static final String LAST_UPDATED_DATE_TEMPLATE_VAR_NAME = "LAST_UPDATED";
+    public static final String PARTICIPANT_FIRST_NAME = "DDP_PARTICIPANT_FIRST_NAME";
+    public static final String PARTICIPANT_LAST_NAME = "DDP_PARTICIPANT_LAST_NAME";
+    public static final String DASHED_DATE = "DDP_DASHED_DATE";
 }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormInstance.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormInstance.java
@@ -20,13 +20,16 @@ import org.broadinstitute.ddp.content.HtmlConverter;
 import org.broadinstitute.ddp.content.I18nContentRenderer;
 import org.broadinstitute.ddp.content.I18nTemplateConstants;
 import org.broadinstitute.ddp.content.Renderable;
+import org.broadinstitute.ddp.db.dao.UserProfileDao;
 import org.broadinstitute.ddp.exception.DDPException;
 import org.broadinstitute.ddp.model.activity.types.ActivityType;
 import org.broadinstitute.ddp.model.activity.types.BlockType;
 import org.broadinstitute.ddp.model.activity.types.FormType;
 import org.broadinstitute.ddp.model.activity.types.ListStyleHint;
+import org.broadinstitute.ddp.model.user.UserProfile;
 import org.broadinstitute.ddp.pex.PexException;
 import org.broadinstitute.ddp.pex.PexInterpreter;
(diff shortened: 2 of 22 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nTemplateConstants.java
```diff
@@ -2,4 +2,7 @@
 
 public class I18nTemplateConstants {
     public static final String LAST_UPDATED_DATE_TEMPLATE_VAR_NAME = "LAST_UPDATED";
+    public static final String PARTICIPANT_FIRST_NAME = "DDP_PARTICIPANT_FIRST_NAME";
+    public static final String PARTICIPANT_LAST_NAME = "DDP_PARTICIPANT_LAST_NAME";
+    public static final String DASHED_DATE = "DDP_DASHED_DATE";
```
yufengwng, line 7: Here are the new special vars we can drop into templates.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormInstance.java
```diff
@@ -220,6 +225,26 @@ public void renderContent(Handle handle, I18nContentRenderer renderer, long lang
...
+                context.put(I18nTemplateConstants.PARTICIPANT_FIRST_NAME, profile.getFirstName());
+            }
+            if (profile.getLastName() != null) {
+                context.put(I18nTemplateConstants.PARTICIPANT_LAST_NAME, profile.getLastName());
+            }
+        }
+
+        return context;
+    }
+
```
yufengwng, line 247: Here is where we fill in values for those special vars.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/MailingAddressComponent.java
```diff
@@ -1,42 +1,77 @@
...
+            registry.accept(titleTemplateId);
+        }
+        if (subtitleTemplateId != null) {
+            registry.accept(subtitleTemplateId);
+        }
     }
 
     @Override
     public void applyRenderedTemplates(Provider<String> rendered, ContentStyle style) {
-        // Implement this method once we decide to use bulk rendering for this component
```
yufengwng, line 40: Updated mailing-address component so it hooks into the bulk rendering process and can take advantage of those special vars.

study-builder/studies/testboston/snippets/consent-section4.conf
```diff
@@ -95,6 +95,29 @@
...
+          <div>
+            <h3>Last Name</h3>
+            <p>$DDP_PARTICIPANT_LAST_NAME</p>
+          </div>
+          <div>
+            <h3>Date</h3>
+            <p>$DDP_DASHED_DATE</p>
+          </div>
+        """,
+        "variables": []
```
yufengwng, line 116: Here is an example of activity definition using those special vars. Note that no variable objects are added to the `variables` list for the special vars.

charlottedye (approved):

MocanaAtBroad (approved):
Looks good. Does precisely what it needs to do. Nice!

MocanaAtBroad:
Ok. Do we need to add templates (internationalized and with support for variables) in a bunch of other places? A crazy idea:  Could we look at centralizing the rendering of template strings and making the rendering annotation-driven? So in the case of the MailAddressComponent (not a good example because you have that inner class but let me try anyways, you would do something like:
```java
public class MailingAddressComponent extends FormComponent {
    @SerializedName("titleText")
    @Template("titleTemplateId")
    private String titleText;
    @SerializedName("subtitleText")
    @Template("subtitleTemplateId")
    private String subtitleText;
}
When time comes to serialize the object (perhaps from a route, perhaps, from the DAO), you calls something like:
```java
Renderer renderer = new Renderer(context); // context will know about the study, the participant the operator, etc
TemplateRenderer.render(mailAddressComponent);
```
And it would do all the template look ups and replacements.
So would help keep these model/like objects more like POJOs, not have to add special methods to each to do replacements, etc.

Sergei-Poliakov (approved):

PR #232, opened 2020-05-29
DDP-4353 Composite child answer api fix

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4353

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiCompositeQuestion.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiCompositeQuestion.java
@@ -68,6 +68,10 @@ public interface JdbiCompositeQuestion extends SqlObject {
     @SqlQuery("queryCompositeQuestionIdByChildQuestionStableId")
     Optional<Long> findParentQuestionIdByChildQuestionStableId(@Bind("childQuestionStableId") String childQuestionStableId);
 
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryCompositeQuestionIdByChildQuestionId")
+    Optional<Long> findParentQuestionIdByChildQuestionId(@Bind("childQuestionId") long childQuestionId);
+
     class RowReducer implements LinkedHashMapRowReducer<Long, CompositeQuestionDto> {
         @Override
         public void accumulate(Map<Long, CompositeQuestionDto> map, RowView rowView) {
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiCompositeQuestion.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiCompositeQuestion.java
@@ -64,6 +64,10 @@ public interface JdbiCompositeQuestion extends SqlObject {
     @UseRowReducer(RowReducer.class)
     Optional<CompositeQuestionDto> findParentDtoByChildQuestionId(@Bind("childQuestionId") long childQuestionId);
 
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryCompositeQuestionIdByChildQuestionStableId")
+    Optional<Long> findParentQuestionIdByChildQuestionStableId(@Bind("childQuestionStableId") String childQuestionStableId);
+
     class RowReducer implements LinkedHashMapRowReducer<Long, CompositeQuestionDto> {
         @Override
         public void accumulate(Map<Long, CompositeQuestionDto> map, RowView rowView) {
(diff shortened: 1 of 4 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiCompositeQuestion.java
```diff
@@ -64,6 +64,14 @@ void insertChild(@Bind("parentQuestionId") long parentQuestionId, @Bind("childQu
     @UseRowReducer(RowReducer.class)
     Optional<CompositeQuestionDto> findParentDtoByChildQuestionId(@Bind("childQuestionId") long childQuestionId);
 
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryCompositeQuestionIdByChildQuestionStableId")
+    Optional<Long> findParentQuestionIdByChildQuestionStableId(@Bind("childQuestionStableId") String childQuestionStableId);
```
ssettipalli, line 69: not used. used it initially and changed to other new method. left it as additional convenience method for any future use.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -175,6 +175,16 @@ public Object handle(Request request, Response response) {
...
                             instanceGuid, false, languageCodeId);
 
+                    //validation to check if question is a composite child
+                    Optional<Long> parentQuestionId = handle.attach(JdbiCompositeQuestion.class)
+                            .findParentQuestionIdByChildQuestionId(question.getQuestionId());
+                    if (parentQuestionId.isPresent()) {
+                        LOG.warn("Passed question stable ID : " + questionStableId + " is a Composite child question. "
+                                + "Only entire Composite question answer can be updated ");
+                        throw ResponseUtil.haltError(response, HttpStatus.SC_NOT_FOUND, new ApiError(
+                                ErrorCodes.QUESTION_REQUIREMENTS_NOT_MET, "Only entire Composite question answer can be updated"));
```
ssettipalli: ErrorCodes.QUESTION_REQUIREMENTS_NOT_MET or any other constant like: OPERATION_NOT_PERMITTED or something else ?
Also return code: 404 or something else !!

yufengwng: I think `OPERATION_NOT_PERMITTED` makes sense here. How about we go with `400`?

bskinner: I agree with `OPERATION_NOT_PERMITTED`, but I think a `404` would be better suited. There isn't a client error with the request, the child question just doesn't exit from that perspective.

bskinner (dismissed):
👍

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
yufengwng: If we returned `404`, I will probably expect to see something like `... hey, we can't find that question!` But instead here we're saying that whatever the client is trying to do is not permitted and that a composite question answer needs to be updated as a whole, so I'll lean more towards `400`.

ssettipalli: I can average it and make it 402 😁

ssettipalli: In the ticket Implementation sketch, Marco suggested 404 😺

bskinner: a `400` is a bad request- there is nothing semantically, or syntactically, wrong with sending an updated answer for a child question. From the perspective of a client, that question does not exist- it's only exists as a child of some parent.

yufengwng: If folks are vouching for `404`, then let's use `404`. In that case, may I suggest returning `QUESTION_NOT_FOUND` instead of `OPERATION_NOT_PERMITTED`?

ssettipalli: @bskinner Are you ok with return error code: QUESTION_NOT_FOUND ?

bskinner: That works great with me!

ssettipalli: One final look please

yufengwng (approved):

bskinner (approved):

PR #234, opened 2020-05-29
testboston: new activity icons

yufengwng:
New activity instance status icons for `testboston`. I got these from Pavel, which I assume came from Erin.


MocanaAtBroad (approved):
Had not caught on that we were serving these from back-end server. Wouldn't it be simpler in API to say what the state is and delegate to client figure out which icons to use for that purpose?
Not asking for a change. This where we are, but if there is a simple reason why this better, please fill me in.

yufengwng:
No, this is not better but a relic of legacy choices. One of the tech debt items is to NOT serve these icons from backend.

bskinner:
I am on-board with this idea!

> Had not caught on that we were serving these from back-end server. Wouldn't it be simpler in API to say what the state is and delegate to client figure out which icons to use for that purpose?
> Not asking for a change. This where we are, but if there is a simple reason why this better, please fill me in.

fat23cat:
Or, we can store icons in the Google Bucket and API can return the icon link. We do the same for angio, brain, mbc consent steps.

fat23cat (approved):

yufengwng:
Thanks! They're all good points. But I'm sorry I don't have bandwidth for that right now.

PR #236, opened 2020-05-29
Enable local registration for studies in other tenants

yufengwng:
1. When we do local registration, `auth0Domain` is not required and backend will figure it out. However, we're hardcoding it to whatever value we have in our config file. In most cases, this is the `ddp-dev.auth0.com` tenant. Other studies, like `testboston`, has a different tenant so local registration will fail. I changed things so we look for the tenant using the study instead.
2. Added support for custom fields in local registration. Auth0 custom fields are stored in user metadata, and the only way to get at it is using Auth0's Management API. I don't think we want to have to use Management API in frontend local registration flow, so I added support to read user metadata in backend.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
@@ -103,14 +103,20 @@ public class UserRegistrationRoute extends ValidatedJsonInputRoute<UserRegistrat
 
         return TransactionWrapper.withTxn(handle -> {
             auth0UserId.set(payload.getAuth0UserId());
-            StudyDto study = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
+            String auth0Domain = payload.getAuth0Domain();
+
+            StudyDto study = null;
+            if (doLocalRegistration && StringUtils.isBlank(auth0Domain)) {
+                study = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
+            } else {
+                study = handle.attach(JdbiUmbrellaStudy.class).findByDomainAndStudyGuid(auth0Domain, studyGuid);
+            }
             if (study == null) {
                 ApiError err = new ApiError(ErrorCodes.STUDY_NOT_FOUND, "Could not find study with guid " + studyGuid);
                 LOG.warn(err.getMessage());
                 throw ResponseUtil.haltError(response, HttpStatus.SC_NOT_FOUND, err);
             }
 
-            String auth0Domain = payload.getAuth0Domain();
             if (doLocalRegistration && StringUtils.isBlank(auth0Domain)) {
                 auth0Domain = handle.attach(JdbiAuth0Tenant.class).findById(study.getAuth0TenantId())
                         .map(Auth0TenantDto::getDomain)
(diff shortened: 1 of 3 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -130,6 +140,19 @@ public Object handle(Request request, Response response, UserRegistrationPayload
                         auth0ClientSecret, payload.getRedirectUri());
                 auth0UserId.set(Auth0Util.getVerifiedAuth0UserId(auth0RefreshResponse.getIdToken(), auth0Domain));
                 LOG.info("Successfully exchanged auth0 code for auth0UserId {} for local registration", auth0UserId);
+
+                // Look for potential user metadata.
+                var auth0User = auth0Util.getAuth0User(auth0UserId.get(), mgmtClient.getToken());
```
ssettipalli, line 145: I believe we should be checking if (payload.getFirstName() == null || payload.getLastName() == null) before making Auth0 call to avoid unnecessary Auth0 call if this data is already passed in payload.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -102,18 +104,26 @@ public Object handle(Request request, Response response, UserRegistrationPayload
...
-
-            StudyDto study = handle.attach(JdbiUmbrellaStudy.class).findByDomainAndStudyGuid(auth0Domain, studyGuid);
             if (study == null) {
                 ApiError err = new ApiError(ErrorCodes.STUDY_NOT_FOUND, "Could not find study with guid " + studyGuid);
                 LOG.warn(err.getMessage());
                 throw ResponseUtil.haltError(response, HttpStatus.SC_NOT_FOUND, err);
             }
 
+            if (doLocalRegistration && StringUtils.isBlank(auth0Domain)) {
+                auth0Domain = handle.attach(JdbiAuth0Tenant.class).findById(study.getAuth0TenantId())
```
ssettipalli, line 121: As per current database design/constraints, I don't believe queried auth0Domain can be null but to be future proof/extra cautious should we check for null and throw exception ?

yufengwng: I think auth0Domain is pretty crucial to our current model, so I don't think it will be made null. Also, this is only in local registration codepath so I'm not too worried about that.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: Name is unlikely to be passed in current local registration flows, but I'll add the check nonetheless.

yufengwng: Actually, I'm deciding not to add the check. In local registration flow, we're almost always not going to get name in the payload, so we're almost always making the call and the check becomes irrelevant.

yufengwng: If we do end up passing first/last name in the payload, there's going to be other (real production) code changes required as well. Otherwise there will be a disparity between real registration vs local regstration. I don't see that happening anytime soon.

ssettipalli (approved):

PR #244, opened 2020-06-04
Cleanup hidden answers on submit

yufengwng:
## Context

This is related to bug DDP-4518 and TestBoston.

When questions are hidden due to conditional toggling, we're currently not doing anything with those hidden answers. This means a few things
* When user reloads the activity and _retoggle_ the hidden questions, they will see their old answers. Whether this is confusing to end-users is debatable.
* Hidden answers are being exported to Elasticsearch since there's no other filtering on these.
* The exported answers are visible in DSM via Data Browser. Whether this is confusing to study staff is debatable.

I think the issue is exacerbated when you have conditional questions in a "write-once" activity, like the Consent we have in TestBoston. Once you have submitted it, you pretty much can't change it afterwards. Imagine a user is on the Consent and flipping back-and-forth between conditional questions (e.g. filling in both self signature and staff signature). In data export, we'll see all of these answers even those some of them are supposedly hidden. Since it's write-once, you can't reload the activity and edit out / "erase" your other answers.

One proposal is to cleanup hidden answers upon submit. Intuitively, when user submits, they're saying they are done. So we go through the activity and delete all the hidden answers. Since this is only on submit, one drawback is that while the user _hasn't_ submitted (e.g. they're still editing / in progress) you will continue to see a mix of answers in data export / Data Browser. But nice thing here is it plays a little nicer with "auto-save" feature (e.g. if user reloads, the hidden answers are still there).

NOTE: since hidden answers are only deleted on submit, current hidden answers in the system will not be deleted, unless user happens to log back in and submits again.

## Checklist

## FUD Score

## Testing

## Release


MocanaAtBroad (dismissed):
I think this is the right thing to do. Counting on user consumers downstream to properly do this filtering does not make sense to me. Bound to introduce errors in data analysis.

ssettipalli:
I would vote for deleting hidden answers during submission rather than during patch'ing because user is still going back and forth and we don't want to delete user entered data before user submits the data.

yufengwng:
@ssettipalli This PR is indeed deleting hidden answers during submission (aka the PUT Answers route).

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormInstance.java
```diff
@@ -337,4 +339,42 @@ public int setDisplayNumbers() {
...
+        }
+        return hidden;
+    }
+
+    private List<Answer> collectAnswers(FormBlock block) {
+        if (block.getBlockType() == BlockType.QUESTION) {
+            return ((QuestionBlock) block).getQuestion().getAnswers();
+        } else if (block.getBlockType() == BlockType.CONDITIONAL) {
+            return ((ConditionalBlock) block).getControl().getAnswers();
+        } else {
```
MocanaAtBroad, line 376: Do you have to worry BlockType.GROUP here?

yufengwng: Yup, because a `group` block may contain a nested `question` block.

ssettipalli (approved):
looks good to me

PR #246, opened 2020-06-05
Fix client sql query

yufengwng:
Missed a table alias for a column in a sql query.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
@@ -132,7 +132,7 @@ public interface JdbiClient extends SqlObject {
             @Bind("auth0TenantId") long auth0TenantId
     );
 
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiClient.java
```diff
@@ -132,7 +132,7 @@ int deleteByAuth0ClientIdAndAuth0TenantId(
             @Bind("auth0TenantId") long auth0TenantId
     );
 
-    @SqlQuery("SELECT c.*, (select auth0_domain from auth0_tenant as t where t.auth0_tenant_id = c.auth0_tenant_id) as auth0_domain"
+    @SqlQuery("SELECT c.*, (select t.auth0_domain from auth0_tenant as t where t.auth0_tenant_id = c.auth0_tenant_id) as auth0_domain"
```
ssettipalli, line 135: since column: autho_domain doesn't exist in both the tables, alias is not a must. 
Its little cleaner to have it though

ssettipalli (approved):

PR #233, opened 2020-05-29
DDP-1855 allow multiple templates for notification action

yufengwng:
## Context

Allow studies to configure multiple email templates for an email notification event action, so that we can sent out email in a different language. When we build the email message in Housekeeping, we will determine the appropriate email to use based on user's preferred language and study's supported languages.

The database migration script will take care of migrating existing study configurations to this new schema, so no additional intervention needed. Study-builder is also updated to support the old format of a single email template, so old config files do not need to be updated.

## Checklist

- [x] I have labeled the type of changes involved using the `C-*` labels.
- [x] I have assessed potential risks and labeled using the `R-*` labels.
- [x] I have considered error handling and alerts, and added `L-*` labels as needed.
- [x] I have considered security and privacy, and added `I-*` labels as needed
- [x] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

_Overall, how are you feeling about these changes?_

- [x] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [x] They are user-visible in dev as a regular user journey and require no additional instructions.
- [ ] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [x] I have written automated positive tests
- [ ] I have written automated negative tests
- [x] I have tried existing studies locally and emails seem to work
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [x] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #249, opened 2020-06-08
DDP-4706 dynamic email templates support

ssettipalli:
look at @https://broadinstitute.atlassian.net/browse/DDP-4706

schema change to track if its dynamic template or legacy template.
change related sql, dao, dto, Message and in EmailNotificationHandler if its dynamic template build dynamic data else for legacy: do substitutions.


pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -199,7 +237,9 @@ boolean messageShouldBeIgnored(Handle apisHandle, String studyGuid, String parti
         subs.put(DDP_PARTICIPANT_LAST_NAME, message.getParticipantLastName());
 
         for (NotificationTemplateSubstitutionDto sub : message.getTemplateSubstitutions()) {
-            subs.put(sub.getVariableName(), sub.getValue());
+            if (sub.getVariableName().startsWith("-")) {
```
ssettipalli: if we think we need to support legacy substitution variables that doesn't follow : "-ddp." , we can remove this check

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/NotificationTemplateVariables.java
```diff
@@ -77,4 +77,21 @@
      * The invitation guid
      */
     public static final String DDP_INVITATION_ID = "-ddp.invitationId-";
+
+    //Dynamic template variables
+    public static final String PARTICIPANT_FIRST_NAME = "participant_firstName";
```
yufengwng, line 82: Maybe group these together so we don't get confused as to what is what?
```
public static final class Dynamic {
  ... PARTICIPANT_FIRST_NAME ...
  ... PARTICIPANT_LAST_NAME ...
}

File: pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/PubSubMessageBuilder.java [hunk 3]
```diff
@@ -188,6 +188,7 @@ public PubsubMessage createMessage(String ddpMessageId,
                         queuedNotificationDto.getNotificationType(),
                         queuedNotificationDto.getNotificationServiceType(),
                         queuedNotificationDto.getTemplateKey(),
+                        queuedNotificationDto.isDynamicTemplate(),
```
yufengwng: After this is updated with PR #233, this flag should come from `NotificationTemplate`.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -149,6 +162,31 @@ public void handleMessage(NotificationMessage message) {
...
+        dynamicData.put(SALUTATION, salutation);
+        dynamicData.put(BASE_WEB_URL, message.getWebBaseUrl());
+        dynamicData.put(STUDY_GUID, message.getStudyGuid());
+        dynamicData.put(PARTICIPANT_GUID, message.getParticipantGuid());
+        dynamicData.put(PARTICIPANT_FIRST_NAME, message.getParticipantFirstName());
+        dynamicData.put(PARTICIPANT_LAST_NAME, message.getParticipantLastName());
+
+        for (NotificationTemplateSubstitutionDto sub : message.getTemplateSubstitutions()) {
+            if (!sub.getVariableName().startsWith("-")) {
+                dynamicData.put(sub.getVariableName(), sub.getValue());
```
yufengwng: I don't think this will do what we want? The queued event message will have template substitutions like `-ddp.link-` that needs to go in the dynamic template as well.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -125,16 +125,20 @@ private long insertEventTrigger(Handle handle, Config triggerCfg) {
     private long insertEventAction(Handle handle, Config actionCfg) {
         EventActionDao actionDao = handle.attach(EventActionDao.class);
         String type = actionCfg.getString("type");
+        boolean isDynamicTemplate = false;
+        if (actionCfg.hasPath("is_dynamic")) {
+            isDynamicTemplate = actionCfg.getBoolean("is_dynamic");
```
yufengwng: Nice! I like how this is backwards-compatible. Let's also update the `study-example.conf` file as well to document this.

yufengwng (commented):
Looking good!

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventActionSql.java
```diff
@@ -73,7 +74,13 @@ int insertUserNotificationAction(
     default long findOrInsertNotificationTemplateId(String templateKey, String langCode) {
         return findNotificationTemplate(templateKey, langCode)
                 .map(NotificationTemplate::getId)
-                .orElseGet(() -> insertNotificationTemplate(templateKey, langCode));
+                .orElseGet(() -> insertNotificationTemplate(templateKey, langCode, false));
```
yufengwng: Maybe call `findOrInsertNotificationTemplateId(templateKey, langCode, false)`?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/NotificationTemplate.java
```diff
@@ -37,4 +40,8 @@ public long getLanguageId() {
     public String getLanguageCode() {
         return languageCode;
     }
+
+    public boolean isDynamicTemplate() {
```
yufengwng: This one can probably be called `isDynamic`.

pepper-apis/src/test/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandlerTest.java
```diff
@@ -152,19 +152,19 @@ public void testMessageShouldBeIgnored_byDoNotContactPreference() {
     public void testBuildSubstitutions() {
         var msg = new NotificationMessage(
                 NotificationType.EMAIL, NotificationServiceType.SENDGRID,
-                "template", List.of("to@ddp.org"), "first", "last", "guid",
+                "template", false, List.of("to@ddp.org"), "first", "last", "guid",
                 "study", "pepper", "from@ddp.org", "key", "salutation",
-                List.of(new NotificationTemplateSubstitutionDto("foo", "bar"),
-                        new NotificationTemplateSubstitutionDto("alice", "bob")),
+                List.of(new NotificationTemplateSubstitutionDto("-ddp.foo", "bar"),
+                        new NotificationTemplateSubstitutionDto("-ddp.alice", "bob")),
```
yufengwng: How about using a bit more proper syntax here? `-ddp.foo-`

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -149,6 +162,29 @@ public void handleMessage(NotificationMessage message) {
...
+
+        dynamicData.put(SALUTATION, salutation);
+        dynamicData.put(BASE_WEB_URL, message.getWebBaseUrl());
+        dynamicData.put(STUDY_GUID, message.getStudyGuid());
+        dynamicData.put(PARTICIPANT_GUID, message.getParticipantGuid());
+        dynamicData.put(PARTICIPANT_FIRST_NAME, message.getParticipantFirstName());
+        dynamicData.put(PARTICIPANT_LAST_NAME, message.getParticipantLastName());
+
+        for (NotificationTemplateSubstitutionDto sub : message.getTemplateSubstitutions()) {
+            dynamicData.put(sub.getVariableName(), sub.getValue());
```
yufengwng: This is still not quite right to me...

Imagine the `message` will have template substitutions like so:
```
[
  { "name": "-ddp.link-", "value": "..." }
]
```
If we just put that directly into `dynamicData`, then we end up with a data object like:
```
{
  "participant_firstName": "...",
  "participant_lastName": "...",
  "-ddp.link-": "...",
  ...
}
```
which doesn't look right...

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/NotificationTemplateVariables.java
ssettipalli: just wanted to keep it like it was in legacy.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -149,6 +162,29 @@ public void handleMessage(NotificationMessage message) {
         }
     }
 
+    private Map<String, Object> getDynamicData(NotificationMessage message) {
+        Map<String, Object> dynamicData = new HashMap<>();
```
yufengwng, line 166: Probably not the best, but here's an approach that will give us a nice structured dynamic data object and support all the existing legacy template variables. Idea is to use Typesafe Config's support for "object paths".
```java
Map<String, Object> map = ConfigFactory.empty()
        .withValue("ddp.study.guid", ConfigValueFactory.fromAnyRef("testboston"))
        .withValue("ddp.baseWebUrl", ConfigValueFactory.fromAnyRef("http://localhost:4200"))
        .withValue("ddp.participant.firstName", ConfigValueFactory.fromAnyRef("foo"))
        .withValue("ddp.participant.lastName", ConfigValueFactory.fromAnyRef("bar"))
        .withValue("ddp.link", ConfigValueFactory.fromAnyRef("ticket"))
        .root().unwrapped().get("ddp");
```
1. First, we build an empty config object.
2. We fill it with values. For the keys, we basically take the template variable name and strip out the dashes. We can even strip out the `ddp.` prefix as well.
3. The last line there will convert the config object to a map object.

It's a bit of a hack and abuse of the config library, but the support for "dotted paths" makes it easy to build the dynamic data object. And in this way, we don't need to add a new set of variable names in `NotificationTemplateVariables.java`.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Once we have dynamic templates  may be clients (invoking code) can start setting new variables like "link" into dynamic template {{link}} right ? 
OR is the expectation that existing client code / invoking code will only set existing legacy template variables like "-ddp.link-" ? If so probably strip "-ddp." and "-" which doesn't look clean but works.
I think its cleaner to enhance invoking code to start setting dynamic variables like "link" rather than "-ddp.link-" once we start using dynamic templates !! May be for backward compatibility set both when we get there  !!

ssettipalli: OR invest time in migrating legacy templates to dynamic to be future proof and convert client code to set variables like "link" !!!

yufengwng: Just so we're on the same page, what is "client" here? The backend server is the one setting these variables. For example, `ActivityStatusEventDao` sets the `-ddp.activityInstanceGuid-` variable.

ssettipalli: yes I meant backend server, which is the send mail invoking code/setting the substitutions.

ssettipalli: Any additional variables set that don't exist in the template will be ignored.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Tried to keep it simple and  handle any new future substitution variables opted for String parsing.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Tried to keep it simple and to handle any new future substitution variables opted for String parsing.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Tested with WELCOME_TEST dynamic template that I created.

yufengwng (approved):
LGTM

PR #251, opened 2020-06-08
Generate config file for Pepper-backend GAE deployment

MocanaAtBroad:
## Context

Setting different parameters for prod deployment of pepper backend.
**Update June 9, 2020**
I added the GCP profiling agent code to be loaded in dev deployments. 
I think this just ought to be a permanent change.

And out of scope, but doing it anyways: setting a max-age on OPTION calls.

## Checklist

- [ X] I have labeled the type of changes involved using the `C-*` labels.
- [X ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance boXttlenecks and written a brief summary in this PR.
- [X ] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.

## FUD Score

- [ X] :relaxed: All good, business as usual!

See if we get more performance complaints.

## Testing
N/A

## Release


pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -0,0 +1,25 @@
...
+{{if eq $environment "prod"}}
+basic_scaling:
+  max_instances: 6
+  idle_timeout: 45m
+{{else}}
+basic_scaling:
+  max_instances: 4
+  idle_timeout: 20m
+{{end}}
+
```
MocanaAtBroad, line 15: More conservative and generous settings for prod than in other environments.
Stretched out the idle_timeout (we were using 10m before) as even during a workday it appear/disappear multiple times a day with just 10min.

.circleci/config.yml
```diff
@@ -270,13 +273,14 @@ commands:
...
             mkdir -p $DEPLOY_DIR
             cd $DEPLOY_DIR
             cp -r ../$TCELL_DIR .
             cp ../output-config/tcell_agent.config $TCELL_DIR
             cp ../target/<<parameters.jar_base_name>>.jar .
-            cp ../src/appengine/<<parameters.jar_base_name>>.yaml .
+            cp "$DEPLOY_CONFIG_FILE_DIR/<<parameters.jar_base_name>>.yaml" .
             cp ../output-config/application.conf .
       - auth-gcp-service-account
       - run:
```
MocanaAtBroad, line 286: This command is only used for back-end and housekeeping. Housekeeping appears to be fine.

yufengwng (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -590,6 +590,9 @@ private static void enableCORS(final String origin, final String methods, final
             if (accessControlRequestMethod != null) {
                 response.header("Access-Control-Allow-Methods", accessControlRequestMethod);
             }
+            if (accessControlRequestMethod != null || accessControlRequestHeaders != null) {
+                response.header("Access-Control-Max-Age", "172800");
+            }
```
MocanaAtBroad, line 601: This is an overdue change! Will prevent a separate OPTION call for before everything  single PATCH, GET, POST etc. There is max value that varies from browser to browser. Apparently for Firefox it maxes out at 24hours. Chromimum browsers max out in shorter periods. I went for 2 days here.

pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -0,0 +1,29 @@
...
+network:
+  instance_tag: study-server # necessary for firewall rules so that DSM will accept https requests from pepper
+{{if and (eq $environment "dev") (eq $gae "true") }}
+entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=pepper-backend,-cprof_cpu_use_per_thread_timers=true -Xmx1640m -Xmx1640m -jar DataDonationPlatform.jar
+{{else}}
+entrypoint: java -javaagent:tcell/tcellagent.jar -Xmx1640m -Xmx1640m -jar DataDonationPlatform.jar
```
MocanaAtBroad: Noted that I added the agent to collect profiler data. I am inclined to make this change permanent for dev deployments until we have a better profiling option.

pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -0,0 +1,29 @@
...
+{{else}}
+basic_scaling:
+  max_instances: 4
+  idle_timeout: 20m
+{{end}}
+
+network:
+  instance_tag: study-server # necessary for firewall rules so that DSM will accept https requests from pepper
+{{if and (eq $environment "dev") (eq $gae "true") }}
+entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=pepper-backend,-cprof_cpu_use_per_thread_timers=true -Xmx1640m -Xmx1640m -jar DataDonationPlatform.jar
```
yufengwng: Did we mean to use `-Xms` for the first non-standard option? Sounds like `-Xms` controls the memory pool initial size, and `-Xmx` the maximum size. Is our intent to set the same value for these two?

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
yufengwng: Nice!

pepper-apis/config/DataDonationPlatform.yaml.ctmpl
MocanaAtBroad: That was a typo! Good catch! Had same mistake in the the housekeeping setup as well.
b5fd7888
07caf5df

yufengwng (approved):

PR #257, opened 2020-06-12
DDP-4378. Add dynamic templates support to StudyBuilder

ssettipalli:
look at @https://broadinstitute.atlassian.net/browse/DDP-4738

### DIFF
--- a/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EmailBuilder.java
+++ b/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EmailBuilder.java
@@ -100,8 +100,9 @@ public class EmailBuilder {
         }
         String subject = emailCfg.getString("subject");
         String name = emailCfg.getString("name");
+        boolean isDynamic = emailCfg.getBoolean("isDynamic");
 
-        var templateResult = sendGrid.createTemplate(name);
+        var templateResult = sendGrid.createTemplate(name, isDynamic);
         templateResult.rethrowIfThrown(e -> new DDPException("Error while creating email " + key, e));
         if (templateResult.hasError()) {
             LOG.error("Error while creating email {}: {}", key, templateResult.getError());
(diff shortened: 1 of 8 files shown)


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EmailBuilder.java
```diff
@@ -100,8 +100,9 @@ private EmailInfo createSendGridEmail(Config emailCfg) {
         }
         String subject = emailCfg.getString("subject");
         String name = emailCfg.getString("name");
+        boolean isDynamic = emailCfg.getBoolean("isDynamic");
```
ssettipalli, line 103: wondered if we wanted to do:
boolean isDynamic = false;
if (emailCfg.hasPath("isDynamic")) {
    isDynamic = emailCfg.getBoolean("isDynamic");
}

This way we don't need to update all existing studies email configs with isDefault = false, but may be its little more cleaner having the property set explicitly rather than defaulting.

yufengwng: I think I will have preferred defaulting, but since you have already made the changes let's just keep it the way you have it.

study-builder/studies/angio/sendgrid_emails.conf
```diff
@@ -5,133 +5,152 @@
...
+      "render": true,
+      "isDynamic": false
     },
     {
       "key": "resendEmail",
       "name": "ANGIO_RESEND_EMAIL",
       "subject": "Link to Your Current Form",
       "filepath": "emails/resend_email.html",
-      "render": true
+      "render": true,,
```
yufengwng: Extra comma. Also in the last object as well.

study-builder/studies/angio/sendgrid_emails.conf
```diff
@@ -5,133 +5,152 @@
       "name": "ANGIO_BLOOD_SENT_REMINDER",
       "subject": "Angiosarcoma Project Check-In",
       "filepath": "emails/blood_sent_reminder.html",
-      "render": true
+      "render": true,
+      "isDynamic": false
```
yufengwng, line 9: Can you also update the `study-example.conf` file and add in this new flag?

yufengwng (approved):

PR #258, opened 2020-06-12
DDP-4737 tooltip support

yufengwng:
## Context

Did a few things:

* Updated database schema to store tooltips.
* Added tooltip to questions.
* Added tooltip to picklist options.
* Update `testboston` study to use tooltips in the Covid Survey.

Most of the changes are just plumbing to pass the tooltip through the various layers/classes/tests.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1113,6 +1115,16 @@ ProfileModel:
       lastName:
         type: string
         example: Barium
+
```
yufengwng, line 1119: CC @fat23cat @bskinner @Sergei-Poliakov 

Question tooltip:
Screen Shot 2020-06-12 at 3 03 47 PM

Picklist tooltip:
Screen Shot 2020-06-12 at 3 04 07 PM

study-builder/studies/testboston/snippets/question-contact-diagnosed.conf
```diff
@@ -13,18 +13,26 @@
...
-        "translations": [
-          {
-            "language": "en",
-            "text": ${i18n.en.contact_diagnosed_tip}
-          }
-        ]
       }
     ]
   },
+  "tooltip": {
```
yufengwng: Here is how we use it for testboston.

pepper-apis/src/main/resources/db-changes/schema/DDP-4737-tooltip.xml
```diff
@@ -0,0 +1,40 @@
...
+                <constraints primaryKey="true" primaryKeyName="tooltip_pk"/>
+            </column>
+            <column name="text_template_id" type="bigint">
+                <constraints nullable="false"
+                             references="template(template_id)"
+                             foreignKeyName="tooltip_text_template_fk"/>
+            </column>
+        </createTable>
+    </changeSet>
+
```
MocanaAtBroad: Do we need a new table for this? Could it just simply be a new column called tooltip_text_template_id in the tables that need it?

MocanaAtBroad (dismissed):
Had a question about the schema, but it all looks good.

pepper-apis/src/main/resources/db-changes/schema/DDP-4737-tooltip.xml
yufengwng: I'm not sure if just a text template will suffice in representing tooltips. After a recent Slack convo about A-T's tooltip requirement, it seems like we don't need anything more. But I'm suspecting we might want other things related to tooltips down the line, so making it a table to give us room to grow. I can simplify this if folks feel that I'm overthinking this.

MocanaAtBroad: I can't think of anything else that we would want. Some logic on when it is shown? Support for HTML? My suggestion is if you are not aware of specific very likely requirement that would necessitate additional properties, we should get rid of it. But it is your call.

yufengwng: From what I can tell, other aspects related to tooltips are:
* Positioning, e.g. left, right, top, bottom, etc
* Timing, e.g. show after a delay of 1.5 secs, keep it on-screen for another second, etc
* Most recommend to keep tooltip as plain text, but perhaps we might want some "richer" text
* I think there are some where the tooltip shows a modal popup sort of thing, but I guess at that point it's probably not considered a "tooltip" anymore is it...

Arguably, some of these are presentation elements or browser/OS specifics, but it wouldn't be the first time we added presentation elements to the model... Also, in particular, I will like to hear from @fat23cat on implementation of tooltips on Angular side and if we need anything more (e.g. icon url).

MocanaAtBroad: Will express hope we don't put any of the stuff on the server, except maybe "richer" text. Maybe. I don't think it is worth dedicating a new table to holding the one foreign key. But your call.

yufengwng: Thanks for the feedback! Since this is the second time folks had questioned the design, I'll go ahead and simplify it so we have a new column instead of a new table, and change API to have just new string property instead of object.

bskinner: I'm with @MocanaAtBroad on keeping the presentation options on the front-end. As for making the item be a `string` instead of an `object`- it sounds good to me!

MocanaAtBroad (dismissed):
Nice!

fat23cat:
Folks, if everything is good, can we merge PR? I want to check frontend side.

bskinner (approved):
:+1:

PR #262, opened 2020-06-16
DDP-4744 track user timezone

yufengwng:
## Context

This PR enables us to start tracking user's timezone. This value is set when user is registered and is provided by client. If not provided, no timezone is set in profile. If timezone is there, server will use it when rendering the `$ddp.date()` activity substitution, otherwise fallback to UTC which is the current behavior.

## Checklist

## FUD Score

## Testing

## Release


MocanaAtBroad (dismissed):

pepper-apis/docs/specification/src/components/request-bodies.yml
```diff
@@ -65,6 +65,20 @@ RegisterRequestBody:
...
+            type: string
+            description: user's last name, if available
+          languageCode:
+            type: string
+            description: user's preferred language, as an ISO code
+            example: en
+          timeZone:
+            type: string
+            description: user's time zone, as an IANA time zone region identifier
+            example: America/New_York
```
yufengwng, line 81: @fat23cat @bskinner

MocanaAtBroad (dismissed):

pepper-apis/docs/specification/src/components/request-bodies.yml
```diff
@@ -65,6 +65,20 @@ RegisterRequestBody:
...
             description: if present, server will associate the presumably account-less user with the auth0UserId after verifying the invitation
+          firstName:
+            type: string
+            description: user's first name, if available
+          lastName:
+            type: string
+            description: user's last name, if available
+          languageCode:
+            type: string
+            description: user's preferred language, as an ISO code
```
bskinner: This needs additional details. What type of ISO code is allowed here? Is it `ISO 639-1`? Are `ISO 639-2` and `ISO 639-3` codes allowed as well? Are we using some other ISO standard?

pepper-apis/src/main/java/org/broadinstitute/ddp/json/Profile.java
```diff
@@ -50,6 +50,7 @@ public Profile(UserProfile other) {
         this.preferredLanguage = other.getPreferredLangCode();
         this.firstName = other.getFirstName();
         this.lastName = other.getLastName();
+        // NOTE: timezone is currently not exposed in Profile API JSON.
```
bskinner, line 53: I think we need to prioritize adding this feature. It is not a safe assumption that people will always be signing up from their preferred timezone- due to travel, or a host of reasons, especially given the prevalence of mobile devices- and we're going to need a way to update it.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -531,12 +549,14 @@ private void initializeProfile(Handle handle, User user, UserRegistrationPayload
         String lastName = payload.getLastName();
         lastName = StringUtils.isNotBlank(lastName) ? lastName.trim() : null;
         long languageId = determineUserLanguage(handle, payload).getId();
+        ZoneId timeZone = parseUserTimeZone(payload.getTimeZone());
```
bskinner, line 557: I'm thinking we should fail the request if the user input an invalid IANA code- I'm thinking a `400 BAD REQUEST` code.

We're expecting a specific type of data- an IANA identifier- not a generic string. I think it's ok, and a good thing for the API if we're strict here. If it's optional, `null`ing it out, or excluding the value should be fine, but if an invalid identifier is sent we need to let the client know, especially since we aren't exposing the TZ anywhere. Having a silent failure that subtly modifies server behavior, _even deterministically_, creates a very unfriendly API.

pepper-apis/docs/specification/src/components/request-bodies.yml
```diff
@@ -65,6 +65,20 @@ RegisterRequestBody:
...
+            example: en
+          timeZone:
+            type: string
+            description: user's time zone, as an IANA time zone region identifier
```
bskinner, line 80: This is a great description- clearly indicates exactly what we're looking for!

bskinner (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/json/Profile.java
yufengwng: Updating user's timezone is a topic that came up. However, I don't think we fully understand the implications yet (e.g. what is the UX? what are the APIs? when and were can user update? etc). I suppose that needs input from rest of the team. I'll add a ticket for now.

yufengwng: https://broadinstitute.atlassian.net/browse/DDP-4760

pepper-apis/docs/specification/src/components/request-bodies.yml
yufengwng: Added `ISO 639-1` to description.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
yufengwng: Added a check for it much earlier in the request.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/Profile.java
bskinner: Sounds good- I'll toss some comments into the ticket. That said, we should be more focused on the developer experience than the user one. The user never interacts directly with the API, the client software and developers do.

bskinner (approved):
Thank you for getting those changes in- Looks good to me. 👍

PR #266, opened 2020-06-18
DDP-4757: Update covid and address surveys titles

fat23cat:
Details are here - https://broadinstitute.atlassian.net/browse/DDP-4757

### DIFF
--- a/study-builder/studies/testboston/i18n/en.conf
+++ b/study-builder/studies/testboston/i18n/en.conf
@@ -394,12 +394,12 @@
 
   "covid_survey": {
     "name": "Covid Survey",
-    "title": "Please tell us about your experience with COVID-19",
+    "title": """<span>Covid-19 Survey <p class="activity-header__text">Please tell us about your experience with COVID-19</p></span>""",
   },
 
   "address": {
     "name": "Mailing Address",
-    "title": "Mailing Address",
+    "title": """<span>Mailing Address <p class="activity-header__text">Where should we send your COVID-19 Test Kit?</p></span>""",
     "address_title": "Your Mailing Address"
   },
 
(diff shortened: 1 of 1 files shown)


study-builder/studies/testboston/i18n/en.conf
```diff
@@ -394,12 +394,12 @@
...
   "covid_survey": {
     "name": "Covid Survey",
-    "title": "Please tell us about your experience with COVID-19",
+    "title": """<span>Covid-19 Survey <p class="activity-header__text">Please tell us about your experience with COVID-19</p></span>""",
   },
 
   "address": {
     "name": "Mailing Address",
-    "title": "Mailing Address",
+    "title": """<span>Mailing Address <p class="activity-header__text">Where should we send your COVID-19 Test Kit?</p></span>""",
```
yufengwng: Actually, I think we should use `subtitle` in activity definition instead.

1) Add new `subtitle` here.
2) In `covid-survey.conf`, add:
```
  "translatedSubtitles": [
    {
      "language": "en",
      "text": ${i18n.en.covid_survey.subtitle}
    }
  ],
```

fat23cat: No, it should be a single element. If we want to do it as title and subtitle and show it as on mockups we should change the Angular SDK activity component, because title and subtitle are different sections in the Angular SDK activity component. I'll replace `<p></p>` to make it more systematically correct.

fat23cat: Now, when we used `span` instead of `p` browser will be perceived title text as a whole text.

yufengwng: Based on those screenshots, I thought they looked similar to subtitles that's used in other studies. But if that doesn't fit here, then this is fine.

yufengwng (approved):

fat23cat:
please run tests

PR #269, opened 2020-06-24
Respond to GAE start call only after server is started

yufengwng:
Fix GAE start handling.


MocanaAtBroad (approved):
A reasonable and concise solution. Let's see it in action!

PR #275, opened 2020-06-25
No more liquibase at DDP startup

yufengwng:
## Context

_Issue 1: 404s_

Somehow GAE is still continuing to pass requests through to the server even thought it hasn't fully booted yet. I suspect that one of the `Spark.*` methods early in the startup process is initializing the Spark HTTP server and it's receiving the `/_ah/start` before our handler is even mounted yet. In this case, it's returning `404` for `/_ah/start` and that is enough for GAE to start passing along requests, which will mostly get `404`s as well since our other route handlers are not mounted yet. My attempt to solve this is to move the early `Spark.*` method calls and group them together so hopefully our GAE handlers get mounted as close to Spark server initialization as possible.

_Edit: On closer inspection, I think `JettyConfig.setupJetty()` has something to do with it. I think the way it played out in the old code is that 1) Jetty server is started, 2) we run liquibase which gives us a good 10-15 second window, 3) GAE `/_ah/start` call comes in within this window, 4) Jetty picks that up and returns `404`, 5) GAE starts unleashing other requests which will all `404` as well. So maybe moving the GAE route handlers closer to the Jetty setup call is the right move..._

_Issue 2: Startup Lag Time_

The last time I measured, I saw the startup time (from moment GAE started instance to getting to `ddp startup complete`) took about 30 seconds. Liquibase migrations seem to take up a large portion of that. To speed this up, I have completely removed the liquibase migration step from DDP. Instead, I move the step over to Housekeeping. I think this makes sense because liquibase migrations need only be ran once, and Housekeeping can accomplish that. Also, GAE starts/stops the DDP instance somewhat frequently, so now we're not paying for the liquibase cost. With this change, startup time gone down to around 15 seconds.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,7 @@ commands:
             develop)
               DEPLOY_ENV=dev
             ;;
-            GAE_Backend_Deployment_Config)
+            gae-start)
               DEPLOY_ENV=dev
             ;;
             test)
@@ -679,15 +679,17 @@ jobs:
       - attach_workspace:
           at: *root_path
       - setup-shared-env
+      # Deploy Housekeeping first so hopefully liquibase migrations
+      # are ran before DDP boots up.
+      - build-jar:
+          pom_file: housekeeping-pom.xml
+      - deploy-jar:
+          jar_base_name: Housekeeping
       - build-jar:
           pom_file: pom.xml
       - deploy-jar:
           jar_base_name: DataDonationPlatform
           deploy_config_file_dir: ../output-config
-      - build-jar:
-          pom_file: housekeeping-pom.xml
-      - deploy-jar:
-          jar_base_name: Housekeeping
       - update-dispatch-rules
 
   deploy-stored-jar-job:
@@ -698,15 +700,17 @@ jobs:
       - checkout:
           path: *repo_path
       - setup-shared-env
+      # Deploy Housekeeping first so hopefully liquibase migrations
+      # are ran before DDP boots up.
+      - retrieve-jar:
+          jar_base_name: Housekeeping
+      - deploy-jar:
+          jar_base_name: Housekeeping
       - retrieve-jar:
           jar_base_name: DataDonationPlatform
       - deploy-jar:
           jar_base_name: DataDonationPlatform
           deploy_config_file_dir: ../output-config
-      - retrieve-jar:
-          jar_base_name: Housekeeping
-      - deploy-jar:
-          jar_base_name: Housekeeping
       - update-dispatch-rules
 
   deploy-docs-job:
@@ -746,7 +750,7 @@ workflows:
             branches:
               only:
                 - develop
-                - GAE_Backend_Deployment_Config
+                - gae-start
                 - &rc /^rc.*/
                 - &hotfix /^hotfix.*/
       - parallel-compile-and-test-job:
@@ -760,7 +764,7 @@ workflows:
               only:
                 - develop
                 - scaling_experiment
-                - GAE_Backend_Deployment_Config
+                - gae-start
           requires:
             - parallel-compile-and-test-job
             - compile-and-run-test-suite-job
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -435,6 +435,8 @@ commands:
               circleci tests split --split-by=timings --timings-type=classname > $THIS_NODE_TESTS_PATH
 
             echo "Number of test classes circleci wants to execute on this node: $(wc -l < $THIS_NODE_TESTS_PATH)"
+            echo "The included test classes are:"
+            cat $THIS_NODE_TESTS_PATH
 
             IGNORE_CLASS_LIST_PATH="${TEST_DATA_PATH}/surefire_classnames_ignore_list"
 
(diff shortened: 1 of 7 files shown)


zyme:
Another idea for `ah/start` is at the [study-manager GAE PR](https://github.com/broadinstitute/ddp-study-manager/pull/33/files#r445805385).  Could we fold that approach in here as well?  I like moving liquibase to housekeeping,  and to deal with the remaining boot lag and GAE prematurely sending traffic to the backend, the study-manager backed approach should buy us some robustness.

.circleci/config.yml
```diff
@@ -679,15 +679,17 @@ jobs:
       - attach_workspace:
           at: *root_path
       - setup-shared-env
+      # Deploy Housekeeping first so hopefully liquibase migrations
+      # are ran before DDP boots up.
```
zyme, line 682: We should start trying to make deployments forward and backwards compatible as far as schema changes go--if we want zero downtime, there will be a window where the schema changes but an "older" version of code will be deployed.

MocanaAtBroad (commented):
1. I think this is a really good problem to tackle.
2. One thing that is impacted by moving when and where we do Liquibase migrations is our parallel test execution. To make them work properly, I had to ensure that migrations occurred before any of the tests executed: one of the tricks about the parallel tests is that CircleCi keeps stats about how long each test takes to run so that the test execution can be distributed so that each VM has more or less the same amount of work to do. e.g. if test A take 7 seconds, B  4 seconds,  C 3 seconds, and D 6, then we would schedule A and C in one process and C and D in another.
In our original setup, the time computation would assign to whatever the first test executed the liquibase migration time, throwing off completely the parallel assignments.
So just be aware that we want to make sure the migrations are not done as part of the execution of a test.
3. We should look at why this is so slow in the first place. Local migrations are much faster. Maybe our MySQL configuration is sub-optimal?

yufengwng:
@zyme I have reproduced, for the most part, the approach that study-manager took.

@MocanaAtBroad 

2. I don't know all the inner workings of how parallel tests are done, but the behavior should be the same for the most part. Before, the liquibase migrations are ran as part of booting up the server in the integration test suit. Now, the test suite is making an explicit `runServerDatabaseMigrations()` call.

3. I think it will be good to know why it's slow as well. Regardless, I don't think it's beneficial to have migrations run on every DDP startup, at least not anymore in the GAE world where our start/stop is at the whims of GAE.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -240,51 +242,27 @@ private static void start() {
...
-        threadPool(-1, -1, requestThreadTimeout);
         Config sqlConfig = ConfigFactory.load(ConfigFile.SQL_CONFIG_FILE);
         initSqlCommands(sqlConfig);
 
-        if (doLiquibase) {
-            LOG.info("Running liquibase migrations against " + dbUrl);
-            LiquibaseUtil.runLiquibase(dbUrl, TransactionWrapper.DB.APIS);
+        JettyConfig.setupJetty(preferredSourceIPHeader);
+        if (appEnginePort != null) {
+            port(Integer.parseInt(appEnginePort));
```
zyme, line 262: It might be worth moving the `port()` call up a bit--the faster we hit that call, the sooner we can block on `ah/start` and avoid premature GAE traffic.

MocanaAtBroad (dismissed):

MocanaAtBroad (commented):
Would it be super difficult to add a check to make sure the DataDonationPlatform executable is running against the "correct" schema; i.e., the last liquibase change in the same git commit as the executable has been applied to the database? I just worry about some head-scratching bugs in GAE or, more likely locally, that turns out they were simply caused by forgetting to first run housekeeping.

yufengwng:
@MocanaAtBroad I don't know what's the proper solution to ensure StudyServer is running against the "correct" schema, or how difficult it is. I agree with your point about local development. My solution for now is to introduce a new flag `doLiquibaseInStudyServer` that is only used in local development config file. This way, local development can mostly stay the same. In deployed environment, only Housekeeping will run the migrations.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -164,13 +164,11 @@ public static void setup() {
...
 
         initializeDatabase();
-
         if (!isTestServerRunning()) {
             startupTestServer();
-            insertTestData();
-        } else {
-            RouteTestUtil.setupDatabasePool();
         }
+        insertTestData();
```
yufengwng, line 171: Before, the parallel tests seem to fail consistently with either a NullPointerException or some kind of deadlock. This change seems to make parallel tests pass now, although I'm not sure of exactly why that is. I do suspect that before we weren't running this -> thus miss data -> thus null pointer. I'm still confused about the deadlock and didn't make any insights in that investigation.

MocanaAtBroad (approved):
Let's see what happens!

PR #276, opened 2020-06-26
Medical History. progress bar

Sergei-Poliakov:
## Context

[ATX-225.](https://broadinstitute.atlassian.net/browse/ATX-225)
In AT case we need to implement the Step Tracker for tracking progress during Medical History activity.
The progress bar is displayed only for Medical History activity now.
When the user clicks on the “Next” button patch request is sent.

Main approach:
- each section possesses equal '%' weight,
- the application calculates '%' on section complete (clicking 'next' button), include all the conditional pages as well.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/components/request-bodies.yml
```diff
@@ -170,3 +170,24 @@ JoinMailingListPayload:
...
+  description: to update user's last visited section
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        required:
+          - userGuid
+          - studyGuid
+          - instanceGuid
```
yufengwng: These identifiers can be taken from the path parameters, so we don't need then in the payload.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -174,6 +175,9 @@ Activity:
     isFollowup:
       type: boolean
       description: whether this activity is considered a "follow-up"
+    lastVisitedActivitySection:
```
yufengwng: Maybe name this `lastVisitedSection` to be more consistent?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -174,6 +175,9 @@ Activity:
...
+      type: integer
```
yufengwng, line 179: Is this index 1-based or 0-based? Can you add `minimum: 0` here?

pepper-apis/docs/specification/src/endpoints/user.studies.last-visited-section.yml
```diff
@@ -0,0 +1,17 @@
...
+  summary: Updates last visited section
+  description: with given user Guid, study's unique identifier and instance Guid Updates user's last visited section
+  requestBody:
+    $ref: '../pepper.yml#/components/requestBodies/UpdateUserLastVisitedSectionRequestBody'
+  responses:
+    200:
+      description: user's last visited section sucessfully updated
+    401:
+      description: missing credentials or temporary user is invalid
+    404:
```
yufengwng: Should we also have a `400` here? E.g. if someone tried to patch a `-1`? Also, these should also have documentation of response object. See file `studies.invitation-check.yml` for example of what that looks like.

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -96,6 +96,8 @@ paths:
     $ref: 'endpoints/user.studies.activity.yml'
   /user/{user}/studies/{study}/activities/{instance}/answers:
     $ref: 'endpoints/user.studies.activity.answers.yml'
+  /user/{user}/studies/{study}/activities/{instance}/lastVisitedSection:
```
yufengwng: Can we rename the path to `/last-visited-section`, aka use dash syntax instead of camel case?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchLastVisitedActivitySectionRoute.java
```diff
@@ -0,0 +1,77 @@
...
+    public PatchLastVisitedActivitySectionRoute(ActivityInstanceService actInstService) {
+        this.actInstService = actInstService;
+    }
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String userGuid = request.params(RouteConstants.PathParam.USER_GUID);
+        String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
+        String instanceGuid = request.params(RouteConstants.PathParam.INSTANCE_GUID);
+        JsonElement data = new JsonParser().parse(request.body());
```
yufengwng: I recommend subclassing from `ValidatedJsonInputRoute` instead of doing json parsing by hand. Here's the ceremonial boilerplate:

1. Create json class to represent request payload. E.g.
```java
package ...ddp.json;

public class PatchLastVisitedSectionPayload {
  @PositiveOrZero
  private int lastVisitedSection;
  ...
}
```
(Note the annotation `@PositiveOrZero` gives you verification on input for free when you use `ValidatedJsonInputRoute`.)

2. Subclass: `extends ValidatedJsonInputRoute<PatchLastVisitedSectionPayload>`.

3. Override the default:
```java
    @Override
    protected int getValidationErrorStatus() {
        return HttpStatus.SC_BAD_REQUEST;
    }
```

Now in your handler you can just do `payload.getLastVisitedSection()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchLastVisitedActivitySectionRoute.java
```diff
@@ -0,0 +1,77 @@
...
+            LOG.info("Using ddp content style {} to format activity content", style);
+
+            LOG.info("Attempting to find a translation for the following language: {}", isoLangCode);
+            Optional<ActivityInstance> inst = actInstService.getTranslatedActivity(
+                    handle, userGuid, instanceDto.getActivityType(), instanceGuid, isoLangCode, style
+            );
+
+            if (!inst.isPresent()) {
+                response.status(HttpStatus.SC_NOT_FOUND);
+            }
```
yufengwng: I don't see the point of this whole block of code. I think it's unnecessary for this route to do its work. The `RouteUtil.findAccessibleInstanceOrHalt()` call will have already determined if activity instance exists or not and return `404`.

pepper-apis/src/main/resources/db-changes/schema/add-last-visited-activity-section-column-to-activity-instance.xml
```diff
@@ -0,0 +1,22 @@
...
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+    <changeSet id="20200528-add-last-visited-activity-section-to-activity-instance" author="sergei-poliakov">
+
+        <addColumn tableName="activity_instance">
+            <column name="last_visited_section" type="int" defaultValue="0">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+
+        <update tableName="activity_instance">
```
yufengwng: I don't think this `update` is necessary. I have found that the combination of `defaultValue` and `nullable=false` (which you have above) is sufficient and will backfill existing rows appropriately.

yufengwng (commented):
Thanks for posting this PR! 🎉 

This looks like a nice little feature to have. However, I couldn't figure out the bigger picture of how this relates to progress tracker. Can you elaborate on how clients will be using this?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchLastVisitedActivitySectionRoute.java
Sergei-Poliakov: if I understand correctly, I can just write
```suggestion
        TransactionWrapper.useTxn(handle -> {
            int lastVisitedSection = dataObject.getLastVisitedSection();

            handle.attach(ActivityInstanceDao.class)
                    .updateLastVisitedSectionByInstanceGuid(instanceGuid, lastVisitedSection);
        });
```
and that's it.
btw, it means we don't need to `401` & `404` responses anymore.
Am I right?

yufengwng: I think you need the `RouteUtil.findAccessibleInstanceOrHalt()` in there as well. And we'll need `404` since that's one of the things `RouteUtil.findAccessibleInstanceOrHalt()` returns.

pepper-apis/docs/specification/src/components/request-bodies.yml
```diff
@@ -170,3 +170,15 @@ JoinMailingListPayload:
...
         $ref: '../pepper.yml#/components/schemas/JoinMailingListPayload'
+UpdateUserLastVisitedSectionRequestBody:
+  description: to update user's last visited section
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        required:
+          - lastVisitedSection
```
MocanaAtBroad: Since we don't have section identifiers, we are probably referring to the sectional ordinal index. Maybe we should call it `lastVisitedSectionIndex` and document that it starts at 0? Or 1?

MocanaAtBroad: Also, if the user is viewing, say, a four-part activity, the first time they go up to section 3.
Then the next day they come back, load the activity, and hit the next button up to section 2.
What do we want this value to show?

MocanaAtBroad: If we are going to use PATCH to URL that includes /last-visited-section, then including the lastVisitedSection property in body seem redundant.
Might be cleaner to PATCH to the activity instance URL without the /last-visited part in the path. Leave door open to PATCHing other properties in future as well.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
```diff
@@ -25,6 +25,7 @@
     private String sessionId;
     private Long firstCompletedAt;
     private boolean allowUnauthenticated;
+    private int lastVisitedSection;
```
MocanaAtBroad: With an `int` here the unitialized value I believe will be 0. This might not be what we want if 0 is the first section and no one has ever seen the the activity instance.
Perhaps `Integer` would be better to allow for NULL to be a valid value.
On serialized JSON API, we have unfortunately do not make much use of OPTIONAL properties, so we probably need to allow for this to be serialized as NULL value for activity instances  that have not been looked at all/just created.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/PatchLastVisitedSectionPayload.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.ddp.json;
+
+import javax.validation.constraints.PositiveOrZero;
+
+public class PatchLastVisitedSectionPayload {
+
+    @PositiveOrZero
+    private int lastVisitedSection;
```
MocanaAtBroad: OK! Looks like you are looking for zeros and positive values. So assuming you want Java/JavaScript-style array index position

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormInstance.java
```diff
@@ -85,7 +86,8 @@ public FormInstance(
             Long firstCompletedAt,
             Long lastUpdatedTextTemplateId,
             LocalDateTime activityDefinitionLastUpdated,
-            boolean isFollowup
+            boolean isFollowup,
+            int lastVisitedSection
```
MocanaAtBroad: See my comment above. Maybe you want to allow this to be a NULL, then you might want an Integer for type.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchLastVisitedSectionRoute.java
```diff
@@ -0,0 +1,49 @@
...
+        LOG.info("Request to update last visited section on instance {} for participant {} in study {}",
+                instanceGuid, userGuid, studyGuid);
+
+        TransactionWrapper.useTxn(handle -> {
+            RouteUtil.findAccessibleInstanceOrHalt(response, handle, userGuid, studyGuid, instanceGuid);
+
+            int lastVisitedSection = dataObject.getLastVisitedSection();
+
+            handle.attach(ActivityInstanceDao.class)
+                    .updateLastVisitedSectionByInstanceGuid(instanceGuid, lastVisitedSection);
```
MocanaAtBroad: Should we check that section actually exists?

pepper-apis/src/main/resources/db-changes/schema/add-last-visited-activity-section-column-to-activity-instance.xml
```diff
@@ -0,0 +1,17 @@
...
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+    <changeSet id="20200528-add-last-visited-activity-section-to-activity-instance" author="sergei-poliakov">
+
+        <addColumn tableName="activity_instance">
+            <column name="last_visited_section" type="int" defaultValue="0">
+                <constraints nullable="false"/>
```
MocanaAtBroad: We need to nail down semantics of 0. Never visited or visited first section only?

Sergei-Poliakov: in terms of user experience, it doesn't matter – he'll see the first section anyway

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
Sergei-Poliakov: sound reasonable. @yufengwng what do you think?

yufengwng: I tend to prefer to avoid `null` as much as possible. For a newly created activity instance that user hasn't physically "looked" at yet, having a default of 0 seems to make some sense, in my opinion. If the notion of "haven't seen it yet" is something useful to capture, then you'll want it be null-able. My opinion is to start with `int` -- it'll be easier to go from `int` to `Integer` than the other way around.

pepper-apis/docs/specification/src/components/request-bodies.yml
Sergei-Poliakov: > Since we don't have section identifiers, we are probably referring to the sectional ordinal index. Maybe we should call it `lastVisitedSectionIndex` and document that it starts at 0? Or 1?

will do

 

> Also, if the user is viewing, say, a four-part activity, the first time they go up to section 3.
> Then the next day they come back, load the activity, and hit the next button up to section 2.
> What do we want this value to show?

depends on what the client sends. In the current task, we need to track the progress of the activity, so we'll update the parameter for each step forward between sections(next button). 


> If we are going to use PATCH to URL that includes /last-visited-section, then including the lastVisitedSection property in body seem redundant.
> Might be cleaner to PATCH to the activity instance URL without the /last-visited part in the path. Leave door open to PATCHing other properties in future as well.

will change the name of the endpoint to `/user/{user}/studies/{study}/activities/{instance}/section:`

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchLastVisitedSectionRoute.java
Sergei-Poliakov: could you suggest the better way to check, please?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.yml
```diff
@@ -41,3 +41,37 @@ get:
       $ref: '../pepper.yml#/components/responses/ActivityDetailNotFoundResponse'
     default:
       $ref: '../pepper.yml#/components/responses/ErrorResponse'
+patch:
+  operationId: patchActivity
+  tags:
+    - Activities
+  summary: updates section index
+  description: with given user Guid, study's unique identifier and instance Guid Updates user's section index
```
yufengwng: Can you make these a bit more generic? E.g. `summary` should be something like `update activity instance properties`, and `description` should say something about updating properties of activity instance but that currently only section index is supported.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/PatchSectionPayload.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.ddp.json;
+
+import javax.validation.constraints.PositiveOrZero;
+
+public class PatchSectionPayload {
+
+    @PositiveOrZero
+    private int index;
+
+    public PatchSectionPayload(@PositiveOrZero int index) {
```
yufengwng: We shouldn't need the annotation in the constructor.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchActivityInstanceRoute.java
```diff
@@ -0,0 +1,60 @@
...
+
+        LOG.info("Request to update section index on instance {} for participant {} in study {}",
+                instanceGuid, userGuid, studyGuid);
+
+        TransactionWrapper.useTxn(handle -> {
+            RouteUtil.findAccessibleInstanceOrHalt(response, handle, userGuid, studyGuid, instanceGuid);
+            int sectionsSize = activityInstanceDao.getActivityInstanceSectionsSize(handle, userGuid, studyGuid, instanceGuid);
+            int index = payload.getIndex();
+
+            if (sectionsSize < index) {
```
yufengwng: Looks like an off-by-one? We might want to do `index != 0 && index >= sectionsSize`?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchActivityInstanceRoute.java
```diff
@@ -0,0 +1,60 @@
...
+                String msg = String.format("Activity %s has sections size %s less than index %s", instanceGuid, sectionsSize, index);
+                LOG.error(msg);
+                throw ResponseUtil.haltError(response, HttpStatus.SC_BAD_REQUEST, new ApiError(ErrorCodes.INVALID_REQUEST, msg));
```
yufengwng: I'll recommend just using `ErrorCodes.BAD_PAYLOAD` here.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchActivityInstanceRoute.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            handle.attach(ActivityInstanceDao.class).updateSectionIndexByInstanceGuid(instanceGuid, index);
+        });
```
yufengwng, line 51: Can we add another log statement somewhere before returning to indicate that we have finished updating to that index?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchActivityInstanceRoute.java
```diff
@@ -0,0 +1,60 @@
...
+        String userGuid = request.params(RouteConstants.PathParam.USER_GUID);
+        String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
+        String instanceGuid = request.params(RouteConstants.PathParam.INSTANCE_GUID);
+
+        LOG.info("Request to update section index on instance {} for participant {} in study {}",
+                instanceGuid, userGuid, studyGuid);
+
+        TransactionWrapper.useTxn(handle -> {
+            RouteUtil.findAccessibleInstanceOrHalt(response, handle, userGuid, studyGuid, instanceGuid);
+            int sectionsSize = activityInstanceDao.getActivityInstanceSectionsSize(handle, userGuid, studyGuid, instanceGuid);
```
yufengwng, line 40: Is there a way to find `sectionsSize` without doing another database query? I think we can grab this number from the activity definition, which should have already been cached in memory. I believe we have some code to do this in PR #282. Can you work with @MocanaAtBroad to see if we can integrate that here?

MocanaAtBroad: @yufengwng would this give us the right value? 
https://github.com/broadinstitute/ddp-study-server/blob/9b7b64da097d880e394b8568b77dda3bb8958533/pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/FormActivityDef.java#L40

yufengwng: I don't think so, because there may be an introduction or closing section (we will want to count those as well right?). We can call `getAllSections().size()` or along those lines.

yufengwng (approved):

PR #279, opened 2020-06-30
Added more logging when looking up participant  status

zyme:

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/client/DsmClient.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/client/DsmClient.java
@@ -162,6 +162,7 @@ public class DsmClient {
         String path = PATH_PARTICIPANT_STATUS
                 .replace(PathParam.STUDY_GUID, studyGuid)
                 .replace(PathParam.USER_GUID, userGuid);
+        String responseBody = null;
         try {
             String auth = RouteUtil.makeAuthBearerHeader(token);
             var request = HttpRequest.newBuilder()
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/client/DsmClient.java
```diff
@@ -172,12 +173,14 @@ private String generateToken() {
...
             if (statusCode == 200) {
-                ParticipantStatus status = new Gson().fromJson(response.body(), ParticipantStatus.class);
+                responseBody = response.body();
+                ParticipantStatus status = new Gson().fromJson(responseBody, ParticipantStatus.class);
                 return ApiResult.ok(statusCode, status);
             } else {
                 return ApiResult.err(statusCode, null);
             }
         } catch (JWTCreationException | IOException | InterruptedException | JsonSyntaxException e) {
+            LOG.error("Trouble looking up status of {} in study {}.  Response was {}", userGuid, studyGuid, responseBody);
```
bskinner, line 183: We should limit the response body to 32/64/128/... characters. That'll at least give us an idea of what's going on and logging an arbitrarily long string is dangerous if we don't have the opportunity to go back and disable this after.

bskinner (approved):
Looks good for debugging, but we should either limit the maximum size of responseBody, or make sure we go back and disable this.

pepper-apis/src/main/java/org/broadinstitute/ddp/client/DsmClient.java
zyme: We might actually need to start logging full payloads at some point.  Is your main concern OOM with big strings?  Looks like we could use some [logback magic](http://logback.qos.ch/manual/layouts.html#formatModifiers) to help us truncate.

bskinner: Memory management, general concern for management & log spelunking, and security, especially with potentially sensitive information. Not a huge issue with the drug or cancer list endpoints, but I'm not sure if the tracking info is considered sensitive or not.

PR #280, opened 2020-07-01
DDP-4772 add GetInvitations API

yufengwng:
## Context

Add new API to get list of participant's invitations in a study. Can be used by clients to find and display a participant's invite code in their dashboard.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1859,6 +1859,53 @@ AddressFieldError:
       description: the address field, where `address` means the general address instead of a single field
       example: street1
 
+#####
+#
+# Invitations
+#
+#####
+Invitation:
+  type: object
+  required:
+    - invitationType
+    - invitationId
+    - createdAt
+    - voidedAt
+    - verifiedAt
+    - acceptedAt
+    - notes
+  properties:
+    invitationType:
+      description: the type of invitation
+      type: string
+      enum:
+        - AGE_UP
+        - RECRUITMENT
+      example: RECRUITMENT
+    invitationId:
+      description: the id of the invitation
+      type: string
+    createdAt:
+      description: timestamp of when invitation was created, in ISO-8601 UTC format
+      type: string
+      example: 2020-05-13T17:58:33.887Z
+    voidedAt:
+      description: timestamp of when invitation is voided, in ISO-8601 UTC format
+      type: string
+      nullable: true
+    verifiedAt:
+      description: timestamp of when invitation is verified, in ISO-8601 UTC format
+      type: string
+      nullable: true
+    acceptedAt:
+      description: timestamp of when invitation is accepted, in ISO-8601 UTC format
+      type: string
+      nullable: true
+    notes:
+      description: the notes for the invitation
+      type: string
+      nullable: true
+
 MedicalProvider:
   type: object
   required:
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1873,7 +1873,6 @@ Invitation:
     - voidedAt
     - verifiedAt
     - acceptedAt
-    - notes
   properties:
     invitationType:
       description: the type of invitation
@@ -1890,21 +1889,20 @@ Invitation:
       type: string
       example: 2020-05-13T17:58:33.887Z
     voidedAt:
-      description: timestamp of when invitation is voided, in ISO-8601 UTC format
+      description: timestamp of when invitation was voided, making the invitation invalid, in ISO-8601 UTC format
       type: string
       nullable: true
+      example: null
     verifiedAt:
-      description: timestamp of when invitation is verified, in ISO-8601 UTC format
+      description: timestamp of when invitation was verified, indicating user has acknowledged receiving the invitation, in ISO-8601 UTC format
       type: string
       nullable: true
+      example: 2020-05-13T18:58:33.887Z
     acceptedAt:
-      description: timestamp of when invitation is accepted, in ISO-8601 UTC format
+      description: timestamp of when invitation was accepted, in ISO-8601 UTC format
       type: string
       nullable: true
-    notes:
-      description: the notes for the invitation
-      type: string
-      nullable: true
+      example: 2020-05-13T19:58:33.887Z
 
 MedicalProvider:
   type: object
--- a/pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
+++ b/pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
@@ -32,16 +32,11 @@ post:
               - type: object
                 required:
                   - userGuid
-                  - userHruid
                   - userLoginEmail
                   - notes
                 properties:
                   userGuid:
                     description: guid of associated user, if invitation is accepted
-                    type: string
-                    nullable: true
-                  userHruid:
-                    description: hruid of associated user, if invitation is accepted
                     type: string
                     nullable: true
                   userLoginEmail:
--- a/pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
+++ b/pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
@@ -27,53 +27,26 @@ post:
       content:
         application/json:
           schema:
-            type: object
-            required:
-              - invitationId
-              - createdAt
-              - voidedAt
-              - verifiedAt
-              - acceptedAt
-              - notes
-              - userGuid
-              - userHruid
-              - userLoginEmail
-            properties:
-              invitationId:
-                description: the id of the invitation
-                type: string
-              createdAt:
-                description: timestamp of when invitation was created, in ISO-8601 UTC format
-                type: string
-                example: 2020-05-13T17:58:33.887Z
-              voidedAt:
-                description: timestamp of when invitation is voided, in ISO-8601 UTC format
-                type: string
-                nullable: true
-              verifiedAt:
-                description: timestamp of when invitation is verified, in ISO-8601 UTC format
-                type: string
-                nullable: true
-              acceptedAt:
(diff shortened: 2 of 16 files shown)


pepper-apis/docs/specification/src/endpoints/user.studies.invitations.yml
```diff
@@ -0,0 +1,24 @@
+get:
+  operationId: getInvitations
```
yufengwng, line 2: CC @fat23cat @bskinner

pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
```diff
@@ -27,53 +27,26 @@ post:
...
-              userHruid:
-                description: hruid of associated user, if invitation is accepted
-                type: string
-                nullable: true
-              userLoginEmail:
-                description: login email of associated user, if invitation is accepted and user has a login account
-                type: string
-                nullable: true
+            allOf:
+              - $ref: '../pepper.yml#/components/schemas/Invitation'
```
yufengwng, line 31: In the spirit of reuse, we're now extending the `Invitation` object. This has the side-effect of pulling in a new `invitationType` property. It's not a breaking change and it shouldn't affect current callers of this API, so it should be fine.

ssettipalli (dismissed):
looks good to me

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1859,6 +1859,53 @@ AddressFieldError:
...
+      example: RECRUITMENT
+    invitationId:
+      description: the id of the invitation
+      type: string
+    createdAt:
+      description: timestamp of when invitation was created, in ISO-8601 UTC format
+      type: string
+      example: 2020-05-13T17:58:33.887Z
+    voidedAt:
+      description: timestamp of when invitation is voided, in ISO-8601 UTC format
```
bskinner: Have we defined what 'voided' means anywhere in the docs?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1859,6 +1859,53 @@ AddressFieldError:
...
+      type: string
+      nullable: true
+    verifiedAt:
+      description: timestamp of when invitation is verified, in ISO-8601 UTC format
```
bskinner: What does 'verified' mean in this context?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1859,6 +1859,53 @@ AddressFieldError:
...
+      type: string
+      nullable: true
+    acceptedAt:
+      description: timestamp of when invitation is accepted, in ISO-8601 UTC format
+      type: string
+      nullable: true
```
bskinner: Let's add some examples to these other date-formatted fields for consistency.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1859,6 +1859,53 @@ AddressFieldError:
...
+    notes:
+      description: the notes for the invitation
```
bskinner: Where are the notes coming from? Are the researcher, provided by the user, etc? What's the context here?

bskinner (commented):
The proposed API looks good to me- the iOS client does not use the APIs, so this change will not impact me at the moment. Added a couple of comments on documentation additions that would make the spec more helpful for folks outside of the dev team.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: The `notes` are meant to be a place where study staff can write arbitrary notes.

bskinner: I wasn't looking for an answer for my own knowledge, that's something that should be in the docs.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/invitation/Invitation.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.gson.annotations.SerializedName;
+import org.broadinstitute.ddp.db.dto.InvitationDto;
+import org.broadinstitute.ddp.model.invitation.InvitationType;
+
+public class Invitation {
+
+    @SerializedName("invitationType")
+    private InvitationType invitationType;
+    @SerializedName("invitationId")
+    private String invitationGuid;
```
MocanaAtBroad, line 18: We have some annotations that convert fields with value to date format.  Since we store date/timestamps in our database using different data types, I think they serve the purpose of making sure that our date serialization in our APIs is consistent. Consider using of our GSON adapters or creating a new one if we don't have one for `Instant`

yufengwng: I didn't know what/where to look for for these magical gson adapters so I just went with a `String`. I'll take a second look.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Yup, that's understood. I simply have the tendency to respond and engage in GitHub comments.

yufengwng: Thinking more about this, it doesn't seem the `notes` property will be useful or appropriate in the participant-visible API. I have removed it and only added it in the admin side of things.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/invitation/Invitation.java
yufengwng: I have added a new `InstantStringAdapter` class to handle this.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: That is even better! It seems like we might want to consider refactoring the API docs out into client-facing, and management-facing during our next maintenance period.

ssettipalli (dismissed):
looks good to me

pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
```diff
@@ -27,53 +27,31 @@ post:
...
+              - type: object
+                required:
+                  - userGuid
+                  - userHruid
```
MocanaAtBroad: Are we sure we want to expose this attribute? Don't think we use it for anything else. If we can help it, let's exclude.

pepper-apis/src/main/java/org/broadinstitute/ddp/transformers/InstantStringAdapter.java
```diff
@@ -0,0 +1,34 @@
...
+            return Instant.parse(jsonReader.nextString());
+        }
+    }
+
+    @Override
+    public void write(JsonWriter jsonWriter, Instant instant) throws IOException {
+        if (instant == null) {
+            jsonWriter.nullValue();
+        } else {
+            jsonWriter.value(instant.toString());
```
MocanaAtBroad: Could we rename to something like `InstantToIsoDateAdapter` to make intent more explicit?

MocanaAtBroad (commented):
Overall looks good. Just a couple of comments.

pepper-apis/docs/specification/src/endpoints/admin.studies.invitation-lookup.yml
yufengwng: I had thought we might need it, but turns out we don't need it. Let me remove it as part of this PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/transformers/InstantStringAdapter.java
yufengwng: The format seems to be called "iso instant format"... I guess I'll call it `InstantToIsoInstant`?

MocanaAtBroad: In my mind `Instant` is just a Java thing. I think idea is to transform to a consumable JSON format. The format we are targeting is ISO-8601, so the IsoDateAdapter suffix seemed more descriptive to me. Also aligns with the some of the other Adapters that have similar mission. But this is an opinion thing. Use your best judgement.

MocanaAtBroad (approved):
Left clarifying comment on naming of class, but regardless looks good.

pepper-apis/src/main/java/org/broadinstitute/ddp/transformers/InstantStringAdapter.java
yufengwng: For me, "iso date" isn't quite right. There is both a "date" and a "time" part. Not exactly sure what ISO-8601 calls it, maybe just "DateTime". I think "iso instant" captures the idea as well, so I'll stick with that for now.

MocanaAtBroad: Java JDK calls it ISO Date Time https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME

yufengwng: Unfortunately, the Java `Instant` class won't be able to parse the ISO Date Time format that's permitted in that URL link. It only does the `Z` timezone, as indicated by the related but different [`ISO_INSTANT` formatter](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html#ISO_INSTANT).

bskinner: I agree with @MocanaAtBroad on this- `IsoDateTimeAdaptor` is a name more apt to its use. It's converting a string to and from an ISO-8601 DateTime format. The external representation here is more important than the internal one.

PR #284, opened 2020-07-08
DDP-4792: Fixed issue where new Hebrew language code wasn't recognized.

charlottedye:
## Context

This ticket covers DDP-4792.  The StudyLanguageResolutionFilter uses Java's locale class to get a language from the locale.  Java's Locale class has some behavior where it converts a few newer language codes ("he", "yi", and "id") to older versions ("iw", "ji", and "in") to maintain compatibility with older things, despite the fact that the "newer" language codes aren't particularly new anymore.  This means that if a study is configured using the correct language code for one of these three languages, even though the new language code is provided, Java will instead substitute the older language code and when users try to see their surveys, an error will happen and prevent them from doing so.  This fixes the problem by converting the old language codes back to the newer language codes, which means it will no longer block DDP-4436.

The performance impact should be minimal.  If the initial language code works, we don't do anything different.  If it doesn't work, we try again with the newer language code, but we only do that once.

## Checklist

- [x] n/a If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [x] n/a If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

## FUD Score

## Testing
- [x] I have written an automated test but am not sure if it is a positive or negative test

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
@@ -26,6 +26,12 @@ import spark.Response;
 public class StudyLanguageResolutionFilter implements Filter {
 
     public static final String USER_LANGUAGE = "USER_LANGUAGE";
+    private static final String HEBREW_OLD = "iw";
+    private static final String HEBREW_NEW = "he";
+    private static final String YIDDISH_OLD = "ji";
+    private static final String YIDDISH_NEW = "yi";
+    private static final String INDONESIAN_OLD = "in";
+    private static final String INDONESIAN_NEW = "id";
 
     private static final Logger LOG = LoggerFactory.getLogger(StudyLanguageResolutionFilter.class);
     private static final String STUDY_GUID_REGEX = "/studies/(\\w+)";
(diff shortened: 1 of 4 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
```diff
@@ -80,6 +80,25 @@ private static Locale getPreferredLocale(Handle handle, String acceptLanguageHea
...
+
+    private static String getNewerLanguageCode(String oldCode) {
+        // Java converts some language codes to older versions.  If the language is one of these older codes, try
+        // converting it to the newer version
+        if ("iw".equals(oldCode)) {
+            return "he";
+        } else if ("ji".equals(oldCode)) {
+            return "yi";
+        } else if ("in".equals(oldCode)) {
+            return "id";
```
yufengwng: Any way to make these constants? Unlike `en` or `fr`, it's a bit difficult to make out what these are at first glance.

yufengwng (dismissed):
👍

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/StudyLanguageResolutionFilter.java
charlottedye: Sure.  Good idea.

yufengwng (approved):

PR #288, opened 2020-07-15
Synchronize when mutating LanguageStore

yufengwng:
## Context

We got [this error](https://console.cloud.google.com/errors/CJzojLnlkbO3ygE?time=P1D&project=broad-ddp-dev) that indicates an issue mutating the `HashMap` in `LanguageStore`, so this PR is adding `synchronized` to methods that mutate the language store.

## Checklist

## FUD Score

## Testing

## Release


charlottedye (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
```diff
@@ -24,7 +24,7 @@ public static LanguageDto getDefault() {
         return languages.get(DEFAULT_LANG_CODE);
     }
 
-    public static LanguageDto getOrCompute(Handle handle, String isoCode) {
+    public static synchronized LanguageDto getOrCompute(Handle handle, String isoCode) {
```
bskinner: Handling synchronization like this is complex and error-prone. This is not going to protect against the case where the backing hash is mutated while another class is attempting to iterate over the collection. We should be treating the Map as an immutable collection.

bskinner (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
yufengwng: We're not exposing the backing HashMap to callers so there's no collection to iterate over? We do have the `get()` methods that I purposely did not `synchronized` -- I was thinking it will be fine if caller is just doing a "read" while other threads is mutating underlying map so we save some synchronization costs, but if this is not recommended let me know and I'll tack on `synchronized` on those as well!

yufengwng: The list of languages is a pretty small list, so what we can do alternatively is load up the whole list once and make the backing HashMap immutable. That should also solve concurrency issues.

bskinner: That'd be great! I'm onboard with that fix

charlottedye: Do we think we'd ever end up with enough languages that the current approach would be worth it?  If not, it seems like loading the languages once might also simplify the code.

yufengwng: I think there's like 100 or so of these? Which is small potatoes...

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
```diff
@@ -14,31 +17,26 @@
...
+    private static Map<String, LanguageDto> languages = Collections.emptyMap();
 
-    public static LanguageDto getDefault() {
-        return languages.get(DEFAULT_LANG_CODE);
+    public static synchronized void init(Handle handle) {
+        Map<String, LanguageDto> map = new HashMap<>();
+        handle.attach(JdbiLanguageCode.class)
+                .findAll()
+                .forEach(lang -> map.put(lang.getIsoCode(), lang));
+        languages = Map.copyOf(map);
```
bskinner, line 27: The `copyOf` call isn't necessary here- assigning the `HashMap` instance we create in the first line should be sufficient.

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
charlottedye: 30 files changed!  I can see why this wasn't your original solution

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
```diff
@@ -14,31 +17,26 @@
...
     }
 
-    public static LanguageDto getOrCompute(Handle handle, String isoCode) {
-        return languages.computeIfAbsent(isoCode, key ->
-                handle.attach(JdbiLanguageCode.class).findLanguageDtoByCode(isoCode));
+    public static LanguageDto get(@Nullable String isoCode) {
```
bskinner, line 30: Have we run into cases where `isoCode` is passed a `null` value?

bskinner (approved):
Looks good, just a couple of minor comments. Thank for hopping in and getting this squared away!

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
yufengwng: Yes, I ran into issue where some of the Profile route code is calling it with `null` and apparently the underlying immutable map method will barf on a null...

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
yufengwng: True, but it spells out that we're making an immutable map and I believe there's some optimizations under-the-hood too.

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
yufengwng: Actually, I wasn't farsighted enough to predict the 30 files -- rather, I was basing my original design choice on how we have done other similar "cache stores", which is to use synchronization blocks. But thinking more about this particular one, it appears that its nature/lifetime is a bit different. The 30 files is just a manifestation of how widespread we're querying the languages!

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
bskinner: Just brushed up on those docs- that's _nice_. I take my comment back- this is a great use of `copyOf`!

charlottedye:
I was almost ready to click the approve button and then my cat insisted on playing with the laser pointer, but for the record I thought this was good.

PR #291, opened 2020-07-16
Initialize LanguageStore in more places

yufengwng:
@ssettipalli brought to my attention that study-builder is running into issue due to recent LanguageStore changes I have made. I then realized that we need to initialize the LanguageStore in more places.


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
```diff
@@ -208,6 +209,7 @@ private Path resolvePathToStudyConfigFile(String filepathOrKey) {
 
     private void execute(Consumer<Handle> task, boolean isDryRun) {
         TransactionWrapper.useTxn(handle -> {
+            LanguageStore.init(handle);
```
yufengwng, line 212: Putting this call here should cover all usages of study-builder.

charlottedye (approved):

PR #292, opened 2020-07-16
DDP-4807-add-script-to-retreive-address-pluscode

Rinchan20:
Have an ops request to change an address - one of the the columns in a mail address is for the pluscode. This is just to add a script to enable quick use/retrieval of pluscode that has the same precision we already use. I haven't added a test for it since OLCServiceTest seems to take care of testing precision in general.

## Context

## Checklist

- [X] I have labeled the type of changes involved using the `C-*` labels.

## FUD Score

- [X] :relaxed: All good, business as usual!

## Testing

- [X] I have written zero automated tests but have poked around locally to verify proper functionality

## Release

- [X] These changes require no special release procedures--just code!


pepper-apis/src/main/java/org/broadinstitute/ddp/AddressPlusCode.java
```diff
@@ -0,0 +1,32 @@
+package org.broadinstitute.ddp;
+
+import com.typesafe.config.Config;
+import org.broadinstitute.ddp.constants.ConfigFile;
+import org.broadinstitute.ddp.service.OLCService;
+import org.broadinstitute.ddp.util.ConfigManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AddressPlusCode {
```
yufengwng: Does this have to live under `src/main`? I'm a bit concerned that folks might get confused and start using it in real code. How about we move to this `src/test`? Maybe just encapsulate this as a `@Test` in `OLCServiceTest` or another test class?

yufengwng: It sounds like we just want to run this locally every once in a while right? This should probably do the trick:
```java
public class OLCServiceTest ... {
  ...

  // Uncomment to run this!
  // @Test
  public void computePluscode() {
    ...
  }
}

Rinchan20: It doesn't need to live under `src/main`, `src/test` sounds good - I was not thinking of that concern, I'll make sure to keep it in mind in the future.

I like the solution proposed! I'll do it that way.

yufengwng (approved):
👍 Looks good!

zyme (commented):
As  long as we're not running this at a utility when pointed at production, this is okay, but it  would be safer for this to run as  a  `main()` instead of a test so that it does not apply database changes via liquibase which:
1) are not needed for this test
2) if run on a non-prod branch while the config points at production, would apply database changes to prod as a side effect, which would put production in an inconsistent and potentially unusable state

MocanaAtBroad (approved):

PR #306, opened 2020-07-24
circleci: cache testing user token and handle rate limit

yufengwng:
## Context

Test suite kept failing. After a bit of digging, it looks like we're hitting Auth0's rate limit. This seems to happen when we're trying to fetch/login our testing user in order to grab an auth token for the tests. I also noticed that we're doing this login each and every time, so I tried to cache the token.

I did two main things:
* Updated our CircleCI pipeline so we cache the auth tokens for the tests. Due to limitations in circleci, I ended up using both the "cache" feature and the artifacts feature. We use the cache to save/restore the `.ddp-cache` directory which holds the cached tokens. I then generated a `.ddp-cache-md5.txt` file in order to use it as the cache key. The `.ddp-cache-md5.txt` needs to live somewhere that we can fetch from before restoring the cache, so I ended up storing it as a build artifact.
* Replaced the `getAuth0User()` method to do some primitive rate limit handling. I ended up having to do a bit of yak-shaving and refactors. I haven't seen the rate limit issue yet since I have put in this fix, but hopefully it will help.

## Checklist

## FUD Score

## Testing

## Release


.circleci/config.yml
```diff
@@ -584,6 +589,49 @@ commands:
...
+            done
+            echo "Contents of [$ddp_cache_md5]:"
+            cat "$ddp_cache_md5"
+      - save_cache:
+          key: ddp-cache-{{ checksum "/home/circleci/.ddp-cache-md5.txt" }}
+          paths:
+            - /home/circleci/.ddp-cache
+      - store_artifacts:
+          path: /home/circleci/.ddp-cache-md5.txt
+          destination: ddp-cache-md5.txt
```
MocanaAtBroad, line 650: Not really understanding how storing the m5s is helping us. Would storing and then restoring the cache in whatever the last state it was be sufficient? Maybe just use the TestUserUtil.java file to generate a checksum like we do for pom.xml file to save the Java libraries? And if we have cache misses and have to get create users, could we just make sure that we enforce we handle the rate limiting (e.g., we get the rate limiting error, we try again a couple of seconds later).

.circleci/config.yml
```diff
@@ -584,6 +589,49 @@ commands:
...
+
+  save-ddp-cache:
+    description: Save ddp cache
+    steps:
+      - run:
+          name: Compute ddp-cache-md5
+          command: |
+            ddp_cache_dir='/home/circleci/.ddp-cache'
+            ddp_cache_md5='/home/circleci/.ddp-cache-md5.txt'
+            echo -n "" > "$ddp_cache_md5"
```
MocanaAtBroad, line 638: Consider refactoring paths to shared variables. We do that elsewhere in file.

.circleci/config.yml
```diff
@@ -584,6 +589,49 @@ commands:
...
+            echo "Contents of [$ddp_cache_md5]:"
+            cat "$ddp_cache_md5"
+      - save_cache:
+          key: ddp-cache-{{ checksum "/home/circleci/.ddp-cache-md5.txt" }}
+          paths:
+            - /home/circleci/.ddp-cache
+      - store_artifacts:
+          path: /home/circleci/.ddp-cache-md5.txt
+          destination: ddp-cache-md5.txt
+
```
MocanaAtBroad, line 651: Not getting how the md5s are helping us here. Wonder if just simpler to save_cache just like you have it here and using the TestingUserUtil.java to generate the key suffix checksum. You already have the retry when we rate-limited.

yufengwng: First, let's clarify what we're caching. We want to cache the auth0 tokens. Those are saved to JSON files in `.ddp-cache` directory. When those tokens change (e.g. they were renewed after being expired) we want to cache those new tokens.

So the problem becomes what to use for the cache key. The key need to be unique and need to be updated whenever the cache contents (the tokens) change. This means we pretty much have to use CircleCI's `{{ checksum ... }}` templating. I'm not definitively sure on this, but it seems `checksum` only takes in a filename, and it must exist on the circleci filesystem. So we can't just have `checksum` do the whole `.ddp-cache` directory. My solution here is to "hash" the contents in the cache. I did this by computing md5 of each file and saving it into a text file. We can then run `checksum` on this file. Note that it doesn't have to be md5, it can be some other hash.

---

> using the TestingUserUtil.java to generate the key suffix checksum

I'm assuming your idea is to use `{{ checksum "...TestingUserUtil.java" }}`? In other words, we're hashing that particular Java source code file? This will not work. What we want to cache is the tokens, not the code.

.circleci/config.yml
yufengwng: CircleCI's caching feature is great for caching dependencies. The `pom.xml` file specifies exactly what the cache contents are, so doing `{{ checksum "pom.xml" }}` gives us the cache behavior we want. But it's a different story when you want to cache some arbitrary stuff. So what I ended up doing was computing the md5 of the cache content into a file so we can use that file in the `checksum`.

.circleci/config.yml
yufengwng: Yeah, I wanted to do something like that. I thought it's a CircleCI feature but didn't find any docs on it. I'm just realizing now that it's a YAML feature. Let me get up to speed on yaml references and refactor.

yufengwng: Switched to yaml references... but I have to say, that was quite a bit of boilerplate, especially for getting a string into snippets of shell script.

MocanaAtBroad: I agree! I keep going back to YAML and Circle docs looking for a better option but no dice. At least the refactor prevents changing a character in one place and breaking the script because we did not paste the same character elsewhere.

.circleci/config.yml
MocanaAtBroad: I guess my concern here is that it is not clear to me that the keys you are generating would be able to be recreated in follow up executions of the script, in which case they can't be retrieved. Looks like you have resolved this problem by storing as artifacts. Why not just use the artifacts?
Regardless, from your code looks to me that the Auth0 retries with waits will work no matter what. Did you. consider just ditching the caching/storing altogether? Shorter scripts/easier to grokk? (Question, not request).

yufengwng: The trouble I have with artifacts is that they seem to get overwritten / cleared out if the latest build fails. This means the next build that comes along will not find any artifacts. This is also the reason why my solution here is not optimal since it relies on artifact for the cache key. But at the very least it's good enough for now. Also, the benefit of keeping with cache is, unlike artifacts, there is _something_ cached, so you can retrieve it, albeit it being a bit older.

The trouble with the Auth0 retries is that I have only put it in one place. I was not able to track down the other places where we're hitting the rate limit yet. Regardless, to answer your question, I think it's better to cache things if we can instead of needlessly always calling Auth0 for a token.

MocanaAtBroad (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -174,6 +174,10 @@ public static void clearAfterHandler() {
     }
 
     public static void main(String[] args) {
+        start(args, null);
+    }
+
+    public static void start(String[] args, SendGridSupplier sendGridSupplier) {
```
yufengwng, line 180: Not too comfortable with the fact that I have to change real code just to facilitate testing. I'm hoping we can make Housekeeping easier to test in a more natural way when we seriously start refactoring it. For now, this change should be harmless.

MocanaAtBroad: D.I.

MocanaAtBroad (approved):

PR #307, opened 2020-07-28
github action to trigger circle for PR

yufengwng:
This PR adds a [GitHub Action](https://github.com/features/actions) to automatically trigger CircleCI on every push, so we don't always have to manually invoke `scripts/run_ci_tests.sh`. I think we're on the [free tier](https://github.com/pricing) which gives us 2,000 minutes/month. I hope that will be plenty, but if not we can just trigger CircleCI manually like usual.

The downside is that I couldn't get CircleCI API request to work with a project API token. I had to resort to using a personal API token. For now, I have added my own personal token to the repo to use for this action.

### DIFF
--- a/.github/workflows/trigger-circle-for-pr.yml
+++ b/.github/workflows/trigger-circle-for-pr.yml
@@ -1,0 +1,31 @@
+# Action to automatically trigger build/tests for a PR on every push.
+#
+# This is done by directly calling CircleCI's API to trigger an on-demand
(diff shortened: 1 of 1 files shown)


charlottedye:
This looks good to me but I've also never actually done anything with any of these .yml files so I could definitely be missing things.

yufengwng:
Going to leave this PR here in case someone wants to resurrect it.

PR #310, opened 2020-07-29
MPC picklist group question export changes and some minor study config changes

ssettipalli:
changes to support picklist Group question in DataExport (both csv and elastic)
some minor study config and migration changes.

### DIFF
--- a/elasticsearch/templates/activity_definition.json
+++ b/elasticsearch/templates/activity_definition.json
@@ -41,6 +41,26 @@
             "dateFields": {
               "type": "keyword"
             },
+            "groups": {
+              "properties": {
+                "groupStableId": {
+                  "type": "keyword"
+                },
+                "groupText": {
+                  "type": "text"
+                },
+                "options": {
+                  "properties": {
+                    "optionStableId": {
+                      "type": "keyword"
+                    },
+                    "optionText": {
+                      "type": "text"
+                    }
+                  }
+                }
+              }
+            },
             "options": {
               "properties": {
                 "optionStableId": {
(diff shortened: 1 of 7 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/PicklistQuestionFormatStrategy.java
```diff
@@ -50,6 +60,28 @@
...
+                stableIdTxt.put("optionStableId", optionDef.getStableId());
+                stableIdTxt.put("optionText", HtmlConverter.getPlainText(optionDef.getOptionLabelTemplate().render("en")));
+                options.add(stableIdTxt);
+            }
+            groupDef.put("groupStableId", groupStableId);
+            groupDef.put("groupText", groupTxt);
+            groupDef.put("options", options);
+            groups.add(groupDef);
+        }
+        if (!groups.isEmpty()) {
```
ssettipalli, line 81: Makes "groups" optional element in elastic json. Will exist only for PL Groups.
wondered if it should be empty list element for non-PL Group questions like groups: []

elasticsearch/templates/activity_definition.json
```diff
@@ -41,6 +41,26 @@
             "dateFields": {
               "type": "keyword"
             },
+            "groups": {
```
yufengwng, line 44: So when we do a release, we'll need to update the template in ES and re-upload all the activity definitions?

yufengwng (approved):
👍

elasticsearch/templates/activity_definition.json
ssettipalli: yes yufeng

PR #328, opened 2020-08-15
Time delayed asynchronous activity instance creation feature

ssettipalli:
testboston https://broadinstitute.atlassian.net/browse/DDP-4852 have a requirement to create activity instance with some specified time delay.
Added pepper support to be able to create asynchronously by Housekeeping with specified time delay. Event is not submitted to pubsub, processed by Housekeeping as queued event with specified time delay.


pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -665,6 +679,22 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
         }
     }
 
+    private static void createActivityInstance(Handle apisHandle, EventConfigurationDto eventConfDto, QueuedEventDto pendingEvent) {
+        User participant = apisHandle.attach(UserDao.class)
+                .findUserByGuid(pendingEvent.getParticipantGuid())
+                .orElse(null);
+        Optional<StudySettings> settings = apisHandle.attach(
+                StudyDao.class).findSettings(pendingEvent.getStudyGuid());
```
yufengwng: It's unfortunate that we don't have the study_id already. For now, it might be better to grab the `StudyDto` from `JdbiUmbrellaStudy` since those are cached.

pepper-apis/src/main/resources/db-changes/schema/DDP-4852-queued-event-add-column.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="sampath" id="20200815-queued-event-add-column">
+        <addColumn tableName="queued_event">
+            <column name="study_activity_id" type="bigint(20)">
```
yufengwng: Adding activity_id to queued_event seems awkward and out-of-place. And more prone to error since someone can just add an arbitrary activity_id even though the event configuration has a specific activity specific. I don't think we need to add this id. The `EventConfigurationDto` should have the `activityInstanceCreationStudyActivityId` that you can use.

ssettipalli: oops. How did I miss that. That would have saved me good time modifying unnecessary classes. Thank you.

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -665,6 +681,21 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
         }
     }
 
+    private static void createActivityInstance(Handle apisHandle, EventConfigurationDto eventConfDto, QueuedEventDto pendingEvent) {
+        User participant = apisHandle.attach(UserDao.class)
+                .findUserByGuid(pendingEvent.getParticipantGuid())
+                .orElse(null);
+        StudyDto studyDto = apisHandle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(pendingEvent.getStudyGuid());
```
MocanaAtBroad: Might as well use the cached version of this:
```suggestion
        StudyDto studyDto = new JdbiUmbrellaStudyCached(apisHandle).findByStudyGuid(pendingEvent.getStudyGuid());
```

yufengwng: I think this needs to be `new JdbiUmbrellaStudyCached(...)` to take advantage of caching?

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -352,7 +360,15 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
...
                                     JdbiMessage jdbiMessage = housekeepingHandle.attach(JdbiMessage.class);
-                                    if (jdbiEvent.shouldHandleEvent(
+                                    if (pendingEvent.getActionType() == EventActionType.ACTIVITY_INSTANCE_CREATION) {
+                                        Optional<EventConfigurationDto> eventConfOptDto =
+                                                eventDao.getEventConfigurationDtoById(pendingEvent.getEventConfigurationId());
+                                        if (!eventConfOptDto.isPresent()) {
+                                            LOG.error("No event configuration found for ID: {} . skipping queued event : {} ",
+                                                    pendingEvent.getEventConfigurationId(), pendingEvent.getQueuedEventId());
+                                        }
+                                        createActivityInstance(apisHandle, eventConfOptDto.get(), pendingEvent);
```
yufengwng: Can you move this block of code to _outside_ of `TransactionWrapper` call? So that we don't unnecessarily open a `DB.HOUSEKEEPING` database connection?

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -357,18 +357,21 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
                             }
 
                             if (hasMetPrecondition) {
+                                if (pendingEvent.getActionType() == EventActionType.ACTIVITY_INSTANCE_CREATION) {
+                                    Optional<EventConfigurationDto> eventConfOptDto =
+                                            eventDao.getEventConfigurationDtoById(pendingEvent.getEventConfigurationId());
+                                    if (!eventConfOptDto.isPresent()) {
```
ssettipalli, line 363: wondering if we need this eventConfOptDto.isPresent() check ? Unless manual db edits we probably will never run into it !!!

yufengwng (approved):

PR #331, opened 2020-08-17
DDP-4860 followup changes

ssettipalli:
Added two new questions to followup


yufengwng (commented):
This doesn't look like it's finished. Can you double-check?

study-builder/studies/testboston/followup-survey.conf
```diff
@@ -151,6 +151,27 @@
...
+          "shownExpr": null
+        },
+        {
+          "question": ${_includes.symptoms_survey.sample_collect_date} {
+            "stableId": ${id.q.followup_sample_collect_date}
+          }
+          "blockType": "QUESTION",
+          "shownExpr": null
+        },
+        {
```
ssettipalli, line 169: side effect of making time as separate question is that: 
For users its one question and backend two questions and question counts that we display on the dashboard will be off by 1. 
Trade off: Work around is not hideNumber and Time question will have a number

ssettipalli: image

fat23cat: How about splitting this question into 2 questions?
7. What date did you collect your sample using the home collection kit?
8. What time did you collect your sample using the home collection kit?

To be honest, I don't like it very much and probably should be discussed with Jen

study-builder/studies/testboston/snippets/picklist-question-yndk-prefernotanswer.conf
```diff
@@ -0,0 +1,93 @@
+{
```
ssettipalli, line 1: Actually this file can go into higher level snippets for use in other studies.  If needed we can add translation texts rather than substitution variables.

study-builder/studies/testboston/followup-survey.conf
yufengwng: I think you can hide number for just that one question.

ssettipalli: Yes I can hide it but question count in dashboard shows up as 8 and users don't see Q#8

study-builder/studies/testboston/snippets/picklist-question-yndk-prefernotanswer.conf
yufengwng: I think it can stay here, since it uses the languages and translations specific to TestBoston.

study-builder/studies/testboston/subs.conf
```diff
@@ -114,6 +120,7 @@
       "household_diagnosed": { include required("snippets/question-household-diagnosed.conf") },
       "contact_diagnosed": { include required("snippets/question-contact-diagnosed.conf") },
       "interested_continuing": { include required("snippets/question-interested-continuing.conf") },
+      "flu_shot": { include required("snippets/question-flu-shot.conf") },
```
yufengwng: This question is only in the longitudinal followup survey right? Why is it included under `covid_survey`?

study-builder/studies/testboston/snippets/question-followup-sample-collect-time-am-pm.conf
```diff
@@ -0,0 +1,60 @@
...
+          { "language": "en", "text": "" },
+          { "language": "es", "text": "" },
+          { "language": "ht", "text": "" }
+        ]
+      }
+    ]
+  },
+  "picklistOptions": [
+    {
+      "stableId": "TIME_AM",
```
yufengwng: Just do `AM`, `PM`.

study-builder/studies/testboston/snippets/question-followup-sample-collect-time-am-pm.conf
```diff
@@ -0,0 +1,60 @@
...
+          }
+        ]
+      }
+    },
+  ],
+  "validations": [
+    {
+      "ruleType": "REQUIRED",
+      "allowSave": true,
+    }
```
yufengwng, line 58: I have noticed that these required rules are making it so you can submit the form without filling in a time, even if you didn't have to fill in date or symptoms... Same case in Baseline Symptom survey. Not sure if that's what we want.

study-builder/studies/testboston/followup-survey.conf
yufengwng: Oh! Yeah, that's an issue...

fat23cat: @ssettipalli, could you please send me in the email the questions' JSON, I want to test the styling for these questions.

study-builder/studies/testboston/snippets/question-followup-sample-collect-time-am-pm.conf
ssettipalli: Date and Time (hour & Min) were made required, hence this is also required. Will double check and probably make all optional.

study-builder/studies/testboston/subs.conf
ssettipalli: I guess I didn't want to add followup_survey entry for just this question. will change it.

study-builder/studies/testboston/snippets/question-followup-sample-collect-time-am-pm.conf
ssettipalli: confirmed with jen, its required.

yufengwng (approved):

study-builder/studies/testboston/snippets/question-followup-sample-collect-time-am-pm.conf
yufengwng: It doesn't look like the date part is marked required. I can fix that once this is merged.

PR #345, opened 2020-08-25
Try to fix Redis timeout error

MocanaAtBroad:
## Context
Observed a couple of errors related to Redis timeouts in logs in dev and staging environments. This PR tries to follow advice in the provided error message.
I do not know if this fixes the errors! Have not seen the error again, but I am not 100% sure what the cause of the error is.

The error:
```
Error while figuring out the user language Redis server response timeout (30000 ms) occured after 3 retry attempts. Increase nettyThreads and/or timeout settings. Try to define pingConnectionInterval setting. Command: (HKEYS), params: [{LOCAL_CACHE:studyIdToLanguageCache}:disabled-keys], channel: [id: 0xbed0f542, L:/169.254.8.1:16143 - R:/10.248.94.251:6379]
```

Note that in staging environment the error occurs while accessing a different cache, so I don't think the error is caused by the code, which pretty simple and straightforward.

This error appears to be transient as it has not happened again. I have accessed routes that would have pinged this cache and no errors have occurred without this fix.

I have another code branch off rc_v1.9 with the same commit. 

## Checklist

- [ X] I have considered security and privacy, and added `I-*` labels as needed
- [ X] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [x ] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

## FUD Score

N/A

## Testing
Executed tests use same settings we are adding here to DataDonationPlatform runtime.

## Release
Just template changes. Will be applied in the course of a normal deployment.


pepper-apis/src/test/resources/redisson-test-jcache.yaml
```diff
@@ -16,6 +16,7 @@ singleServerConfig:
   database: 0
   dnsMonitoringInterval: 5000
 threads: 16
-nettyThreads: 32
+nettyThreads: 64
+pingConnectionInterval: 3000
```
yufengwng, line 20: Is there a difference putting this at the top-level or under `singleServerConfig`?

MocanaAtBroad: I am not sure. I based the original configuration on what they have for examples in the repo:
https://github.com/redisson/redisson/wiki/2.-Configuration#262-single-instance-yaml-config-format

yufengwng: Okay. Was just asking because the two files are different. If the `redisson-jcache.yaml.ctmpl` file above looks right to you, then let's ship it.

yufengwng (approved):

PR #350, opened 2020-08-31
DDP 4842 mpc followup pdf

ssettipalli:
DDP-4842 MPC followup activity pdf generation needs composite question answers substituted into pdf. 
Added pepper platform level support for composite question answer pdf substitution.
Each composite question is substituted as one separate pdf page. 
There are things that can be improved but this is probably good enough for now.

Few things to implement in future:
Figure out a way to execute javascript in pdfs from iText during Pdf generation
Remove empty template pages if dependency questions or entire composite question is not answered
Remove dependency questions from generated pdfs (can be handled if javascript can be triggered from java/iText)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfSql.java
```diff
@@ -137,15 +137,19 @@ default int insertPhysicianInstitutionTemplate(PhysicianInstitutionTemplate temp
...
-            + "  where qsc.stable_id = :stableId"
-            + "    and qsc.umbrella_study_id = (select study_id from study_activity where study_activity_id = :activityId))")
+    @SqlUpdate("insert into pdf_answer_substitution (pdf_substitution_id, question_stable_code_id, parent_question_stable_code_id) "
+            + "(select :subId, "
+            + "(select question_stable_code_id from question_stable_code where stable_id = :stableId "
+            + "    and umbrella_study_id = (select study_id from study_activity where study_activity_id = :activityId)) "
+            + " , "
+            + "(select question_stable_code_id from question_stable_code where stable_id = :parentStableId "
+            + "    and umbrella_study_id = (select study_id from study_activity where study_activity_id = :activityId)) "
+            + "   from dual) ")
```
yufengwng: How about cleaning up the statement to use regular `insert .. values (:subId, (select ..), (select ..))` instead of selecting from `dual`?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -720,6 +707,128 @@ private void convertSubstitutionToPdf(AnswerSubstitution substitution,
...
+                break;
+
+            default:
+                errors.add("tried to use an unsupported answer type " + substitution.getQuestionType());
+                return;
+        }
+    }
+
+    private void substituteBoolean(BooleanAnswerSubstitution substitution, Answer answer, PdfFormField field) {
+        BooleanAnswerSubstitution booleanSubstitution = substitution;
```
yufengwng: Seems like a redundant assignment.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -720,6 +707,128 @@ private void convertSubstitutionToPdf(AnswerSubstitution substitution,
...
+            if (booleanSubstitution.checkIfFalse()) {
+                shouldCheck = !boolValue;
+            } else {
+                shouldCheck = boolValue;
+            }
+            setIsChecked(field, shouldCheck);
+        }
+    }
+
+    private void substituteDate(PdfFormField field, Answer answer) {
```
yufengwng: Can you swap the order of these two parameters to be consistent with other similar methods?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -720,6 +707,128 @@ private void convertSubstitutionToPdf(AnswerSubstitution substitution,
...
+
+            PdfAcroForm compositeForm = PdfAcroForm.getAcroForm(renderedCompositePdf, true);
+            compositeForm.setGenerateAppearance(true);
+
+            List<Answer> childAnswers = answerRow.getValues();
+
+            for (Answer childAnswer : childAnswers) {
+                if (childAnswer != null) {
+                    for (AnswerSubstitution sub : compositeSubstitution.getChildAnswerSubstitutions()) {
+                        if (sub.getQuestionStableId().equalsIgnoreCase(childAnswer.getQuestionStableId())) {
```
yufengwng: We should do an exact match on the stable_id to be on the safe side.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -720,6 +707,128 @@ private void convertSubstitutionToPdf(AnswerSubstitution substitution,
...
 
+    private void substituteComposite(CompositeAnswerSubstitution substitution, PdfAcroForm form, CustomTemplate template,
+                                     List<String> errors, Answer answer) throws IOException {
+        if (answer == null) {
+            return;
+        }
+
+        //passed form contains custom composite template pages. just remove those empty pages
+        int pageCount = form.getPdfDocument().getNumberOfPages();
+        IntStream.range(1, pageCount + 1).forEach(i -> form.getPdfDocument().removePage(i));
```
yufengwng, line 718: Not understanding what this block of code does. Why are they empty? Why do we need to remove them?

study-builder/studies/mpc/study-pdfs.conf
```diff
@@ -171,6 +171,167 @@
...
+                "street": "street",
+                "city": "city",
+                "state": "state",
+                "zip": "zip",
+                "country": "country",
+                "phone": "phone"
+              }
+            },
+            {
+              "filepath": "pdfs/followup_curr_medication.pdf",
```
yufengwng, line 216: I couldn't add a comment to the PDF files on github, but wanted to ask: these and the "details" PDF files already has `FALSE` and `NO` text in the file itself, will that cause any issues?

yufengwng (commented):
Looks good, just a few questions.

study-builder/studies/mpc/study-pdfs.conf
ssettipalli: After some back and forth  with Mary and Lauren I had a default value for one question (FALSE) . 
Since we implemented these Yess/No (True/False) questions as Picklists , Mary wanted value if not checked.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
ssettipalli: Each composite question is one pdf page.  Form already includes empty composite template/page (existing code). Code in this method does not use that already included page. Iterates through answers and creates 1 or more pages depending on how many answers were provided (List of Answers), hence deleting those empty pages.

yufengwng: I see. So the caller, `generateCustomPdf()`, had initialized the `form` with the template and we're clearing that out before adding in rendered composite pages.

yufengwng (approved):
Looks good

PR #351, opened 2020-09-01
DDP-4908 allow using proxy for sengrid api requests

yufengwng:
## Context

This PR allows setting up SendGrid API calls to be routed through our proxy, so that we can whitelist that IP in our SendGrid account. In terms of performance, sendgrid API is only used by Housekeeping and the frequency of sending emails is moderate to low, so any user-facing degradation is not expected.

**Deployment**: we'll need to rebuild the `pepper-squid` docker image and push up new configs. Also need to add proxy url to vault. Then whitelist the proxy VM's IP in SendGrid account.

## Checklist

## FUD Score

## Testing

- [x] Tested the squid config, will test more after we put this on `dev`.

## Release

### DIFF
--- a/pepper-apis/gae-egress-proxy/squid.conf
+++ b/pepper-apis/gae-egress-proxy/squid.conf
@@ -1,6 +1,8 @@
 # Access control definitions.
+acl dest_ports port 443                     # Regular SSL port
 acl dest_ports port 9243                    # Elasticsearch port
 acl allowed_domains dstdomain .cloud.es.io  # Elasticsearch host
+acl allowed_domains dstdomain api.sendgrid.com  # SendGrid API host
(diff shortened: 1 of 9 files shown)


MocanaAtBroad (approved):

pepper-apis/gae-egress-proxy/squid.conf
```diff
@@ -1,6 +1,8 @@
 # Access control definitions.
+acl dest_ports port 443                     # Regular SSL port
```
bskinner, line 2: Clarification on the port numbering- this is the regular HTTP SSL port (squid does support other protocols)

bskinner (approved):
One comment, but looks good othetwisr

PR #356, opened 2020-09-02
testboston: minor translation fixes

yufengwng:
Minor translation changes from translation company shared by Jen. Also fixes DDP-4914.

### DIFF
--- a/study-builder/studies/testboston/snippets/question-live-with-other-people.conf
+++ b/study-builder/studies/testboston/snippets/question-live-with-other-people.conf
@@ -9,9 +9,37 @@
       {
         "name": "live_with_other_people_prompt",
         "translations": [
-          {"language": "en", "text": ${i18n.en.live_with_other_people_prompt}},
-          {"language": "es", "text": ${i18n.es.live_with_other_people_prompt}},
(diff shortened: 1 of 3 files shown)


study-builder/studies/testboston/snippets/question-live-with-other-people.conf
```diff
@@ -9,9 +9,37 @@
...
+  "falseTemplate": {
+    "templateType": "TEXT",
+    "templateText": "$live_with_other_people_no",
+    "variables": [
+      {
+        "name": "live_with_other_people_no",
+        "translations": [
+          { "language": "en", "text": ${i18n.en.option.no} },
+          { "language": "es", "text": ${i18n.es.option.no} },
+          { "language": "ht", "text": ${i18n.ht.option.no} }
```
MocanaAtBroad, line 42: Would it make more sense to put the yes/no translations in the the snippet that is getting imported:
studies/snippets/bool-question-yes-no.conf

yufengwng: Perhaps, but since the import snippet is also imported by other studies, what if the other studies doesn't support the same set of languages that we do?

MocanaAtBroad (approved):

PR #357, opened 2020-09-03
testboston: base skeleton study

yufengwng:
We can setup skeleton study without changing code, but I'll rather not have to fiddle with anything when doing a prod deploy. I pulled out the minimal set of configs out to a `study-base.conf` file that we can run to setup TestBoston skeleton. With this change, we can just run a single command for the skeleton:

```
java -Dconfig.file=output-config/application.conf -jar target/StudyBuilder.jar \
    --vars output-config/testboston-vars.conf \
    --substitutions studies/testboston/subs.conf \
    studies/testboston/study-base.conf
```

### DIFF
--- a/study-builder/studies/testboston/study-base.conf
+++ b/study-builder/studies/testboston/study-base.conf
@@ -1,0 +1,57 @@
+{
+  "tenant": {
+    "domain": ${auth0.domain},
+    "mgmtClientId": ${auth0.mgmtClientId},
+    "mgmtSecret": ${auth0.mgmtClientSecret}
+  },
+
+  "umbrella": {
+    "name": "TestBoston",
+    "guid": ${id.study}
+  },
+
+  "study": {
+    "name": "TestBoston",
+    "guid": ${id.study},
+    "studyEmail": ${contact.email},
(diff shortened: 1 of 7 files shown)


study-builder/studies/testboston/study-base.conf
```diff
@@ -0,0 +1,57 @@
...
+      "secret": ${auth0.crcClientSecret},
+      "passwordRedirectUrl": null
+    },
+  ],
+
+  "adminUser": {
+    "guid": "TESTBOSTONADMINUSER1"
+  },
+
+  "supportedLanguages": [
```
yufengwng, line 41: I wanted to be able to have languages in the skeleton, so that the PIs can see the different languages on the frontend.

study-builder/studies/testboston/study-base.conf
```diff
@@ -0,0 +1,57 @@
+{
```
yufengwng, line 1: These are the only things that will be initialized in the skeleton.

MocanaAtBroad (approved):
Guess it _**should**_ work...

PR #364, opened 2020-09-11
Fix i18n key in testboston translation files

MocanaAtBroad:
Was trying to read the study builder files for studybuilder-ui parser and found this problem.


yufengwng (approved):
Yup, my bad. I fixed this in my other PR, but I don't mind merging this in separately.

PR #366, opened 2020-09-14
DDP-4931 testboston activity config changes

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-4931

Added new DB column: study_activity.hide_instances to track activities for which existing instances should be hidden.
DB column is used to Hide all existing adhoc-symptom survey instances once new OnDemand adhoc-symptom survey instance is created from DSM.


pepper-apis/src/main/resources/db-changes/schema/DDP-4931-activity-hide-instances.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ssettipalli, line 1: would have preferred to use some existing db column or config rather than add new column to track if existing instances should be hidden for study activity. 
Felt like over loading with new columns. We have study_activity.exclude_from_display but that seem to be for different purpose. 
Wondering if we can consolidate these db columns and purposes
study_activity.exclude_from_display
study_activity.hide_instances
activity_instance.is_hidden

yufengwng: I do think they serve different purposes and so there's different properties, but certainly agree that we should look for opportunities to consolidate.

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
```diff
@@ -84,6 +84,7 @@
         public static final String EXCLUDE_STATUS_ICON_FROM_DISPLAY = "exclude_status_icon_from_display";
         public static final String ALLOW_UNAUTHENTICATED = "allow_unauthenticated";
         public static final String IS_FOLLOWUP = "is_followup";
+        public static final String HIDE_INSTANCES = "hide_instances";
```
yufengwng: I understand the purpose of this flag, but how about a different name? `hide_instances` seems to imply that we're always hiding all instances. Maybe `hide_existing_instances_on_creation`?

study-builder/studies/testboston/study-events.conf
```diff
@@ -301,7 +328,6 @@
           || !user.studies["testboston"].forms["BASELINE_COVID"].questions["INTERESTED_CONTINUING"].isAnswered()
           || user.studies["testboston"].forms["BASELINE_COVID"].questions["INTERESTED_CONTINUING"].answers.hasFalse()
           """,
-      "maxOccurrencesPerUser": 1,
```
yufengwng: We shouldn't need to delete this line. It's true that you can only submit the baseline_symptom once, but it wouldn't hurt to also limit this event occurrence to 1. Note that this "max occurrence" is for the event and not the number of activity instances.

study-builder/studies/testboston/study-events.conf
```diff
@@ -194,6 +194,20 @@
...
+    {
+      "trigger": {
+        "type": "DSM_NOTIFICATION",
+        "dsmEvent": "TESTBOSTON_DELIVERED"
+      },
+      "action": {
+        "type": "HIDE_ACTIVITIES",
+        "activityCodes": [${id.act.longitudinal}],
+      },
+      "cancelExpr": """!user.studies["testboston"].forms["LONGITUDINAL_COVID"].hasInstance()""",
```
yufengwng: Having this cancel expr here seems redundant -- if there's no instances of longitudinal_covid, then none will be updated.

study-builder/studies/testboston/study-events.conf
```diff
@@ -345,6 +371,40 @@
...
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
+        "activityCode": ${id.act.longitudinal},
+        "statusType": "CREATED"
+      },
+      "action": {
+        "type": "HIDE_ACTIVITIES",
+        "activityCodes": [${id.act.baseline_symptom}],
+      },
+      "cancelExpr": """!user.studies["TESTBOSTON"].forms["BASELINE_SYMPTOM"].hasInstance()""",
```
yufengwng: cancelExpr appears redundant.

study-builder/studies/testboston/study-events.conf
```diff
@@ -374,5 +434,51 @@
...
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
+        "activityCode": ${id.act.result_report},
+        "statusType": "CREATED"
+      },
+      "action": {
+        "type": "HIDE_ACTIVITIES",
+        "activityCodes": [${id.act.baseline_symptom}],
+      },
+      "cancelExpr": """!user.studies["TESTBOSTON"].forms["BASELINE_SYMPTOM"].hasInstance()""",
```
yufengwng: I think this cancelExpr is redundant.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/DsmTriggerOnDemandActivityRoute.java
```diff
@@ -86,8 +87,12 @@ public Object handle(Request request, Response response, TriggerActivityPayload
                 return;
             }
 
-            ActivityInstanceDto instanceDto = handle.attach(ActivityInstanceDao.class)
-                    .insertInstance(activityDto.getActivityId(), participantGuid, participantGuid,
+            ActivityInstanceDao instanceDao = handle.attach(ActivityInstanceDao.class);
+            if (activityDto.isHideInstances()) {
```
yufengwng: Is `hide_instances` only applicable to on-demand activities? What happens when an activity instance is created through `INSTANCE_CREATION` event? Do we need to respect this flag in `ActivityInstanceCreationEventAction` as well?

study-builder/studies/testboston/study-events.conf
```diff
@@ -238,6 +252,19 @@
...
       "order": 1
     },
+    {
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
+        "activityCode": ${id.act.baseline_symptom},
+        "statusType": "CREATED"
+      },
+      "action": {
+        "type": "HIDE_ACTIVITIES",
```
yufengwng, line 261: Do you think there's an alternative approach here, of extending the `HIDE_ACTIVITIES` event instead of adding `hide_instances` flag? What if we add a new flag to `HIDE_ACTIVITIES` to "hide existing instances except the latest one". This way, we can chain this event after an `INSTANCE_CREATION` event to hide any old instances while keeping the latest instance (the one just created) visible.

The implementation will be totally different, and probably a bit more work. But the good thing is we can use this for both adhoc_symptom and longitudinal_covid activities. Two approaches I see:

1. Add a flag to the event action itself. We'll need to create a new table for `HIDE_ACTIVITIES` event action to hold this new boolean flag because we don't have a table yet. The drawback here is that this flag will apply for every activity listed in `activityCodes`.

2. We make it granular and specific to each activity we list. So we deprecate `activityCodes` and add `activities` instead, which will make it look like this:
    ```
    "action": {
      "type": "HIDE_ACTIVITIES",
      "activities": [
        { "activityCode": "foobar", "excludeLatest": true },
        ...
      ]
    }
    ```
    For this, we can add a column to the `event_action_target_activity` table. The downside here is that the `event_action_target_activity` table is also used in the `MARK_ACTIVITIES_READ_ONLY` event action. But if we named the column `exclude_latest_instance` then it should also make sense for `MARK_ACTIVITIES_READ_ONLY` too? (we don't have to support excluding latest for `MARK_ACTIVITIES_READ_ONLY` right now but can do so later).

ssettipalli: Approach I took seem to be more cleaner & less confusing to me than hacking existing events.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/DsmTriggerOnDemandActivityRoute.java
ssettipalli: If we have a need yes I would enhance INSTANCE_CREATION event to respect this flag.

study-builder/studies/testboston/study-events.conf
ssettipalli: tried to be extra safe but I am ok to remove them

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
ssettipalli: renamed

pepper-apis/src/main/java/org/broadinstitute/ddp/route/DsmTriggerOnDemandActivityRoute.java
ssettipalli: Handled this flag in ActivityInstanceCreationEventAction for any future use

yufengwng (approved):

PR #375, opened 2020-09-19
Refactor housekeeping tasks receiver

yufengwng:
## Context

Refactor the Housekeeping tasks receiver so we no longer need to provider the subscription to match on. There will only be one instance of Housekeeping and so we'll only setup one pubsub subscription for it. Before, it was a big hassle because you have to first go look for the subscription name before sending a task message to Housekeeping.

A new value needs to go into Vault, and we also need to setup the subscription, both of which I have already done on `dev` environment.

## Checklist

## FUD Score

- [x] Getting dev into a state where this is user-visible requires some tech fiddling, which I have done.

## Testing

## Release


ssettipalli (approved):
looks good to me.

PR #379, opened 2020-09-22
DDP-5012 support activity second names

yufengwng:
## Context

Adds support for configuring the name of activity instances. The `name` can now be rendered using "content substitutions" allow for more customization. We also add a "second name" that can be used for the second instance and newer.

Included in this PR is adding a "second name" for the `result_report` activity for TestBoston.

## Checklist

## FUD Score

- [x] Getting dev into a state where this is user-visible requires some tech fiddling. For TestBoston.

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
@@ -68,6 +68,7 @@ public class SqlConstants {
         public static final String ID = "study_activity_id";
         public static final String CODE = "study_activity_code";
         public static final String NAME_TRANS = "activity_name_trans";
+        public static final String SECOND_NAME_TRANS = "activity_second_name_trans";
         public static final String TITLE_TRANS = "activity_title_trans";
         public static final String SUBTITLE_TRANS = "activity_subtitle_trans";
         public static final String DESCRIPTION_TRANS = "activity_description_trans";
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
@@ -28,6 +28,7 @@ public class RenderValueProvider {
     private LocalDate date;
     private String testResultCode;
     private Instant testResultTimeCompleted;
+    private Integer activityInstanceNumber;
 
     private RenderValueProvider() {
         // Use builder.
@@ -120,6 +121,17 @@ public class RenderValueProvider {
         }
     }
 
+    /**
+     * Returns the activity instance number, if available.
+     */
+    public String activityInstanceNumber() {
+        if (activityInstanceNumber == null) {
+            return null;
+        } else {
+            return String.valueOf(activityInstanceNumber);
+        }
+    }
+
     // Get provided values as a map to save as snapshot. Should not be called within templates.
     public Map<String, String> getSnapshot() {
         var snapshot = new HashMap<String, String>();
@@ -197,6 +209,11 @@ public class RenderValueProvider {
             return this;
         }
 
+        public Builder setActivityInstanceNumber(Integer activityInstanceNumber) {
+            provider.activityInstanceNumber = activityInstanceNumber;
+            return this;
+        }
+
         public Builder withSnapshot(Map<String, String> snapshot) {
             String value = snapshot.get(I18nTemplateConstants.Snapshot.PARTICIPANT_GUID);
             if (value != null) {
@@ -251,6 +268,7 @@ public class RenderValueProvider {
             copy.date = provider.date;
             copy.testResultCode = provider.testResultCode;
             copy.testResultTimeCompleted = provider.testResultTimeCompleted;
+            copy.activityInstanceNumber = provider.activityInstanceNumber;
             return copy;
         }
     }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
@@ -132,6 +132,30 @@ public class RenderValueProvider {
         }
     }
 
+    /**
+     * Provides more flexibility for how to display an activity instance number, if available. The activity instance
+     * number will first be adjusted by subtracting the given offset. Then, if the adjusted number is less than the
+     * given cutoff number, then no number will be displayed. The cutoff number effectively serves as the first number
+     * to be displayed after the offset adjustment. The prefix is useful for optionally adding additional text when a
+     * number is displayed (e.g. prepending a space or a "#" symbol).
+     *
+     * @param offsetToSubtract subtract this amount from the number
+     * @param numberCutoff     if adjusted number is less than this cutoff number, no number will be displayed
+     * @param prefix           a prefix to prepend
+     * @return adjusted number
+     */
+    public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCutoff, String prefix) {
+        if (activityInstanceNumber == null) {
+            return null;
+        }
+        int adjustedNumber = activityInstanceNumber - offsetToSubtract;
+        if (adjustedNumber < numberCutoff) {
+            return "";
+        } else {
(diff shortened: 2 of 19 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
```diff
@@ -378,35 +380,60 @@ public Long getActivityInstanceIdByGuid(Handle handle, String guid) {
...
-            var instanceDao = handle.attach(org.broadinstitute.ddp.db.dao.ActivityInstanceDao.class);
-            var substitutions = instanceDao.bulkFindSubstitutions(instanceIds)
-                    .collect(Collectors.toMap(wrapper -> wrapper.getActivityInstanceId(), wrapper -> wrapper.unwrap()));
-            var sharedSnapshot = I18nContentRenderer
-                    .newValueProviderBuilder(handle, userId)
-                    .build().getSnapshot();
-            for (var summary : summaries) {
-                if (StringUtils.isBlank(summary.getActivitySummary())) {
-                    continue;
+    public void renderActivitySummary(Handle handle, long userId, List<ActivityInstanceSummary> summaries) {
```
yufengwng, line 393: This method has the most significant changes. The rest of the PR is plumbing to pass through the new configuration data.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateActivityNamingDetails.java
```diff
@@ -60,16 +60,17 @@ public void run(Handle handle) {
                     .collect(Collectors.toMap(ActivityI18nDetail::getIsoLangCode, Functions.identity()));
 
             Map<String, String> names = collectTranslatedText(definition, "translatedNames");
+            Map<String, String> secondNames = collectTranslatedText(definition, "translatedSecondNames");
```
yufengwng, line 62: The `UpdateActivityNamingDetails` task can now also deal with "second names".

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
```diff
@@ -73,6 +73,7 @@
         public static final String INSTANTIATE_UPON_REGISTRATION = "instantiate_upon_registration";
         public static final String MAX_INSTANCES_PER_USER = "max_instances_per_user";
         public static final String NAME_TRANS = "activity_name_trans";
+        public static final String SECOND_NAME_TRANS = "activity_second_name_trans";
```
yufengwng, line 71: Anyone got a better name than "second name"?

yufengwng:
@charlottedye I recall Eugene had asked for functionality similar to this. The use-case was to use a different name for the second instance of the "medical survey" for Prion. Will this new feature satisfy the use-case in Prion?

charlottedye:
Just asked Gene again about what the original goal was...  He says it was to have the medical surveys increment like this: Medical Questionnaire, Medical Questionnaire Update, Medical Questionnaire Update 2, etc.  Is that how this code would work?

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
bskinner: `name_pattern`? It looks like it's being used to provide an auto-naming scheme when there's multiple activities with the same code.

As implemented, this has potential i18n issues (you cannot always meaningfully add the number as a suffix) so I would recommend either adding printf-style substitution ("use %d to mark where the activity number should be added"), or a named parameter (a string search & replace with something like `{{count}}` could work). It's a common pattern in iOS development to use format strings in the translation files.

yufengwng:
@charlottedye Got it, thanks for the information. In that case, what's implemented here WILL NOT satisfy that. I'll see if I can somehow work that in.

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
yufengwng: Substitution is how it is working. If there is a "second name" available, we don't automatically append a number to the end. Instead, we render the text so you can put `$ddp.activityInstanceNumber()` anywhere  inside that text. I'm keeping the old behavior of "automatically append number" since that's how existing studies work -- if we want to get rid of  that, we need to go back and add/update activity definitions.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
```diff
@@ -378,35 +380,60 @@ public Long getActivityInstanceIdByGuid(Handle handle, String guid) {
...
+            context.put(I18nTemplateConstants.DDP, provider);
+
+            // Render the name.
+            final String nameText;
+            if (summary.getInstanceNumber() > 1) {
+                if (StringUtils.isNotBlank(summary.getActivitySecondName())) {
+                    nameText = renderer.renderToString(summary.getActivitySecondName(), context);
+                } else {
+                    String originalName = renderer.renderToString(summary.getActivityName(), context);
+                    nameText = originalName + " #" + summary.getInstanceNumber();
```
MocanaAtBroad, line 427: Are we going to have the first second instance have a #2 suffix (we start at 1 right? This might be good to annotate somewhere). Might be a little weird to have a #2 without a #1

yufengwng: Actually, it has always been that way. First instance has no number, second instance has `#2`, third instance has `#3`, etc. I'm keeping that behavior so we don't change anything for existing studies.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -120,6 +121,41 @@ public String testResultTimeCompleted(String format) {
...
+     * given cutoff number, then no number will be displayed. The cutoff number effectively serves as the first number
+     * to be displayed after the offset adjustment. The prefix is useful for optionally adding additional text when a
+     * number is displayed (e.g. prepending a space or a "#" symbol).
+     *
+     * @param offsetToSubtract subtract this amount from the number
+     * @param numberCutoff     if adjusted number is less than this cutoff number, no number will be displayed
+     * @param prefix           a prefix to prepend
+     * @return adjusted number
+     */
+    public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCutoff, String prefix) {
```
yufengwng, line 147: @charlottedye My solution is to provide this substitution: `$ddp.activityInstanceNumberDisplay(offset, cutoff, prefix)`. This should allow us to achieve the numbering scheme desired by Prion. Using TestBoston as a playground, here's some examples:

1. Example of naming desired by Prion. Here, I used:
```
name: Medical Questionaire
second_name: Medical Questionaire Update${ddp.activityInstanceNumberDisplay(1, 2, " #")}
```
act_inst_num_prion

2. Another example, which is closer to what TestBoston might use:
```
name: Medical Questionaire
second_name: Medical Questionaire Month ${ddp.activityInstanceNumberDisplay(0, 2, "")}
```
act_inst_name_2

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
charlottedye: Maybe something like follow up name?  Not sure that's really any better.

MocanaAtBroad (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
charlottedye: Nice!

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/SqlConstants.java
charlottedye: Or SECOND_INSTANCE_NAME_TRANS.

charlottedye (dismissed):

MocanaAtBroad (approved):

PR #390, opened 2020-09-30
DDP-5032 update TestBoston Spanish translations

yufengwng:
## Context

In order to support updating the translation text, I ended up needing to make a ton of code changes in both pepper and study-builder. Summary:

* Updated `UpdateActivityBaseSettings` task to support updating status summaries.
* Removed `UpdateActivityNamingDetails` task and consolidated it into `UpdateActivityBaseSettings`.
* Updated a ton of DAO, DTO, and activity definition classes so we're querying out template-related data.
* Added `UpdateTemplatesInPlace` task to iterate through and update all templates in study.

Going forward, whenever we update a template or the i18n strings, we can just run `UpdateTemplatesInPlace` if it makes sense for the study and the particular change. This task does make several assumptions, and only works in certain situations, but at least it's better than not having anything.

**Performance:** There is a performance penalty because the new code is querying out more template objects. This will affect the "boot time" because on boot there are no activity definitions cached. There is also more memory footprint too because we're pulling out more template objects and storing them in the activity definition cache. I plan to optimize the activity definition queries in a separate ticket/PR.

**To try this locally:** First, checkout `develop` and setup TestBoston study locally. Then, checkout this branch and run both the `UpdateActivityBaseSettings` and `UpdateTemplatesInPlace` study-builder tasks.

## Checklist

## FUD Score

- [x] Getting dev into a state where this is user-visible requires running the study-builder tasks

## Testing

## Release


MocanaAtBroad (approved):
Don't see anything that jumps at me as being wrong.

PR #398, opened 2020-10-01
DDP-5049 count activity summaries after filtering

yufengwng:
## Context

Part 1 of optimizing performance of listing activities that's typically used in dashboard. There's more to come in separate PRs. In brief, key insight is to minimize the amount of work we're doing. Counting questions & answers takes up a good portion of time, so we try doing that only after filtering down the list.

**Performance:** I tried timing the list activities API endpoint locally using TestBoston as an example. In my local setup, I have 3 activities in TestBoston dashboard. Before this PR, the first load of activity list took ~ 1.9 secs and subsequent loads took 300 - 400 ms. After this PR, first load took ~ 1.5 secs and subsequent loads took 100 - 200 ms. First load is slow because of needing to query and cache activity definition objects. Timings are rough estimates using print statements and `System.currentTimeMillis()`. 

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
```diff
@@ -266,17 +266,9 @@ public Long getActivityInstanceIdByGuid(Handle handle, String guid) {
...
-        List<IconBlobDto> iconBlobs = jdbiFormTypeActivityInstanceStatusType.getIconBlobs(studyGuid);
-        // Transforming a List of icon blobs into a Map having form type and activity instance status type
-        // concatenated with "-" as a delimiter as keys and icon blobs as values, e.g. { "CONSENT-5": "<icon blob>" }
-        iconBlobs.forEach(blob -> formTypeAndStatusTypeToIcon.put(
-                blob.getFormType() + "-" + blob.getStatusTypeCode(), blob.getIconBlob()
-                )
-        );
-        Collection<ActivityInstanceSummary> activitySummaries = new ArrayList<>();
+        ActivityDefStore activityDefStore = ActivityDefStore.getInstance();
+        Map<String, Blob> formTypeAndStatusTypeToIcon = activityDefStore.findActivityStatusIcons(handle, studyGuid);
```
yufengwng, line 270: Key insight here is to cache the icons since they rarely change (if at all) instead of querying for them every time we invoke this method. This saves several milliseconds.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
```diff
@@ -379,6 +356,51 @@ public Long getActivityInstanceIdByGuid(Handle handle, String guid) {
...
+     * @param activitySummaries the list of activity summaries
+     */
+    public void countActivitySummaryQuestionsAndAnswers(Handle handle, String userGuid, String studyGuid,
+                                                        List<ActivityInstanceSummary> activitySummaries) {
+        ActivityDefStore activityDefStore = ActivityDefStore.getInstance();
+
+        Set<String> instanceGuids = activitySummaries.stream()
+                .map(ActivityInstanceSummary::getActivityInstanceGuid)
+                .collect(Collectors.toSet());
+        Map<String, FormResponse> instanceResponses = handle
```
yufengwng, line 374: Before, each call to `countQuestionsAndAnswers()` will query for the instance answer data. Key insight here is to do a single query for all the instance answer data we need for the activity summaries rather than individual queries. And now that we have the data loaded, there's an opportunity to leverage those. That will be in upcoming optimizations.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserActivityInstanceListRoute.java
```diff
@@ -60,6 +60,7 @@ public UserActivityInstanceListRoute(ActivityInstanceDao activityInstanceDao) {
             // IMPORTANT: do numbering before filtering so each instance is assigned their correct number.
             performActivityInstanceNumbering(summaries);
             summaries = filterActivityInstancesFromDisplay(summaries);
+            activityInstanceDao.countActivitySummaryQuestionsAndAnswers(handle, userGuid, studyGuid, summaries);
```
yufengwng, line 63: Some of the instances will get filtered out if they are hidden or otherwise not shown. So the key insight here is to do the counting after the filtering, so to save us some work.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
yufengwng: Upon closer look locally, the performance here of querying all instances and then doing the counting seems reasonable. The counting part dominants the runtime of the list-activities API, but locally it's averaging < 100 ms on a warm cache, so I think I will leave it for now and do other optimizations instead.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityInstanceDao.java
```diff
@@ -379,6 +356,51 @@ public Long getActivityInstanceIdByGuid(Handle handle, String guid) {
...
+                FormActivityDef formActivityDef = activityDefStore.getActivityDef(studyGuid, activityCode, versionTag);
+                if (formActivityDef == null) {
+                    ActivityDto activityDto = handle.attach(JdbiActivity.class)
+                            .findActivityByStudyGuidAndCode(studyGuid, activityCode).get();
+                    formActivityDef = handle.attach(FormActivityDao.class).findDefByDtoAndVersion(
+                            activityDto, versionTag, versionId, revisionStart);
+                    activityDefStore.setActivityDef(studyGuid, activityCode, versionTag, formActivityDef);
+                }
+
+                Pair<Integer, Integer> questionAndAnswerCounts = activityDefStore.countQuestionsAndAnswers(
```
ssettipalli, line 395: Skipping question answer counts for studies that don't need this feature should contribute  to some performance improvement.

ssettipalli (approved):
looks good to me

PR #402, opened 2020-10-05
DDP-5049 wip

yufengwng:
This PR builds on PR #398, and aims to speed up loading of activity definitions. The main point is to refactor queries to do bulk queries of sections, blocks, questions, templates, etc. In order to do this, I had to refactor a number of DAO and DTO objects, which causes a ripple effect of having update a large number of files.

Still in the middle of the refactor, but want to open a draft PR to kick off code review. I recommend starting with `FormActivityDao` and working your way through `SectionBlockDao` etc.


yufengwng:
This PR is too big. I will close and break it up.

PR #418, opened 2020-10-11
Add code to test redis connection when an exception has occurred

MocanaAtBroad:
## Context

Two things:
First, we were using 2 different configuration for caches. And one of them was using a very long timeout (30s) which has been causing problems in production.
So one change makes sure we use the same settings for all Redis caches.

Second, going through logs, it appears frequently when a Redis problem occurs, there are other errors in the logs that suggest we have network connectivity issues; have seen Redis errors coinciding with tcell not being able to read from its own servers and logback logging not being able to contact Slack.

Wherever we detect a RedisException we try to run a simple test: using a completely different Redis library, we try to read and write a value.

If a RedisException occurs, but we have connectivity, then we can't blame the VPC or any other networking system.

Code is self-explanatory.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,6 +124,9 @@ commands:
             develop)
               DEPLOY_ENV=dev
             ;;
+            RedisConnectionHandling)
+              DEPLOY_ENV=dev
+            ;;
             test)
               DEPLOY_ENV=test
             ;;
@@ -824,6 +827,7 @@ workflows:
                 - develop
                 - &rc /^rc.*/
                 - &hotfix /^hotfix.*/
+                - RedisConnectionHandling
       - parallel-compile-and-test-job:
           parallelism: *test_parallelism
           <<: *compile_and_test_filters
@@ -834,6 +838,7 @@ workflows:
             branches:
               only:
                 - develop
+                - RedisConnectionHandling
           requires:
             - parallel-compile-and-test-job
             - compile-and-run-test-suite-job
--- a/pepper-apis/config/redisson-jcache.yaml.ctmpl
+++ b/pepper-apis/config/redisson-jcache.yaml.ctmpl
@@ -10,12 +10,12 @@ singleServerConfig:
   retryInterval: 1500
   password: null
   subscriptionsPerConnection: 5
-  clientName: null
+  clientName: pepper-backend
   address: "redis://{{$conf.Data.redis.host}}:{{$conf.Data.redis.port}}"
   subscriptionConnectionMinimumIdleSize: 1
   subscriptionConnectionPoolSize: 50
   connectionMinimumIdleSize: 24
-  connectionPoolSize: 86
+  connectionPoolSize: 64
   database: 0
   dnsMonitoringInterval: 5000
   pingConnectionInterval: 15000
--- a/pepper-apis/config/redisson-jcache.yaml.ctmpl
+++ b/pepper-apis/config/redisson-jcache.yaml.ctmpl
@@ -10,7 +10,7 @@ singleServerConfig:
   retryInterval: 1500
   password: null
   subscriptionsPerConnection: 5
-  clientName: pepper-backend
+  clientName: null
   address: "redis://{{$conf.Data.redis.host}}:{{$conf.Data.redis.port}}"
   subscriptionConnectionMinimumIdleSize: 1
   subscriptionConnectionPoolSize: 50
--- a/pepper-apis/parent-pom.xml
+++ b/pepper-apis/parent-pom.xml
@@ -452,6 +452,11 @@
             <version>3.13.4</version>
         </dependency>
         <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>3.3.0</version>
+        </dependency>
+        <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-common</artifactId>
             <version>4.1.50.Final</version>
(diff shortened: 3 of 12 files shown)


.circleci/config.yml
```diff
@@ -834,6 +838,7 @@ workflows:
             branches:
               only:
                 - develop
+                - RedisConnectionHandling
           requires:
```
MocanaAtBroad, line 842: This stuff is running on dev environment. These changes should be removed.

pepper-apis/config/redisson-jcache.yaml.ctmpl
```diff
@@ -15,7 +15,7 @@ singleServerConfig:
   subscriptionConnectionMinimumIdleSize: 1
   subscriptionConnectionPoolSize: 50
   connectionMinimumIdleSize: 24
-  connectionPoolSize: 86
+  connectionPoolSize: 64
   database: 0
```
MocanaAtBroad, line 19: We have two separate Redis clients being created. Let's back off a bit on the connections. There has been no evidence that this was the problem in the first place.

pepper-apis/parent-pom.xml
```diff
@@ -451,6 +451,11 @@
             <artifactId>redisson</artifactId>
             <version>3.13.4</version>
         </dependency>
+        <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>3.3.0</version>
+        </dependency>
```
MocanaAtBroad, line 458: Completely separate library to talk to Redis. Just to eliminate possibility that using the same library (Redisson) could mess up our test. Can go away if we verify that connectivity is not a problem

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/CacheService.java
```diff
@@ -68,12 +69,13 @@ private CacheService() {
             }
             URI redissonConfigPathUri = redissonConfigPath.toUri();
             cacheManager = buildCacheManager(redissonConfigPathUri);
-            String redisAddress = ConfigManager.getInstance().getConfig().getString(ConfigFile.REDIS_SERVER_ADDRESS);
-            Config redissonConfig = new Config();
-            redissonConfig.useSingleServer()
-                    .setTimeout(30000)
```
MocanaAtBroad, line 74: This is where the 30s timeouts where coming from. We were also picking up all the default Redisson configuration settings for the local caches instead of the tweaked settings we were trying out to get past the errors.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/AnswerCachedDao.java
```diff
@@ -104,6 +105,7 @@ public void deleteAllByInstanceIds(Set<Long> instanceIds) {
             } catch (RedisException e) {
                 LOG.warn("Failed to retrieve value from Redis cache: " + idToAnswerCache.getName() + " key lookedup:"
                         + answerId + "Will try to retrieve from database", e);
+                TestRedisConnection.doTest();
```
MocanaAtBroad: Added this line wherever we catch RedisException

pepper-apis/src/main/java/org/broadinstitute/ddp/util/TestRedisConnection.java
```diff
@@ -0,0 +1,37 @@
...
+        String redisAddress = ConfigManager.getInstance().getConfig().getString(ConfigFile.REDIS_SERVER_ADDRESS);
+        String host = StringUtils.substringBetween(redisAddress, "//", ":");
+        String portString = StringUtils.substringAfterLast(redisAddress, ":");
+        doTest(host, Integer.parseInt(portString));
+    }
+
+    public static void doTest(String host, int port) {
+        String testHash = "testhash";
+        String valToWrite = Math.random() + "";
+        try (var jedis = new Jedis(host, port)) {
```
MocanaAtBroad: Using Jedis instead or Redisson. Jedis does not have a cache implementation. (Though Spring caching uses Jedis in their Redis cache implementation).

pepper-apis/src/main/java/org/broadinstitute/ddp/cache/CacheService.java
zyme: Nice!

pepper-apis/src/main/java/org/broadinstitute/ddp/util/TestRedisConnection.java
```diff
@@ -0,0 +1,37 @@
+package org.broadinstitute.ddp.util;
+
+import org.apache.commons.lang.StringUtils;
+import org.broadinstitute.ddp.constants.ConfigFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import redis.clients.jedis.Jedis;
+
+public class TestRedisConnection {
```
zyme: How about `RedisConnectionValidator`?  Not sure if other folks are using their own class name regexes for things like intellij test profiles--starting or ending a class name with `Test` might be confusing.

zyme (dismissed):
Looks good!

pepper-apis/src/main/java/org/broadinstitute/ddp/util/TestRedisConnection.java
MocanaAtBroad: 1ff7ed3e02d726d6247aab104ad3151737017fa5

zyme (approved):

PR #423, opened 2020-10-14
Ping Redis when ERROR is logged

MocanaAtBroad:
## Context
Sorry this includes a lot of the same files from the previous PR.
Only 2 files are different and matter for this PR (the other stuff is the same)
```
logback.xml
RedisConnectionValidatorAppender
```
**Note**: Have not seen any Redis errors in dev env since these changes were deployed mid-day Tuesday. And weirdly, no Redis errors in prod either. For that matter, no 500 for PATCH or PUTs calls either.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/log/RedisConnectionValidatorAppender.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/log/RedisConnectionValidatorAppender.java
@@ -1,0 +1,12 @@
+package org.broadinstitute.ddp.log;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.AppenderBase;
+import org.broadinstitute.ddp.util.RedisConnectionValidator;
+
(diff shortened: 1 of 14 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/log/RedisConnectionValidatorAppender.java
```diff
@@ -0,0 +1,12 @@
...
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.AppenderBase;
+import org.broadinstitute.ddp.util.RedisConnectionValidator;
+
+public class RedisConnectionValidatorAppender extends AppenderBase<ILoggingEvent> {
+    @Override
+    protected void append(ILoggingEvent eventObject) {
+        RedisConnectionValidator.doTest();
+    }
+}
```
MocanaAtBroad, line 12: Not a real appender. But if Redisson library logs an ERROR (particularly the libraries own PINGs which we do not call explicitly), we will do our own ping of Redis.

zyme (approved):
Looks good!

PR #425, opened 2020-10-14
More test boston translations

ssettipalli:
Added more translations for spanish and creole
Added 6 translations. Translations for some 63 entries are missing in the Document for spanish.
Added some 120+ translations for Creole.
Used google translator to translate "Other" and "Please Specify"  for Creole.
Added translations to creole emails


MocanaAtBroad (approved):
LGTM. Let's see what happens when we put it in dev.

PR #430, opened 2020-10-17
Increase Java Heap and turn on profiling for housekeeping in prod

MocanaAtBroad:
In case we want to pursue this angle further, I have created this PR.

There might be a memory leak rearing its head in prod.
Turn on GCP profiler to investigate.
Also memory setting was way low. Increased to match settings used in pepper-backend.


zyme (approved):
Let's give it a shot!

PR #443, opened 2020-10-21
DDP-5182: limit fetch size of kit queries

yufengwng:
## Context

Enable fetch size on MySQL connector and limit size for kit queries.

By default, MySQL does not support fetch size, so you have to manually enable it. When fetch size is set to > 0, I believe this uses the so-called "cursor-based fetching", in which rows are buffered on the db server side in a temporary table and a set number of rows are fetched each time. See [the `useCursorFetch` option here](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html), [the ResultSet section here](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-implementation-notes.html), and [this SO answer](https://stackoverflow.com/a/36986532).

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
@@ -372,6 +372,7 @@ public class TransactionWrapper {
         config.addDataSourceProperty("prepStmtCacheSize", "250");
         config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
         config.addDataSourceProperty("cacheServerConfiguration", "true");
+        config.addDataSourceProperty("useCursorFetch", "true");
         config.setInitializationFailTimeout(0); // don't fail to  create the pool if the db is not available
         config.setAutoCommit(true); // will be managed by jdbi, which expects autcommit to be enabled initially
         config.setTransactionIsolation("TRANSACTION_READ_COMMITTED");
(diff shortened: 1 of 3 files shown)


zyme (approved):
Overall looks good!  Let's keep the performance schema enabled on prod for a while after this goes out and see if we notice any unintended side effects with temp tables.

Would it be helpful to add some logging around the use of this query (starting/completed) to help troubleshooting?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -372,6 +372,7 @@ private HikariDataSource createDataSource(int maxConnections, String dbUrl, DB d
         config.addDataSourceProperty("prepStmtCacheSize", "250");
         config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
         config.addDataSourceProperty("cacheServerConfiguration", "true");
+        config.addDataSourceProperty("useCursorFetch", "true");
```
MocanaAtBroad, line 375: Can't figure out what this option does when we don't use the setFetchSize option for queries. Do you know?

yufengwng: Based on the documentation that I linked in the PR description, it uses cursor-based fetching when `setFetchSize() > 0`. So when we don't have `setFetchSize()` I'll assume it uses default behavior, which is fetching all the rows in resultset.

PR #451, opened 2020-10-22
DDP-5181 Housekeeping email bug and reducing memory use from invitations in export

zyme:
Bundling some already reviewed changes into a single hotfix


yufengwng (approved):
LGTM

PR #453, opened 2020-10-23
DDP-5183 use first sent timestamp as initial kit sent time

yufengwng:
## Context

Remove tracking of initial kit in Pepper and instead use the first kit sent event as initial kit sent time for scheduling next kits.

## Checklist

## FUD Score

## Testing

## Release


ssettipalli (approved):
looks good to me.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ReceiveDsmNotificationRoute.java
```diff
@@ -79,6 +87,11 @@ public Object handle(Request request, Response response, DsmNotificationPayload
                 testResult = parseTestResult(response, payload);
             }
 
+            if (eventType == DsmNotificationEventType.TESTBOSTON_SENT) {
```
zyme, line 90: How specific is this to tbos, and should we consider using a lookup here to determine that it's for tbos?  My concern here is the potential for a patten of study-specific events over the long haul.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ReceiveDsmNotificationRoute.java
```diff
@@ -124,4 +137,25 @@ private TestResult parseTestResult(Response response, DsmNotificationPayload pay
...
+                        studyDto.getGuid(), kitType.getName())));
+
+        var kitScheduleDao = handle.attach(KitScheduleDao.class);
+        var record = kitScheduleDao.findRecord(user.getId(), kitConfigId).orElse(null);
+
+        if (record != null && record.getInitialKitSentTime() == null) {
+            kitScheduleDao.updateRecordInitialKitSentTime(record.getId(), timestamp);
+            LOG.info("Updated initial kit sent time to {} for participant={}, study={}, kitConfigurationId={}",
+                    timestamp, user.getGuid(), studyDto.getGuid(), kitConfigId);
+        }
```
zyme, line 159: Maybe a `LOG.error()` in the else if we get a message here > 1 for a kit that already has a sent time?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ReceiveDsmNotificationRoute.java
yufengwng: `TESTBOSTON` is a kit type so this is specific to kit types rather than studies. We have other kit types like `SALIVA` and `BLOOD`. We do have the `SALIVA_SENT` event, but no `BLOOD_SENT` event. We can add support for those other kit types when they get used for scheduling longitudinal kits.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ReceiveDsmNotificationRoute.java
yufengwng: Actually, we're not tracking the status _per kit_. Pepper doesn't really need to know that a particular kit is sent. It just needs to know that the first kit, whichever it is, is sent so it can schedule the longitudinal kits. The first kit may be a regular kit or it may be an uploaded kit that only DSM knows about. Also, we WILL be getting > 1 `TESTBOSTON_SENT` events since we have longitudinal kits, so I think adding `log.error()` will be a false positive.

zyme: Ah, got it.  Thanks for helping me understand.

yufengwng: In terms of tracking kit status in order to filter out duplicates and such, I agree on that. I had created a ticket for it a while back: DDP-5004.

PR #456, opened 2020-10-24
Save Housekeeping Java Heap Dumps to Google Bucket

MocanaAtBroad:
This PR enables ~~a POST route~~ a task that runs when a run task request is received via pub/sub message. Housekeeping that will call a java heap dump programmatically, copy the dump to a bucket, and then delete local copy.

For review. Can decide whether we want to use or not.

~~You can generate a heap dump with following curl command:~~
You can generate a heap dump with a `gcloud` command like the one below:
```
gcloud --project=broad-ddp-dev pubsub topics publish housekeeping-tasks --attribute 'type=GENERATE_HEAP_DUMP' 
```

Has been working on dev. Here are the heaps that we have saved so far in dev environment:
https://console.cloud.google.com/storage/browser/broad-ddp-dev-heap-dumps/heap_dumps;tab=objects?authuser=1&project=broad-ddp-dev&prefix=&forceOnObjectsSortingFiltering=false

Unfortunately(?) housekeeping has been pretty stable there and after examining the heaps in YourKit tool it is not clear if there is a leak. Could be that production (and maybe staging?) is performing tasks that dev is not?

**Important!** Heap is saved to local /tmp directory. My understanding is that /tmp in GAE is a fake inmemory file system, so I modified Java memory setting at launch to be smaller to leave space for dump file. Seeing sizes of 100 to 140MB in dev. Gonna guess in prod they will be larger.


.circleci/config.yml
```diff
@@ -124,7 +124,7 @@ commands:
             develop)
               DEPLOY_ENV=dev
             ;;
-            RedisConnectionHandling)
+            TrySavingHeaps)
               DEPLOY_ENV=dev
```
MocanaAtBroad, line 128: Building/deploying on develop to try and get this right.

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -511,6 +512,32 @@ private static void respondToGAEStartHook(String envPort) {
             response.status(200);
             return "";
         });
+        Spark.post("/heapdump", (request, response) -> {
+            String postedPassword = request.queryParams("password");
+            if (StringUtils.isNotEmpty(postedPassword)) {
+                String configuredPwd = ConfigManager.getInstance().getConfig().getString(ConfigFile.HEALTHCHECK_PASSWORD);
+                if (!configuredPwd.equals(postedPassword)) {
```
MocanaAtBroad: Wanted a little bit of security if we put this in production. Healthcheck password was already in config so why not?

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -511,6 +512,32 @@ private static void respondToGAEStartHook(String envPort) {
...
+                }
+            } else {
+                LOG.error("Missing password for heapdump call");
+                response.status(404);
+                return "";
+            }
+            try {
+                String projectId = ConfigManager.getInstance().getConfig().getString(ConfigFile.GOOGLE_PROJECT_ID);
+                String bucketName = projectId + "-heap-dumps";
+                new JavaHeapDumper().dumpHeap(projectId, bucketName);
```
MocanaAtBroad: Each env will get their corresponding bucket. Did not want to mess with permissions to allow prod to save to dev buckets or vice-versa. Only dev bucket exists at this time.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/JavaHeapDumper.java
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+
+    }
+
+    private String generateDumpFileName() {
+        String gcpFileName = generateGcpFileName();
+        return gcpFileName != null ? gcpFileName : nonGcpFileName();
+    }
+
```
MocanaAtBroad, line 69: Wanted to be able to try locally first.

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -525,9 +552,9 @@ private static void respondToGAEStartHook(String envPort) {
                 LOG.warn("Wait interrupted", e);
             }
         }
-
-        Spark.stop();
-        Spark.awaitStop();
+        // Commenting out while we keep heapdump functionality
+        //        Spark.stop();
+        //        Spark.awaitStop();
```
yufengwng: Instead of opening an HTTP endpoint, how about we add a `HEAP_DUMP` task instead? It seems to fit as an "operational task". You should be able to easily add it to `HousekeepingTaskReceiver`. Security-wise, only folks who has the right access in GCP can publish to the pubsub topic. See https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/1261797381/Housekeeping+Operational+Tasks.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/JavaHeapDumper.java
```diff
@@ -0,0 +1,81 @@
...
+    private static Logger LOG = LoggerFactory.getLogger(JavaHeapDumper.class);
+    public static final String DEFAULT_LOCAL_PATH = "/tmp";
+    public static final String DEFAULT_BUCKET_PATH = "heap_dumps";
+
+    public void dumpHeapToLocalFile(String filePath) throws IOException {
+        MBeanServer server = ManagementFactory.getPlatformMBeanServer();
+        HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy(
+                server, "com.sun.management:type=HotSpotDiagnostic", HotSpotDiagnosticMXBean.class);
+        LOG.info("About to create local file at:" + filePath);
+        mxBean.dumpHeap(filePath, true);
```
zyme, line 31: Woah!  Did not realize one could use mxbean stuff to dump heap.  Very cool!

pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
yufengwng: Production Housekeeping does not have a SparkJava HTTP server running, while it does have a pubsub subscriber for these tasks. So I'll recommend going this route to get a better reflection of memory usage.

MocanaAtBroad: I agree using pub sub would be more in the spirit of Housekeeping. And the more I think about it, it is probably worth the additional effort. @zyme @yufengwng would you guys agree that keeping the on-demand heap dumps as permanent part of system is worthwhile? I had been thinking of this as tactical/do it and forget it capability, but I now believe we should keep this available for the next time too.

MocanaAtBroad: 3e73c4b34b38a4da66fff83a4471a559c71588b5

zyme: Having this as a permanent option seems great!

yufengwng (approved):
LGTM

PR #458, opened 2020-10-26
DDP-5196 paginate kit query and narrow transactions

yufengwng:
## Context

Paginate kit query results so we control our memory usage. Narrowing the scope of transactions so we better handle failures.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
@@ -86,9 +86,9 @@ public class KitCheckService {
 
             LOG.info("Processing potential kit recipients for study {} and kit configuration {}", studyGuid, kitConfigId);
 
-            int processed = 0;
+            int fetched = 0;
             while (true) {
-                int offset = processed;
+                int offset = fetched;
                 Queue<PotentialRecipient> batch = withAPIsTxn(handle -> new ArrayDeque<>(
                         findPotentialKitRecipients(handle, studyGuid, kitTypeId, offset, batchSize)));
                 int fetchedSize = batch.size();
@@ -103,11 +103,11 @@ public class KitCheckService {
                         LOG.error("Error while checking potential kit recipient {}, continuing", candidate.getUserGuid(), e);
                     }
                 }
-                processed += fetchedSize;
+                fetched += fetchedSize;
             }
 
             LOG.info("Finished processing {} potential kit recipients for study {} and kit configuration {}",
-                    processed, studyGuid, kitConfigId);
+                    fetched, studyGuid, kitConfigId);
         }
 
         return kitCheckResult;
@@ -187,9 +187,9 @@ public class KitCheckService {
             String studyGuid = kitConfig.getStudyGuid();
             LOG.info("Checking kit schedule records for study {} and kit configuration {}", studyGuid, kitConfigId);
 
-            int processed = 0;
+            int fetched = 0;
             while (true) {
-                int offset = processed;
+                int offset = fetched;
                 Queue<PendingScheduleRecord> batch = withAPIsTxn(handle -> new ArrayDeque<>(
                         handle.attach(KitScheduleDao.class)
                                 .findPendingScheduleRecords(kitConfig.getId(), offset, batchSize)));
@@ -205,11 +205,11 @@ public class KitCheckService {
                         LOG.error("Error while checking kit schedule record for participant {}, continuing", record.getUserGuid(), e);
                     }
                 }
-                processed += fetchedSize;
+                fetched += fetchedSize;
             }
 
             LOG.info("Finished processing {} kit schedule records for study {} and kit configuration {}",
-                    processed, studyGuid, kitConfigId);
+                    fetched, studyGuid, kitConfigId);
         }
 
         return kitCheckResult;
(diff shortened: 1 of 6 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
```diff
@@ -63,123 +62,162 @@ public KitCheckService(int batchSize) {
...
      *         participants.
      */
-    public KitCheckResult checkForInitialKits(Handle apisHandle) {
+    public KitCheckResult checkForInitialKits() {
         KitCheckResult kitCheckResult = new KitCheckResult();
 
-        KitScheduleDao kitScheduleDao = apisHandle.attach(KitScheduleDao.class);
-        DsmKitRequestDao kitRequestDao = apisHandle.attach(DsmKitRequestDao.class);
-        KitConfigurationDao kitConfigurationDao = apisHandle.attach(KitConfigurationDao.class);
+        Queue<KitConfiguration> kitConfigs = withAPIsTxn(handle -> new ArrayDeque<>(
```
yufengwng, line 72: Using a `Queue` so we remove things as we go instead of storing the whole list in memory the whole time. `ArrayDeque` seems to be a fast implementation of a queue, and it's more memory-efficient since it doesn't need to do any copies during a `remove()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
```diff
@@ -406,4 +449,8 @@ public void add(KitCheckResult other) {
             }
         }
     }
+
+    <R, X extends Exception> R withAPIsTxn(HandleCallback<R, X> callback) throws X {
+        return TransactionWrapper.withTxn(TransactionWrapper.DB.APIS, callback);
+    }
```
yufengwng, line 455: I had to do this thing of wrapping the open transaction call so we can mock it in the tests.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
MocanaAtBroad: Maybe this approach of using limit and offset could be packaged up into a utility method that you give it the query and it gives you a query?

pepper-apis/src/main/resources/org/broadinstitute/ddp/service/KitCheckService.sql.stg
```diff
@@ -27,4 +27,6 @@ select user.user_id,
           and kit.study_id = study.umbrella_study_id
        )
    and valid_to is null
+ order by user.user_id
+ limit :limit offset :offset
```
MocanaAtBroad, line 31: Since you will be consuming the users in incrementally instead of one big transaction, it could happen that your overall result set (i.e., the set of all rows) you are trying to consume could change while you are iterating through them.
Consider using user_study_enrollment.valid_from as the ORDER BY key . Presumably that will put new enrollees at the end of the result set. `user_id` might or might not.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
```diff
@@ -63,123 +62,162 @@ public KitCheckService(int batchSize) {
...
-            if (!batch.isEmpty()) {
-                scheduleNextKitsForBatch(apisHandle, kitCheckResult, kitConfig, batch);
+                while (!batch.isEmpty()) {
+                    PendingScheduleRecord record = batch.remove();
+                    try {
+                        scheduleNextKitForParticipant(kitCheckResult, kitConfig, record);
+                    } catch (Exception e) {
+                        LOG.error("Error while checking kit schedule record for participant {}, continuing", record.getUserGuid(), e);
+                    }
+                }
```
MocanaAtBroad, line 207: Might be possible to package this type of iteration of set over multiple queries and multiple transactions into a utility. No need to do it now, but if this becomes a recurring pattern...

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
yufengwng: Perhaps. But at first glance the tricky thing seems to be how we do order things. Some queries might order by `user_id`, while others should be ordered by something totally different. Not sure if that can be made generic. Also, having a utility for limit/offset seems a bit overkill, maybe it's just me.

pepper-apis/src/main/resources/org/broadinstitute/ddp/service/KitCheckService.sql.stg
yufengwng: I liked ordering by `user_id` since that's deterministic. But I see your point about new enrollments. I had assumed that new users coming into study will be new user (i.e. new user ids). Perhaps there might be a chance that old users (hence, old user id) can come in to. I guess it will be safer to BOTH `order by valid_from asc, user_id asc`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
yufengwng: Yup, it's certainly something I have thought about since I've written it a few times. But haven't found the right pattern to expose this as a utility yet.

pepper-apis/src/main/resources/org/broadinstitute/ddp/service/KitCheckService.sql.stg
yufengwng: Just saw your other comment in other PR. I like `user_study_enrollment_id`, I'll use that instead.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
yufengwng: I have a vague idea of exposing this kind of pattern as an `Iterator`, but I'll flush that out in a separate refactor later.

MocanaAtBroad (approved):

PR #465, opened 2020-10-28
Close all JDBI streams

MocanaAtBroad:
Continuing to try to find source of memory leaks in housekeeping I kept finding JDBC objects in heap dumps.

Looking at the docs below, it appears that we might have been mishandling JDBI result streams:

Found this interesting tidbit in JDBI docs: https://jdbi.org/#__sqlquery
Screen Shot 2020-10-27 at 10 32 43 PM

I tried to find every place where we generate a stream of JDBI results and make sure it is closed.

To be clear: I don't know if this the root cause of our memory leak, but it appears to be a problem.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/StudyGovernanceDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/StudyGovernanceDao.java
@@ -17,6 +17,7 @@ import org.jdbi.v3.sqlobject.CreateSqlObject;
 import org.jdbi.v3.sqlobject.SqlObject;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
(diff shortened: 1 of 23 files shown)


yufengwng (approved):
Let's give it a try!

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/StudyGovernanceDao.java
```diff
@@ -99,6 +100,7 @@ default int removeAgeUpCandidates(Set<Long> candidateIds) {
     @UseStringTemplateSqlLocator
     @SqlQuery("queryAllAgeUpCandidatesByStudyId")
     @RegisterConstructorMapper(AgeUpCandidate.class)
+    @FetchSize(300)
```
zyme, line 103: Just flagging this since it wasn't mentioned in the overview.  Seems like a nice incidental improvement!

zyme (approved):
I'm optimistic!  Thanks for finding that jdbi nugget!

PR #467, opened 2020-10-28
DDP-5224 cache Elasticsearch clients

yufengwng:
## Context

I saw ES export in `test` environment timeout a few times. I'm not sure root cause, but I realized that the ES clients holds quite a few underlying resources and that we're creating a new client on every batch of participants. Comparing memory usage and "time to complete exports", I didn't notice any major differences at first glance, so I don't have any hard evidence that this helps improve things. However, this intuitively feels like a beneficial thing to do.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
@@ -3,8 +3,10 @@ package org.broadinstitute.ddp.export;
 import static org.broadinstitute.ddp.model.activity.types.ComponentType.MAILING_ADDRESS;
 
 import java.io.BufferedWriter;
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.Writer;
+import java.net.MalformedURLException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Instant;
@@ -124,7 +126,7 @@ import org.jdbi.v3.core.Handle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class DataExporter {
+public class DataExporter implements Closeable {
 
     public static final String TIMESTAMP_PATTERN = "MM/dd/yyyy HH:mm:ss";
     public static final DateTimeFormatter TIMESTAMP_FMT = DateTimeFormatter
@@ -140,6 +142,7 @@ public class DataExporter {
     private Gson gson;
     private Set<String> componentNames;
     private PdfService pdfService;
+    private RestHighLevelClient esClient;
 
     public static String makeExportCSVFilename(String studyGuid, Instant timestamp) {
         DateTimeFormatter fmt = DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmssX").withZone(ZoneOffset.UTC);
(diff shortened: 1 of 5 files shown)


ssettipalli (dismissed):
looks good

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -140,6 +142,7 @@
     private Gson gson;
     private Set<String> componentNames;
     private PdfService pdfService;
+    private RestHighLevelClient esClient;
```
zyme: Maybe make this `final` since `close()` assumes non-null?

ssettipalli (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -124,7 +126,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class DataExporter {
+public class DataExporter implements Closeable {
```
MocanaAtBroad, line 129: Not seeing DataExporter being closed . Intentional?

MocanaAtBroad: And if you do intend to close() DataExporter, then consider instantiating a DataExporter each time you `run` a Job as opposed to hanging on to the instance time after time and allow GC to reclaim whatever resources are linked back to this.
(Looking at heap dump, I think we are indirectly creating lots of threads that don't  go away. The underlying http client here might be part of this problem).

MocanaAtBroad (commented):
Sorry for belated comments!

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
yufengwng: > Not seeing DataExporter being closed . Intentional?

The DataExporter is mostly used within background jobs that exports data, so those literally hang around for the lifetime of Housekeeping. I'm implicitly letting this get closed by the JVM when Housekeeping gets destroyed.

> consider instantiating a DataExporter each time you run a Job
> ...
> I think we are indirectly creating lots of threads that don't go away

Actually, I want to do the opposite. I don't want to create new DataExporter on every `run` of a job. It's wasteful and I suspect that is causing the `indirectly creating lots of threads that don't go away` that you might have observed. Instead, we're creating the DataExporter just once (or just a few times..) and reusing that.

MocanaAtBroad: OK. My concern more specifically: the elasticsearch client.
Underneath implementation it creates asynchronous apache http clients.
The worry here is that each elastic search client ends up indirectly creating a separate thread pool. We are creating more threads as we go. Heap analysis is showing me that.
So two ways to protect from building up threads unnecessarily:
1) Go with a true singleton elastic search client (Apparently that is their recommendation) 
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_changing_the_client_8217_s_initialization_code.html
2) Create it use and dispose of it to reduce chance of having multiple ones running at same time.

yufengwng: Good find! That can be our next step in improvement! I tried digging deeper into the implementation but wasn't too familiar with the async/reactor/NIO stuff. I saw that there was a limit of "max connections" and "max connections per route", so I though there was a upper limit on that. I might be wrong there. I think I'll vote for option 1 of having a singleton.

MocanaAtBroad: Yeah. That sounds like the better choice.

yufengwng: Great! I'll create a separate PR for that.

PR #472, opened 2020-10-30
Brain pos-consent race changes

ssettipalli:
Look at https://broadinstitute.atlassian.net/browse/DDP-5168 and
https://broadinstitute.atlassian.net/browse/DDP-5040

Self/Adult POSTCONSENT version 2 (Patch) with updated RACE question (nested picklist options).

Update CHILD_POSTCONSENT (not yet in prod) RACE question with nested picklist options.

Update status icons


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/ActivityBuilder.java
```diff
@@ -239,6 +239,35 @@ private void insertActivityStatusIcons(Handle handle) {
         }
     }
 
+    public void updatetActivityStatusIcons(Handle handle) {
```
yufengwng: Typo: `updatet`.

yufengwng (dismissed):

yufengwng (approved):

PR #474, opened 2020-10-30
DDP-5110 slack unfurling and dialing back slack overload

zyme:
## Context

DDP-5510 removes link unfurling and moves `SlackAppender` to a lossy, limited way to emit less noise to our slack alert channels.  Along the way, since @MocanaAtBroad nudged me in this direction, I ripped out the `jslack` and replaced with some simple http calls.

I also made a better test, which necessitated upgrading our mockserver version, which had a knock-on effect of needing to upgrade apache commons.

Since my loaner laptop is a mess, I have not been able to run all the tests to see what else breaks as a result of the pom changes.

## Checklist

- [ x] I have labeled the type of changes involved using the `C-*` labels.
- [ x] I have assessed potential risks and labeled using the `R-*` labels.
- [ x] I have considered error handling and alerts, and added `L-*` labels as needed.
- [ x] I have considered security and privacy, and added `I-*` labels as needed
- [ x] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

## Testing

## Release


pepper-apis/parent-pom.xml
```diff
@@ -134,7 +129,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.4</version>
+            <version>3.11</version>
```
zyme, line 132: Needed to bump this for the upgrade of mockserver.

pepper-apis/parent-pom.xml
```diff
@@ -351,7 +346,23 @@
         <dependency>
             <groupId>org.mock-server</groupId>
             <artifactId>mockserver-netty</artifactId>
-            <version>5.5.1</version>
+            <version>5.11.0</version>
```
zyme, line 349: Upgraded here so I could call `hasStarted()` during the test.

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
```diff
@@ -115,49 +149,71 @@ private String getStringifiedStackTrace(ILoggingEvent e) {
...
+                    .uri(slackHookUrl)
+                    .POST(HttpRequest.BodyPublishers.ofString(new Gson().toJson(payload)))
+                    .header(HttpHeaders.CONTENT_TYPE, ContentType.APPLICATION_JSON.getMimeType())
+                    .build();
         try {
-            response = slack.send(slackHookUrl, payload);
-            if (response.getCode() != 200) {
-                throw new IOException("Could not post " + payload + " to slack.  Hook returned " + response
-                        .getMessage());
+            var httpClient = HttpClient.newHttpClient();
```
zyme: @MocanaAtBroad taking your advice here and just dropping down to the API directly...

zyme: I did attempt a different approach earlier that involved closing up 3 different potential resources in `okhttp`, wrapped up in `jslack`.  This is much simpler.

pepper-apis/src/main/java/org/broadinstitute/ddp/thread/ThreadFactory.java
```diff
@@ -0,0 +1,24 @@
+package org.broadinstitute.ddp.thread;
+
+/**
+ * Simple {@link java.util.concurrent.ThreadFactory} that lets
+ * you set the name for the group and the priority.
+ */
+public class ThreadFactory implements java.util.concurrent.ThreadFactory {
```
zyme, line 7: Brought over from Birch--nice to de-propritize slack thread stuff so that pepper can continue doing "real work" when it's crunch time.

pepper-apis/src/test/java/org/broadinstitute/ddp/SlackAppenderTest.java
```diff
@@ -1,43 +0,0 @@
-package org.broadinstitute.ddp;
```
zyme, line 1: Moved this to the right package: `log`.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GsonUtilTest.java
```diff
@@ -4,13 +4,18 @@
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import org.broadinstitute.ddp.transformers.Exclude;
 import org.broadinstitute.ddp.util.GsonUtil;
-import org.json.JSONException;
```
zyme, line 9: Looks like the `org.json` was a transitive dep that was coming from `jslack`.  Once that library was gone, we needed to find something else to use here, so I went for `gson`.

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
MocanaAtBroad: Oh! My suggestion was switching to Slack's own official Java library instead the third-party library given how many difficulties it was creating.
But looks like you managed to do it using REST, so awesome! Less stuff to worry about!

MocanaAtBroad: I believe this httpclient instances are reusable so you can hang on to it and not instantiate another one.

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
```diff
@@ -43,45 +53,69 @@
...
+        }
+
+        this.channel = slackChannel;
+        if (StringUtils.isNumeric(queueSize)) {
+            this.queueSize = Integer.parseInt(queueSize);
+        } else {
+            this.queueSize = 10;
+        }
+        if (StringUtils.isNumeric(intervalInSeconds)) {
+            this.intervalInMillis = Integer.parseInt(intervalInSeconds);
```
MocanaAtBroad: Unit mismatch?

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
```diff
@@ -43,45 +53,69 @@
...
 
-    /**
-     * Used for testing.  Actual deployments use {@link #SlackAppender() the no-arg constructor}
-     */
-    public SlackAppender(Slack slack, String slackHookUrl, String slackChannel) {
-        this.slack = slack;
-        this.slackHookUrl = slackHookUrl;
-        this.channel = slackChannel;
-    }
+    private List<SlackMessagePayload> messagesToSend = Collections.synchronizedList(new ArrayList<>());
```
MocanaAtBroad, line 56: Consider using one of these instead:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ArrayBlockingQueue.html
I would expect less worries with `synchronized`

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
yufengwng: If we instantiate the `HttpClient` in the constructor, it should be possible to pass one in (e.g. dependency injection). This way, it should be possible to get rid of dependency on mockserver in the tests. I did something similar in `DsmClientTest`.

zyme:
@MocanaAtBroad the intent here is to make the bounded queue non-blocking, so it's lossy.  I suspect that blocking on this appender could cause some weirdness under load.

zyme:
@yufengwng I had attempted to mock without mockserver, but this gets into `BodyPublisher`stuff that seemed more complex than just running mockserver, since a key verification is checking that the right payload went to slack.  Maybe I'm missing a simpler way to check the request payload?

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
```diff
@@ -43,45 +53,72 @@
...
             canLog = false;
         }
-        this.slackHookUrl = slackHook;
-        this.channel = slackChannel;
         if (canLog) {
-            LOG.info("Slack alerts will be sent to " + slackChannel);
+            LOG.info("At most {} slack alerts will be sent to {} every {} ms", queueSize, slackChannel, intervalInMillis);
         }
+
+        ScheduledThreadPoolExecutor executorService = new ScheduledThreadPoolExecutor(1,
```
yufengwng: As a small optimization, perhaps we should move initialization of `httpclient` and `executorService` inside the `if (canLog)` block above? So local development can completely turn of slack appender.

yufengwng (approved):
LGTM

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
zyme: Done!

PR #477, opened 2020-10-31
Make sure Redisson does not log to Slack

MocanaAtBroad:
Forgot to include the magical logback `additivity` attribute. No more Slack ing of PING errors!


zyme (approved):
Woo hoo!

PR #484, opened 2020-11-03
Remove Xms param from Housekeeping java startup command

MocanaAtBroad:
## Context
A new possible scenario for how we could be running out of memory: amount of heap space allocated for Java is more than enough, but there is not enough memory available for non-heap space Java memory and whatever else can reside in the 2048mb that we supposedly have available in GAE instances,

I did a test in dev with the headergrabber app, setting only the -Xmx flag (no Xms flag) just high enough for the simple operations and artificially created an OutOfMemoryError by allocating lots of objects to overwhelm Xmx setting.
The OutOfMemory error was actually reported in GAE logs before it died.

By contrast we have yet to see an OutOfMemory error in GAE (housekeeping or pepper-backend).

So it is possible that housekeeping Xmx heap max is never reached, but since we have been setting Xms to same value, only room for non-heap memory in 
GAE instance is  =2048m - Xms
and it is the non-heap (Java or non-Java. I don't know) that GAE is detecting and using as reason to kill instance.

This is a no risk change: at worst it will be some additional memory management CPU cycles trying to handle the necessary memory allocations.

### DIFF
--- a/pepper-apis/config/Housekeeping.yaml.ctmpl
+++ b/pepper-apis/config/Housekeeping.yaml.ctmpl
@@ -12,9 +12,9 @@ manual_scaling:
 # todo arz give housekeeping and backends different app names for tcell
 
 {{if and (or (eq $environment "prod") (eq $environment "staging") (eq $environment "dev") (eq $environment "test")) (eq $gae "true") }}
-entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=housekeeping,-cprof_cpu_use_per_thread_timers=true,-logtostderr -Xms1340m -Xmx1340m -jar Housekeeping.jar
+entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=housekeeping,-cprof_cpu_use_per_thread_timers=true,-logtostderr -Xmx1340m -jar Housekeeping.jar
 {{else}}
-entrypoint: java -javaagent:tcell/tcellagent.jar -Xms1640m -Xmx1640m -jar Housekeeping.jar
+entrypoint: java -javaagent:tcell/tcellagent.jar -Xmx1640m -jar Housekeeping.jar
 {{end}}
 
 # GAE does not like the "=" character in entrypoint and does not support "." in environment variables


yufengwng:
I have always been suspicious of the `Xms` flag, but your description doesn't quite match my own understanding.

For me, the `Xmx` flag means the maximum heap memory allocated for JVM, while the `Xms` flag means the _initial_ amount of heap at boot. We have always had set `Xms` to the same value as `Xmx`, which I didn't quite understand at first. But I reasoned that our desire was to have the JVM start with the maximum allowed heap right out of the gate, so JVM has memory to spare and spends less time doing memory allocations and more time doing "real work". If we remove the `Xms` flag, then I suppose the JVM will just start with a small amount and eventually climb up to the max allowed heap size.

Like you have mentioned, I'm also aware that the JVM will allocate other memory besides the heap. Obviously there is the stack and JVM's own internal memory needs. I googled these flags just now to refresh myself and reread this [SO answer](https://stackoverflow.com/a/14763095). It mentions that other off-heap memory include thread stacks and native memory. So if anything, I think threads and native methods are more likely suspects rather than `Xms`.

I think the equation you provided should be revised to:
```
off_heap = gae_max - jvm_max
off_heap = (2048m) - (Xmx)
```
Now that I think about it: Since our `Xmx` is set to 1640m, I think that means we have < 500m of off-heap memory for the JVM? Maybe the answer here is that we're exceeding that 500m off-heap memory (e.g., too many threads?) and that's where GAE is crashing our services? Perhaps we should try lowering the `Xmx` flag instead?

bskinner:
One additional fine point to this- `-Xms` also sets the _minimum_ heap size for the JVM, in addition to the _initial_ heap size. Even if the RAM is unused, the JVM can't shrink the heap any further.

> For me, the `Xmx` flag means the maximum heap memory allocated for JVM, while the `Xms` flag means the _initial_ amount of heap at boot. We have always had set `Xms` to the same value as `Xmx`, which I didn't quite understand at first. But I reasoned that our desire was to have the JVM start with the maximum allowed heap right out of the gate, so JVM has memory to spare and spends less time doing memory allocations and more time doing "real work".

yufengwng:
@bskinner Aha! Maybe that explains why the memory usage in App Engine never really goes down that much!?

bskinner:
Yep! Another thing to take a look at is tweaking the -XX:MaxDirectMemorySize. By default, if that flag is not set or the value is <= 0, then the MaxDirectMemorySize (used for off-heap allocation) is set to the same value as -Xmx, so we could be implicitly setting a much higher limit than what's actually available on the VM.

I have not checked to see if we're setting that flag or not, so ignore this if it's already been tested.

> @bskinner Aha! Maybe that explains why the memory usage in App Engine never really goes down that much!?

MocanaAtBroad:
**OK. Thanks for trying to think this through this with me. Some comments below:**
> I have always been suspicious of the `Xms` flag, but your description doesn't quite match my own understanding.
> 
> For me, the `Xmx` flag means the maximum heap memory allocated for JVM, while the `Xms` flag means the _initial_ amount of heap at boot. We have always had set `Xms` to the same value as `Xmx`, which I didn't quite understand at first. But I reasoned that our desire was to have the JVM start with the maximum allowed heap right out of the gate, so JVM has memory to spare and spends less time doing memory allocations and more time doing "real work". If we remove the `Xms` flag, then I suppose the JVM will just start with a small amount and eventually climb up to the max allowed heap size.
> 
**There is default Xmx value. Looks like it is dependent on JVM implementation/environment etc. I still think it is a good idea to set it explicitly.
Regarding our previous use of Xms, your understanding of what it does and the reason to set it match my understanding.**

> Like you have mentioned, I'm also aware that the JVM will allocate other memory besides the heap. Obviously there is the stack and JVM's own internal memory needs. I googled these flags just now to refresh myself and reread this [SO answer](https://stackoverflow.com/a/14763095). It mentions that other off-heap memory include thread stacks and native memory. So if anything, I think threads and native methods are more likely suspects rather than `Xms`.

**To be more explicit about my reasoning for dropping Xms: there is no evidence that we are running out heap space in housekeeping (my experiment seems to show that if you run out of heap, an OutOfMemoryError is reported in GAE).
But GAE is killing process because says we are using too much memory. So if it is not heap-space, then it could be metaspace, stack space, JNI libraries allocations (note: tcell has its own JNI library that it loads), etc.**

**So I am concerned that Xms immediately reserves the specified amount of memory for heap and never relinquishes it to off-heap needs.**

**So my thinking is : We don't know how much heap space is needed, but it is less than Xms. So let's not try to force it. Let VM grab what it needs when it needs it**

> 
> I think the equation you provided should be revised to:
> 
> ```
> off_heap = gae_max - jvm_max
> off_heap = (2048m) - (Xmx)
> ```
**I think your equation might be right for limits, but not appropriate to describe actual operational usage of memory.**
> 
> Now that I think about it: Since our `Xmx` is set to 1640m, I think that means we have < 500m of off-heap memory for the JVM? Maybe the answer here is that we're exceeding that 500m off-heap memory (e.g., too many threads?) and that's where GAE is crashing our services? Perhaps we should try lowering the `Xmx` flag instead?

**OK. I think we agree on what is the possible problem we are dealing with (i.e., not enough off-heap memory available).
This is part of reason I have been concerned with having an excess of threads in housekeeping for connection pools, Redis access, http asyn, etc.**

**But a clarification: Xmx does not allocate any memory. It just sets the limit over how much the VM is allowed to grab for heap. By contrast Xms DOES grab the associated amount immediately.**

**So I am expecting dropping Xms will allow VM to take just the amount it needs for heap and no more, making the rest available for these other suspects that we have less control over.**

MocanaAtBroad:
> @bskinner Aha! Maybe that explains why the memory usage in App Engine never really goes down that much!?

**Right. This is why our processes memory usage immediately shoot up to between 1500 and 1800MB. We are grabbing these amounts explicitly right from the start.**

yufengwng:
@MocanaAtBroad Sweet, this is starting to make so much more sense! This change looks good to me!

yufengwng (approved):

PR #487, opened 2020-11-04
DDP-5248 fix encoding issue in PostPasswordResetRoute

yufengwng:
## Context

Main change here is fixing the query parameter encoding issue for `PostPasswordResetRoute`. Also fixed other tests that was failing or had potential to fail. See comments for more details.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/endpoints/post-password-reset.yml
+++ b/pepper-apis/docs/specification/src/endpoints/post-password-reset.yml
@@ -10,7 +10,7 @@ get:
     configured client application URL using a 302 redirect.
 
     The redirect will provide the following query parameters:
-    - `email` (string): if provided by auth0
+    - `email` (string): url-encoded email, if provided by auth0
     - `errorCode` (string): will be set to `PASSWORD_RESET_LINK_EXPIRED` if
       password reset flow failed
   parameters:
@@ -42,7 +42,7 @@ get:
     302:
       description: password reset is successful
     400:
-      description: client id  is not provided or not unique
+      description: client id is not provided or not unique
       content:
         application/json:
           schema:
--- a/pepper-apis/docs/specification/src/pepper.yml
+++ b/pepper-apis/docs/specification/src/pepper.yml
@@ -2,7 +2,7 @@ openapi: 3.0.0
 servers: []
 info:
   description: 'Pepper API specification'
-  version: 1.9.1
+  version: 1.9.5
   title: Pepper
   contact:
     email: info@datadonationplatform.org
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
@@ -1,11 +1,15 @@
 package org.broadinstitute.ddp.route;
 
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.Optional;
 
-import okhttp3.HttpUrl;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpStatus;
+import org.apache.http.client.utils.URIBuilder;
 import org.broadinstitute.ddp.constants.ErrorCodes;
 import org.broadinstitute.ddp.constants.RouteConstants.QueryParam;
 import org.broadinstitute.ddp.db.TransactionWrapper;
@@ -40,7 +44,7 @@ public class PostPasswordResetRoute implements Route {
             throw ResponseUtil.haltError(response, HttpStatus.SC_BAD_REQUEST, new ApiError(ErrorCodes.REQUIRED_PARAMETER_MISSING, errMsg));
         }
 
-        HttpUrl clientPwdResetUrl = null;
+        URI clientPwdResetUrl = null;
         boolean isDomainSpecified = auth0Domain != null && !auth0Domain.isBlank();
         if (isDomainSpecified) {
             clientPwdResetUrl = getWebClientPasswordResetRedirectUrl(auth0ClientId, auth0Domain, response);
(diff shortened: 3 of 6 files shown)


pepper-apis/src/test/java/org/broadinstitute/ddp/route/PostPasswordResetRouteTest.java
```diff
@@ -56,27 +56,23 @@ public static void setup() {
...
     @Test
     public void test_WhenRouteIsCalledWithValidClientId_ItRespondsWithCorrectHttpRedirect() {
-        Map<String, String> queryParams = new HashMap<>();
-        queryParams.put(QueryParam.AUTH0_CLIENT_ID, auth0ClientId);
-        queryParams.put(QueryParam.AUTH0_DOMAIN, auth0Domain);
-        queryParams.put(QueryParam.EMAIL, testEmail);
-        queryParams.put(QueryParam.SUCCESS, "true");
-        HttpUrl fullUrl = buildEncodedUrl(url, queryParams);
         given(TestUtil.RestAssured.nonFollowingRequestSpec())
-                .when().get(fullUrl.toString()).then().assertThat()
```
yufengwng, line 66: We were using `HttpUrl`, which was from a transitive dependency of `okhttp v3.9.1` that was pulled in by the `jslack` library. This version of `HttpUrl` did not do proper encoding of query parameters. After removing the `jslack` library, the version we're using became `okhttp v3.12.0` which came from the `google-maps-services` library. This new version of `HttpUrl` have updated rules for encoding query params.

What this all means is that:
* Previously, this test was NOT encoding the query params. So we pass in `fullUrl` which is not encoded. The `RestAssured` library will do a round of encoding itself, so nothing was wrong here. The `PostPasswordResetRoute` received properly encoded query params, SparkJava does the right thing and decodes it, and we're able to find the client and such. All good.
* However, now, `fullUrl` is getting encoded, and because `RestAssured` also does encoding, we have issue of **double encoding**. The end result is that in `PostPasswordResetRoute`, after SparkJava decodes, we still have an encoded value, causing client not found and the tests to fail.

My solution here is to not use `HttpUrl`, and use `RestAssured`'s own support for adding query params instead to avoid double-encoding.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PostPasswordResetRouteTest.java
```diff
@@ -37,7 +36,8 @@
     private static String auth0Domain;
     private static final String testEmail = "test_user@datadonationplatform.org";
     private static final String testRedirectUrl = "http://www.datadonationplatform.org/default-password-reset-page/";
-    private static final String testRedirectUrlWithEmail = testRedirectUrl + "?" + QueryParam.EMAIL + "=" + testEmail;
+    private static final String testRedirectUrlWithEmail = testRedirectUrl + "?" + QueryParam.EMAIL + "="
+            + URLEncoder.encode(testEmail, StandardCharsets.UTF_8);
```
yufengwng, line 40: Notice here that we didn't test for proper encoding of the email (e.g., the `@` character should be encoded). Because we used `HttpUrl` in the `PostPasswordResetRoute` implementation, it was not encoded. However, now, it is being encoded. it's a breaking change, but I think that's the way it should be -- the query value should be encoded. So I updated the test here to reflect that.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
```diff
@@ -62,15 +66,15 @@ public Object handle(Request request, Response response) throws Exception {
             }
         }
 
-        HttpUrl.Builder urlBuilder = clientPwdResetUrl.newBuilder();
+        var urlBuilder = new URIBuilder(clientPwdResetUrl);
```
yufengwng, line 69: Since the new `HttpUrl` does proper encoding now, I took the opportunity to replace it with `URIBuilder` which also does proper encoding. The advantage here is that `URIBuilder` comes from the Apache `httpclient` library which we have a direct dependency for versus `HttpUrl` which is a transitive dependency.

pepper-apis/docs/specification/src/endpoints/post-password-reset.yml
```diff
@@ -10,7 +10,7 @@ get:
     configured client application URL using a 302 redirect.
 
     The redirect will provide the following query parameters:
-    - `email` (string): if provided by auth0
+    - `email` (string): url-encoded email, if provided by auth0
```
yufengwng, line 13: Updated documentation to indicate the API change.

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -2,7 +2,7 @@ openapi: 3.0.0
 servers: []
 info:
   description: 'Pepper API specification'
-  version: 1.9.1
+  version: 1.9.5
```
yufengwng, line 5: Took  the opportunity to bump up the version in the documentation.

pepper-apis/src/test/java/org/broadinstitute/ddp/log/SlackAppenderTest.java
```diff
@@ -60,10 +57,4 @@ public void testSuccessfulLoggingToSlack() throws Exception {
             Assert.fail("Mock slack not running");
         }
     }
-
-    @After
-    public void shutDown() {
-        mockServerClient.stop();
```
yufengwng, line 66: Shutting this down seems to cause issues with other tests that uses mockserver. There's no need to do a stop here since we're using `MockServerRule` which integrates with JUnit so JUnit will handle cleaning up those resources.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/ReceiveDsmNotificationRouteTest.java
```diff
@@ -327,8 +330,10 @@ public void testEvents_setInitialKitSentTime() {
                 assertNotNull(record.getInitialKitSentTime());
             });
         } finally {
-            TransactionWrapper.useTxn(handle ->
-                    DBUtils.checkDelete(1, handle.attach(KitScheduleSql.class).deleteRecord(recordId)));
+            TransactionWrapper.useTxn(handle -> {
+                DBUtils.checkDelete(1, handle.attach(KitScheduleSql.class).deleteRecord(recordId));
+                DBUtils.checkDelete(1, handle.attach(KitConfigurationDao.class).deleteConfiguration(kitConfigId.get()));
```
yufengwng, line 335: Some of the kit tests will fail depending on the test order, and the cause seems to be that we're not cleaning up the kit configuration setup here.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PostPasswordResetRoute.java
zyme: Nice!

pepper-apis/src/test/java/org/broadinstitute/ddp/log/SlackAppenderTest.java
zyme: I was wondering about this--thank you for tracking this down.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PostPasswordResetRouteTest.java
zyme: Wow--the double encoding thing is interesting.  Thanks for this thorough summary of the findings!

zyme (approved):
Looks good!  All tests pass locally for me in intellij and maven.

PR #492, opened 2020-11-05
study-builder: support passing arguments to tasks

yufengwng:
Added plumbing to pass command-line arguments to study-builder tasks.

The "double dash" is a command-line convention that is supported by the Apache CLI library. The library will mostly "ignore" the arguments after the double-dash. We take the full list of arguments and trim it down to the stuff after the double-dash, then pass that in to the task. The `CustomTask` interface is updated with a default method so we don't break the set of existing tasks. Tasks may choose to handle the arguments however they want -- they may require positional arguments or optional flags.

I added `UpdateActivityTemplatesInPlace` as an example that requires a positional argument, and as a nice little utility. You can run it like so:

```bash
$ java -Dconfig.file=/path/to/config/file -jar target/StudyBuilder.jar \
  testboston \
  --vars /path/to/vars/file \
  --substitutions studies/testboston/subs.conf \
  --run-task UpdateActivityTemplatesInPlace -- BASELINE_SYMPTOM
```


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateActivityTemplatesInPlace.java
```diff
@@ -0,0 +1,96 @@
...
+import org.broadinstitute.ddp.model.user.User;
+import org.broadinstitute.ddp.studybuilder.ActivityBuilder;
+import org.jdbi.v3.core.Handle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * General task to update templates for an activity in-place. Need to provide the `--activity-code` argument.
+ */
+public class UpdateActivityTemplatesInPlace implements CustomTask {
```
ssettipalli, line 29: Ability to run UpdateTemplates for a chosen activity will be useful

ssettipalli (approved):
looks good to me

PR #495, opened 2020-11-09
Reduce memory footprint for PDF generation

MocanaAtBroad:
## Context
More can be done to reduce footprint (and will if we need to go further) but this ought to help.
Existing code was retrieving the all the individual PDF templates in a "configuration" as well as the generated PDF and keeping in memory the generated PDF documents.

Strategy here is to retrieve the templates one at a time and as needed, generate the corresponding document, use it and go on to next one.

This should reduce memory usage by good chunk for our multi template PDF generation. These are the biggest problems.

More can be done if needed.

Setting this up to run in dev environment. GCP profiler highlighted this functionality as a heap usage hotspot.

## Checklist

- [ X] I have labeled the type of changes involved using the `C-*` labels.
- [ X] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ X] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc
## FUD Score

GCP profiler will show if helps.

## Testing
Existing tests cover functionality. Some refactoring was needed.
- [X ] I have written zero automated tests but have poked around locally to verify proper functionality

## Release


.circleci/config.yml
```diff
@@ -850,7 +850,7 @@ workflows:
             branches:
               only:
                 - develop
-                - DownsizeRedissonInHousekeeping
+                - DDP-5236_ReduceMemoryPdfHandling
           requires:
```
MocanaAtBroad, line 854: Want to try out this on dev before merge.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -143,19 +144,19 @@ private void insertAnswerSubstitution(AnswerSubstitution substitution) {
      * @param config the pdf configuration
      * @return id of newly created pdf configuration
      */
-    default long insertNewConfig(PdfConfiguration config) {
+    default long insertNewConfig(PdfConfiguration config, List<PdfTemplate> templates) {
         PdfSql pdfSql = getPdfSql();
```
MocanaAtBroad, line 150: One of consequences of splitting PdfConfiguration and templates.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -143,19 +144,19 @@ private void insertAnswerSubstitution(AnswerSubstitution substitution) {
...
 
         config.getVersion().setConfigId(configId);
         long versionId = insertConfigVersionWithDataSources(config.getVersion());
 
-        List<PdfTemplate> templates = config.getTemplates();
         for (int i = 0, size = templates.size(); i < size; i++) {
             int order = i + 1;
             long templateId = insertTemplate(templates.get(i));
             pdfSql.assignTemplateToVersion(versionId, templateId, order);
+            config.addTemplateId(templateId);
```
MocanaAtBroad, line 161: The configuration only refers to the id. The real template object will have to be queried separately. More about this...

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -268,13 +269,13 @@ private void deleteAnswerSubstitution(AnswerSubstitution substitution) {
...
 
         PdfSql pdfSql = getPdfSql();
-        DBUtils.checkDelete(templates.size(), pdfSql.unassignTemplatesFromVersion(version.getId()));
+        DBUtils.checkDelete(templateIds.size(), pdfSql.unassignTemplatesFromVersion(version.getId()));
 
-        for (PdfTemplate template : templates) {
-            deleteTemplate(template);
+        for (Long templateId : templateIds) {
+            findFullTemplateByTemplateId(templateId).ifPresent(template -> deleteTemplate(template));
         }
```
MocanaAtBroad, line 282: This is a rarely invoked method, so was satisfied with not writing code that would delete templates with just one operation.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -317,6 +322,14 @@ default int deleteAllConfigVersions(long configId) {
...
     List<PdfTemplate> findBaseTemplatesByVersionId(@Bind("versionId") long versionId);
 
+    @UseStringTemplateSqlLocator
+    @SqlQuery("findBaseTemplatesByTemplateId")
+    @RegisterConstructorMapper(PdfTemplateDto.class)
+    @RegisterConstructorMapper(value = MailingAddressTemplateDto.class, prefix = "m")
+    @RegisterConstructorMapper(value = PhysicianInstitutionTemplateDto.class, prefix = "p")
+    @UseRowReducer(BaseTemplatesReducer.class)
+    Optional<PdfTemplate> findBaseTemplateByTemplateId(@Bind("templateId") long templateId);
+
```
MocanaAtBroad, line 330: New way to retrieve templates.

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -273,20 +272,22 @@ boolean messageShouldBeIgnored(Handle apisHandle, String studyGuid, String parti
...
-                InputStream pdfStream = pdfBucketService.getPdfFromBucket(blobName).orElse(null);
+            try (
+                    InputStream pdfStreamFromBucket = pdfBucketService.getPdfFromBucket(blobName).orElse(null)) {
+                InputStream pdfStream = pdfStreamFromBucket;
                 if (pdfStream == null) {
                     // todo: remove the generateIfMissing feature
                     LOG.info("Could not find {} in bucket {}, generating", blobName, pdfBucketService.getBucketName());
-                    byte[] pdfBytes = pdfGenerationService.generateFlattenedPdfForConfiguration(
+                    pdfStream = pdfGenerationService.generateFlattenedPdfForConfiguration(
                             pdfConfig,
```
MocanaAtBroad, line 281: Returning stream instead of bytes. Does not make a difference in terms of memory right now, but could if we do further refactoring.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfConfiguration.java
```diff
@@ -12,15 +13,16 @@
...
     private List<PdfTemplate> templates;
+    private List<Long> templateIds;
 
-    public PdfConfiguration(PdfConfigInfo info, PdfVersion version, List<PdfTemplate> templates) {
+    public PdfConfiguration(PdfConfigInfo info, PdfVersion version, Long[] templateIds) {
         this.info = info;
         this.version = version;
-        this.templates = templates;
+        this.templateIds = new ArrayList<>(Arrays.asList(templateIds));
     }
```
MocanaAtBroad, line 21: One of the runtime memory problems trying to fix here: a configuration could have a list of templates associated with it. And each template has a reference to the the PDF blob. Objective here is to handle only one PDF blob at a time. So switching from list of templates to list of template ids.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,49 +139,42 @@ private static void setIsChecked(PdfFormField field, boolean isChecked) {
...
                 PdfWriter pdfWriter = new PdfWriter(renderedStream);
                 PdfDocument mergedDoc = new PdfDocument(pdfWriter)) {
-            // concatenate each rendered doc to a master doc
-
             int counter = 0;
-            for (byte[] pdfStream : individualPdfs) {
-                copyPdfToMasterDoc(pdfStream, counter, mergedDoc);
+            for (int pdfOrderIndex = 0; pdfOrderIndex < configuration.getTemplateIds().size(); pdfOrderIndex++) {
+                Long templateId = configuration.getTemplateIds().get(pdfOrderIndex);
+                PdfTemplate template = handle.attach(PdfDao.class)
```
MocanaAtBroad, line 151: Retrieve on pdf template at a time for processing.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,49 +139,42 @@ private static void setIsChecked(PdfFormField field, boolean isChecked) {
...
-
+        byte[] pdf;
         try (
                 ByteArrayOutputStream renderedStream = new ByteArrayOutputStream();
                 PdfWriter pdfWriter = new PdfWriter(renderedStream);
                 PdfDocument mergedDoc = new PdfDocument(pdfWriter)) {
-            // concatenate each rendered doc to a master doc
-
             int counter = 0;
-            for (byte[] pdfStream : individualPdfs) {
```
MocanaAtBroad, line 180: Avoid holding the individual generated pdfs in templates in memory. Add it to working document and let it go.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/PdfGenerationServiceTest.java
```diff
@@ -69,12 +70,15 @@
 
     @BeforeClass
     public static void setup() throws Exception {
+        System.out.println("About to sleep during setup");
+        Thread.sleep(30000);
```
yufengwng: Are we sure we want to sleep during the test suite? And 30 seconds at that...

pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfConfiguration.java
```diff
@@ -55,11 +55,12 @@ public PdfVersion getVersion() {
...
 
-    public List<PdfTemplate> getTemplates() {
-        return templates;
+    public List<Long> getTemplateIds() {
+        return templateIds;
     }
 
-    public void addTemplate(PdfTemplate template) {
-        templates.add(template);
+    public void addTemplateId(Long templateId) {
```
yufengwng: Let's use primitive `long` to avoid possibility of `null`s.
```suggestion
    public void addTemplateId(long templateId) {
```

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -268,13 +269,13 @@ private void deleteAnswerSubstitution(AnswerSubstitution substitution) {
...
+        List<Long> templateIds = config.getTemplateIds();
 
         PdfSql pdfSql = getPdfSql();
-        DBUtils.checkDelete(templates.size(), pdfSql.unassignTemplatesFromVersion(version.getId()));
+        DBUtils.checkDelete(templateIds.size(), pdfSql.unassignTemplatesFromVersion(version.getId()));
 
-        for (PdfTemplate template : templates) {
-            deleteTemplate(template);
+        for (Long templateId : templateIds) {
+            findFullTemplateByTemplateId(templateId).ifPresent(template -> deleteTemplate(template));
```
yufengwng: Might be better to do `orElseThrow()` and calling `deleteTemplate()` separately.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/PdfDao.sql.stg
```diff
@@ -103,10 +103,26 @@ select base.pdf_base_template_id    as template_id,
...
 
+findBaseTemplatesByTemplateId() ::= <<
+<selectBaseTemplates()>
+where pvt.pdf_base_template_id = :templateId
+>>
+
+findTemplateIdsByVersionId() ::= <<
+    select pvt.pdf_base_template_id
+    from pdf_version_template as pvt
+    join pdf_base_template as base on base.pdf_base_template_id = pvt.pdf_base_template_id
```
yufengwng: Doesn't look like we need the join to the `pdf_base_template` table.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,49 +139,42 @@ private static void setIsChecked(PdfFormField field, boolean isChecked) {
...
-                copyPdfToMasterDoc(pdfStream, counter, mergedDoc);
+            for (int pdfOrderIndex = 0; pdfOrderIndex < configuration.getTemplateIds().size(); pdfOrderIndex++) {
+                Long templateId = configuration.getTemplateIds().get(pdfOrderIndex);
```
yufengwng:
```suggestion
                long templateId = configuration.getTemplateIds().get(pdfOrderIndex);
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,49 +139,42 @@ private static void setIsChecked(PdfFormField field, boolean isChecked) {
...
+                                configuration.getStudyGuid(), handle);
+                        break;
+                    default:
+                        errors.add("Tried to use a template type (" + template.getType() + ") that is unsupported.");
+                        pdf = null;
+                        break;
+                }
+                if (pdf != null && pdf.length != 0) {
+                    copyPdfToMasterDoc(pdf, counter, mergedDoc);
+                }
```
yufengwng, line 174: Do you think we should add a `pdf = null` statement here? So that we "drop" that reference at the end of this loop iteration, so that in the next loop iteration, if the JVM needs to run GC before loading in the result of `findFullTemplateByTemplateId()` then it can free up the `byte[]`? We could also simply just move the `byte[] pdf` variable declaration inside the `for` loop. That should drop that byte array reference after each iteration. Or am I totally misunderstanding things here?

yufengwng (commented):
Very nice!

pepper-apis/src/test/java/org/broadinstitute/ddp/service/PdfGenerationServiceTest.java
MocanaAtBroad: Those should not be there. Removed them (was holding up test so I could attach Profiler my test code executed).
b24e1e52

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
MocanaAtBroad: Sure 0892b1c0

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/PdfDao.sql.stg
MocanaAtBroad: Right! Dropping join.
7bb503c8

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
MocanaAtBroad: Nulling it variable explicitly ought to be unnecessary. Assigning a new object to that variable and stack frame for this method getting unloaded both ought to de-reference the PDF object and make it candidate for GC reaping.
But moving the variable into the loop is a good idea. Probably marginal benefit, but definitely in the spirit of what I am trying to do in PR.
7137e2eb

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetPdfRoute.java
```diff
@@ -99,29 +99,30 @@ public Object handle(Request request, Response response) {
...
                         configInfo.getFilename()));
-
                 HttpServletResponse raw = response.raw();
-                raw.getOutputStream().write(pdfBytes);
-                raw.getOutputStream().flush();
-                raw.getOutputStream().close();
+
+
+                try (OutputStream outputStream = raw.getOutputStream()) {
+                    IOUtils.copy(pdfInputStream, outputStream);
```
zyme, line 123: `IOUtils` is great!

pepper-apis/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -273,20 +272,22 @@ boolean messageShouldBeIgnored(Handle apisHandle, String studyGuid, String parti
                     pdfConfig.getConfigName(),
                     pdfConfig.getVersion().getVersionTag());
 
-            try {
-                InputStream pdfStream = pdfBucketService.getPdfFromBucket(blobName).orElse(null);
+            try (
+                    InputStream pdfStreamFromBucket = pdfBucketService.getPdfFromBucket(blobName).orElse(null)) {
+                InputStream pdfStream = pdfStreamFromBucket;
```
zyme: Should we `close()` this stream?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfService.java
```diff
@@ -126,6 +126,6 @@ public String generateAndUpload(Handle handle,
                 participantGuid,
                 pdfConfiguration.getConfigName(),
                 pdfConfiguration.getVersion().getVersionTag(),
-                new ByteArrayInputStream(pdf));
+                pdfStream);
```
zyme: Should we `close()` `pdfStream` prior to return?

zyme (dismissed):
Overall looks good to me.  CC @Rinchan20 for some testing here both with DSM and with emailed consents.  @MocanaAtBroad I added a few notes about maybe closing some streams.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfService.java
MocanaAtBroad: Yes we should. Thanks! d0afd6bd

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -504,13 +498,13 @@ private int mergeAllProviderPdfsIntoOutput(ByteArrayOutputStream output, List<Me
...
-        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
-        byte[] freeSans;
         try {
-            freeSans = IOUtils.toByteArray(classLoader.getResourceAsStream(FONT_PATH));
-            return PdfFontFactory.createFont(freeSans, PdfEncodings.IDENTITY_H, true);
+            if (!PdfFontFactory.getRegisteredFonts().contains(INTERNATIONAL_FONT_NAME)) {
+                PdfFontFactory.register(INTERNATIONAL_FONT_PATH);
+            }
+            return PdfFontFactory.createRegisteredFont(INTERNATIONAL_FONT_NAME, PdfEncodings.IDENTITY_H, true);
         } catch (IOException e) {
```
MocanaAtBroad, line 506: Changed this based on observation running unit tests locally: the font loading/creation is the most memory intensive part of running the pdf generation unit tests.
Don't think that overall this change improves memory situation: the change uses the itext font caching so it is not reloaded each time is needed. It MIGHT reduce the size of documents.

yufengwng (approved):
LGTM

PR #501, opened 2020-11-10
DDP-5267 MGU: "The requested user GUID is not managed" error for newl…

YuriyPavlovBI:
Fix for JWT to DDPAuth cache invalidation during governed user creation


yufengwng (approved):
Looks good.

PR #510, opened 2020-11-17
Content model 2.0 sketch with velocity

zyme:


pepper-apis/src/test/java/org/broadinstitute/ddp/VelocitySketchTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+    static VelocityEngine velocityEngine = null;
+
+    @BeforeClass
+    public static void setUp() {
+        velocityEngine = new VelocityEngine();
+        // in production, we will probably use more flexible
+        // ways to load macros, such as VelocityEngine.evaluate()
+        Properties velocityProps = new Properties();
+        velocityProps.put("file.resource.loader.path", "/sandbox/ddp-study-server/pepper-apis/src/test/resources");
```
MocanaAtBroad, line 33: test had problems finding the templates until I replaced the path here with my local absolute path for the test/resources directory.

MocanaAtBroad:
Sure! Something like this would give us more control over emails. Would also facilitate building a WYSIWYG UI.

Trimax:
Discussed with @zyme. Not going to be merged.

PR #511, opened 2020-11-17
circleci: skip job if no meaningful changes

yufengwng:
Updated our CircleCI pipeline configuration to allow skipping jobs. Skipping is based on matching file patterns: if there's a match we'll continue, otherwise the job exits early with a success status. Skipping is only allowed on the "on-demand" workflow, so that we can skip for PRs but continue to run everything for the `develop` / RC / hotfix branches.

Even though we're skipping things, we're still incurring a wait time of about 2 minutes. This is because CircleCI has to setup the docker environment for the job (which takes forever, especially the parallel job) before checking the files. I haven't found a way to conditionally spin up jobs yet (I don't think CircleCI supports it..).

Despite that, PRs that only changed study configuration files should be less painful now since wait time is a bit lower and PR won't be affected by flaky tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,7 @@ commands:
 
             # untarring will create a "tcell" dir
             tar --no-same-owner -xvf $TAR_FILE_PATH
+
   build-jar:
     description: Build jar
     parameters:
@@ -104,6 +105,7 @@ commands:
           name: Run maven package
           command: |
             mvn package -DskipTests -file << parameters.pom_file >> -Dcheckstyle.skip -Dmaven.repo.local=<< parameters.m2_repo >>
+
   setup-shared-env:
     description: "Setup shared ENV vars used by multiple scripts"
     steps:
@@ -261,6 +263,7 @@ commands:
             mkdir -p $(dirname $JAR_FILE_PATH)
             gsutil cp  $JAR_FILE_URL $JAR_FILE_PATH
             echo ""
+
   deploy-jar:
     parameters:
       jar_base_name:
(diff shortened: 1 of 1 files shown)


MocanaAtBroad (dismissed):
Looks good! Maybe we should include the circleci config file as one of the files that each build depends on?

MocanaAtBroad:
Regarding the spin-up time of images in CircleCi: this was a source of frustration to me when I was working on making builds as fast as possible. Ultimately all I could do was parallelize everything that could be parallelized and try to create Docker images that are small as they could be or use Docker images that belong to circleci to maximize chance they could be found in their local cache.
Would be slower to do job that decides whether to do all the parallel stuff or not. Don't think I didn't consider that!

yufengwng:
@MocanaAtBroad Good point about the CircleCI config file. I've added it so that we run everything if that file changed to ensure the pipeline works.

MocanaAtBroad (approved):

.circleci/config.yml
```diff
@@ -681,12 +718,28 @@ jobs:
...
+        default: false
     steps:
       - checkout:
           path: *repo_path
+      # When any code or configuration changes, run compile step below.
+      - when:
+          condition: << parameters.allow_skip_job >>
+          steps:
+            - match-file-patterns-or-halt:
+                file_patterns: "pepper-apis/config pepper-apis/src study-builder/src pom.xml checkstyle.xml"
```
zyme, line 733: Nice, so code changes pass, but config file changes will be skipped!

zyme (approved):

PR #515, opened 2020-11-19
DDP 5322 elastic changes

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-5322
Add nested picklist options to participant_structured elastic index .
Changes is needed for DSM to show nested options in POSTCONSENT: RACE question of Brain Pediatrics
nested options already exists in activity_definition but DSM prefers it in  participant_structured index along with answers for ease of displaying them.

Added nestedOptions API doc changes

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/export/json/structured/PicklistQuestionRecord.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/export/json/structured/PicklistQuestionRecord.java
@@ -16,7 +16,7 @@ import org.broadinstitute.ddp.model.activity.types.QuestionType;
 public final class PicklistQuestionRecord extends QuestionRecord {
 
     @SerializedName("answer")
-    private List<Object> selected = new ArrayList<>();
+    private List<String> selected = new ArrayList<>();
     @SerializedName("optionDetails")
     private List<Map<String, String>> optionDetails = new ArrayList<>();
     @SerializedName("nestedOptions")
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/export/json/structured/PicklistQuestionRecord.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/export/json/structured/PicklistQuestionRecord.java
@@ -2,21 +2,25 @@ package org.broadinstitute.ddp.export.json.structured;
 
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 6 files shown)


pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.sql.stg
```diff
@@ -66,6 +67,8 @@ select ai.activity_instance_id   as a_instance_id,
   left join numeric_answer as na on na.answer_id = ans.answer_id
   left join picklist_option__answer as pa on pa.answer_id = ans.answer_id
   left join picklist_option as po on po.picklist_option_id = pa.picklist_option_id
+  left join picklist_nested_option as npo on npo.nested_option_id = pa.picklist_option_id
```
yufengwng, line 70: Since we're only pulling one column from the parent picklist option, have you considered doing a subquery instead of left joins? Since some answer rows are picklists while other answer rows are not picklist, I'm wondering what the performance is for subquery versus left-join. Same question for `AnswerDao.sql.stg`.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/json/structured/PicklistQuestionRecord.java
```diff
@@ -17,14 +19,21 @@
...
+    private Map<String, List<String>> nestedOptions = new HashMap<>();
 
     public PicklistQuestionRecord(String stableId, List<SelectedPicklistOption> answer) {
         super(QuestionType.PICKLIST, stableId);
         if (CollectionUtils.isNotEmpty(answer)) {
             List<SelectedPicklistOption> selected = new ArrayList<>(answer);
             selected.sort(Comparator.comparing(SelectedPicklistOption::getStableId));
             for (SelectedPicklistOption option : selected) {
-                this.selected.add(option.getStableId());
+                if (StringUtils.isNotBlank(option.getParentStableId())) {
```
yufengwng, line 31: Nice! I like approach 3 of keeping the answer array the top-level selected options, while the nested options are separated out.

yufengwng (approved):
LGTM

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceSql.sql.stg
ssettipalli: I believe optimizer handles joins efficiently. Will check it out

PR #523, opened 2020-11-24
DDP-5349 Allow usage of variables in i18n substitutions

YuriyPavlovBI:
Currently it is not supported to have tags like $ddp.participantName() in the substitutions, they won't be processed during rendering.
However, for different languages, name etc. could be at different positions in the sentence.
Therefore, it is necessary to have second level of template processing in order to support i18n.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
@@ -289,7 +289,15 @@ public class I18nContentRenderer {
         VelocityContext ctx = new VelocityContext(context);
         StringWriter writer = new StringWriter();
         engine.evaluate(ctx, writer, TEMPLATE_NAME, template);
-        return writer.toString();
+        String result = writer.toString();
+        if (result.contains("$")) {
+            // Here we have a second pass in case of variables in the substitution values,
+            // e.g. participantName() with locale-dependent position in the sentence.
+            writer = new StringWriter();
+            engine.evaluate(ctx, writer, TEMPLATE_NAME, result);
+            result = writer.toString();
+        }
+        return result;
     }
 
     private String convertToString(Object obj) {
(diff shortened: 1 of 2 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -289,7 +289,15 @@ public String renderToString(String template, Map<String, Object> context) {
         VelocityContext ctx = new VelocityContext(context);
         StringWriter writer = new StringWriter();
         engine.evaluate(ctx, writer, TEMPLATE_NAME, template);
-        return writer.toString();
+        String result = writer.toString();
+        if (result.contains("$")) {
```
yufengwng, line 293: Can you give an example, or even better, write a test case?

YuriyPavlovBI: Sure. Test provided.

pepper-apis/src/test/java/org/broadinstitute/ddp/content/I18nContentRendererTest.java
```diff
@@ -64,6 +67,33 @@ public void testRenderContentSuccess() {
...
+            JdbiRevision jdbiRev = handle.attach(JdbiRevision.class);
+
+            RenderValueProvider valueProvider = new RenderValueProvider.Builder().setParticipantFirstName("John").build();
+
+            Map<String, Object> context = new HashMap<>();
+            context.put(I18nTemplateConstants.DDP, valueProvider);
+
+            Template tmpl = new Template(TemplateType.HTML, null, "<em>$question_name</em>");
+            tmpl.addVariable(new TemplateVariable("question_name", Collections.singletonList(
+                    new Translation("en", "Your name is $ddp.participantFirstName()?"))));
```
yufengwng, line 83: Ah ha! This makes sense!

yufengwng (approved):

PR #536, opened 2020-12-01
DDP-5360 don't use keep-alive connections to proxy

yufengwng:
## Context

My attempt at fixing DDP-5360.

To deploy this, there are two steps:
* Deploy the new code like normal.
* Copy new Squid config file to proxy VM and reboot the proxy docker container. These can be done using the `pepper-apis/gae-egress-proxy/pepper-squid.sh` script.

## Checklist

## FUD Score

    - Requires updating the squid config file and rebooting proxy. 

## Testing

- [x] I am unable to reproduce or trigger the issue to verify this change.

## Release

    - Requires updating the squid config file and rebooting proxy.

### DIFF
--- a/pepper-apis/gae-egress-proxy/squid.conf
+++ b/pepper-apis/gae-egress-proxy/squid.conf
@@ -25,6 +25,9 @@ http_access deny all
 # The port the proxy listens on.
 http_port 3128
 
+# Prefer IPv4 for DNS.
+dns_v4_first on
+
 # Turn off caching.
 cache deny all
 cache_dir null /tmp
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/client/SendGridClient.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/client/SendGridClient.java
@@ -22,6 +22,7 @@ import com.sendgrid.Response;
 import com.sendgrid.SendGrid;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.http.HttpHost;
+import org.apache.http.impl.NoConnectionReuseStrategy;
 import org.apache.http.impl.client.HttpClients;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -76,6 +77,7 @@ public class SendGridClient {
                 throw new IllegalArgumentException("proxy needs to be a valid url");
             }
             httpClientBuilder.setProxy(new HttpHost(proxyUrl.getHost(), proxyUrl.getPort(), proxyUrl.getProtocol()));
+            httpClientBuilder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
             LOG.info("Using SendGrid proxy: {}", proxy);
         }
         var client = new Client(httpClientBuilder.build());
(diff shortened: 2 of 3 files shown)


pepper-apis/gae-egress-proxy/squid.conf
```diff
@@ -25,6 +25,9 @@ http_access deny all
 # The port the proxy listens on.
 http_port 3128
 
+# Prefer IPv4 for DNS.
+dns_v4_first on
```
yufengwng, line 29: This change is based on [this SO question](https://serverfault.com/questions/644604/squid-https-tunnelling-using-connect-very-slow).

I have found that adding this makes the `connection reset` issue show up less often, but it doesn't fully fix the issue. I don't have concrete evidence to back this up.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ElasticsearchServiceUtil.java
```diff
@@ -71,6 +72,7 @@ public static synchronized RestHighLevelClient getElasticsearchClient(Config cfg
                     httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider);
                     if (proxyUrl != null) {
                         httpClientBuilder.setProxy(new HttpHost(proxyUrl.getHost(), proxyUrl.getPort(), proxyUrl.getProtocol()));
+                        httpClientBuilder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
```
yufengwng, line 75: This change is based on [this](https://stackoverflow.com/questions/52997697/how-to-get-around-connection-reset-by-peer-when-using-elasticsearchs-restclie) and [this](https://stackoverflow.com/questions/39755456/how-to-configure-squid-connectionreadwrite-timeout-per-request-instead-of-glob).

My theory is that the HTTP client here is keeping connections around when on the other side the Squid proxy had already closed the connection. I suspect we're not seeing the `connection reset` issue on every single request is because the HTTP client cycles through a pool of connections, and we just so happens to miss out on the closed connection sometimes. From those 2 SO questions, it sounds like for HTTPS traffic, Squid is not able to inspect the keep-alive headers and so will automatically close connections. Meanwhile, the Apache HTTP client has some fancy built-in logic for doing keep-alive, and when there's no keep-alive header (as is the case for Elasticsearch) it will keep connections around indefinitely. I don't know what the keep-alive time is for these various systems, so I opted to not reuse connections.

Previously, we're creating and closing the ES client on every single request. So moving to "no connection reuse" will effectively give us the same behavior as before. However, we still get some benefits since there's underlying resources like thread pools so we don't create those every single time.

pepper-apis/src/main/java/org/broadinstitute/ddp/client/SendGridClient.java
```diff
@@ -76,6 +77,7 @@ public SendGridClient(String apiKey, @Nullable String proxy) {
                 throw new IllegalArgumentException("proxy needs to be a valid url");
             }
             httpClientBuilder.setProxy(new HttpHost(proxyUrl.getHost(), proxyUrl.getPort(), proxyUrl.getProtocol()));
+            httpClientBuilder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
```
yufengwng, line 80: I have seen some of the SendGrid requests fail with connection issue, especially the request to get template version. I am less certain about this change here, even less than the change to the Elasticsearch HTTP client. However, I think the SendGrid client is in the same situation, so not reusing connections to proxy should hopefully help.

MocanaAtBroad (approved):
Worth trying! Can't hurt.

PR #539, opened 2020-12-02
MPC fixes

ssettipalli:
Add country validation to MPC about-you 
Text fix
https://broadinstitute.atlassian.net/browse/DDP-4776

followup expression fix


study-builder/studies/mpc/study-workflows.conf
```diff
@@ -56,11 +56,6 @@
             && user.studies["cmi-mpc"].forms["ABOUTYOU"].questions["COUNTRY"].answers.hasAnyOption("US", "CA", "PR", "GU", "VI", "MP", "AS")
             """
```
yufengwng, line 57: The ticket mentions:
> allow those who select another country → submit → and then go back and select a valid location to be able to have a consent generated for them

I don't see how this PR allows that. After they submit AboutYou, status is `complete`. When they go back to AboutYou and change country to U.S. and submit again, events will not run again so if consent is created by event it will not be created. And the workflow configuration here will only go to consent if it is created. So I don't see where consent is created for users who changes their country from "not u.s." to "u.s.".

ssettipalli: PR addresses the issue with slightly different approach.
Currently users can submit the form even with non US-Canada country and we are showing International Users page/ message and NOT creating consent.
Drawback is once user updates country since form is already submitted (COMPLETE), it doesn't trigger consent creation event since it's already in COMPLETE status.
To fix this issue and make it more clean I changed it as validation with relevant message and blocks submission of about-you, similar to what we do in some other cmi studies . 
Mary is fine with this validation and message.
Basically avoiding submission with invalid country.

yufengwng: Okay, got it.

yufengwng (approved):

PR #541, opened 2020-12-02
DDP-5368 Post-enrollment submission functionality per activity. - Some fixes & improvements

YuriyPavlovBI:
## Context

DDP-5368 Post-enrollment submission functionality per activity.
Fixed study configuration and readOnly computation.
Added previous activity instance GUID into activity summary.
PEX updated to support previous instance predicate

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/json/activity/ActivityInstanceSummary.java
```diff
@@ -55,6 +55,9 @@
     @SerializedName("isFollowup")
     private boolean isFollowup;
 
+    @SerializedName("prevInstanceGuid")
```
yufengwng: Since this will show up in JSON response, how about we rename it to `previousInstanceGuid`? Also, can you update the docs in `pepper-apis/docs/specification`? Thanks!

pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserActivityInstanceListRoute.java
```diff
@@ -83,8 +83,13 @@ private void performActivityInstanceNumbering(
             summariesWithTheSameCode.sort(Comparator.comparing(ActivityInstanceSummary::getCreatedAt));
             // Number items within each group.
             int counter = 1;
+            String prevGuid = null;
             for (var summary : summariesWithTheSameCode) {
+                if (prevGuid != null) {
+                    summary.setPrevInstanceGuid(prevGuid);
+                }
```
yufengwng: Very nice! I like that there's no additional db lookup.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
```diff
@@ -23,13 +24,20 @@ public static boolean isReadonly(Handle handle, QuestionDto dto, String activity
...
                         "Could not find activity instance with guid=%s while getting question with id=%d and stableId=%s",
                         activityInstanceGuid, dto.getId(), dto.getStableId())));
 
-        return InstanceStatusType.COMPLETE.equals(statusType);
+        InstanceStatusType statusType = activityInstanceDto.getStatusType();
+        if (InstanceStatusType.COMPLETE.equals(statusType)) {
+            return true;
+        }
+
+        // If it is not the first instance - block editing
```
yufengwng, line 37: I see. Typically, if question is write-once, then we will have already marked it read-only when status is completed. However, since there can be multiple instance of activity, the second/third/etc instances becomes a "loophole" where user can edit the question answer. So this will prevent that.

I think there is an interesting case where question is configured write-once BUT answers are not copied from previous instance. So in that case, user somehow somewhere gets a new blank instance, and then for that particular question they won't be able to answer it since it's now read-only. We can probably improve on this and fix that edge-case down the line.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
```diff
@@ -23,13 +24,20 @@ public static boolean isReadonly(Handle handle, QuestionDto dto, String activity
...
+        var instanceDao = handle.attach(ActivityInstanceDao.class);
+        Long previousInstanceId = instanceDao.findMostRecentInstanceBeforeCurrent(activityInstanceDto.getId())
```
yufengwng, line 39: If we have multiple questions in the same activity defined as write-once, then we're potentially doing this lookup multiple times even though it's the same instance id. I think it will be ideal if we do this lookup once and pass it in. However, looking through the code, I don't really see a good place for that. The refactor needed will be significant. So unless there's an easy way to minimize this lookup, I think the way you have it here is fine.

yufengwng (commented):
Overall looks reasonable. Just missing API documentation update.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/activity/ActivityInstanceSummary.java
YuriyPavlovBI: Property renamed, doc schema updated

yufengwng (approved):

PR #557, opened 2020-12-08
Ddp 5262 family history spike - do not merge

MocanaAtBroad:
**Not meant to be merged as is!**
Just a couple of files that demonstrate how we would map activity data to serialized model of family history.
Added comments to files.

The API JSON model with example is here:
https://github.com/MocanaAtBroad/FamilyHistory

Note that the code and examples are aiming to enable generating a family history model that corresponds to these mockups (you might or might not have access to mockups. I believe Erin controls access):
https://projects.invisionapp.com/share/MSZCDB0Z45E#/screens/436961372?browse

---
This change is [Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-study-server/557)

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/FamilyHistoryRelation.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/FamilyHistoryRelation.java
@@ -1,11 +1,14 @@
 package org.broadinstitute.ddp.model.familyhistory;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
 import org.broadinstitute.ddp.model.user.UserProfile.SexType;
 
+/**
+ * Try to add some additional semantics to the labels used to refer to family relations
+ * We define sex and parents when the label suggests who they might be relative to another relation
+ */
 public enum FamilyHistoryRelation {
     PATERNAL_GRANDFATHER(SexType.MALE, List.of()),
     PATERNAL_GRANDMOTHER(SexType.FEMALE, List.of()),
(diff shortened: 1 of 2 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/familyHistoryMapping.conf
```diff
@@ -0,0 +1,96 @@
...
+              "activityCode": "CHILDMEDHISTORY"
+            }
+          }
+        },
+        {
+          "relationDiscriminator": {
+            "stableId": "AUNTUNCLETYPE",
+            "medicalHistory": {
+              "type": "ACTIVITY",
+              "activityCode": "AUNTUNCLEMEDHISTORY"
```
zyme, line 60: It could be possible for all relationships to share `activityCode`, right?  Or if we wanted to ask different questions of different relationships, we could do that too.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/FamilyHistoryRelation.java
```diff
@@ -0,0 +1,100 @@
...
+ * We define sex and parents when the label suggests who they might be relative to another relation
+ */
+public enum FamilyHistoryRelation {
+    PATERNAL_GRANDFATHER(SexType.MALE, List.of()),
+    PATERNAL_GRANDMOTHER(SexType.FEMALE, List.of()),
+    MATERNAL_GRANDFATHER(SexType.MALE, List.of()),
+    MATERNAL_GRANDMOTHER(SexType.FEMALE, List.of()),
+    MOTHER(SexType.FEMALE,List.of(MATERNAL_GRANDFATHER,MATERNAL_GRANDMOTHER)),
+    FATHER(SexType.MALE,List.of(PATERNAL_GRANDFATHER, PATERNAL_GRANDMOTHER)),
+    PATERNAL_AUNT(SexType.FEMALE, List.of(), List.of(FATHER)),
```
zyme, line 19: I'm guessing the `enum` here is for readability at a POC stage--would this sort of concept be something we manage relationally?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/familyHistoryMapping.conf
```diff
@@ -0,0 +1,96 @@
...
+          "relation": "MOTHER",
+          "medicalHistory": {
+            "type": "ACTIVITY",
+            "activityCode": "MOTHERMEDHISTORY"
+          }
+        },
+        // The family relation now comes from an answer to a question.
+        // In mockups, the relations for siblings come from pull-downs.
+        // Assuming here that the pull-down values match the enum names
+        // Note that each instance of the activity will result in a new member added to family
```
zyme, line 34: Should be standardize on some kind of `name` for each relationship so that a user can keep track of their relationships informally?  Can that be something baked into the model or would this be expressed as a common field to various `activityCode`s?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/FamilyHistoryRelation.java
MocanaAtBroad: For sure, this same relationships could be defined in the database. However, my sense overall, is that they are easier to work with as enums. Either way, if these definitions become unwieldy in code or would be desirable to allow end users to define their own, they can be easily migrated. Note that I tried to cover here all the relationships in Erin's mockups.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/familyHistoryMapping.conf
MocanaAtBroad: Yes! We could potentially have just one activity that could be used to enter the medical history for every relation. The every activity **instance** would provide the medical history for the relation that was entered.
The mappings I used here are specifically trying to show how we would map Erin's family history mockups. There is a different survey for mother,  father, siblings, etc. Siblings of all types (full sister,paternal half-brother,etc) share the same survey.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/familyhistory/familyHistoryMapping.conf
MocanaAtBroad: The enums above are my attempt to standardize the relation codes. We can (and probably should) have internationalized names for them (if that is what you were asking?)

Trimax:
Discussed with @zyme. Not going to be merged.

PR #563, opened 2020-12-11
DDP-5396 testboston: allow admin login from google-oauth2

yufengwng:
## Context

This PR adds support for google-oauth2 authentication for TestBoston study admins. This is achieved purely using Auth0 configuration without any changes to backend codebase. Basic idea is to use an Auth0 database as the source-of-truth for admins logging in from google, and link the google identity to this database identity. If google email is in Auth0 database we proceed with login, otherwise we error. We can then follow our regular SOP to add an `auth0_user_id` to our database manually to enable a study admin.

To deploy this, we can use the Auth0 Deploy CLI to push these changes up to Auth0 tenant. We then need to configure several things in Auth0 as well as in GCP. I'll document these separately.

## Checklist

## FUD Score

    - Need to setup up Auth0 configs.

## Testing

## Release

    - Need to setup Auth0 configs.

### DIFF
--- a/study-builder/tenants/testboston/rules/link-admin-accounts.js
+++ b/study-builder/tenants/testboston/rules/link-admin-accounts.js
@@ -104,6 +104,8 @@ function(user, context, callback) {
             return callback(new Error(body));
           }
 
+          // Note that once identity linking is done, Auth0 will actually delete the google profile
+          // and merge it into the database profile.
           console.log(`Successfully linked identity ${user.user_id} to admin user ${adminUser.user_id}`);
 
           // Setting the primary user seems to be required so that we designate the original admin
--- a/study-builder/tenants/testboston/rules/link-admin-accounts.js
+++ b/study-builder/tenants/testboston/rules/link-admin-accounts.js
@@ -28,13 +28,12 @@ function(user, context, callback) {
   }
 
   request({
-    url: auth0.baseUrl + '/users',
+    url: auth0.baseUrl + '/users-by-email',
     headers: {
       Authorization: 'Bearer ' + auth0.accessToken
     },
     qs: {
-      search_engine: 'v3',
-      q: `identities.connection:"${adminConnection}" AND email:"${user.email}"`
+      email: user.email
     }
   }, function(err, response, body) {
     if (err) {
@@ -45,13 +44,19 @@ function(user, context, callback) {
 
     if (response.statusCode !== 200) {
       console.log('Failed to search for admin user with email:', user.email);
-      console.log(body);
+      console.log('Status code:', response.statusCode);
+      console.log('Body:', body);
       return callback(new Error(body));
     }
 
-    const data = JSON.parse(body);
+    const data = JSON.parse(body).filter(obj => obj.identities
+      .findIndex(identity => identity.connection === adminConnection) >= 0);
+
     if (data.length !== 1) {
-      console.log('Expected to find one admin user but found', data.length);
+      // Failing from this rule should halt the auth pipeline so we don't run Pepper registration,
+      // which should be the next rule. Auth0 will still create a profile but at least it won't be
+      // created in Pepper yet.
+      console.log(`Expected to find one admin user but found ${data.length} using email ${user.email}`);
       return callback(new Error("Could not find a corresponding study admin account"));
     }
 
@@ -92,9 +97,10 @@ function(user, context, callback) {
             return callback(err);
           }
 
-          if (response.statusCode >= 400) {
+          if (response.statusCode !== 201) {
             console.log('Failed to link identities');
-            console.log(body);
+            console.log('Status code:', response.statusCode);
+            console.log('Body:', body);
             return callback(new Error(body));
           }
 
--- a/study-builder/tenants/testboston/rules/link-admin-accounts.js
+++ b/study-builder/tenants/testboston/rules/link-admin-accounts.js
@@ -1,0 +1,119 @@
+// This rule will look for admin account and link it to the Google social account. Account link is
+// determined by using the email. This is only executed for the admin client and only if the
+// accounts are not already linked.
+//
+// This was derived from Auth0's rule template from here:
+// https://github.com/auth0/rules/blob/b381ef4a1d5df16c1716ea066f46cff87b88f7f1/src/rules/link-users-by-email-with-metadata.js
+//
+// This assumes the rule runtime is Node 12, which is configured in Auth0 Dashboard Tenant Settings.
+function(user, context, callback) {
+  const _ = require('lodash');
+  const request = require('request');
+  const adminConnection = 'study-admin-authentication';
+  const googleConnection = 'google-oauth2';
+
+  // Note that client metadata key/values are all strings.
+  context.clientMetadata = context.clientMetadata || {};
+  const isAdminClient = context.clientMetadata.isAdminClient === 'true';
+
+  if (!isAdminClient || context.connection !== googleConnection) {
+    console.log('Either not admin client or not google connection, skipping account linking');
+    return callback(null, user, context);
+  }
+
+  const adminIdentity = user.identities.find(identity => identity.connection === adminConnection);
+  if (adminIdentity) {
+    console.log('Current user already has admin identity, meaning google identity has already been linked');
+    return callback(null, user, context);
+  }
+
+  request({
+    url: auth0.baseUrl + '/users',
+    headers: {
+      Authorization: 'Bearer ' + auth0.accessToken
+    },
+    qs: {
+      search_engine: 'v3',
+      q: `identities.connection:"${adminConnection}" AND email:"${user.email}"`
+    }
+  }, function(err, response, body) {
+    if (err) {
+      console.log('Error while searching for admin user with email:', user.email);
+      console.log(err);
+      return callback(err);
+    }
+
+    if (response.statusCode !== 200) {
+      console.log('Failed to search for admin user with email:', user.email);
+      console.log(body);
+      return callback(new Error(body));
+    }
+
+    const data = JSON.parse(body);
+    if (data.length !== 1) {
+      console.log('Expected to find one admin user but found', data.length);
+      return callback(new Error("Could not find a corresponding study admin account"));
+    }
+
+    const adminUser = data[0];
+    const provider = user.identities[0].provider;
+    const providerUserId = user.identities[0].user_id;
+
+    console.log('Current user:', user);
+    console.log('Found admin user:', adminUser);
+
+    const mergeCustomizer = function(objectValue, sourceValue) {
+      if (_.isArray(objectValue)) {
+        return sourceValue.concat(objectValue);
+      }
+    };
+    const mergedUserMetadata = _.merge({}, user.user_metadata, adminUser.user_metadata, mergeCustomizer);
+    const mergedAppMetadata = _.merge({}, user.app_metadata, adminUser.app_metadata, mergeCustomizer);
+
+    console.log('Merged user_metadata:', mergedUserMetadata);
+    console.log('Merged app_metadata:', mergedAppMetadata);
+
+    auth0.users.updateUserMetadata(adminUser.user_id, mergedUserMetadata)
+      .then(auth0.users.updateAppMetadata(adminUser.user_id, mergedAppMetadata))
+      .then(function() {
+        request.post({
+          url: auth0.baseUrl + '/users/' + adminUser.user_id + '/identities',
+          headers: {
+            Authorization: 'Bearer ' + auth0.accessToken
+          },
+          json: {
+            provider: provider,
+            user_id: String(providerUserId)
+          }
+        }, function(err, response, body) {
+          if (err) {
+            console.log('Error while linking identities');
+            console.log(err);
+            return callback(err);
+          }
+
+          if (response.statusCode >= 400) {
+            console.log('Failed to link identities');
+            console.log(body);
+            return callback(new Error(body));
+          }
+
+          console.log(`Successfully linked identity ${user.user_id} to admin user ${adminUser.user_id}`);
+
+          // Setting the primary user seems to be required so that we designate the original admin
+          // user is the primary account.
+          context.primaryUser = adminUser.user_id;
+
+          // We assume the next rule in the pipeline is Pepper registration. We want the registration
+          // to work with the original admin user id, so we pass along the admin user object rather
+          // than the current user.
+          callback(null, adminUser, context);
+        });
+      })
+      .catch(function(err) {
+        console.log('Error while updating metadata');
+        console.log(err);
+        callback(err);
+      });
+  });
+}
--- a/study-builder/tenants/testboston/tenant.yaml
+++ b/study-builder/tenants/testboston/tenant.yaml
@@ -4,14 +4,21 @@ tenant:
 databases:
   - name: study-admin-authentication
     strategy: auth0
+    # Idea is to NOT enable this database for any client, so that clients can't add users to it.
+    # However, you need to enable database for AT LEAST one client, otherwise Auth0 will mark it as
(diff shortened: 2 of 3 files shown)


study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+    if (response.statusCode !== 200) {
+      console.log('Failed to search for admin user with email:', user.email);
+      console.log(body);
+      return callback(new Error(body));
+    }
+
+    const data = JSON.parse(body);
+    if (data.length !== 1) {
+      console.log('Expected to find one admin user but found', data.length);
+      return callback(new Error("Could not find a corresponding study admin account"));
```
yufengwng, line 60: This means anyone trying to login will fail, unless we find them in the `study-admin-authentication` database. And users need to be manually added to that database, so this gives us fine control of who gets in.

Also, failing here will cause the auth pipeline to halt. Since this is the first rule (see `tenant.yaml`) it means we won't run the `register-user-in-pepper.js` rule. Auth0 will still create a profile for them, but at the very least, it won't hit Pepper so no users will be created in backend.

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+            return callback(err);
+          }
+
+          if (response.statusCode >= 400) {
+            console.log('Failed to link identities');
+            console.log(body);
+            return callback(new Error(body));
+          }
+
+          console.log(`Successfully linked identity ${user.user_id} to admin user ${adminUser.user_id}`);
```
yufengwng, line 109: Note that once identity linking is done, Auth0 will actually delete the google profile and merge it into the database profile. ("profile" means user profile in Auth0)

study-builder/tenants/testboston/tenant.yaml
```diff
@@ -1,12 +1,28 @@
...
   default_directory: Username-Password-Authentication
 
+databases:
+  - name: study-admin-authentication
+    strategy: auth0
+    enabled_clients:
+      - Auth0 Management API
+    is_domain_connection: false
+    options:
+      disable_signup: true
```
yufengwng, line 16: This means we can only manually add users to this database, giving us fine control.

study-builder/tenants/testboston/tenant.yaml
```diff
@@ -1,12 +1,28 @@
 tenant:
   default_directory: Username-Password-Authentication
 
+databases:
+  - name: study-admin-authentication
+    strategy: auth0
+    enabled_clients:
+      - Auth0 Management API
```
yufengwng, line 12: Idea is to NOT enable this database for any client, so that clients can't add users to it. However, you need to enable database for AT LEAST one client, otherwise Auth0 will mark it as inactive and you can't even manually add users. Auth0 Management client is an arbitrary choice, but seems fine for now.

study-builder/tenants/testboston/tenant.yaml
```diff
@@ -44,6 +60,7 @@ clients:
       ##LOCALHOST_CALLBACKS##
     allowed_logout_urls:
       - '##BASE_URL##'
+      - '##BASE_URL##/error'
```
yufengwng, line 70: When someone attempts to login with google-oauth2, they will fail login if they're not allowed, so we redirect them to an error page. Need to whitelist the URL otherwise Auth0 won't redirect to it after logout.

Since they're likely not logged in when this happens, frontend code doesn't really know if we're dealing with admin client or not, so need to whitelist URL in both clients.

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
+// This rule will look for admin account and link it to the Google social account. Account link is
+// determined by using the email. This is only executed for the admin client and only if the
+// accounts are not already linked.
+//
+// This was derived from Auth0's rule template from here:
+// https://github.com/auth0/rules/blob/b381ef4a1d5df16c1716ea066f46cff87b88f7f1/src/rules/link-users-by-email-with-metadata.js
```
yufengwng, line 6: It sounds like Auth0 used to provide this as a starting template when you create a new rule from the dashboard, but apparently they removed it (not sure why). I found their code through the Auth0 community forum and their git repo history.

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  const adminIdentity = user.identities.find(identity => identity.connection === adminConnection);
+  if (adminIdentity) {
+    console.log('Current user already has admin identity, meaning google identity has already been linked');
+    return callback(null, user, context);
+  }
+
+  request({
+    url: auth0.baseUrl + '/users',
```
bskinner: Auth0's docs recommend not using the endpoint during the authentication process as it is not immediately consistent, and references the `GET /api/v2/users-by-email` endpoint. Will the alternate endpoint serve an equivalent purpose for us here?

https://auth0.com/docs/users/user-search/retrieve-users-with-get-users-by-email-endpoint

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }, function(err, response, body) {
+    if (err) {
+      console.log('Error while searching for admin user with email:', user.email);
+      console.log(err);
+      return callback(err);
+    }
+
+    if (response.statusCode !== 200) {
+      console.log('Failed to search for admin user with email:', user.email);
```
bskinner, line 46: I'd consider logging the status code we received here, as well. Might save some digging later on.

study-builder/tenants/testboston/rules/link-admin-accounts.js
bskinner: That's great- sounds like it'll save us a few headaches later. Given we're squashing the commits together, could you add a comment noting that behavior?

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+      console.log(body);
+      return callback(new Error(body));
+    }
+
+    const data = JSON.parse(body);
+    if (data.length !== 1) {
```
bskinner, line 55: Would there be any valid cases where there is more than 1 result returned from the search?

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+            user_id: String(providerUserId)
+          }
+        }, function(err, response, body) {
+          if (err) {
+            console.log('Error while linking identities');
+            console.log(err);
+            return callback(err);
+          }
+
+          if (response.statusCode >= 400) {
```
bskinner: It looks like a `201` is the status code for a successful request to this endpoint, so a `!== 201` would net us a bit more safety against false positives here.
```suggestion
          if (response.statusCode !== 201) {
```

study-builder/tenants/testboston/rules/link-admin-accounts.js
```diff
@@ -0,0 +1,119 @@
...
+            return callback(new Error(body));
+          }
+
+          console.log(`Successfully linked identity ${user.user_id} to admin user ${adminUser.user_id}`);
+
+          // Setting the primary user seems to be required so that we designate the original admin
+          // user is the primary account.
+          context.primaryUser = adminUser.user_id;
+
+          // We assume the next rule in the pipeline is Pepper registration. We want the registration
```
bskinner, line 115: This is a very helpful comment- nice!

study-builder/tenants/testboston/tenant.yaml
bskinner: Where is this error page being generated from? Would it be a better option to return a machine-readable error code in the response from Auth0 and let the front-end handle the error entirely on its end?

bskinner (commented):
Looks good! A couple comments which need resolution, but it looks good to me once those are addressed.

study-builder/tenants/testboston/tenant.yaml
yufengwng: That is actually what we have. Auth0 will return an error, and the apps decide what they want to do. The error page (even it's URL) is defined by the apps. This `tenant.yaml` is specific to the TestBoston tenant used by the TestBoston apps, and its error paged is defined to be located at `/error`.

study-builder/tenants/testboston/rules/link-admin-accounts.js
bskinner: Saw this ordering clearly specified in the tenant setup file further down- my comment can be disregarded.

study-builder/tenants/testboston/rules/link-admin-accounts.js
yufengwng: Good catch! I'll double-check the docs and switch to the `/users-by-email` endpoint.

study-builder/tenants/testboston/rules/link-admin-accounts.js
yufengwng: Hmmm, I couldn't think of any valid cases for having more than 1 result -- given how things are setup, that seems to me like someone messed up adding an admin.

study-builder/tenants/testboston/rules/link-admin-accounts.js
yufengwng: Switched!

study-builder/tenants/testboston/rules/link-admin-accounts.js
yufengwng: Done.

study-builder/tenants/testboston/rules/link-admin-accounts.js
yufengwng: Yup!

MocanaAtBroad:
Have a couple of questions that I am not sure about by reading code:
1) The process of adding an admin user. Am I right in understanding that we would create the user in the Auth0 tenant and add them to the study-admin-connection using as an email their Google email address that they would later use to login. Do they need the password? Better question: what is the admin registration process from perspective of the user?
2) And if above is correct, the rule would create a second social-login user record when they try to login using the Google authenticated Auth0 application?
3) After the linking is performed by the rule, can they then login with password if they choose to?

yufengwng:
@MocanaAtBroad Thanks for the questions! Here's my reply:

1. As I posted in Slack, [this wiki page](https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/1934458957/Add+New+TestBoston+Study+Admin) outlines the SOP for adding new admin. Your understanding is correct that this is almost all manual. There is not really such a thing as "admin registration process from user perspective". The registration process is asking us to add them, either via Jen or Sarah. They do not use password. They are only allowed to do Google login.
2. Technically, this new rule will not create social-login user record -- that has already been created by Auth0 by default. What this rule does is link that social-login user record to their Auth0 database record that we manually added. In fact, after linking is done, this social-login user record is "deleted" by Auth0 and "merged" into the other record.
3. As mentioned in wiki page, they do not login with username/password, only Google login.

Regarding the question about passwords, note that the Auth0 database connection is NOT ENABLED for the admin client, so they are not even allowed to enter in username/password.

MocanaAtBroad:
OK. Thanks for the links.
The part I am wondering about is whether is it really necessary to add the user manually into Auth0 AND manually into the database.

I am unclear on whether when the link admin rule executes if the registration rule also executes.

Presumably it would be possible to register the admin user like all the other users; via the registration route if we can be sure that the caller of the route is Auth0 and can be trusted to tell us this person was entered as an admin in Auth0.

Is this something you considered and dismissed as technically not doable or too hard?

> @MocanaAtBroad Thanks for the questions! Here's my reply:
> 
> 1. As I posted in Slack, [this wiki page](https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/1934458957/Add+New+TestBoston+Study+Admin) outlines the SOP for adding new admin. Your understanding is correct that this is almost all manual. There is not really such a thing as "admin registration process from user perspective". The registration process is asking us to add them, either via Jen or Sarah. They do not use password. They are only allowed to do Google login.
> 2. Technically, this new rule will not create social-login user record -- that has already been created by Auth0 by default. What this rule does is link that social-login user record to their Auth0 database record that we manually added. In fact, after linking is done, this social-login user record is "deleted" by Auth0 and "merged" into the other record.
> 3. As mentioned in wiki page, they do not login with username/password, only Google login.
> 
> Regarding the question about passwords, note that the Auth0 database connection is NOT ENABLED for the admin client, so they are not even allowed to enter in username/password.

yufengwng:
> The part I am wondering about is whether is it really necessary to add the user manually into Auth0 AND manually into the database.

There's trade-offs here. We can go without the manual step in Auth0, but the drawback here is that any Google email will be able to login. They will see the Prism interface, although they won't be able to do anything since they are not assigned as an admin. The login flow will also call Pepper and a user will be created in the database. And, given how the code is right now, there are also other undesirable effects as well. If we want to avoid those, we'll need to update the backend code. So yes, it's not necessary, but I have opted to go the other route, which gives us fine control of which email can login, no code change needed, no undesirable database changes.

Since this is only for TestBoston, I figured that it'll be best to try this out in Auth0 configurations rather than making code changes. If we find a better option, we can switch then!

> I am unclear on whether when the link admin rule executes if the registration rule also executes.

If link-admin rule finished successfully, the registration rule will run next, as I alluded to in the code comments. Code comments also point out that if link-admin rule fails, it halts the process and registration rule won't run. The execution order is specified in the `tenant.yaml` file.

> Presumably it would be possible to register the admin user like all the other users; via the registration route if we can be sure that the caller of the route is Auth0 and can be trusted to tell us this person was entered as an admin in Auth0.

I did ponder this briefly, but I'm hesitate on changing the `UserRegistrationRoute` as it's a bit complex. Also, I don't know what will be a good "admin registration process" will look like yet, so I don't want to prematurely change critical backend code. My impression is that we're not adding 100s of study admins, so a manual process should be fine.

bskinner:
>Presumably it would be possible to register the admin user like all the other users; via the registration route if we can be sure that the caller of the route is Auth0 and can be trusted to tell us this person was entered as an admin in Auth0.

At that point you're still trusting user input to indicate "I'd like to create an admin user." That is exceedingly risky- we'd need more information to make sure that request is coming from Auth0 than just the IP filtering. This solution works, and keeps the logic inside Auth0. I could see Auth0 roles or user metadata playing a role here eventually, since we can query Auth0 for that from the backend using our SA.

MocanaAtBroad (approved):
Works as advertised! LGTM.

MocanaAtBroad:
@bskinner : sorry. edited your comment by mistake instead of quote replying!
> > Presumably it would be possible to register the admin user like all the other users; via the registration route if we can be sure that the caller of the route is Auth0 and can be trusted to tell us this person was entered as an admin in Auth0.
> 
> > At that point you're still trusting user input to indicate "I'd like to create an admin user." That is exceedingly risky- we'd need more information to make sure that request is coming from Auth0 than just the IP filtering. This solution works, and keeps the logic inside Auth0. I could see Auth0 roles or user metadata playing a role here eventually, since we can query Auth0 for that from the backend using our SA.
> 
We have mechanisms in place to ensure that it is the Auth0 client that is initiating the user registration API call.
My thinking is that we are already delegating (and entrusting) authentication to Auth0. The next logical step is to delegate to Auth0 role management. It is a function that Auth0 is designed to support.

bskinner:
> The next logical step is to delegate to Auth0 role management. It is a function that Auth0 is designed to support.

I never said don't use it- The last sentence in my previous message calls out using Auth0 for authorization. All I'm saying is IP address validation is a security measure, not authentication.

PR #559, opened 2020-12-09
DDP-5385

YuriyPavlovBI:
## Context

_Say a few words about the bigger picture beyond this PR. How do your changes improve life for our users? Link directly to the jira ticket (just type "DDP-..." and it'll automatically link) to help your teammates quickly get up to speed on the motivation for these changes._

## Checklist

- [ ] I have labeled the type of changes involved using the `C-*` labels.
- [ ] I have assessed potential risks and labeled using the `R-*` labels.
- [ ] I have considered error handling and alerts, and added `L-*` labels as needed.
- [ ] I have considered security and privacy, and added `I-*` labels as needed
- [ ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [ ] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ ] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

_If unsure or need help with any of the above items, add the `help wanted` label. For items that starts with `If applicable`, if it is not applicable, check it off and add `n/a` in front._

## FUD Score

_Overall, how are you feeling about these changes?_

- [ ] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [ ] They are user-visible in dev as a regular user journey and require no additional instructions.
- [ ] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [ ] I have written automated positive tests
- [ ] I have written automated negative tests
- [ ] I have written zero automated tests but have poked around locally to verify proper functionality
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [ ] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #565, opened 2020-12-15
Initial version of File Answer API changes

MocanaAtBroad:
**Updated 1/4/2020:**
Extending activity instance API to allow for file uploads. At broad strokes, proposed API changes:

- File Question is introduced.
- Client will POST a request for a file upload URL
- Client will upload directly to bucket using resumable or non-resumable upload
- Client will PATCH the upload id to pepper-backend to indicate conclusion of upload.
- **GET** activity instance will include in response body the name of the file and the size for client to display.
- A **DELETE** /answer/answerGuid end-point has been added to allow for client to remove file.

Note that this is a bit different then outcome of discussion during PRty time before break. The proposal was to have client obtain upload URL with answer PATCH.

After some exchanges on Slack and considering more carefully the back-end implementation I have concluded that having the client POST to obtain an upload URL followed by PATCHing the upload id when upload is completed is the better choice.

Trying to obtain the signed URL via PATCH call would leave it up to the server to figure out if/when the file upload actually occurred. 

Three possible solutions to this problem that I can think off:
- Having the client PATCH again to inform of the upload completion (IMO inconsistent with other answers/2 different PATCH formats for one answer), 

- Having the server listen from bucket for notification (IMO would complicate back-end implementation, introducing asynchronicity to the answer handling and final PUT. Listeners would need to be added to pepper-backend or would need to somehow ensure housekeeping is handling the notification immediately.

- Another possible solution would be to delay checking file upload until form SUBMIT,  but this would cause problems if activity is unsubmitted: what is the answer information we would display in the web page in case of a page reload?).


pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -463,9 +463,15 @@ CreateFileUploadUrlResponse:
           url:
             type: string
             description: Signed URL which needs to be used for file upload
-          guid:
+          httpMethod:
```
bskinner: Stated this in Yuriy's PR, but this is encoding implementation details about Google's Bucket API into our API, and may be misleading given there are more behaviors than just the verb. Developers using this API should be familiar with the use of the Bucket APIs.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -378,6 +398,12 @@ DateValue:
       type: integer
       nullable: true
       minimum: 0
+FileAnswerValue:
```
bskinner, line 403: Given that this is sent via the `PATCH`, how about we style this after the `FileUploadRequest` and return a `FileUploadResponse` in the `PATCH` response? Saves clients from making another upload request to an alternate endpoint, the server has the bucket id

pepper-apis/docs/specification/src/components/responses.yml
MocanaAtBroad: It does look like URL and the resumable boolean is enough for client to figure out what to do.
577f5cfd

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: Not quite following your suggestion or understand your point about saving another upload request. Could you clarify?  Just to be super-clear about intention here: in the response to the  "create file upload URL request", we include an `uploadId` (along with the Google Bucket signed URL and resumable flag). The intent of client PATCH ing the upload id is to inform server that client has completed the upload that it authorized earlier and link in the back-end the file location.

bskinner: Gist to the rescue!

https://gist.github.com/bskinner/fc715e80cad6e615a84a805feab85e84

While I was typing this up, I had another thought, and realized I might be approaching it from the wrong direction- what if we used a `201 CREATED` instead? Stick the upload URL in the `Location` header?

Downside, we'd lose the `validationFailures` and `blockVisibility` updates for that pass. Given that the only things needed are a `name` and `fileSize`, I think we can get by without the `validationFailures`. It's more likely the `blockVisibility` _might_ change, though- other blocks could change their state based on if a file was uploaded or not.

I think that could be mitigated by adding the platform requirement that a `File` question is not able to affect another `block`'s visiblity. It feels a bit hacky, but good enough for a version 1.

MocanaAtBroad: So the idea would be that the upload URL would be obtained by client PATCHing  file name, size, etc and upload URL would be in either the PATCH response body or in Location header?

bskinner: Yep! My thoughts are:
* It reduces number of requests the client needs
* Feels like it works within our current model
* Gives both the server and client a little more context around the file upload (could be useful for caching down the road, for example)
* In Pepper, it might help the implementation of the bucket creation to be a little more isolated

I am leaning more towards the "modify the `PATCH` body" approach, though. It's a less drastic departure from the current API, and I feel it fits in place a bit better. The `201` would need to be re-designed from its base if we ever enable submitting batched answers in a `PATCH`

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -351,6 +351,28 @@ Answer.Picklist:
           type: array
           items:
             $ref: '../pepper.yml#/components/schemas/SelectedOption'
+Answer.File:
```
bskinner, line 354: Can you expand on what's going on here? It doesn't look like there's enough info for a `PATCH` to connect an answer to a file upload in either the `Answer.File` or `FileInfo` structures.

Edit: I think I see the flow here- the question's stable ID is referenced in the `CreateFileUploadUrlRequestBody` so there never is a `PATCH`, correct?

MocanaAtBroad: The proposal is that the PATCH will have the same shape of a text or boolean question answer, but the `value` property will be the upload id that was. obtained in the same POST body that provided the upload URL. The screenshot below captures the intent:
Screen Shot 2021-01-05 at 2 03 42 PM

The chunk of YAML here is mean to describe the answer value that would appear when we GET the activity instance contents and we have a File answer.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.answer.yml
```diff
@@ -0,0 +1,44 @@
...
+      type: string
+  - in: path
+    name: answerGuid
+    required: true
+    description: the answer guid
+    schema:
+      type: string
+
+delete:
+  operationId: deleteAnswer
```
yufengwng: Regarding idea about `DELETE /answer` endpoint: How about we allow a client to patch a `null` for file answer `value` in the PATCH? I propose that if server sees a `null` then it will somehow "delete" the FileAnswer and mark the FileUpload as "outdated" or similar so the bucket file can be cleaned up.

Related question: What happens when client PATCH a new file upload to a file question that already has a file answer? I assume new file upload should somehow "replace" the old one. What happens to the old FileAnswer then? I propose the old FileAnswer be deleted somehow and its associated FileUpload be marked "outdated" for cleanup. This way, the behavior here will be similar to if client patched `null`, so I think we'll have some consistency?

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: That's an awkward interaction. There shouldn't be 2 types of "Answer" objects.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.answer.yml
MocanaAtBroad: > Regarding idea about `DELETE /answer` endpoint: How about we allow a client to patch a `null` for file answer `value` in the PATCH? I propose that if server sees a `null` then it will somehow "delete" the FileAnswer and mark the FileUpload as "outdated" or similar so the bucket file can be cleaned up.
> 
Deleting answer with a null PATCH value is consistent with other answer types, so I agree we can drop `DELETE` (the DELETE operation I believe was necessary when PATCH was being used for the purpose of getting the upload URL).

Also agree about handling "deleted" files.

> Related question: What happens when client PATCH a new file upload to a file question that already has a file answer? I assume new file upload should somehow "replace" the old one. What happens to the old FileAnswer then? I propose the old FileAnswer be deleted somehow and its associated FileUpload be marked "outdated" for cleanup. This way, the behavior here will be similar to if client patched `null`, so I think we'll have some consistency?

I think that we should aim to reproduce the semantics of PATCH that we are using for existing questions. So if the activity instance is still writeable (i.e., not readonly), then previous file answers should be able to be replaced.

The proposed implementation of file answer is to have a reference (foreign key) to a file upload object. The file uploads will have a status. 
There is also supposed to be a job that cleans out unused uploads. This same job could also remove unreferenced upload records in the database and corresponding files in bucket.

So, yeah, a null for answer or a new file answer ought to trigger similar handling for existing/previous file uploads.

yufengwng: Sounds good.

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: I recognize there is a bit of asymmetry between the PATCH body and the JSON we are including as the answer for File question in the activity instance.
Root cause is the difference between what the client needs to tell the server ("I just completed that upload you authorized me earlier to do") vs what the client needs to know to display something reasonable in the completed activity instance web page (file name and file size).  I think is fine. The file answer handling was always going to be a little different than the other answers.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.answer.yml
MocanaAtBroad: 0d300fcd
Screen Shot 2021-01-05 at 3 42 29 PM

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -351,6 +351,28 @@ Answer.Picklist:
...
+  allOf:
+    - $ref: '../pepper.yml#/components/schemas/Answer'
+    - type: object
+      required:
+        - value
+      properties:
+        value:
+          type: object
+          $ref: '../pepper.yml#/components/schemas/FileInfo'
+FileInfo:
```
bskinner, line 364: This should include the `uploadId` as well, if one exists. The client might need to reference back to it.

bskinner: Marking this resolved- walked through a couple uploads and there doesn't seem to be a reason for the client to hold onto the upload id once it's been `PATCH`ed.

yufengwng (approved):
Looks like a good start. I'm assuming that this PR is just to serve as discussion and we're not actually merging this in?

MocanaAtBroad:
> Looks like a good start. I'm assuming that this PR is just to serve as discussion and we're not actually merging this in?

I was thinking of merging it into Yuriy's PR https://github.com/broadinstitute/ddp-study-server/pull/559

He has implemented a some of this functionality already and now having an agreement on the API should provide clarity to how we want this thing t. work. Make sense to you?

bskinner:
> > Looks like a good start. I'm assuming that this PR is just to serve as discussion and we're not actually merging this in?
> 
> I was thinking of merging it into Yuriy's PR #559
> 
> He has implemented a some of this functionality already and now having an agreement on the API should provide clarity to how we want this thing t. work. Make sense to you?

Doesn't that PR implement this as an unauthenticated API? I thought we had quashed that.

MocanaAtBroad:
> > > Looks like a good start. I'm assuming that this PR is just to serve as discussion and we're not actually merging this in?
> > 
> > 
> > I was thinking of merging it into Yuriy's PR #559
> > He has implemented a some of this functionality already and now having an agreement on the API should provide clarity to how we want this thing t. work. Make sense to you?
> 
> Doesn't that PR implement this as an unauthenticated API? I thought we had quashed that.

We should go back and look at what he has done in light of our File API discussions. At the very least he was generating signed upload URLs via a POST route, so that is at least a start/has figured out how to create these URLs.

But the actual problem he was actually trying to solve was the unauthenticated emails with attachments. 
Maybe we should try to figure out what we want for that problem too? We talked a lot about it, including possibly using Cloud Functions, but did not reach a final conclusion.

bskinner:
> We should go back and look at what he has done in light of our File API discussions. At the very least he was generating signed upload URLs via a POST route, so that is at least a start/has figured out how to create these URLs.

Sounds reasonable- my concern is primarily giving an implicit green-light to resuming work in that direction.

bskinner:
Something that came to mind here, what is the expected behavior if a client `PATCH`s over an `uploadId` that has not been completed yet? If the server requires that the upload be completed first, this would force an implementation of the mobile apps that prevents the user from continuing until the file has been completely uploaded.

yufengwng:
@bskinner Re: "what is the expected behavior if a client PATCHs over an uploadId that has not been completed yet".

Based on the design I'm reading here, PATCH will result in error since file is not finished uploading yet. Implication is that client will need to do upload first before PATCH-ing.

yufengwng:
Another question for folks: can we trust the authenticated user to not upload malware?

I feel we should implement malware scanning even though the files are coming from authenticated users -- anyone can sign up for a study, so who knows who's coming into our system and uploading stuff. I believe this should be solely a server-side implementation detail, and should not affect client-facing API surface, but not too sure. Basic idea is to scan files "in the background" and mark file upload objects as "pending/quarantined/scanned". I think we can still operate on the file objects even though they haven't been scanned, so that as far as a client is concerned there's no changes in terms of API. However, I'm not too sure yet what happens if file is quarantined and we need to delete it.

bskinner:
> Another question for folks: can we trust the authenticated user to not upload malware?

No. User input should never be implicitly trusted- I'm all for quarantining and scanning files before we do anything with them.

Edit: To clarify, this has very little to do with trusting the user. It's entirely possible a file could carry some sort of malware without the user's knowledge.

On a similar topic, all image uploads should likely have their metadata stripped while in quarantine.

MocanaAtBroad:
For implementation, we should pursue the scanning that Denis talked about https://www.clamav.net/ .
Stripping metadata from images is a good idea too.
Don't think this affects API, but we should figure out what we want to have happen when a document is rejected. Do we notify the user somehow? Do we change the status of the activity or the answer? 
There is always a twist.
Regardless, I suggest we deal with the notification issue separately.

MocanaAtBroad:
> @bskinner Re: "what is the expected behavior if a client PATCHs over an uploadId that has not been completed yet".
> 
> Based on the design I'm reading here, PATCH will result in error since file is not finished uploading yet. Implication is that client will need to do upload first before PATCH-ing.

➕

PR #566, opened 2020-12-15
DDP-5234 allow study admins to view and edit hidden/read-only activities

yufengwng:
## Context

This PR updates the various Activity APIs to allow study admins to view as well as edit activity data even when they are in hidden or read-only mode. Other APIs (such as Address and Medical Providers) does not seem to impose restrictions on hidden or read-only, so no changes was made there. We also return a new `isHidden` property so clients can choose to display appropriate indicators.

See https://github.com/broadinstitute/ddp-angular/pull/394 for the Angular SDK changes and how TestBoston uses the new features.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -116,6 +116,7 @@ Activity:
     - lastUpdatedText
     - lastUpdated
     - isFollowup
+    - isHidden
```
yufengwng, line 119: @fat23cat @bskinner We're adding new properties to APIs. There's actually two different flags for "hidden" but I don't think that's a detail client needs to know, so I settled on one `isHidden` flag.

bskinner: What is the expected behavior if the flag is set?

zyme: I agree that internal to the server, the two flags are okay (although kind of confusing) but that client should one `isHidden`.  @yufengwng are we only including an activity if `isHidden` is true _and_ the user has admin access?  If not, we should do that.  Otherwise we are leaking an activity and asking the client to hide it.  So my two cents @bskinner is that the mobile client should ignore this for now because we haven't thought through the study staff UX when using the masquerade-type feature that this is used for in the prism route (currently only for testboston).

zyme: I shall answer my own question: we are filtering things appropriately (`filterActivityInstancesFromDisplay()`)

zyme: So @bskinner I think it's safe to take no action client-wise, but I think the intended behavior here should be a comment in the schema to the effect of "If true, client should indicate to the user that this activity is not visible to the participant, still make it visible"?

bskinner: That seems contradictory on its face- Is that from the perspective of the operator?

pepper-apis/src/test/java/org/broadinstitute/ddp/route/GetActivityInstanceRouteStandaloneTest.java
```diff
@@ -735,6 +736,30 @@ public void test_whenIsHidden_thenNotFound() {
         }
     }
 
+    @Test
+    public void testStudyAdmin_canRetrieveInstances() {
```
zyme, line 740: Niiiice!

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -168,7 +170,7 @@ public Object handle(Request request, Response response) {
                 return res;
             }
 
-            if (ActivityInstanceUtil.isReadonly(formActivityDef.getEditTimeoutSec(), instanceDto.getCreatedAtMillis(),
+            if (!isStudyAdmin && ActivityInstanceUtil.isReadonly(formActivityDef.getEditTimeoutSec(), instanceDto.getCreatedAtMillis(),
```
zyme, line 173: I wonder if we should add an `isStudyAdmin` param to various `isReadonly()` and `isHidden()` methods?  This would centralize the decision about whether a thing is read-only given the additional context of the user.

zyme (commented):
Looks good overall.  I am wondering if we should pass `isStudyAdmin` into `ActivityInstanceDao.listActivityInstancesForUser()` so that we reduce the chance of other calls (of which I admit there are few) getting mixed up and forgetting to do an auth check after returning the list.  Failure to do so could result in a participant seeing something that they shouldn't see.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
yufengwng: From what I can tell, `ActivityInstanceUtil.isReadonly()` is used in lower-level DAO methods. I'm hesitate to pass in `isStudyAdmin` all the way down there because:

1. It will need to be pass through several layers, resulting in a handful of refactors.
2. And most importantly, I don't think these two should be tied together. "read-only" is a property of the activity instance, and so the computation for that property ideally shouldn't involved other things like the authentication status. `isStudyAdmin` is really just a way for us to bypass that "read-only" property -- the activity instance itself is still inherently read-only.

yufengwng:
@zyme Re: `ActivityInstanceDao.listActivityInstancesForUser()`.

Again, I'm hesitate to tie `isStudyAdmin` into that method. Given how things are, we actually need to fetch all the activity instances (hidden or not) because there's a separate step for numbering/rendering.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Re: the two flags. So this evolved by design. One flag hides a single instance, while the other flag hides all instances of an activity regardless of the other flag, so it gives us control over hiding.

Re: expected behavior. `isHidden` means whether activity was hidden _from the participant_. It relates to the participant, which is my implicit understanding but I should have put that in words in the documentation. If flag is set, then as Zim mentioned client should prefer to show some indicator that it's hidden _from the participant_. Sometimes, such as in the case of study admin, the operator is not the participant, so it may be hidden from participant but not from operator.

zyme (dismissed):

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Ok, so let me parity check my understanding:

* If `isHidden` is `false`, the participant _may_ see the instance, or activity, as normal.
* If `isHidden` is `true`, the participant _may not_ see the instance, or activity. Users with management privileges _may_ see the participants instances, or activites, but should have a visible indicator they are hidden from the participant.

Sound about right?

bskinner (approved):

PR #568, opened 2020-12-16
Set default value for application param

MocanaAtBroad:
## Context
If used as specified in --help, was not generating a config file. Set a default value for it


pepper-apis/api-build.sh
```diff
@@ -223,7 +223,7 @@ function docker_nginx_push {
 }
 
 function print_usage() {
-    echo "usage: $NAME <version> <environment> <dir> <option>..."
+    echo "usage: $NAME <version> <environment> <dir> [housekeeping] g<option>..."
```
bskinner: Is the `...] g<option>` a typo?

pepper-apis/api-build.sh
```diff
@@ -293,6 +293,7 @@ DOCS_PROXIED_HOST=$DOCS_PROXIED_HOST
 VERSION=$1
 ENVIRONMENT=$2
 DIR=$3
+APPLICATION=unknown
```
bskinner, line 296: What is the expected behavior if the application is not passed in the arguments?

pepper-apis/api-build.sh
MocanaAtBroad: 8c547f80

pepper-apis/api-build.sh
MocanaAtBroad: This `APPLICATION` value filters its way to template generation.
Up to recently housekeeping and pepper-backend shared exact application.conf template .
Had need to generate slightly different .conf files for each.
So if APPLICATION=housekeeping , there is a conditional statement in a template that renders a param value in application.conf . If it is not housekeeping, then it render the default value for param.

I had made the param optional in the interest of keeping everything backwards compatible (i.e., previous usage was preserved). I introduced bug misjudging how GO templates would handle missing value.

bskinner: Thank you!

bskinner (approved):

PR #569, opened 2020-12-17
DDP-5455 fix read-only calculation in Patch Answers

yufengwng:
## Context

Fixes DDP-5455.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
@@ -1,10 +1,7 @@
 package org.broadinstitute.ddp.db.dto;
 
-import java.util.Optional;
-
 import org.broadinstitute.ddp.model.activity.types.ActivityType;
 import org.broadinstitute.ddp.model.activity.types.InstanceStatusType;
-
 import org.jdbi.v3.core.mapper.reflect.ColumnName;
 import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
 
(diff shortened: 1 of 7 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
```diff
@@ -95,10 +95,6 @@ public Long getFirstCompletedAt() {
         return firstCompletedAt;
     }
 
-    public boolean isReadonly() {
-        return Optional.ofNullable(isReadonly).orElse(false);
```
yufengwng, line 99: I removed this method so we don't default read-only to `false` and cause another issue with bypassing check.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -169,9 +169,8 @@ public Object handle(Request request, Response response) {
             }
 
             if (ActivityInstanceUtil.isReadonly(formActivityDef.getEditTimeoutSec(), instanceDto.getCreatedAtMillis(),
-                    instanceDto.getStatusType().name(), formActivityDef.isWriteOnce(), instanceDto.isReadonly())) {
-                String msg = "Activity instance with GUID " + instanceGuid
-                        + " is read-only, cannot submit answer(s) for it";
+                    instanceDto.getStatusType().name(), formActivityDef.isWriteOnce(), instanceDto.getReadonly())) {
```
yufengwng, line 172: `isReadonly()` -> `getReadonly()` is the fix.

MocanaAtBroad (dismissed):

MocanaAtBroad:
You sure you want to allow getReadOnly() to be a null value?  Would it make sense to have a default value of false in the database? 3 possible values for a boolean (even a Boolean) is a little strange.

yufengwng:
@MocanaAtBroad I recall that it was a deliberate choice. The read-only on an activity instance acts as a "master" switch. If we want to override everything else and mark an instance as NOT read-only, we'll set it to `false`. So with that, we need to allow `null` or else every instance will bypass read-only check...

MocanaAtBroad (approved):

PR #571, opened 2020-12-17
DDP-5454 limit search query pagination size for auth0 user emails

yufengwng:
## Context

Potentially fixes DDP-5454.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/util/Auth0Util.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/util/Auth0Util.java
@@ -367,14 +367,14 @@ public class Auth0Util {
             return new HashMap<>();
         }
 
-        int maxPerPage = 100;
+        int maxPerPage = 50;
         Map<String, String> results = new HashMap<>();
         List<String> ids = new ArrayList<>(auth0UserIds);
         ManagementAPI auth0Mgmt = new ManagementAPI(baseUrl, mgmtApiToken);
 
-        // IMPORTANT: in order to satisfy certain limits and restrictions, especially URL length limits, we "pagination"
-        // through the authUserIds. 100 auth0UserIds in the Lucene query seems like a safe limit. And it works out that
-        // the max items per page from the API is 100.
+        // IMPORTANT: In order to satisfy certain limits and restrictions, especially URL length limits, we "pagination"
+        // through the auth0UserIds. We had issues with 100 auth0UserIds, so 50 in the Lucene query seems like a safe bet.
+        // 50 also happens to be Auth0's default for this parameter.
         for (int i = 0; i < ids.size(); i += maxPerPage) {
             int end = Math.min(i + maxPerPage, ids.size());
             List<String> subset = ids.subList(i, end);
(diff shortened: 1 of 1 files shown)


yufengwng:
Data exports for Prion is failing with a stacktrace that looks like this:
```
Error while retrieving auth0 user emails, continuing pagination No content to map due to end-of-input
at [Source: (String)""; line: 1, column: 0]
at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4146)
at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4001)
at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3043)
at com.auth0.net.CustomRequest.createResponseException(CustomRequest.java:103)
at com.auth0.net.CustomRequest.parseResponse(CustomRequest.java:57)
at com.auth0.net.BaseRequest.execute(BaseRequest.java:37)
at org.broadinstitute.ddp.util.Auth0Util.getUserPassConnEmailsByAuth0UserIds(Auth0Util.java:393)
at org.broadinstitute.ddp.export.DataExporter.fetchAndCacheAuth0Emails(DataExporter.java:166)
...
```

This suggests that it's failing during the Auth0 API call to fetch user emails. Specifically, it got an empty response back and the Auth0 library is failing to deserialize it. One can argue that it's likely a bug in the Auth0 library (I haven't looked if there's a new version of library). I did, however, was able to reproduce this locally and I believe we can fix this on our end.

I ran a Prion data export locally pointed at the `test` environment. I confirmed that we are indeed getting back an empty response, which looks something like this (there is nothing after the headers):

Screen Shot 2020-12-17 at 6 13 55 PM

Auth0's [docs](https://auth0.com/docs/api/management/v2#!/Users/get_users) for the API in question specifies a few scenarios for status code 400. We are using the v3 search engine, and we're not running over the record limit since we're always fetching the first page. So by process of elimination, it likely means something is wrong with our query string.

My suspicion is that we're running over the _URL length limit_. I believe browsers (and probably servers too!) have a limit on how long a URL can be. Since this Auth0 API is a `GET` call, the auth0_user_ids that we're searching for are all jammed into the URL query string. Taking a look at the auth0_user_ids in Prion (either test, staging, or production), we see an odd pattern like:
```
auth0|auth0|12345...
```
Typically, auth0_user_ids are `auth0|12345...`. Based on what @charlottedye told me, this was due to an undesirable artifact of how import/export was done for Prion migration. This seems fine for Auth0 -- we're able to lookup and work with the user just fine. However, because there's an extra part of `auth0|` in front of each auth0_user_id of imported Prion users, the query string now becomes longer for requests related to Prion. Coincidentally, we only see this issue with Prion data export and not other studies.

So in short, I believe we're hitting a general URL length limit issue, which was surfaced by the fact that Prion imported users have longer ids. The fix here is to reduce the number of auth0_user_ids we include in each search request.

ssettipalli (approved):
May be right  way is to clean it up.
Update auth0 userid in database and auth0 to remove extra auth0 from auth0|auth0|5d9247f1a2e6850dfcxxxx . 
Reducing url size should fix the issue though. :+1:

yufengwng:
@ssettipalli I think cleaning up the data will be ideal, but definitely a more extensive operation than what I can cover in a PR. I'll put in a note about it and discuss with Charlotte once she gets back from vacation.

PR #577, opened 2021-01-05
DDP-5484 fix caching of picklist question options

yufengwng:
## Context

Fixes DDP-5484.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionCachedDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionCachedDao.java
@@ -13,7 +13,7 @@ import org.jdbi.v3.core.Handle;
 import org.redisson.client.RedisException;
 
 public class PicklistQuestionCachedDao extends SQLObjectWrapper<PicklistQuestionDao> implements PicklistQuestionDao {
-    private static Cache<Long, GroupAndOptionDtos> questionIdToGroupAndOptionsCache;
+    private static Cache<String, GroupAndOptionDtos> questionVersionKeyToGroupAndOptionsCache;
 
     public PicklistQuestionCachedDao(Handle handle) {
         super(handle, PicklistQuestionDao.class);
@@ -21,8 +21,9 @@ public class PicklistQuestionCachedDao extends SQLObjectWrapper<PicklistQuestion
     }
 
     private void initializeCache() {
-        if (questionIdToGroupAndOptionsCache == null) {
-            questionIdToGroupAndOptionsCache = CacheService.getInstance().getOrCreateCache("questionIdToGroupAndOptionsCache",
+        if (questionVersionKeyToGroupAndOptionsCache == null) {
+            questionVersionKeyToGroupAndOptionsCache = CacheService.getInstance().getOrCreateCache(
+                    "questionVersionKeyToGroupAndOptionsCache",
                     new Duration(),
                     ModelChangeType.STUDY,
                     this.getClass());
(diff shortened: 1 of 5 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionCachedDao.java
```diff
@@ -56,33 +57,37 @@ public TemplateDao getTemplateDao() {
...
             return delegate.findOrderedGroupAndOptionDtos(questionIds, timestamp);
         } else {
             Set<Long> missingQuestionIds = new HashSet<>();
             Map<Long, GroupAndOptionDtos> result = new HashMap<>();
 
             for (var questionId : questionIds) {
+                String key = questionId + ":" + timestamp;
                 try {
-                    GroupAndOptionDtos cachedDto = questionIdToGroupAndOptionsCache.get(questionId);
+                    GroupAndOptionDtos cachedDto = questionVersionKeyToGroupAndOptionsCache.get(key);
```
yufengwng, line 69: The fix is to cache using a composite key (made from `questionId` and a `timestamp` which indirectly represents the version) rather than solely relying on `questionId`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionCachedDao.java
```diff
@@ -56,33 +57,37 @@ public TemplateDao getTemplateDao() {
...
     public Map<Long, GroupAndOptionDtos> findOrderedGroupAndOptionDtos(Iterable<Long> questionIds, long timestamp) {
-        if (isNullCache(questionIdToGroupAndOptionsCache)) {
+        if (isNullCache(questionVersionKeyToGroupAndOptionsCache)) {
             return delegate.findOrderedGroupAndOptionDtos(questionIds, timestamp);
         } else {
             Set<Long> missingQuestionIds = new HashSet<>();
             Map<Long, GroupAndOptionDtos> result = new HashMap<>();
 
             for (var questionId : questionIds) {
+                String key = questionId + ":" + timestamp;
```
yufengwng, line 67: Key is made by joining with a `:` (colon) separator. Other `*CachedDao` classes uses colon so we're just following convention here.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -410,15 +412,19 @@ default PicklistQuestion getPicklistQuestion(PicklistQuestionDto dto,
...
                 .orElseThrow(() -> new DaoException(String.format(
-                        "Could not find activity instance with guid=%s while getting picklist question with id=%d and stableId=%s",
-                        activityInstanceGuid, dto.getId(), dto.getStableId())));
+                        "Could not find activity version for instance with guid=%s using"
+                                + " activityId=%d and createdAt=%d while getting picklist question %s",
+                        activityInstanceGuid, instanceDto.getActivityId(), instanceDto.getCreatedAtMillis(), dto.getStableId())));
 
         PicklistQuestionDao.GroupAndOptionDtos container = getPicklistQuestionDao()
-                .findOrderedGroupAndOptionDtos(dto.getId(), timestamp);
+                .findOrderedGroupAndOptionDtos(dto.getId(), versionDto.getRevStart());
```
yufengwng, line 427: This method is used when calling the `Get Activity Instance` API, which is a separate codepath than loading activity definitions. We ensure we use the _activity definition timestamp_ rather than the instances' created_at timestamp, to avoid an explosion of cached items (since instance created_at timestamp is more-or-less unique!).

pepper-apis/src/main/java/org/broadinstitute/ddp/event/HousekeepingTaskReceiver.java
```diff
@@ -85,6 +89,13 @@ private void handleClearCache(PubsubMessage message, AckReplyConsumer reply) {
         reply.ack();
     }
 
+    private void handleClearRedis(PubsubMessage message, AckReplyConsumer reply) {
+        LOG.info("Clearing redis caches used in Housekeeping...");
+        CacheService.getInstance().resetAllCaches();
```
yufengwng, line 94: I think it will be nice to have this feature to "clear" Redis caches. HOWEVER, note that this does NOT clear ALL Redis caches. If you look through the code, you'll see that this `resetAllCaches()` method will only clear caches that has been _initialized_. So calling this method here will only clear caches that has been used by Housekeeping. Hopefully the log message of `... used in Housekeeping...` makes this explicit.

Also note that this `clear_redis` message is different from the `clear_cache` message. The former is clearing stuff in Redis, while the latter is clearing stuff stored _in memory_.

Here's a screenshot of the caches that this will clear in Housekeeping after a Brain Study export (there might be more if you do `clear_redis` after Housekeeping had used other caching code):

Screen Shot 2021-01-04 at 7 34 47 PM

ssettipalli:
Looks good to me. 
Ina  separate ticket may be we should check other question types too. 
Basically make sure any question with versions is cached correctly (All  versions cached)

ssettipalli (approved):

PR #590, opened 2021-01-12
ATX-215 Submission of DAR Form

YuriyPavlovBI:
I have implemented functionality in accordance to the schema below.
image

It is working now as expected but the following still needs to be clarified:
Are we going to implement any other checks on top of existing ones (file size check, captcha)?
How it is going to be deployed/configured? Should we use GCP CI/CD? Vault?
Also, I have no permissions to deploy these two functions to the DEV ENV:
image


cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.logging.Logger;
+
+@SuppressWarnings("unused")
+public class StudyContactFunction implements HttpFunction {
+    private static final Logger logger = Logger.getLogger(StudyContactFunction.class.getName());
+    private static final Gson gson = new GsonBuilder().disableHtmlEscaping().create();
+
+    private static final String CAPTCHA_KEY = "CAPTCHA_KEY";
+    private static final String GCP_PROJECT = "broad-ddp-dev";
+    private static final String BUCKET_NAME = "dar-form";
```
yufengwng: We should put these into a config file. We can probably render a config file and have it be uploaded as part of the CF during deploy time.

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private static final String BUCKET_FOLDER = "files";
+    private static final char[] allowedChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890".toCharArray();
+
+    private final Storage storage;
+
+    public StudyContactFunction() {
+        GoogleCredentials googleCredentials;
+        try {
+            googleCredentials = GoogleCredentials.getApplicationDefault();
```
yufengwng, line 61: Using default credentials should work for now, but for better security we might consider creating a new separate service-account for the CF.

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+                String guid = generateGuid();
+                String filename = req.getAttachment().getName();
+                String metadataPath = String.format("%s/%s.metadata", BUCKET_FOLDER, guid);
+                String filePath = String.format("%s/%s.pdf", BUCKET_FOLDER, guid);
+
+                Map<String, String> metadata = req.getData();
+                metadata.put("meta_filesize", String.valueOf(req.getAttachment().getSize()));
+                metadata.put("meta_filename", filename);
+                BlobInfo dataFile = BlobInfo.newBuilder(BlobId.of(BUCKET_NAME, metadataPath))
+                        .setMetadata(metadata)
```
yufengwng, line 109: This looks like a nice approach! But is there a reason why we're storing an empty file to hold the metadata instead of setting the metadata on the PDF file itself? Does it have something to do with signing URL?

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+            StudyContactRequest req = gson.fromJson(request.getReader(), StudyContactRequest.class);
+            if (!isCaptchaPassed(req.getCaptchaToken())) {
+                response.setStatusCode(401, "Captcha verification failed");
+            } else {
+                String guid = generateGuid();
+                String filename = req.getAttachment().getName();
+                String metadataPath = String.format("%s/%s.metadata", BUCKET_FOLDER, guid);
+                String filePath = String.format("%s/%s.pdf", BUCKET_FOLDER, guid);
+
+                Map<String, String> metadata = req.getData();
```
yufengwng, line 98: How about we check the size of the map and limit it to, say, at most 50 key/value pairs (or however many we're expecting).

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+            logger.severe("Error parsing JSON: " + e.getMessage());
+        }
+    }
+
+    private boolean isCaptchaPassed(String token) throws IOException {
+        logger.info(token);
+        String url = "https://www.google.com/recaptcha/api/siteverify";
+        String params = "secret=" + CAPTCHA_KEY + "&response=" + token;
+
+        HttpURLConnection http = (HttpURLConnection) new URL(url).openConnection();
```
yufengwng: I think you can use Java 11's `HttpClient`. Might be nicer to work with for HTTP calls.

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+        InputStream res = http.getInputStream();
+        JsonObject response = gson.fromJson(new InputStreamReader(res, StandardCharsets.UTF_8), JsonObject.class);
+        res.close();
+        logger.info(response.toString());
+        return response.has("success") && response.get("success").getAsBoolean();
+    }
+
+
+    private String generateGuid() {
+        Random random = new Random();
```
yufengwng: I recommend importing the `nanoid` library instead of rolling our own random generator. It's a very small library. See the main pepper repo for how we use it.
```
        <dependency>
            <groupId>com.aventrix.jnanoid</groupId>
            <artifactId>jnanoid</artifactId>
            <version>2.0.0</version>
        </dependency>
```

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
```diff
@@ -0,0 +1,136 @@
...
+            if (!isCaptchaPassed(req.getCaptchaToken())) {
+                response.setStatusCode(401, "Captcha verification failed");
+            } else {
+                String guid = generateGuid();
+                String filename = req.getAttachment().getName();
+                String metadataPath = String.format("%s/%s.metadata", BUCKET_FOLDER, guid);
+                String filePath = String.format("%s/%s.pdf", BUCKET_FOLDER, guid);
+
+                Map<String, String> metadata = req.getData();
+                metadata.put("meta_filesize", String.valueOf(req.getAttachment().getSize()));
```
yufengwng: We should do a sanity check on the size right? I think at most 2MB?

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/json/StudyContactRequest.java
```diff
@@ -0,0 +1,35 @@
+package org.broadinstitute.ddp.cf.json;
+
+import java.util.Map;
+
+import com.google.gson.annotations.SerializedName;
+
+public class StudyContactRequest {
```
yufengwng, line 7: How about we also require a `clientDomain` and `clientId` as input? We can check those against values from our config file.

cloud-functions/cf-emailer/src/main/java/org/broadinstitute/ddp/cf/AttachmentLoadedFunction.java
```diff
@@ -0,0 +1,141 @@
...
+    private static final String ATTACHMENT_DISPOSITION = "attachment";
+    private static final String ATTACHMENT_PDF_TYPE = "application/pdf";
+    public static final String PATH_MAIL_SEND = "mail/send";
+
+    private static final String API_KEY = "SG_KEY";
+    private static final String FROM_EMAIL = "FROM_EMAIL";
+    private static final String FROM_NAME = "DEV";
+    private static final String TO_EMAIL = "TO_EMAIL";
+    private static final String TEMPLATE_ID = "d-5ec2c969bbad4fffabb8168abafa470c";
+    private static final String GCP_PROJECT = "broad-ddp-dev";
```
yufengwng: Again, these will go into the config file we're using for the study-contact CF.

cloud-functions/cf-emailer/src/main/java/org/broadinstitute/ddp/cf/AttachmentLoadedFunction.java
```diff
@@ -0,0 +1,141 @@
...
+
+        BlobId contentBlobId = BlobId.of(bucketName, filename);
+        Blob contentBlob = storage.get(contentBlobId);
+        BlobId metadataBlobId = BlobId.of(bucketName, pattern.matcher(filename)
+                .replaceAll("$1.metadata"));
+        Blob metadataBlob = storage.get(metadataBlobId);
+        Map<String, String> metadata = metadataBlob.getMetadata();
+        String originalFilename = metadata.get("meta_filename");
+        if (!contentBlob.getSize().equals(Long.valueOf(metadata.get("meta_filesize")))) {
+            throw new RuntimeException("File size mismatch");
```
yufengwng, line 91: When we get these deployed to `dev`, we should make sure that logging errors and throwing exceptions will be picked up in stackdriver and show up in our Error Reporting so a human is notified of these issues.

yufengwng (commented):
Overall looks good!

> Are we going to implement any other checks on top of existing ones

I added some comments about this. We'll also check with the security team once this is more fleshed out.

> How it is going to be deployed/configured?

I'm not familiar with this aspect yet. But if we can figure out how to do it manually, we can probably automate it with CircleCI.

> Also, I have no permissions to deploy these two functions to the DEV ENV.

I'll look into the permissions when I get a chance. Worse case, I can help deploy these to `dev` when time is right.

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: Yes, that's why I asked how it is going to be deployed and configured

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: The point is that the metadata is being overridden during the actual file upload. So I had to do such a trick.

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: I made it similar to pepper config render procedure

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: AFAIK, it is configurable per function:
image

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: Sounds reasonable. Done

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: Moved to HttpClient

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: Done

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
YuriyPavlovBI: Sure. Validation is in place now

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/json/StudyContactRequest.java
YuriyPavlovBI: Requiring now

cloud-functions/cf-emailer/src/main/java/org/broadinstitute/ddp/cf/AttachmentLoadedFunction.java
YuriyPavlovBI: Config made

cloud-functions/cf-emailer/src/main/java/org/broadinstitute/ddp/cf/AttachmentLoadedFunction.java
YuriyPavlovBI: I'll check and let you know

cloud-functions/cf-study-contact/src/main/java/org/broadinstitute/ddp/cf/StudyContactFunction.java
yufengwng: Nice! We can keep the code as-is then.

yufengwng (approved):
LGTM. I've asked someone to grant you the `cloudfunctions.functions.setIamPolicy` permission so you can set CF to allow unauthenticated invocations. In the meantime, I see that you have the `Editor` role in our `dev` GCP project, so you should have all the permissions you need to deploy and work with Cloud Functions. If you're still missing permissions, please ping folks in Slack with explicit list of permissions you need.

PR #592, opened 2021-01-13
Ddp 5442 auth0 log events

evgeniipr:
## Context

Implemented REST endpoint: POST ../pepper/v1/auth0-log-event which persists the **auth0 log events** data passed from auth0 (this URL should be specified during auth0 stream custom webhook creation). Such webhook needs to be created for each auth0 tenant.
Currently the **auth0 custom webhook** is created in auth0 **tenant** 'ddp-dev' which POSTs **log events** to URL https://pepper-dev.datadonationplatform.org/pepper/v1/auth0-log-event?tenant=ddp-dev
Another webhook parameter which may be configured - **authorization token**. It is passed in header **"Authorization"** and checked by ddp backend: if it's incorrect then log event saving failed. 
If config parameter 'privateApisTokens.auth0LogEventBearerToken' is specified and not empty
then it contains a token which should be compared with token in header 'Authorization'
(in a format: 'Bearer ' + 'privateApisTokens.auth0LogEventBearerToken').
If token check is enabled and specified in config then the same token should be specified in Auth0 Custom Webhook settings
(with prefix 'Bearer ').
 
The backend code organized the following way: 
**Auth0LogEventRoute** delegates the logic of parsing log events and persisting them to **Auth0LogEventService** which utilizes **GsonRecursiveReader** to fetch needed data (to be persisted in DB) from incoming POST payload (JSON).
It's added scripts to create new BD tables: **auth0_log_event, auth0_log_event_code** (it will be filled with data later by these codes: https://auth0.com/docs/logs/log-event-type-codes - needs to implement script for it).

NOTE: when deploying to production it needs to create custom Webhook in auth0 for prod tenant (this can be done any time after the deploy is done: so that we have REST endpoing deployed to prod).. 
And if it needs to protect the API with token then it needs:
- define config parameter 'auth0.auth0LogEventApiAuthorizationToken' containing token value and define same value in the auth0 custom webhook settings.

NOTE: it is to be added added a new section into conf templates and needs to insert appropriate values into vault:
```
       "auth0LogEvents" : {
		"token" : "{{$auth0LogEvents.token}}"
	},
```
The possible default value can be an empty string (in such case the token check is ignored)_

## Checklist

## FUD Score

## Testing

## Release

After RELEASE, in order to activate auth0 log events sending it needs to create Custom WebHook (in Auth0 section streams) where set the following parameters:
- URL for POST request;
- Authorization token (optional, only if it's required by Pepper side), in format: 'Bearer ' + token;
- Content format: JSON Object.


pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -324,10 +330,15 @@ private static void start() {
         if (cfg.getBoolean(ConfigFile.RESTRICT_REGISTER_ROUTE)) {
             whitelist(API.REGISTRATION, cfg.getStringList(ConfigFile.AUTH0_IP_WHITE_LIST));
         }
+        if (cfg.getBoolean(ConfigFile.RESTRICT_AUTH0_LOG_EVENT_ROUTE)) {
```
zyme: Is the list of IPs different for the log events hook, or can we use the same ones that are in `ConfigFile.RESTRICT_REGISTER_ROUTE`?  It looks like the allow list is the same: https://auth0.com/docs/security/allowlist-ip-addresses

evgeniipr: Yes, looks the same. Will fix it.

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
```diff
@@ -0,0 +1,80 @@
...
+            <column name="auth0_log_event_id" type="bigint" autoIncrement="true" startWith="1">
+                <constraints primaryKey="true" primaryKeyName="auth0_log_event_pk"/>
+            </column>
+            <column name="tenant" type="varchar(100)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="log_id" type="varchar(200)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="date" type="datetime">
```
yufengwng: Let's have microseconds precision for the `date`.
```suggestion
            <column name="date" type="datetime(6)">
```

pepper-apis/src/test/resources/auth0-log-event-testdata/auth0-log-event-testdata1.json
```diff
@@ -0,0 +1,130 @@
...
+            {
+              "name": "login",
+              "flow": "login",
+              "initiatedAt": 1610220854143,
+              "completedAt": 1610220857888,
+              "timers": {
+                "rules": 701
+              },
+              "user_id": "auth0|XXX_user_id_XXX",
+              "user_name": "test_name@broadinstitute.org",
```
yufengwng: I recommend using `@datadonationplatform.org` for test dummy emails.

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
```diff
@@ -0,0 +1,80 @@
...
+            <column name="type" type="varchar(100)"/>
+        </createIndex>
+        <createIndex tableName="auth0_log_event" indexName="auth0_log_event_email_idx">
+            <column name="email" type="varchar(200)"/>
+        </createIndex>
+
+    </changeSet>
+
+    <changeSet author="evgenii.pr" id="20210113-create-auth0-log-event-code-table">
+        <createTable tableName="auth0_log_event_code">
```
yufengwng: Hmmm, I thought we were going to use this type table in some way (e.g. at the very least populating it with types). But it looks like we're not doing anything with it... I wonder if it will be better to just disregard this type table and add it later when we do it need...

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/Auth0LogEventDao.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.broadinstitute.ddp.json.auth0.Auth0LogEventNode.resolveDateTimeValue;
+import static org.broadinstitute.ddp.json.auth0.Auth0LogEventNode.resolveStringValue;
+
+import java.util.Map;
+
+
+import com.google.gson.JsonElement;
+import org.jdbi.v3.sqlobject.CreateSqlObject;
+
+public interface Auth0LogEventDao {
```
yufengwng:
```suggestion
public interface Auth0LogEventDao extends SqlObject {
```

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/Auth0LogEventDao.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @CreateSqlObject
+    JdbiAuth0LogEvent getJdbiAuth0LogEvent();
+
+    default long insertAuth0LogEvent(final Map<String, JsonElement> logEvent, String tenant) {
```
yufengwng: Working with JSON objects in the DAO feels awkward. Can you try representing the `logEvent` as a Java object and pass that in? A plain class with fields that parallel the database columns should do nicely.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/auth0/Auth0LogEventNode.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Reject prefix and separator after it.
+     */
+    private static String rejectPrefix(String str, String prefixSep) {
+        int prefixInd = str.indexOf(prefixSep);
+        return prefixInd != -1 ? str.substring(str.indexOf(prefixSep) + 1) : str;
```
yufengwng: If I'm understanding, this will chop off and throw out the `auth0|` part of the string? That will be the opposite of what we want. We consider auth0 user ids to be the whole thing with the `auth0|` as part of it.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @param logEventsJson payload doc coming with POST request (from Auth0 stream).
+     * @return {@code List<Map<String, JsonElement>>} list of maps with parsed log events data
+     */
+    public List<Map<String, JsonElement>> parseAuth0LogEvents(String logEventsJson) {
+        var rootNode = new Gson().fromJson(logEventsJson, JsonObject.class);
+        var logsNode = rootNode.getAsJsonArray(LOGS.nodeName());
+        final List<Map<String, JsonElement>> logEvents = new ArrayList<>();
+        for (JsonElement node : logsNode) {
+            logEvents.add(GsonRecursiveReader.readValues(node, AUTH0_LOG_EVENT_NODE_NAMES));
```
yufengwng, line 78: I'm a bit concerned that this recursive parsing strategy might pick up false positives. What I mean is that this just parses for the first thing that it finds in JSON, without considering the context of the data. For example, imagine if there is a "change email" log entry that has the old email before the new email. We might want the new email, but this will pickup the old email since that's the first one it encounters while parsing. I'm not too concerned, since we're storing the entire raw `data` anyways, but it's something to consider.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        return logEvents;
+    }
+
+    public void logAuth0LogEvent(final Map<String, JsonElement> logEvent, String tenant) {
+        log.info(
```
yufengwng: I'm concerned that logging these entries will produce too much logging output and clog up stackdriver with noise. Might be better to change this to a "debug" log or something.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
```diff
@@ -0,0 +1,112 @@
...
+                resolveStringValue(IP, logEvent),
+                resolveStringValue(EMAIL, logEvent),
+                resolveStringValue(DATA, logEvent));
+    }
+
+    /**
+     * Save log event and tenant data to DB table 'auth0-log-event'
+     */
+    public void persistAuth0LogEvent(final Map<String, JsonElement> logEvent, String tenant) {
+        TransactionWrapper.useTxn(handle -> {
```
yufengwng: Can we pass in a database `Handle` instead? It'll make things a bit easier to test, and allows for easier composition with other operations in a single transaction.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/GsonRecursiveReader.java
```diff
@@ -0,0 +1,86 @@
...
+     * already detected element).
+     *
+     * @param rootNode root element of JSON document where to find elements
+     * @param names    list of elements' names which to search for
+     * @return {@code Map<String, JsonElement>} map with detected node values
+     */
+    public static Map<String, JsonElement> readValues(final JsonElement rootNode, final List<String> names) {
+        final Map<String, JsonElement> results = new HashMap<>();
+        if (names != null && names.size() > 0) {
+            final List<String> yetNotReadNames = new ArrayList<>(names);
```
yufengwng: Are we using a `List` because we want to parse the names in the given order? Curious why not use a `Set`?

yufengwng (commented):
Still reading, but have a few comments.

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
zyme: Agreed, let's set this to `debug`.

pepper-apis/src/test/resources/auth0-log-event-testdata/auth0-log-event-testdata1.json
evgeniipr: Fixed

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
evgeniipr: Created script which fills code data to table "auth0_log_event_code".

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/Auth0LogEventDao.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/Auth0LogEventDao.java
evgeniipr: Created a POJO class Auth0LogEvent, used it instead of map.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/auth0/Auth0LogEventNode.java
evgeniipr: Removed this method as it is not needed..

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
evgeniipr: I agree, but unfortunately I still don't know all possible structures of 'detail' subnode in case of different event types.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
evgeniipr: Changed to debug

pepper-apis/src/main/java/org/broadinstitute/ddp/service/Auth0LogEventService.java
evgeniipr: Sure. Passed Handle

pepper-apis/src/main/java/org/broadinstitute/ddp/util/GsonRecursiveReader.java
evgeniipr: It just a temp List: intitally contains all names and as soon as certain names found in JSON it removed from the list (so we not going to search it further in the JSON). I changed ArrayList to LinkedList for better performance of removing.

pepper-apis/config/testing-inmemorydb.conf.ctmpl
```diff
@@ -30,6 +30,7 @@
     "dsmJwtSigner": "org.broadinstitute.kdux",
     "dbInstanceId": "ddp-dev",
     "restrictRegisterRoute": false,
+    "restrictAuth0LogEventRoute": false,
```
zyme, line 33: @evgeniipr do we have the procedure for changing this to `true` documented in the release plan?

evgeniipr: @zyme actually not yet. Could you give me some example to write a release plan doc?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
```diff
@@ -0,0 +1,139 @@
...
+    public static final String SYSTEM_PROPERTY_AUTH0_LOG_EVENT_TOKEN_CHECK_DISABLE = "auth0LogEvent.tokenCheck.disable";
+    /** By default token check is not disabled */
+    public static final boolean DEFAULT_AUTH0_LOG_EVENT_TOKEN_CHECK_DISABLE = false;
+
+    /** Mandatory parameter in the log event URL. Specifies name of auth0 tenant */
+    public static final String QUERY_PARAM_TENANT = "tenant";
+
+    private final Config config;
+
+    public Auth0LogEventRoute(final Config config) {
```
zyme: Not sure we need to include `final` declaration in method params...do others have thoughts on this?

zyme: Usually we don't store the entire `Config` object as state in the route, but instead pass in only the values that we need.  Passing in the config object can make it harder to understand what the object needs.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
```diff
@@ -0,0 +1,139 @@
...
+        this.config = config;
+    }
+
+    @Override
+    public Object handle(Request request, Response response) throws Exception {
+        String tenant = readTenant(request);
+        checkAuthorizationToken(request);
```
zyme: Can we isolate the auth check to a `before()` filter?  That's more in keeping with our use of filters.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
```diff
@@ -0,0 +1,139 @@
...
+        checkBody(request);
+
+        final var auth0LogEventService = new Auth0LogEventService();
+
+        final var logEvents = auth0LogEventService.parseAuth0LogEvents(request.body());
```
zyme: Coding standards question: usually we only declare fields `final` (or variables that need to be final for the sake of lambdas).  Should we formalize the non-use of `final`s here?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: It just to prevent from occasional reassigning the argument reference.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
```diff
@@ -0,0 +1,139 @@
...
+                }
+                if (!addBearerPrefixToToken(auth0LogEventApiBearerToken).equals(authorizationToken)) {
+                    haltError(SC_UNAUTHORIZED, INVALID_TOKEN, "Invalid authorization token");
+                }
+            }
+        }
+    }
+
+    private boolean isCheckToken() {
+        return !SystemPropertyUtil.readProperty(SYSTEM_PROPERTY_AUTH0_LOG_EVENT_TOKEN_CHECK_DISABLE,
```
zyme: Usually our system properties are read 100% by typesafe config, using `.withFallback()` as needed for defaults.  Can we remove `SystemPropertyUtil`, and pass in these values in the constructor?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Ok, will pass only the needed params. Actually I tried to prevent the situation when the auth0 event params not defined in config and this is checked inside the Route class. But ok, I will move the check out of it.

pepper-apis/src/main/resources/changelog-master.xml
```diff
@@ -242,6 +242,8 @@
     <include file="db-changes/schema/DDP-5310-create-statistics-types.xml" relativeToChangelogFile="true"/>
     <include file="db-changes/seed/add-languages-de-hi-it-ja-pl-tr.xml" relativeToChangelogFile="true"/>
     <include file="db-changes/schema/DDP-5277-revision-activity-naming-details.xml" relativeToChangelogFile="true"/>
+    <include file="db-changes/schema/DDP-5442-auth0-log-event.xml" relativeToChangelogFile="true"/>
```
zyme: We generally append new liquibase files to the very bottom of the file so that merge conflicts are a little easier to resolve when multiple people are making liquibase changes in different tickets.  It's fine to leave this as-is...just mentioning it for future work.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Yes, sure. Will do that.

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event-code.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+
+    <changeSet author="evgeniipr" id="20210113-create-auth0-log-event-code-table">
+        <createTable tableName="auth0_log_event_code">
```
zyme, line 10: We usually use generated primary key columns and avoid using "natural" keys as primary keys.  Can you add a generated `bigint` id column here, use it for the fk from the event table, and change `code` from a primary key to a unique key?

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
```diff
@@ -0,0 +1,65 @@
...
+            <column name="type" type="varchar(100)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="client_id" type="varchar(200)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="connection_id" type="varchar(200)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="user_id" type="varchar(200)">
```
zyme: `user_id` might be misunderstood to be a reference to `user.user_id`.  Can you change this to something like `auth0_user_id`?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Yes, the finals not necessary here (should I better remove it?), I just trying to follow some of existing code practices: assigning final prevents from occasional changing the variable and make it more safe in a concurrent env. I usually set non-final when I really know that a variable could be changed..
Extra `final` keyword makes code a bit more verbose: unfortunately they didn't added in Java11 `val` (which exist in Scala and Kotlin denoting an 'immutable variable').

pepper-apis/src/main/resources/changelog-master.xml
evgeniipr: I think my line occured in the middle after updating (merging) the FB from develop.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Fixed. Implemented filter Auth0LogEventCheckTokenFilter

pepper-apis/src/main/resources/changelog-master.xml
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: I thought may be to remove this system variable at all and in future define a paramter which could be set/unset via Google Cloud Runtime Configuration API ?

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event-code.xml
evgeniipr: Fixed

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
evgeniipr: Fixed. Renamed to auth0_log_event_user_id. Also renamed couple of other fields.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Removed. The token check is disabled until the parameter not added to config file (now it is not there).

pepper-apis/config/testing-inmemorydb.conf.ctmpl
yufengwng: This is just the local testing file so it'll be fine. The one used in real code is rendered from Vault.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiAuth0LogEvent.java
```diff
@@ -0,0 +1,45 @@
...
+            @Bind("userId") String userId,
+            @Bind("userAgent") String userAgent,
+            @Bind("ip") String ip,
+            @Bind("email") String email,
+            @Bind("data") String data
+    );
+
+    @SqlQuery("select c.* from auth0_log_event_code c where c.code=:code")
+    @RegisterRowMapper(Auth0LogEvent.Auth0LogEventCodeDto.Auth0LogEventCodeDtoMapper.class)
+    Optional<Auth0LogEvent.Auth0LogEventCodeDto> findByCode(@Bind("code") String code);
```
yufengwng: Was a bit confused and thought we were querying logs by "code". Can you rename this? Perhaps something like `findTypeByCode`.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/auth0/Auth0LogEventNode.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public enum Auth0LogEventNode {
+
+    LOGS("logs", JSON_NODE),
+    LOG_ID("log_id", STR),
+    DATE("date", DATETIME),
+    TYPE("type", STR),
+    CLIENT_ID("client_id", STR),
+    CONNECTION_ID("connection_id", STR),
+    USER_ID("user_id", STR),
```
yufengwng, line 38: What happened to the prefix thing? I think the previous approach was on the right track, but with this new approach I'm a bit more worried that it will pick up the wrong `user_id` value. For example, below is a snippet of json taken from what you had posted in Slack (`auth_log3.json`).

We'll want `body.user_id` but I think `GsonRecursiveReader` will pickup `body.identities[0].user_id` instead since it looks like implementation is depth-first. If we had that prefix thing from before, it'll at least ensure that we have the `auth0|` prefix to minimize false positives. Or am I misunderstanding?
```
{
      "log_id": "90020210110175027874000812534194750248972703030491217970",
      "data": {
        ...
        "details": {
          "request": {
          ...
          },
          "response": {
            "statusCode": 200,
            "body": {
              "identities": [
                {
                  "user_id": "5ffb3e4df2243e0069f878b4",
                  "provider": "auth0",
                  "connection": "Username-Password-Authentication",
                  "isSocial": false
                }
              ],
              "user_id": "auth0|5ffb3e4df2243e0069f878b4",
              "user_metadata": {
                "language": "en"
              },
              ...
            }
          }
        }
      }
  },
```

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event-code.xml
```diff
@@ -0,0 +1,137 @@
...
+            <column name="auth0_log_event_code_id" type="bigint" autoIncrement="true" startWith="1">
+                <constraints primaryKey="true" primaryKeyName="auth0_log_event_code_pk"/>
+            </column>
+            <column name="code" type="varchar(100)" >
+                <constraints nullable="false"/>
```
yufengwng: You can add attributes `unique="true" uniqueConstraintName="..."` to the `constraints` tag. That should be sufficient and we won't need the `createIndex` below.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/Auth0LogEventServiceTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    public void testAuth0LogEventsParse() throws FileNotFoundException {
+        final var logEvents = auth0LogEventService.parseAuth0LogEvents(readJSONFromFile(AUTH0_LOG_EVENT_TESTDATA1).toString());
+
+        assertEquals(2, logEvents.size());
+
+        final var logEvent = logEvents.get(0);
+
+        assertEquals("s", logEvent.get(TYPE.nodeName()).getAsString());
+        assertEquals("auth0|XXX_user_id_XXX", logEvent.get(USER_ID.nodeName()).getAsString());
```
yufengwng, line 35: Will be interesting to see what `user_id` is for `logEvents.get(1)`.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/GsonRecursiveReader.java
yufengwng: I asked because if order doesn't matter then I would've suggested using `Set`. But this is a small enough thing compared to other things going on in this PR, so let's close this topic for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
yufengwng: I don't feel this is necessary. I'll vote to leave out `final` in method params to be consistent with other code.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Removed `final `keyword in all places.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/Auth0LogEventRoute.java
evgeniipr: Removed `final `keyword in all places

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiAuth0LogEvent.java
evgeniipr: Renamed the method.

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event-code.xml
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/json/auth0/Auth0LogEventNode.java
evgeniipr: Yes, in this doc it will get "user_id" with prefix "auth0|..".
May be it's ok. I checked other possible log records. For example this one:

``` 
{
      "log_id": "90020210111084426740006556785942888001718262650720223298",
      "data": {
        "date": "2021-01-11T08:44:24.064Z",
        "type": "sertft",
        "description": "Successful Refresh Token exchange",
        "connection": "Username-Password-Authentication",
        "connection_id": "con_7q6tTpAKRg1fvS5j",
        "client_id": "64YEu4K0XkuCPft4dx9EgdIm9ChxPe2C",
        "client_name": "evgenii-local",
        "ip": "83.102.216.27",
        "user_agent": "Apache-HttpClient/4.5.1 (Java/11.0.9)",
        "details": {
          "tokenCounter": 1,
          "familyId": "nv-EsDeqQhM89i4acu9xiQ"
        },
        "user_id": "auth0|5ffc0fd73ca027006f4fd97a",
        "user_name": "pavel_20210111_10001@broadinstitute.org",
        "audience": "https://ddp-dev.auth0.com/userinfo",
        "scope": "openid profile",
        "log_id": "90020210111084426740006556785942888001718262650720223298"
      }
    }
```
Here it's only one occurence of "user_id" and it is with prefix.
Seems it always gets "user_id" with prefix and as you said earlier that better to include it in the user_id then user_id data will be consistent. Of course it will be consistent if we always reject prefix (as in 1st implementation). Seems both are ok.

pepper-apis/src/test/java/org/broadinstitute/ddp/service/Auth0LogEventServiceTest.java
evgeniipr: I added the check of logEvents.get(1) and found some problem in GsonRecursiveReader. Fixed this now.

evgeniipr: The problem was that the reader when going through elements of a node process each node recursively immediately as it encountered. I fixed it by splitting into 2 for-circles: 1st one reads values of each element (without recursive processing), 2nd - process each element recursively, if possible. Now it reads the nodes as expected.
Thanks a lot for suggestion to check this!

pepper-apis/src/main/resources/db-changes/schema/DDP-5442-auth0-log-event.xml
yufengwng: In my opinion, this name is a little too verbose... I'm in favor of Zim's suggestion of `auth0_user_id`. It's understandable and consistent with how this piece of data is named in other tables.

evgeniipr: Renamed it to auth0_user_id.

yufengwng (dismissed):
LGTM

yufengwng (approved):

PR #599, opened 2021-01-19
DDP-5231 delete user endpoint

YuriyPavlovBI:
## Context

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/endpoints/user.yml
```diff
@@ -0,0 +1,24 @@
...
+    Remove user.
+
+    Some restrictions apply:
+    _Only the proxy user is allowed to delete one of their managed user._
+    _The user to be deleted cannot have an Auth0 account._
+    _The user cannot yet have reached the ENROLLED status._
+  tags:
+    - Operator & Participant
+  responses:
+    204:
```
yufengwng, line 23: Looks like response can also be `422`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/DsmKitRequestDao.java
```diff
@@ -205,4 +205,8 @@ long createKitRequest(@Bind("guid") String requestGuid, @Bind("studyGuid") Strin
     @SqlUpdate("deleteKitRequestById")
     @UseStringTemplateSqlLocator
     void deleteKitRequest(Long id);
+
+    @SqlUpdate("deleteKitRequestByParticipantId")
+    @UseStringTemplateSqlLocator
+    void deleteKitRequestsByParticipantId(@Bind("participantId") Long participantId);
```
yufengwng: I'm concerned that we're deleting a user that has kit requests -- this will mess up DSM. Can we another check to ensure user doesn't have kit requests before allowing delete?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/DeleteUserRoute.java
```diff
@@ -0,0 +1,103 @@
...
+        this.userService = userService;
+    }
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String userGuid = request.params(USER_GUID);
+        DDPAuth ddpAuth = RouteUtil.getDDPAuth(request);
+        String operatorGuid = ddpAuth.getOperator();
+
+        LOG.info("Trying to delete user with GUID: {}", userGuid);
```
yufengwng: Can we also log who the operator is?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,76 @@
...
+        queueEvent.deleteQueuedNotificationSubstitutionsByUserIds(Collections.singleton(userId));
+        queueEvent.deleteQueuedNotificationsByUserIds(Collections.singleton(userId));
+        queueEvent.deleteQueuedEventsByUserIds(Collections.singleton(userId));
+        // Removing governances
+        UserGovernanceSql userGovernanceSql = handle.attach(UserGovernanceSql.class);
+        userGovernanceSql.deleteAllGovernancesByOperatorUserId(userId);
+        userGovernanceSql.deleteAllGovernancesByParticipantUserId(userId);
+        handle.attach(StudyGovernanceSql.class).deleteAgeUpCandidateByParticipantId(userId);
+        // Removing user itself
+        handle.attach(DataExportDao.class).deleteDataSyncRequestsForUser(userId);
```
yufengwng, line 95: Can you investigate deleting the user's documents in Elasticsearch as well? DSM depends on ES for user data so if we don't delete it from ES, it will not be accurate. Some thoughts:

* Should delete from the `participants` and `participants_structured` indices. Documents are identified by user guid. We should call ES's [Delete API](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-delete.html) for this.
* Should also delete from `users` index. Document is also identified by user guid. HOWEVER, document has relationship between participant and proxy, and each proxy also have their own documents in here. What this means is that once the governed participant is deleted, we should update the proxy's document so the relationship is updated. We might be able to do it by queuing up `queueDataSync(proxy, study)`.
* Should lookup which study this is and use the correct indices for that study.
* You can see the structure of the documents in directory `ddp-study-server/elasticsearch/templates`.
* You can get an ES client using `ElasticsearchServiceUtil`.

Looks like it might be a decent amount of effort. If so, might be good to split into a separate ticket/PR.

pepper-apis/docs/specification/src/endpoints/user.yml
YuriyPavlovBI: Documented 422 as well as 404

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/DsmKitRequestDao.java
YuriyPavlovBI: Sure. Added

pepper-apis/src/main/java/org/broadinstitute/ddp/route/DeleteUserRoute.java
YuriyPavlovBI: Logged

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: It turned out to be not a big piece of code so I kept the Elasticsearch changes in this PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,129 @@
...
+                .collect(Collectors.toSet());
+        userGovernanceDao.deleteAllGovernancesForProxy(userId);
+        userGovernanceDao.deleteAllGovernancesForParticipant(userId);
+        handle.attach(StudyGovernanceSql.class).deleteAgeUpCandidateByParticipantId(userId);
+        // Removing user itself
+        handle.attach(DataExportDao.class).deleteDataSyncRequestsForUser(userId);
+        handle.attach(JdbiUser.class).deleteAllByIds(Collections.singleton(userId));
+
+        // Cleaning up the elasticsearch data
+        RestHighLevelClient esClient = ElasticsearchServiceUtil.getElasticsearchClient(ConfigManager.getInstance().getConfig());
```
yufengwng: Can we pass in this dependency instead of instantiating it here?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,129 @@
...
+        BulkRequest bulkRequest = new BulkRequest().timeout("2m");
+        for (String studyGuid : studyGuids) {
+            StudyDto studyDto = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
+
+            String indexParticipant = ElasticsearchServiceUtil.getIndexForStudy(handle, studyDto,
+                    ElasticSearchIndexType.PARTICIPANTS);
+            String indexParticipantStructured = ElasticsearchServiceUtil.getIndexForStudy(handle, studyDto,
+                    ElasticSearchIndexType.PARTICIPANTS_STRUCTURED);
```
yufengwng: We also need to delete from the `USERS` index for this participant.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,129 @@
...
+
+            bulkRequest.add(new DeleteRequest()
+                    .index(indexParticipantStructured)
+                    .type(REQUEST_TYPE)
+                    .id(userGuid));
+        }
+
+        BulkResponse bulkResponse = esClient.bulk(bulkRequest, RequestOptions.DEFAULT);
+
+        if (bulkResponse.hasFailures()) {
```
yufengwng: What happens if we get a `404` as part of the bulk request? For example, a study may not have a `PARTICIPANTS` index and so the document is not found. Is that considered a failure? How do we handle that situation?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,129 @@
...
+            LOG.error(bulkResponse.buildFailureMessage());
+        }
+        userGovernances.stream().map(Governance::getProxyUserGuid)
+                .forEach(proxyGuid -> handle.attach(DataExportDao.class).queueDataSync(proxyGuid));
```
yufengwng: I'm not sure this will work. I think the issue is that the proxy is not actually "enrolled" in the study, e.g. they don't have a row in the `user_study_enrollment` table, so `DataSyncJob` might skip over them. I think we need to do `queueDataSync(proxyGuid, studyGuid)` in order for the `DataSyncJob` to pick up and process it properly.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Entry-point for back-end operations related to users.
+ */
+public class UserService {
```
yufengwng, line 42: I didn't see any tests for the new code. Did we forget to commit some files?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
yufengwng: I'm okay with having the ES changes in this file. Now that we're using ES in places other than data export, we should consolidate those operations into one place, but that's a separate topic for another day.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: I'm not sure I get it. Static method `getElasticsearchClient` tries to get a client for given config ES credentials from "cache" if already instantiated, otherwise creates and caches it for the future calls. Did you mean something else?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: Added

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: In case of index absence, it will be a failure response. I assume we can't do anything in this case so I just logged this event.
In case of document absence it won't be a failure.
image

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: You're right, I overlooked that.
Fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: The test is on its way.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
yufengwng: How about we pass in a ES client in the constructor for `UserService`? We'll call `getElasticsearchClient()` in `DataDonationPlatform` where UserService is instantiated. I believe this will make testing a bit easier, so we can pass in a mock to UserService.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
yufengwng: Okay, in this case I think it's more common to see document absence, so I think this is fine for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: Since we have to check new route limitations, that would make impossible to bypass ES for tests. I didn't find better solution to have both DeleteUserRouteTest and DataExporterTest working with the same config. Any thoughts?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
YuriyPavlovBI: Created

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserService.java
yufengwng: I will still urge us to consider injecting ES client as dependency. For testing, there are a couple things we can do. One thing to consider is to restructure your code to be more test-able. For example, perhaps encapsulate the route limitation logic into a method that's exposed for testing. Then we can instantiate a "route" object in the test and call that method directly for assertions. Like so:

```
public class MyRoute {
  MyError checkLimits(..) {
    ...
  }
  Object handle(..) {
    var err = checkLimits(..);
    if (err != null) {
      // halt!
    }
    ...
  }
}

public class MyRouteTest {
  void test() {
    var route = new MyRoute(mockServiceObject);
    ... route.checkLimits(mockDatabaseHandle);
  }
}
```

`GetDsmParticipantStatusRoute`, `GetStudyPasswordPolicyRoute`, and `InvitationCheckStatusRoute` are some examples that try to do this. See if that helps.

YuriyPavlovBI: That's done

yufengwng (approved):
LGTM

PR #601, opened 2021-01-19
circleci: deploy dispatch file from same branch

yufengwng:
## Context

Issue: CircleCI script is hardcoded to use the `develop` branch of `ddp-angular` repo for deploying dispatch rules. When we deploy to test/staging/production environments, it means we will pick up changes in `develop` which is not good. If we added new service in `develop` dispatch, it will cause failures when deploying to other environments.

Fix: When deploying to test/staging/production, we ensure we use the same branch in the `ddp-angular` repo so dispatch rules will be synced up and we won't have cross-pollution from `develop`.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -358,6 +358,9 @@ commands:
       config_branch:
         type: string
         default: *github_dispatch_config_branch
+      use_current_branch_name:
+        type: boolean
+        default: false
     steps:
       - run:
           name: Get dispatch.yaml file
@@ -369,6 +372,9 @@ commands:
             FILE_PATH="<< parameters.config_file_path >>/${ENVIRONMENT}/dispatch.yaml"
             FILE_URL="${API_BASE_URL}${API_BASE_PATH}${FILE_PATH}"
             BRANCH="<< parameters.config_branch >>"
+            if [[ '<< parameters.use_current_branch_name >>' == 'true' ]]; then
+              BRANCH="${CIRCLE_BRANCH}"
+            fi
             FULL_URL="${FILE_URL}?ref=${BRANCH}"
             curl \
               -H 'Accept:application/vnd.github.v3.raw' \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -859,7 +865,8 @@ jobs:
       - deploy-jar:
           jar_base_name: DataDonationPlatform
           deploy_config_file_dir: ../output-config
-      - update-dispatch-rules
+      - update-dispatch-rules:
+          use_current_branch_name: true
```
yufengwng, line 869: The `deploy-stored-jar-job` is used for test/staging/production deploys, so we use new parameter here.

.circleci/config.yml
```diff
@@ -358,6 +358,9 @@ commands:
       config_branch:
         type: string
         default: *github_dispatch_config_branch
+      use_current_branch_name:
```
yufengwng, line 361: Technically, we're not using the same branch, since we're pulling file from ddp-angular repo. Hence, the parameter is `use_current_branch_name` instead of `use_current_branch`.

MocanaAtBroad: Hmm. I am not sure about this, Yufeng. Note that in the config repo the dispatch.yml files are already organized in separate folder by environment (dev/test/etc.). So going by branch AND folder seems to me adds confusion.
Will have to remember to create branch of same name in ddp-angular repo when all you are changing is something in pepper-backend .
The current setup (i.e., housing deployment configuration in ddp-angular/develop) is less than ideal; in Circle script we just treat as a place to store files. Perhaps best solution would be to create separate repo for configuration? Was discussed at time but seemed like too much for too little.
I am agreeing that current setup is not great, but not sure about this solution. Maybe a Slack discussion is in order?

yufengwng: I welcome other solutions here. The trade-off is a bit more confusion, but it brings more safety to deployments. The cross-pollution from `develop` is undesirable. Note that this setup ONLY for when deploying branches named `test`, `staging`, and `production`. For those three branches, the branch in both `ddp-study-server` and `ddp-angular` should already be "in sync", if not we probably have other issues.

PR #620, opened 2021-01-28
[DDP-3709] part 2: file questions and answers

yufengwng:
## Context

Part 2 of DDP-3709. This PR contains:

* New file question and answer models.
* Database support for file question/answers.
* Updated `PATCH answer` API to support file answers.

Credits to @MocanaAtBroad for the initial API specs.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -404,7 +404,9 @@ DateValue:
 FileInfo:
   type: object
   nullable: true
-  description: Description of the file uploaded, can be `null` if upload has been cleared.
+  description: |
+    Description of the file uploaded. This can be `null` if no uploaded file
+    is assigned or associated with the answer.
   required:
     - fileName
     - fileSize
@@ -473,7 +475,10 @@ AnswerSubmission.File:
         value:
           type: string
           nullable: true
-          description: File upload guid provided by server when upload URL was requested. A `null` value clears the upload.
+          description: |
+            File upload guid provided by server when upload URL was requested.
+            A `null` value will unselect and unassign the uploaded file from
+            this answer if a file upload was previously assigned.
 AnswerSubmission.Numeric.Integer:
   allOf:
     - $ref: '../pepper.yml#/components/schemas/AnswerSubmission'
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -324,6 +324,15 @@ Answer.Boolean:
       properties:
         value:
           type: boolean
+Answer.File:
+  allOf:
+    - $ref: '../pepper.yml#/components/schemas/Answer'
+    - type: object
+      required:
+        - value
+      properties:
+        value:
+          $ref: '../pepper.yml#/components/schemas/FileInfo'
 Answer.Numeric:
   allOf:
     - $ref: '../pepper.yml#/components/schemas/Answer'
@@ -392,6 +401,20 @@ DateValue:
       type: integer
       nullable: true
       minimum: 0
+FileInfo:
+  type: object
+  nullable: true
+  description: Description of the file uploaded, can be `null` if upload has been cleared.
+  required:
+    - fileName
+    - fileSize
+  properties:
+    fileName:
+      type: string
+      description: Name of the file
+    fileSize:
+      type: integer
+      description: Size of the file in bytes
 AnswerSubmission:
   type: object
   discriminator:
@@ -440,6 +463,17 @@ AnswerSubmission.Date:
       properties:
         value:
           $ref: '../pepper.yml#/components/schemas/DateValue'
+AnswerSubmission.File:
+  allOf:
+    - $ref: '../pepper.yml#/components/schemas/AnswerSubmission'
+    - type: object
+      required:
+        - value
+      properties:
+        value:
+          type: string
+          nullable: true
+          description: File upload guid provided by server when upload URL was requested. A `null` value clears the upload.
 AnswerSubmission.Numeric.Integer:
   allOf:
     - $ref: '../pepper.yml#/components/schemas/AnswerSubmission'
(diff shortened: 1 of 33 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -324,6 +324,15 @@ Answer.Boolean:
       properties:
         value:
           type: boolean
+Answer.File:
```
yufengwng, line 327: @bskinner @YuriyPavlovBI Here's the API changes. Screenshots to help visualize things:

File question:
Screen Shot 2021-01-28 at 5 27 01 PM

File answer:
Screen Shot 2021-01-28 at 5 27 38 PM

PATCH-ing answer:
Screen Shot 2021-01-28 at 5 28 11 PM

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -196,6 +208,12 @@ private void updateAnswerValue(long operatorId, long answerId, Answer newAnswer,
         }
     }
 
+    private void updateAnswerFileValue(long answerId, FileAnswer newAnswer) {
+        FileInfo info = newAnswer.getValue();
+        Long uploadId = info == null ? null : info.getUploadId();
+        DBUtils.checkUpdate(1, getAnswerSql().updateFileValue(answerId, uploadId));
```
yufengwng, line 214: It seemed more straightforward to just overwrite the `file_upload_id` instead of marking it `is_replaced`. The old file will effectively become "un-assigned". Un-assigned files will be cleaned up by a background job, which I'll implement in upcoming PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileUpload.java
```diff
@@ -125,9 +125,9 @@ public Instant getCreatedAt() {
     }
 
     /**
-     * Whether this file upload has been replaced by another upload. Replaced uploads should be removed.
+     * Returns timestamp of when file was finished uploading, null if not yet uploaded.
```
yufengwng, line 128: Time when file was uploaded seems like a very useful piece of data to have. Adding it here so we don't have to lookup the bucket file whenever we need to know this.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -493,6 +505,25 @@ private DateAnswer convertDateAnswer(String stableId, String guid, String actIns
         }
     }
 
+    private FileAnswer convertFileAnswer(Handle handle, Response response, String stableId, String guid,
+                                         String instanceGuid, JsonElement value) {
+        boolean isNull = (value == null || value.isJsonNull());
+        if (isNull || (value.isJsonPrimitive() && value.getAsJsonPrimitive().isString())) {
```
yufengwng, line 511: Since we allow PATCH-ing `null`, there's a few implications which we didn't touch on in the API discussions:

* The first PATCH call to a file question can pass in `null`, which will result in a new FileAnswer being created with a `null` as `value`.
* Client can get back a FileAnswer object with a valid `answerGuid` but `null` for `value`. I think this behavior is okay. It more-or-less means the user has answered the question before but has not "cleared" their answer value. Client can use the `answerGuid` to PATCH again with a new file upload.

```
// Something like this:
"question": {
   ...
   answers: [
    { "answerGuid": "XYZ", "value": null, ... }
  ]
}
```

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -618,6 +649,30 @@ private void checkAnswerConstraints(Response response, Answer answer) {
         }
     }
 
+    private void checkAndPrepareFile(Handle handle, Response response, ActivityInstanceDto instanceDto, FileAnswer answer) {
+        long participantId = instanceDto.getParticipantId();
+        long uploadId = answer.getValue().getUploadId();
+        var upload = handle.attach(FileUploadDao.class).findById(uploadId)
+                .orElseThrow(() -> new DDPException("Could not find file upload with id " + uploadId));
+        var checkResult = fileService.checkAndSetUploadStatus(handle, participantId, upload);
```
yufengwng, line 657: Currently, we only check the file and set `UPLOADED` status if user actual attempts to PATCH with that file upload. If user (for whatever reason) requested a file upload, went ahead to finish the file upload, but doesn't PATCH it, then server won't ever look at that file upload. This means that file upload will continue to stay in `AUTHORIZED` status, and when we implement background job, the job will clean out and delete that file. I think this is fine because if user uploaded file but doesn't "assign" it to an answer, then it's effectively not "used", so cleaning it out should be fine.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -117,6 +133,69 @@ URL generateSignedUrl(String blobName, String mimeType, HttpMethod method) {
...
+     * reported file size, and belong to the participant. Files already marked uploaded will not be checked again.
+     *
+     * @param handle            the database handle
+     * @param participantUserId the participant who this upload will be assigned to
+     * @param upload            the file upload
+     * @return check result
+     */
+    public CheckResult checkAndSetUploadStatus(Handle handle, long participantUserId, FileUpload upload) {
+        if (participantUserId != upload.getParticipantUserId()) {
+            return CheckResult.OWNER_MISMATCH;
```
yufengwng, line 147: There's a few scenarios here, which we didn't discuss in API discussion, so I made some decisions here. One obvious one is that the file should "belong" to the participant. Another scenario:

* What should we do if the file is already "assigned"/"used" by another answer? Can we assign a file to multiple answers?
    * In this implementation, we're allowing it since it makes things easier and doesn't seem like an error.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Can you expand on what 'cleared' means in this context in the documentation?

yufengwng: Updated the documentation to clarify this. I think it means that whatever file was "associated" with the answer will be unassigned/unselected. I welcome suggestions on what vocabulary to use here!

MocanaAtBroad (commented):
Nice!

PR #621, opened 2021-02-02
DDP-5235 - implement 1st version of API for listening DSM tasks 

evgeniipr:
## Context

Implemented 1st version of API for listening PubSub tasks (subscribing to a specifiedsubscription), executing needed actions on server side and publishing response with rewult to a specified topic.
It is implemented API for listening different types of **PubSub Tasks** (listened from pubsub topic "**dsm-to-dss-tasks**"), doing appropriate actions, and sending response with result by publishing them to pubsub topic "**dss-to-dsm-results**".

API package is **org.broadinstitute.ddp.event.pubsubtask.api**. 
The idea of this API (and it's implementation) to gain **loose coupling** of code layers: **API layer** should be isolated and not depended upon **implementation layer**. And all new processors (for new taskType's will be added to impl layer.

Basic API classes:
**PubSubTaskPubSubConnectionCreator** - creates pubsub Publisher, Subscriber and classes for handing PubSub Tasks and sending responses: **PubSubTaskReceiver, PubSubTaskResultSender**;
**PubSubTaskProcessor** - functional interface defining PubSub Task processor (doing some actions for certain **taskType**);
**PubSubTaskProcessorFactory** - interface defining a factory where to register processors for certain **taskType**-s;
**PubSubTaskData** - holds data extracted from PubSub task message (retrieved from topic **"dsm-to-dss-tasks"**;
**PubSubTaskResultData** - holds data to be sent to topic **"dss-to-dsm-results"**.

Concrete implementations (in package **org.broadinstitute.ddp.event.pubsubtask.impl**):
**UpdateProfileProcessor** - implementation of PubSubTaskProcessor for **taskType=UPDATE_PROFILE**;
**PubSubTaskProcessorFactoryImpl** - implementation of PubSubTaskProcessorFactory where registered the processor.

Root class is **PubSubTaskConnectionService**.  It creates **PubSubTask Subscriber** (reading messages from subscription **"dsm-to-dss-tasks-sub"**) and **PubSubTask result Publisher** (publishing results to topic **"dss-to-dsm-results"**).
Also it creates **PubSubTaskReceiver** which receives messages with **PubSub tasks**, parses it and
run corresponding **PubSubTaskProcessor** which does needed actions according to
a task '**taskType**'. After doing an action it is called **PubSubTaskResultSender**
which sends result to registered **Publisher** (to topic "**dss-to-dsm-results**").

## Checklist

**Help wanted: it needs some solution about the suggested architecture of DSM to DSS pubsub communication API.**

## FUD Score

## Testing

## Release

**Release notes:**
- it needs to manually create topics "dsm-to-dss-tasks" (and subscription "dsm-to-dss-tasks-sub") and "dss-to-dsm-results";
- it needs to add to Vault the values:
        "pubSubTasksSubscription": "dsm-to-dss-tasks-sub",
        "pubSubTasksResultTopic": "dsm-to-dss-results",
        "pubSubTaskSubscriberAwaitRunningTimeout": 30


evgeniipr:
@zyme , @yufengwng I have yet couple of questions (besides the overall architecture):
1) **UpdateUserEmailRoute.java** has very similar functionality as I did for email saving: therefore I copy-pasted needed code and modified it. But definitely it needs to move out common code into a separate class and reuse it in both places. Should it be done in a separate ticket (I can create one if needs - to do afterwards)?

2) The same as above regarding **PatchProfileRoute.java**: code taken from it for making update first and last name (should be extracted from separate class and reused in both places - also needs a ticket?).

zyme:
@evgeniipr I suggest adding the refactor into this PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskData.java
```diff
@@ -0,0 +1,64 @@
+package org.broadinstitute.ddp.event.dsmtask.api;
+
+/**
+ * Data which fetched from DsmTask pubsub message
+ * published by DSM to pubsub topic "dsm-to-dss-tasks"
+ */
+public class DsmTaskData {
```
zyme: Can we remove `dsm` and replace with `PubSub`?  Technically anything can post stuff to pubsub, so the source might not necessarily be DSM.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
```diff
@@ -0,0 +1,19 @@
...
+    public DsmTaskResultData processDsmTask(DsmTaskData dsmTaskData) {
+        try {
+            doIt(dsmTaskData);
+        } catch (Exception e) {
+            return new DsmTaskResultData(ERROR, e.getMessage(), dsmTaskData);
+        }
+        return new DsmTaskResultData(SUCCESS, dsmTaskData);
+    }
+
+    protected abstract void doIt(DsmTaskData dsmTaskData);
```
zyme: How about `handleTask()` instead of `doIt()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskPubSubConnectionCreator.java
```diff
@@ -0,0 +1,110 @@
...
+ * runs corresponding {@link DsmTaskProcessor} which does needed actions according to
+ * DSM task 'taskType' (each processor implemented to a specific 'taskType').
+ * After doing an action it is called the {@link DsmTaskResultSender}
+ * which sends result to topic "dss-to-dsm-results".
+ */
+public class DsmTaskPubSubConnectionCreator {
+
+    private static final Logger LOG = getLogger(DsmTaskPubSubConnectionCreator.class);
+
+    private static final long SUBSCRIBER_AWAIT_RUNNING_TIMEOUT = 30L;
```
zyme: Please extract to a config file value.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskPubSubConnectionCreator.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private void createDsmTaskPubSubTopicSubscriber() {
+        dsmTaskSubscriber = pubSubConnectionManager.subscribeBuilder(projectSubscriptionName, dsmTaskReceiver)
+                .setParallelPullCount(1)
+                .setExecutorProvider(InstantiatingExecutorProvider.newBuilder()
+                        .setExecutorThreadCount(1)
+                        .build())
+                .build();
+        try {
+            dsmTaskSubscriber.startAsync().awaitRunning(SUBSCRIBER_AWAIT_RUNNING_TIMEOUT, TimeUnit.SECONDS);
```
zyme: Please add a handler that responds to subscription errors by re-subscribing, [as outlined here](https://cloud.google.com/pubsub/docs/samples/pubsub-subscriber-error-listener).

@pegahtah @SimoneMaiwald this should apply to DSM's subscription for TestBoston as well...

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateFirstLastNameHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final Logger LOG = getLogger(UpdateFirstLastNameHandler.class);
+
+    public void doIt(String userGuid, String firstName, String lastName) {
+        TransactionWrapper.useTxn(DB.APIS, handle -> updateFirstLastName(handle, userGuid, firstName, lastName));
+    }
+
+    private void updateFirstLastName(Handle handle, String userGuid, String firstName, String lastName) {
+        var profileDao = handle.attach(UserProfileDao.class);
+        UserProfile profile = profileDao.findProfileByUserGuid(userGuid).orElse(null);
```
zyme: I'm not sure we want to make a call out to auth0 to see if things need to be changed.  We can speed things up for the majority of cases (where values are changing) by skipping this step and just always issuing an update, even if the values are the same.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskReceiver.java
```diff
@@ -0,0 +1,111 @@
...
+        this.dsmTaskProcessorFactory = dsmTaskProcessorFactory;
+        this.dsmTaskResultSender = dsmTaskResultSender;
+    }
+
+    @Override
+    public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
+        DsmTaskData dsmTaskData = parseMessage(message, consumer);
+        if (dsmTaskData != null) {
+            sendResponse(processDsmTask(dsmTaskData));
+            consumer.ack();
```
zyme: Probably safer to `ack()` this between `processDsmTask()` and `sendResponse()`.  Otherwise, an error sending a complete message will trigger a stream of retries, which I don't think we want.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.broadinstitute.ddp.event.dsmtask.api.DsmTaskResultData.DsmTaskResultType.SUCCESS;
+
+public abstract class DsmTaskProcessorAbstract implements DsmTaskProcessor {
+
+    @Override
+    public DsmTaskResultData processDsmTask(DsmTaskData dsmTaskData) {
+        try {
+            doIt(dsmTaskData);
+        } catch (Exception e) {
+            return new DsmTaskResultData(ERROR, e.getMessage(), dsmTaskData);
```
zyme: Should an error here result in `ack()` or `nack()`?  Or is it up to the implementing class?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public class UpdateProfileProcessor extends DsmTaskProcessorAbstract {
+
+    public static final String TASK_TYPE__UPDATE_PROFILE = "UPDATE_PROFILE";
+
+    @Override
+    public void doIt(DsmTaskData dsmTaskData) {
+        UpdateProfileData updateProfileData = (UpdateProfileData)dsmTaskData.getPayloadObject();
+
+        updateEmail(dsmTaskData.getParticipantGuid(), updateProfileData.getEmail());
```
zyme: Should we be wrapping both update calls in a try/catch that determines whether it should `ack()` or `nack()?`  For example, if there is a transient db or auth0 connectivity issue, it would be nice to `nack()` and use an exponential backoff (configured in the pubsub topic) to retry for a while.  Perhaps `doIt()` at the interface level should return some kind of enum with values like `SUCCESS`, `RETRY`, or `FAILURE` so that the dispatcher can apply the appropriate `ack()` or `nack()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskData.java
evgeniipr: You mean to rename the topic or class name?

evgeniipr: Actually, good idea: so we can do more generic API for sending tasks to DSS..
@zyme what do you think to rename topics to:
"dsm-to-dss-task" to "dss-task"
"dss-to-dsm-task" to "dss-task-result" ?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskData.java
evgeniipr: If we do it a more generic API then I suggest to add extra attribute "taskSource" (and in our case source="DSM")

pepper-apis/config/application.conf.ctmpl
```diff
@@ -58,6 +58,8 @@
     "pubsub": {
         "enableHousekeepingTasks": {{$conf.Data.pubsub.enableHousekeepingTasks}},
         "housekeepingTasksSubscription": "{{$conf.Data.pubsub.housekeepingTasksSubscription}}",
+        "dsmTasksSubscription": "{{$conf.Data.pubsub.dsmTasksSubscription}}",
+        "dsmTasksResultTopic": {{$conf.Data.pubsub.dsmTasksResultTopic}}"
```
yufengwng: Typo! Missing an opening double-quotes.

yufengwng: dev server crashed because of bad config. I fixed it and redeployed to dev. Please pull the latest commit when you get a chance.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+ * <p>Does the following steps:
+ * <ul>
+ *     <li>updates user's email (checking if this email not used by another user);</li>
+ *     <li>updates user's first and last name.</li>
+ * </ul>
+ * NOTE: if any data (passed from Dsm) is null then it not updated.
+ */
+public class UpdateProfileProcessor extends DsmTaskProcessorAbstract {
+
+    public static final String TASK_TYPE__UPDATE_PROFILE = "UPDATE_PROFILE";
```
yufengwng: I think task type should be a `TaskType` enum instead of strings.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Processes Dsm Task of taskType='UPDATE_PROFILE'.
+ * It updates person's email, first, and last name.
+ *
+ * <p>Does the following steps:
+ * <ul>
+ *     <li>updates user's email (checking if this email not used by another user);</li>
+ *     <li>updates user's first and last name.</li>
+ * </ul>
+ * NOTE: if any data (passed from Dsm) is null then it not updated.
```
yufengwng: This doesn't sound like the semantics we wanted? I thought we were going to allow `null` as value, and we only skip updating if the _key_ is not present in json. (Although I'm not sure what allow setting email to `null` will even mean...)

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
yufengwng: Also, should we pass back the entire exception object instead of just the message? So someone can log the exception and it's stacktrace?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskReceiver.java
yufengwng: Yup, I agree. If things look good and we finished processing it, we should `ack()`. And then it becomes our responsibility to send out the result message.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
yufengwng: I see that we have a try/catch in the "abstract" processor class, which will return a "error" result data object. However, similar to Zim's concern, I don't see anywhere handling the error case and calling `nack()`. The "result sender" seems to go straight to publishing the result...

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskPubSubConnectionCreator.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskPubSubConnectionCreator.java
evgeniipr: Added

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateFirstLastNameHandler.java
evgeniipr: Refactored

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskReceiver.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
evgeniipr: The idea is that in any case the request considered as processed (ack) and to the result topic sent a message with results (SUCCESS or ERROR).
@yufengwng are you suggesting to send a result message a stacktrace?

pepper-apis/config/application.conf.ctmpl
evgeniipr: Sorry, my fault! Thank you a lot!

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: I did it for purpose (not using enum): in order to separate api and impl layer: in case of adding type in enum it needs to create this enum in api layer. Currently it is possible no to change api layer when need to add new type/new processor.
I posted more detailed message in Slack..
If you think that it not worth do do efforts for components loose coupling, then I can define types as enums. 
It's not a problem: I just want to understand architect approach in Pepper project and will follow it (I not 'religious' regarding coding patterns:)).

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskProcessorAbstract.java
yufengwng: What I mean was, if there was an exception while processing, we should log it and try again. So someone should log the exception and call `nack()` so pubsub will send the message again and we can try processing again (e.g. imagine if there was a temporary database connection issue).

However, if the message was faulty and we're unable to process it due to "client mistake" then yeah, I can see us returning some sort of "error" and then `ack()` the message. For example, imagine if we got an invalid email -- that will be improper data and a client error.

So I guess what we're missing in implementation is a way to distinguish between "client error" versus "server processing error". In the client case, we should probably `ack()` message and move on. In server case, we should probably `nack()` and try again.

evgeniipr: I added the nack() for the case when internal server error happened in auth0.
About temporary DB connection issue - I need to investigate more what kind of errors are possible..

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: Refactored this code: added exception DsmTaskException which have boolean flag meaning that if this exception thrown with flag set to true, then nack() should be called (i.e. it will try to process the task again).

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/api/DsmTaskData.java
evgeniipr: @zyme I replaced Dsm with PubSub

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: I reimplemented according to the semantics: "we were going to allow null as value, and we only skip updating if the key is not present in json."

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
```diff
@@ -0,0 +1,139 @@
...
+    public void destroy() {
+        if (pubSubTaskSubscriber != null && pubSubTaskSubscriber.isRunning()) {
+            pubSubTaskSubscriber.stopAsync();
+        }
+        if (pubSubTaskResultPublisher != null) {
+            try {
+                pubSubTaskResultPublisher.shutdown();
+                pubSubTaskResultPublisher.awaitTermination(PUBLISHER_AWAIT_TERMINATION_TIMEOUT_MIN, TimeUnit.MINUTES);
+            } catch (Exception e) {
+                LOG.error(errorMsg("Failed to shutdown PubSubTask connection"), e);
```
zyme: Seems like this error and the one on L76 are fatal--perhaps throw a runtime exception instead of logging?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskMessageParser.java
```diff
@@ -0,0 +1,113 @@
...
+        this.projectSubscriptionName = projectSubscriptionName;
+        this.pubSubTaskProcessorFactory = pubSubTaskProcessorFactory;
+    }
+
+    public PubSubTaskMessageParseResult parseMessage(PubsubMessage message, AckReplyConsumer consumer) {
+        String messageId = message.getMessageId();
+        String taskType = message.getAttributesOrDefault(ATTR_TASK_TYPE, null);
+        String participantGuid = message.getAttributesOrDefault(ATTR_PARTICIPANT_GUID, null);
+        String userId = message.getAttributesOrDefault(ATTR_USER_ID, null);
+        String studyGuid = message.getAttributesOrDefault(ATTR_STUDY_GUID, null);
```
zyme: I thought all of these fields (other than `messageId`) were going to be fields in the payload?  My concern here is that it may be difficult to re-run an operation using these as message attributes when compared to deserializing the data payload.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskMessageParser.java
```diff
@@ -0,0 +1,113 @@
...
+                projectSubscriptionName, messageId, taskType, participantGuid, userId, studyGuid, payloadJson);
+
+        var pubSubTaskDescriptor = pubSubTaskProcessorFactory.getPubSubTaskDescriptors(taskType);
+
+        String errorMessage = null;
+        Object payloadObject = null;
+        PubSubTask.PayloadMap payloadMap = new PubSubTask.PayloadMap();
+        if (pubSubTaskDescriptor.isPayloadConvertibleToMap()) {
+            payloadMap.setMap(gson.fromJson(payloadJson, Map.class));
+        }
```
zyme: What happens if this is not convertible to a map?  I think it's safe to assume the data is always going to be json, so let's just use json and dispense with the conversion to a map that requires fields to be strings.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateEmailHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        if (errMsg != null) {
+            throw new PubSubTaskException(errMsg);
+        }
+    }
+
+    private void updateEmailInAuth0(Handle handle, ManagementAPI mgmtAPI, UserDto userDto, String email, String userGuid) {
+        var status = Auth0Util.updateUserEmail(mgmtAPI, userDto, email);
+        String errMsg = null;
+        switch (status.getAuth0Status()) {
+            case SUCCESS:
```
zyme, line 64: I'm not sure we need to add more details in the message for a sub set of auth0 statuses.  Since we know how to find details in auth0, we could just put `status.getAuth0Status()` into the error message directly.

evgeniipr:
Added unit tests

evgeniipr:
Added unit tests

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskMessageParser.java
evgeniipr: @zyme you mean by "re-run "- to call this task not from DSM but for example from command line? What kind of difficulty you mean?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
evgeniipr: Ok. But seems it will be fatal for whole Housekeeping. Is it ok?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskMessageParser.java
evgeniipr: Agree. Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateEmailHandler.java
evgeniipr: This logic taken from UpdateUserEmailRoute.java. May be to make it a common part for both classes (UpdateUserEmailRoute.java and UpdateEmailHandler.java)? Actually you've already confirmed earlier that it needs to be done (this kind of refactoring). But I still not done it.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
zyme: How about wrapping this in a try/catch when it's called in `Housekeeping.java`?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskLogUtil.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.ddp.event.pubsubtask.api;
+
+/**
+ * Adds a prefix to log messages in PubSubTask API.
+ */
+public class PubSubTaskLogUtil {
```
zyme, line 6: I'm not sure we need this?  logback will include the class/method names, which should give us enough information to track this to pubsub.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskDataReader.java
```diff
@@ -0,0 +1,36 @@
...
+ * Reads a task-specific attributes and parses a task-specific payload.
+ * The readerhould be defined for a certain taskType {@link PubSubTaskProcessorFactory}.
+ * The message to be parsed received from subscription
+ * defined by config param "pubsub.pubSubTasksSubscription".
+ * The fetched attributes set to {@link PubSubTask#getAttributes()}.
+ * An object with parsed payload is returned from the method.
+ * If error happens during parsing then {@link PubSubTaskException} will be thrown.
+ */
+@FunctionalInterface
+public interface PubSubTaskDataReader {
```
zyme: This still seems a bit over-engineered to me.  At the end of the day, it's just json as a string.  Can we simplify this so that we don't need to create separate parsers?  The gson tools seem sufficient without having to wrap things and deal with type safety checks.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskException.java
```diff
@@ -0,0 +1,38 @@
...
+        this(message, pubSubTask, false);
+    }
+
+    public PubSubTaskException(String message, PubSubTask pubSubTask, boolean needsToRetry) {
+        super(message);
+        this.pubSubTask = pubSubTask;
+        this.needsToRetry = needsToRetry;
+    }
+
+    public boolean isNeedsToRetry() {
```
zyme: How about `shouldRetry` instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskDataReader.java
zyme: Another reason to not have such tight coupling and type safety is the likelihood that we will change these definitions over time, and have situations in which we see two slightly different payload versions sitting in a topic.  If we force this kind of strongly typed parsing, we have to tightly coordinate deployments on both sending and receiving time, and arrange for quiescence, which can get tricky given the async nature of pubsub.

evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskException.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskLogUtil.java
evgeniipr: It's easier to find log messages related only to this api: `textPayload: "PUBSUB_TASK"`

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
```diff
@@ -45,6 +45,20 @@ public static Integer getIntIfPresent(Config cfg, String key) {
         return cfg.hasPath(key) ? cfg.getInt(key) : null;
     }
 
+    /**
+     * Get value as integer if the key is present and value is not null.
+     * If key not present then return a default value.
+     */
+    public static int getIntIfPresent(Config cfg, String key, int defaultValue) {
```
yufengwng: No change needed here, but just wanted to say that for things that return a default value, I find that naming them with a pattern like `*OrElse` is much clearer: `getIntOrElse()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
```diff
@@ -45,6 +45,20 @@ public static Integer getIntIfPresent(Config cfg, String key) {
...
+        try {
+            return cfg.hasPath(key) ? cfg.getInt(key) : defaultValue;
+        } catch (ConfigException e) {
+            String value = e instanceof ConfigException.WrongType ? "=" + cfg.getValue(key) : "";
```
yufengwng: I'm starting to think that catching an exception and returning a default like this is hiding a potential error. My opinion is we should fail-fast here so folks have to fix up the config.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/SendGridEventService.java
```diff
@@ -24,7 +24,8 @@ public SendGridEventService() {
         return gson.fromJson(sendGridEventsJson, SendGridEvent[].class);
     }
 
-    public void logSendGridEvents(SendGridEvent[] sendGridEvents) {
+    public void logSendGridEvents(SendGridEvent[] sendGridEvents, String sendGridEventsJson) {
+        LOG.info("SendGrid-EVENT-PAYLOAD: " + sendGridEventsJson);  // temporary for events investigation purposes (TODO: to be removed)
```
yufengwng: This looks like an unrelated change. Should this be removed?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    }
+
+    private void createPubSubTaskSubscriber() {
+        try {
+            // Provides an executor service for processing messages.
+            var executorProvider =
+                    InstantiatingExecutorProvider.newBuilder().setExecutorThreadCount(SUBSCRIBER_FAILURE_LISTENER_THREAD_COUNT).build();
```
yufengwng: Doesn't look like this executor is being used... I don't see it being used in `subscribeBuilder()`, and `addListener()` is using `MoreExecutors.directExecutor()` instead.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
```diff
@@ -0,0 +1,139 @@
...
+
+            pubSubTaskSubscriber = pubSubConnectionManager.subscribeBuilder(projectSubscriptionName, pubSubTaskReceiver)
+                    .setParallelPullCount(1)
+                    .setExecutorProvider(InstantiatingExecutorProvider.newBuilder()
```
yufengwng: I think there's also a "system executor" too. Might want to set that too to control the resources we're using.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
```diff
@@ -0,0 +1,139 @@
...
+                            if (!executorProvider.getExecutor().isShutdown()) {
+                                createPubSubTaskSubscriber();
+                            }
+                        }
+                    },
+                    MoreExecutors.directExecutor());
+
+            pubSubTaskSubscriber.startAsync().awaitRunning(subscriberAwaitRunningTimeout, TimeUnit.SECONDS);
+        } catch (TimeoutException e) {
+            // Shut down the subscriber after specified timeout (default=30s). Stop receiving messages.
```
yufengwng: This comment is a bit misleading, can you take a second look?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTask.java
```diff
@@ -0,0 +1,53 @@
+package org.broadinstitute.ddp.event.pubsubtask.api;
+
+import java.util.Map;
+
+/**
+ * Data extracted from PubSubTask message
+ * (published to pubsub topic which DSS subscribed to - subscription is specified by
+ * config parameter "pubsub.pubSubTasksSubscription").
+ *
+ * <p>Data is fetched from pubsub message attributes and from message payload (JSON foc).
```
yufengwng: Typo?
```suggestion
 * <p>Data is fetched from pubsub message attributes and from message payload (JSON doc).
```

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskProcessorFactory.java
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.ddp.event.pubsubtask.api;
+
+/**
+ * Factory for registering {@link PubSubTaskProcessor}'s
+ * for each of known PubSubTask taskType's.
+ */
+public interface PubSubTaskProcessorFactory {
+
+    PubSubTaskDescriptor getPubSubTaskDescriptors(String pubSubTaskType);
```
yufengwng: Should the name be singluar? e.g. `..Descriptor()`. Taking a closer look, can we just return the `PubSubTaskProcessor` and remove a layer of abstraction? Or is there really a need for "descriptor"?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskProcessorAbstract.java
```diff
@@ -0,0 +1,56 @@
+package org.broadinstitute.ddp.event.pubsubtask.api;
```
yufengwng, line 1: Should these "abstract" classes be in the `impl` package instead of the `api` package? Since they're part of implementation? Or am I not understanding the design?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskProcessorAbstract.java
```diff
@@ -0,0 +1,56 @@
...
+
+    protected final Gson gson = GsonUtil.standardGson();
+
+    @Override
+    public PubSubTaskProcessorResult processPubSubTask(PubSubTask pubSubTask) {
+        boolean shouldRetry = false;
+        PubSubTaskResult.PubSubTaskResultType pubSubTaskResultType = SUCCESS;
+        String errorMessage = null;
+        try {
+            LOG.info(infoMsg("Task processing STARTED: taskType={}, data={}"), pubSubTask.getTaskType(), pubSubTask.getPayloadJson());
```
yufengwng: I think we have already logged the type and data payload in the receiver. Maybe trim down on the log message?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskResult.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public PubSubTaskResultPayload getPubSubTaskResultPayload() {
+        return pubSubTaskResultPayload;
+    }
+
+    public PubSubTask getPubSubTask() {
+        return pubSubTask;
+    }
+
+    public String toString() {
```
yufengwng, line 37: Should add a `@Override` here for clarity.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void handleTask(PubSubTask pubSubTask) {
+        validateTaskData(pubSubTask);
+        updateData(pubSubTask);
+    }
+
+    protected void validateTaskData(PubSubTask pubSubTask) {
+        String participantGuid = pubSubTask.getAttributes().get(ATTR_PARTICIPANT_GUID);
+        String userId = pubSubTask.getAttributes().get(ATTR_USER_ID);
+        if (participantGuid == null || userId == null) {
```
yufengwng: Should also check string `.isBlank()` as well.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public void handleTask(PubSubTask pubSubTask) {
+        validateTaskData(pubSubTask);
+        updateData(pubSubTask);
+    }
+
+    protected void validateTaskData(PubSubTask pubSubTask) {
+        String participantGuid = pubSubTask.getAttributes().get(ATTR_PARTICIPANT_GUID);
+        String userId = pubSubTask.getAttributes().get(ATTR_USER_ID);
```
yufengwng, line 37: Is the study staff `user_id` used in any way? Looks like we're not even logging it?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+            throw new PubSubTaskException(format(errorMsg("Some attributes are not specified: participantGuid=%s, userId=%s"),
+                    participantGuid, userId), pubSubTask);
+        }
+        if (StringUtils.isBlank(pubSubTask.getPayloadJson())) {
+            throw new PubSubTaskException(errorMsg("Empty payload"), pubSubTask);
+        }
+    }
+
+    protected void updateData(PubSubTask pubSubTask) {
+        Properties properties = gson.fromJson(pubSubTask.getPayloadJson(), Properties.class);
```
yufengwng, line 48: Is there a benefit of using `Properties` instead of just a regular `Map<String,String>`? Just curious for my own learning purposes.

pepper-apis/src/test/java/org/broadinstitute/ddp/event/pubsubtask/PubSubTaskTestUtil.java
```diff
@@ -0,0 +1,97 @@
...
+    public static final String TEST_USER_ID = "user_id";
+    public static final String TEST_STUDY_GUID = "study_guid";
+
+    public static final String EMAIL = "email";
+    public static final String EDUCATION = "education";
+    public static final String MARITAL_STATUS = "maritalStatus";
+    public static final String FIRST_NAME = "firstName";
+    public static final String LAST_NAME = "lastName";
+
+    public static final String TEST_EMAIL = "test@broadinstitute.org";
```
yufengwng: I recommend `test@datadonationplatform.org`. You never know, someone at Broad might actually have the email `test@broad`!

pepper-apis/src/test/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskCustomMessageTest.java
```diff
@@ -0,0 +1,170 @@
...
+     * check that payload can be converted to a map and properties fetched from it.
+     * Create result pubsub message (successful one) out of the parsed {@link PubSubTask}
+     */
+    @Test
+    public void testCustomMessageProcessing1() {
+        init();
+        var message = buildMessage(TestProcessor1.TEST_TASK_1,
+                format("{'%s':'%s', '%s':'%s', '%s':'%s'}",
+                        EMAIL, TEST_EMAIL, EDUCATION, TEST_EDUCATION, MARITAL_STATUS, TEST_MARITAL_STATUS), true);
+        pubSubTaskReceiver.receiveMessage(message, new TestAckReplyConsumer());
```
yufengwng: I think this is an ideal place to use mocks (using Mockito)!
* We can create a `var mockReply = mock(AckReplyConsumer.class)` and pass that in.
* Then we can actually verify behavior with something like `verify(mockReply, times(1)).ack()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/dsmtask/impl/updateprofile/UpdateProfileProcessor.java
yufengwng: Okay, let's give that a try!

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskMessageParser.java
evgeniipr: @zyme could we discuss this suggestion (to put all data in the payload) together with DSM team? I posted the message about it in Slack channel ddp-5235. Because they also needs to do these changes in case if it will be decided to do it.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
evgeniipr: I agree. I renamed anyway.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/SendGridEventService.java
evgeniipr: Yes, removed it. I added this to deploy to dev the extra logging to check what sent in Sendgrid events payload..

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
evgeniipr: Thank you, my fault: seems first took the sample from Housekeeping then added the listener from google cloud sample.. Fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: Agree. Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: It used by DSM: it needs to return this attribute in result message so that DSM could find user bu user_id.
I added logging of all attributes.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
evgeniipr: Initially I used Map<String, String>, then I added a class PayloadMap (encapsulating this Map<String,String>, now I decided to replace to Properties.
I see 2 benefits:
- it possible to hold not only string but other types because Property holds objects: may be for some other taskTypes it needs to pass some numeric or boolean values in JSON payload,
- it more convenient to manipulate with Properties type than Map<String, String> type - just less verbose code, and may be more reliable.., and no need to invent new custom class like PayloadMap).

pepper-apis/src/test/java/org/broadinstitute/ddp/event/pubsubtask/PubSubTaskTestUtil.java
evgeniipr: Fixed, thanks

pepper-apis/src/test/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskCustomMessageTest.java
evgeniipr: Yes, initially my idea was to mock this but then I wanted to add some test code into custom ack(), nack(), but not added..
Ok, will utilize mockito. Thanks!

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
evgeniipr: Fixed it.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
evgeniipr: Removed try/catch.

pepper-apis/src/test/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskCustomMessageTest.java
evgeniipr: Utilized Mocks instead of custom class creation.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskConnectionService.java
evgeniipr: Removed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTask.java
evgeniipr: Thanks, fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskProcessorFactory.java
evgeniipr: Yes, fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskProcessorAbstract.java
evgeniipr: This particular class have common (api level) code.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskResult.java
evgeniipr: Thanks, added

yufengwng (approved):
This looks good to me so far. I think the rest can be cleaned up and refactored later.

PR #622, opened 2021-02-03
Remove hideProgressIndicator property from API

MocanaAtBroad:
hideProgressIndicator property has been added to support F.H. design.
In latest design, every page in the wizard show the progress indicator.
Not needed so would like to drop.


yufengwng (approved):
Weird. I see a whole bunch of commits. Can you either rebase your branch, or make sure to use the squash-and-merge button?

MocanaAtBroad:
> Weird. I see a whole bunch of commits. Can you either rebase your branch, or make sure to use the squash-and-merge button?

I think this happened because I just reused the original API changes branch that had all the commits created over multiple weeks. Lesson learned! At any rate, on merge to develop we'll squash.

PR #623, opened 2021-02-04
[DDP-3709] part 3: file scanner service

yufengwng:
## Context

Part of DDP-3709. This PR adds a Cloud Function for scanning uploaded files using ClamAV. See the README for all the details. The Cloud resources we need for this implementation is 1 Cloud Function and 2 pubsub topics. The file scanner should be general enough to be leveraged for other Pepper functionality. I'll implement actually consuming the scan result and moving the scanned files in an upcoming PR.

Putting this in draft mode until I work out the rough edges and do more testing.

## Checklist

## FUD Score

    - See README.

## Testing

## Release

    - See README.


yufengwng:
CC @YuriyPavlovBI @bskinner .

yufengwng:
I have a version of this pipeline running on `dev` environment now, so you can try it out. Drop a file into the `ddp-dev-file-uploads` bucket, see the CF gets triggered, and subscribe to the `cf-file-scan-result` topic to get the scan result message.

cf-file-scanner/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Project layout
+
+```
+build/              - This is where the clamav tarball go. Ignored during deploy.
+clamav/             - This is where clamav binaries go.
+config/             - Put your config file in here. Ignored during deploy.
+src/                - Project source.
+application.conf    - Main config file. This will be rendered during deploy.
+freshclam.conf      - Config file that freshclam needs in order to run.
```
MocanaAtBroad: I have to ask. **freshclam** . This is lingo used by clamav or something you came up with?

cf-file-scanner/README.md
```diff
@@ -0,0 +1,109 @@
...
+function. This is so we can trigger the function locally easily using something
+like `curl`. The HTTP request payload follows a certain format. See the
+[docs][call-cf] for more info.
+
+You'll also need Google application default credentials on your local
+environment. You can get this by doing `gcloud auth login`.
+
+1. Install ClamAV on your local machine.
+2. Create subdirectory named `clamav`.
+3. Copy `clamscan` and `freshclam` binaries FROM YOUR LOCAL MACHINE to the `clamav` directory.
```
MocanaAtBroad, line 74: Guess there is my answer regarding **freshclam** . I am a happy clam now.

cf-file-scanner/build-clamav-inside-docker.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo ''
+echo '=> Moving bundle artifact into /build...'
+if [[ ! -d /build ]]; then
+  mkdir /build
+fi
+mv clamav.tar.gz /build
+
+echo ''
+echo '=> Done!'
```
MocanaAtBroad: Oh! Not a Dockerfile?

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+                if (!created) {
+                    throw new RuntimeException("Could not create database directory: " + DB_DIR);
+                }
+            }
+            runFreshclam();
+        }
+    }
+
+    private static CommandResult runCommand(boolean withInput, byte[] input, String... command) {
+        var builder = new ProcessBuilder(command);
```
MocanaAtBroad, line 115: What is the input? Is this meant to be the contents of the file?

cf-file-scanner/config/example.conf
```diff
@@ -0,0 +1,18 @@
...
+  # Unit need to match one of Java's `TimeUnit` enum variants.
+  # These are: DAYS, HOURS, MINUTES, SECONDS, MILLISECONDS.
+  "refreshTime": 1,
+  "refreshUnit": "HOURS",
+
+  # The GCP project. Used for both storage and pubsub.
+  "gcpProjectId": "string",
+
+  # The pubsub topic to publish scan results to.
+  "resultTopic": "string",
```
dinvlad: You may want to have the `resultTopic` defined via an env var for the GCF. This is usually the way to pass customization options to GCFs, because it's easier than having to modify a config file from CI/CD.

`gcpProjectId` should also be read from the env var, which is usually one of these:
> GCP_PROJECT | Reserved: The current GCP project ID.
> GCLOUD_PROJECT | Reserved: The current GCP project ID (deprecated).
> GOOGLE_CLOUD_PROJECT | Reserved: Not set but reserved for internal use.
(from https://cloud.google.com/functions/docs/env-var)

(note that project ID vars may not be defined for the Java GCF runtime, but it's worth defining explicitly when you deploy the function)

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+        long refreshTime = cfg.getLong(CFG_REFRESH_TIME);
+        refreshMillis = unit.toMillis(refreshTime);
+
+        String gcpProjectId = cfg.getString(CFG_GCP_PROJECT_ID);
+        String resultTopic = cfg.getString(CFG_RESULT_TOPIC);
+        try {
+            storage = StorageOptions.newBuilder()
+                    .setCredentials(GoogleCredentials.getApplicationDefault())
+                    .setProjectId(gcpProjectId)
```
dinvlad, line 74: I'm not sure if setting the project here is necessary - I'd double-check if the storage client requires it.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+        // Ensure database is up-to-date before scanning.
+        checkDatabaseFiles();
+
+        logger.info("Scanning file: " + blobId.toString());
+        ScanResult result = runClamscan(blob.getContent());
+
+        String data = message.getData() != null ? message.getData() : "";
+        var resultMessage = PubsubMessage.newBuilder()
+                .putAllAttributes(message.getAttributes())
+                .putAttributes(ATTR_SCAN_RESULT, result.name())
```
dinvlad, line 247: I'm curious if we should include a link to the Cloud Logging console message with the reason for failure (or the entire message as a data field or an attribute, if the space allows). Either way, this could help easily determine why a file didn't pass the scan, in a downstream app (e.g. notification email sender).

cf-file-scanner/deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+echo 'Rendering config file...'
+gcloud --project=$PROJECT_ID secrets versions access latest \
+  --secret=$SECRET_ID > application.conf
+
+echo 'Uploading function...'
+echo 'Note: it might take a few minutes to upload clamav binaries.'
+
+echo ''
+gcloud --project=$PROJECT_ID functions deploy $CF_NAME \
+  --entry-point org.broadinstitute.ddp.cf.FileScanner \
```
dinvlad, line 30: Please set up a dedicated Service Account for this function - it's not great to run it with the "default" SA, which has Editor in the project (and thus defeats the purpose of an isolated function, because if the file is "malicious enough" it could possibly break out of ClamAV jail and wreak havoc in your project..)

dinvlad: This SA should have only Storage Object Viewer role on the bucket, and PubSub Publisher role on the target topic.

cf-file-scanner/config/example.conf
yufengwng: I actually wanted to use a `env.yaml` file for these initially. However, these config values will be different for our different environment, so I'll need a way to render these on deploy. I guess I could put the `env.yaml` file into Secret Manager. But I think this setup is a little easier and more consistent to how Pepper does things. Also, I think it's a lot easier to pass in a config file than dealing with env variables during local development and testing.

cf-file-scanner/deploy.sh
yufengwng: Yup, that's my next step!

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Currently, I don't see us needing that information, but could be worked out in the future!

cf-file-scanner/deploy.sh
yufengwng: Done! I added a new SA with no permissions or roles. Then added the necessary storage and pubsub permissions at the resource-level for those 2 particular resources.

cf-file-scanner/deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+  --trigger-topic $TOPIC_NAME \
+  --service-account $SA_ACCT \
+  --region us-central1 \
+  --runtime java11 \
+  --memory 2048MB \
```
yufengwng, line 35: Function seems to be operating at the upper end of 1.7 GB so we'll need to allocate 2GB here.

Screen Shot 2021-02-04 at 2 38 43 PM

cf-file-scanner/deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+  --timeout 5m \
```
yufengwng, line 36: I tried running a scan on some large files and it ran fine, well within a minute, so I think 5 minute timeout should be good.

12+MB file:
Screen Shot 2021-02-04 at 2 35 04 PM

84+MB file:
Screen Shot 2021-02-04 at 2 37 47 PM

yufengwng (commented):
This PR is ready for thorough review.

cf-file-scanner/deploy.sh
dinvlad: I wonder how much of that is actual Java overhead, and if this becomes a problem then maybe worth looking at Python instead..

cf-file-scanner/config/example.conf
dinvlad: OK, that's fine for now. In general, it just makes function code less "portable" and more dependent on your particular deployment mechanism ;) Aka 12-factor apps :)

dinvlad: I think the traditional approach is to store them in `.env` file and source that when you deploy, then you can just assign the values through the env files at deploy time instead of packaging the `.env`/`env.yaml` config itself with those. It also makes it easier to trace which function was deployed with which values - you can see those right in the Cloud Console!

dinvlad: https://12factor.net/config

dinvlad: I also was thinking more in terms of reusability of this GCF for other teams etc. If you're not looking to make this code useful to others, then it's fine to store config together with your code. I was more in favor of env vars because of reusability concern. So I think AppSec can then make a separate reusable implementation in our own repo, and treat your code as just useful for Pepper only.

yufengwng: Okay, that does sound nice and convincing. I think I can get it to work with env vars. As a general question: what if we have a more secret-like config value, e.g. database url, are those still good to be set as env vars?

dinvlad: For "true" secrets (those containing credentials), definitely those should be set "out of band" (i.e. not stored in any config files..), usually by hand via Cloud Console for Secret Manager. This is what we've done for database connection strings in my previous team ;-)

dinvlad:
Btw did you have a chance to test it on a "suspicious" file? We may want to find one that's used by research community to test virus scanners (so it's "safe" but intentionally contains suspicious content).

One standard example is here: https://www.eicar.org/?page_id=3950 (it's not a PDF, but we should anticipate that users may trick the system into accepting these)

cf-file-scanner/deploy.sh
dinvlad: LGTM! Great analysis here

cf-file-scanner/config/example.conf
yufengwng: All set! I've switched to using env vars. It's really nice to see everything about the function in one place in gcloud console. However, with env vars, I find myself doing something like this in the `deploy.sh` script:
```
source .env
VARS="GCP_PROJECT=$GCP_PROJECT"
VARS+=",RESULT_TOPIC=$RESULT_TOPIC"
...

gcloud ... --set-env-vars $VARS
```
because I need to build a string to pass to `gcloud functions deploy`. So instead of sourcing env file, I ended up rendering a `env.yaml` file from Secret Manager instead. The function itself and local development still uses env vars though.

yufengwng:
@dinvlad I haven't yet, but wanted to! Every time I tried creating a file with the EICAR signature, the SentinelOne program on my work machine immediately deletes the file... but I'm going to try again.

yufengwng:
@dinvlad Thanks for the link! I didn't know they have a ZIP version of that (I was only trying the text file one). I was able to use the zip eicar and looks like the scanner picked it up! I'll do more testing here but so far so good.

Screen Shot 2021-02-04 at 6 54 05 PM

Screen Shot 2021-02-04 at 6 55 17 PM

dinvlad:
@yufengwng Awesome, LGTM, and thanks for finding a workaround for testing! 😀

cf-file-scanner/build-clamav-inside-docker.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+#
+# This script will create a bundle of binaries and shared objects necessary for
+# running clamav scanner. This should be ran inside a fresh image of Ubuntu
```
MocanaAtBroad: Typo:
```suggestion
# running clamav scanner. This should be run inside a fresh image of Ubuntu
```

cf-file-scanner/deploy.sh
```diff
@@ -0,0 +1,38 @@
...
+
+echo 'Uploading function...'
+echo 'Note: it might take a few minutes to upload clamav binaries.'
+
+echo ''
+gcloud --project=$PROJECT_ID functions deploy $CF_NAME \
+  --entry-point org.broadinstitute.ddp.cf.FileScanner \
+  --trigger-topic $TOPIC_NAME \
+  --service-account $SA_ACCT \
+  --region us-central1 \
```
MocanaAtBroad, line 33: Should we be hard-coding region?

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+            process = builder.start();
+        } catch (IOException e) {
+            throw new RuntimeException("Error starting command: " + command[0], e);
+        }
+
+        if (withInput && input != null) {
+            try {
+                process.getOutputStream().write(input);
+                process.getOutputStream().flush();
+                process.getOutputStream().close();
```
MocanaAtBroad: Maybe `finally` or treating stream as closeable resource would make sense here?

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        Instant createdAt = Instant.ofEpochMilli(blob.getCreateTime());
+        logger.info("Found file that was created at: " + createdAt.toString());
+
+        // Ensure database is up-to-date before scanning.
+        checkDatabaseFiles();
+
+        logger.info("Scanning file: " + blobId.toString());
+        ScanResult result = runClamscan(blob.getContent());
```
MocanaAtBroad: Can we stream the file contents instead of reading them all into memory? Would be surprised if clamav cannot take a file stream???

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+        String value = System.getenv(name);
+        if (value == null || value.isBlank()) {
+            throw new RuntimeException("Missing environment variable: " + name);
+        }
+        return value;
+    }
+
+    // This method needs to be thread-safe since it uses shared global state and is used in `accept()`.
+    private static synchronized void checkDatabaseFiles() {
+        if (dbUpdatedAt == null || Instant.now().toEpochMilli() - dbUpdatedAt.toEpochMilli() >= refreshMillis) {
```
MocanaAtBroad, line 94: Don't know much about cloud functions. Does this end up being executed by GCP as a single long-lived process? Trying to figure if dbUpdatedAt is truly a global for the whole system.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,280 @@
...
+        String data = message.getData() != null ? message.getData() : "";
+        var resultMessage = PubsubMessage.newBuilder()
+                .putAllAttributes(message.getAttributes())
+                .putAttributes(ATTR_SCAN_RESULT, result.name())
+                .setData(ByteString.copyFromUtf8(data))
+                .build();
+
+        try {
+            String msgId = publisher.publish(resultMessage).get();
+            logger.info("Published scan result with messageId: " + msgId);
```
MocanaAtBroad, line 253: Have you gotten yet to the handling of files/upload status updates? Is that going to happen in housekeeping?

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Yup, see usage in `runClamscan()` call.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Is there a way to get the contents of a bucket object as a stream? I didn't see anything when looking at the methods on the `Blob` Java class. Happy to stream more things if possible!

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: I'm not too experienced with Cloud Function infrastructure either. But what I know now is:
* GCP will create an "instance" to host your function (I think it's app engine under the hood, not sure)
* Global stuff is ran once. Imagine if GCP called `var fn = new YourFunctionClass()` once.
* The `accept()` method is invoked for each request. E.g. imagine GCP calling `fn.accept()`.
* The same function instance MAY be reused for multiple requests, OR new function instances might be spinned up.
* And each instance is totally separate, so separate compute and memory.

So answer to your question is not really. It's only lives long enough as GCP decides it should.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Just did! See PR #626.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
MocanaAtBroad: Starting from a Blob, I think you can stream stuff using:
https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/Blob.html#reader-com.google.cloud.storage.Blob.BlobSourceOption...-
(I think you might be able to use directly or create an InputStream from it if treated as an nio Channel)

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Done!

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Thanks! I did some digging around and saw the `reader()` thing as well. I also saw that gcloud libraries come with some nifty byte stream utils, so I stitched something together.

I did a quick comparison by running things locally with an 84MB file as input. Overall, I see that using streams does reduce memory usage but increases scanning time. I think this is because behind the scenes the `reader()` from the blob actually makes HTTP calls out to GCP to load chunks of bytes. So with streams there's going to be more roundtrips, roughly 10 seconds more on my local. It feels that memory is a more limited resource when it comes to Cloud Functions, so waiting a handful more seconds seems like an acceptable trade-off.

Before: There's a big spike starting around 6:32:12pm, which is when I think we start loading the blob contents.

Screen Shot 2021-02-05 at 6 34 50 PM
Screen Shot 2021-02-05 at 6 34 37 PM

After: The spike is noticeable smaller and more flat, starting around 6:40:42pm.

Screen Shot 2021-02-05 at 6 42 28 PM
Screen Shot 2021-02-05 at 6 42 20 PM

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
MocanaAtBroad: So is it big deal that we don't persist the timestamp for last check. Obvious consequence is extra checks and extra downloads.

yufengwng: It's not a big deal. It'll just do the download before running a file scan. I added this as a small optimization in case the function instance does get reused (which most likely it will during higher request rates) -- it shaves off roughly 40 seconds.

dinvlad: I think it might be best to keep it as-is (with timestamp) if it allows you to fetch the most recent update. Afaik Google keeps these instances sometimes running for a day or so, so the cached version might actually go out of date. But if you'd like to remove this check, no big deal! Then I'd recommend to just cache it once at the start of the function, and reuse on all subsequent invocations.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
dinvlad: Yea if this substantially reduces memory usage, I'd be in favor of longer scan times tbh - as memory is unfortunately very limited on GCF, and not an "elastic" resource unlike time 😄

PR #616, opened 2021-01-27
[DDP-3709] part 1: file upload support and API

yufengwng:
## Context

Part 1 of DDP-3709. This PR contains:
* File upload database support,
* New file upload service and config file changes,
* New API endpoint for authorizing file uploads.

Credits to @YuriyPavlovBI for initial implementation in #559, which this PR is based off of. API follows spec as outlined in #565. Note that this merges into my feature branch `ddp-3709`: this is part 1 of a series of "stacked" PRs. More PRs will be created to implement other parts of DDP-3709.

**Pre-Deployment Steps (adapt as appropriate to different environments):**
1. Create new service account named `ddp-file-uploader`.
2. Create new JSON key for service account, and save the JSON file.
3. Create new custom role named `DDP File Uploader` and grant these permissions:
    - `storage.objects.create`
    - `storage.objects.delete`
4. Create new bucket to hold uploads, called `ddp-dev-file-uploads`.
5. Grant `ddp-file-uploader` permission to the bucket with the `DDP File Uploader` role.
6. Save service account and configs into Vault.

## Checklist

- [x] I have labeled the type of changes involved using the `C-*` labels.
- [x] I have assessed potential risks and labeled using the `R-*` labels.
- [x] I have considered error handling and alerts, and added `L-*` labels as needed.
- [x] I have considered security and privacy, and added `I-*` labels as needed
- [x] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

_Overall, how are you feeling about these changes?_

- [ ] :relaxed: All good, business as usual!
- [x] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [ ] They are user-visible in dev as a regular user journey and require no additional instructions.
- [x] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [x] I have written automated positive tests
- [ ] I have written automated negative tests
- [x] I have written zero automated tests but have poked around locally to verify proper functionality
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

[...]

PR #624, opened 2021-02-05
DDP-3709 file upload feature

yufengwng:
## Context

See DDP-3709. File upload feature is implemented via a series of PRs. You can see more detailed changes and commentary by viewing the PRs below:

* Part 1: file upload API #616
* Part 2: file questions/answers #620
* ~Part 3: file scanning service #623~ (is now #636)
* Part 4: file upload verification #625
* Part 5: file scan result handler #626
* Part 6: file cleanup job #630 
* Part 7: file upload data export #631

## Cloud Resources

The implementation uses several new paid resources:

* 1 service account
* 1 pubsub subscription
* 3 Google Buckets

## Release
_Note: Replace `$env` with the appropriate environment in the instructions below._

(1) The `cf-file-scanner` service should have already been deployed by #636.

(2) Once file scanner is setup, we can setup Pepper stuff, starting with the buckets.
* Create 3 new buckets named `ddp-$env-file-uploads`, `ddp-$env-file-scanned`, and `ddp-$env-file-quarantine`.
* Use the `cf-file-scanner/init-bucket-event.sh` script to setup bucket notification.
    * `ddp-$env-file-uploads` bucket -> `cf-file-scanner` pubsub topic.
* Grant `cf-file-scanner` SA `Storage Object Viewer` role to the `ddp-$env-file-uploads` bucket, at the resource-level.

(3) Create a subscription to the `cf-file-scan-result` topic.
* Name: `cf-file-scan-result-hkeep-sub`
* Subscription filter: `attributes.bucketId="ddp-$env-file-uploads"`
* No expiration, 300 secs acknowledgement, with exponential backoff.

(4) Create new service account called `ddp-file-uploader`, which will be used solely for authorizing file uploads.
* Create SA and create new JSON key (save this for later)
* Assign `ddp-file-uploader` SA `DDP Storage Objects Create` role to the `ddp-$env-file-uploads` bucket, at the resource-level.

(5) Add new properties to Vault. I recommend the following settings:
```
    "fileUploads": {
        "uploadsBucket": "ddp-$env-file-uploads",
        "scannedBucket": "ddp-$env-file-scanned",
        "quarantineBucket": "ddp-$env-file-quarantine",
        "maxFileSizeBytes": 2097152,  // 2 MB
        "maxSignedUrlMins": 5,
        "removalExpireTime": 1,
        "removalExpireUnit": "DAYS",
        "removalBatchSize": 100,
        "enableScanResultHandler": true,
        "scanResultSubscription": "cf-file-scan-result-hkeep-sub",
        "signerServiceAccount": { ... the ddp-file-uploader JSON key here ... }
    },
    "schedules": {
        "fileUploadCleanup": "0 0 5 ? * * *",  // Every day at 5AM UTC
    }
```

(6) Upload the new `participants_structured` index template to Elasticsearch. Can use the `elasticsearch/elastic.sh` script to do this.

(7) Deploy new code.

(8) Since no existing studies will be using `FILE` questions, we don't have to rebuilt the Elasticsearch indices. Any new studies should pickup the new index template and get the proper data types. If existing studies will use `FILE` questions, then we should consider rebuilding their indices.


yufengwng:
CC @YuriyPavlovBI @SimoneMaiwald

cf-file-scanner/build-clamav-inside-docker.sh
```diff
@@ -0,0 +1,70 @@
...
+# 18.04, which is the image that Cloud Functions use by default for the Java 11
+# runtime.
+#
+# The final tarball will be located in the `/build` directory, so you can run
+# docker with a mount point to grab the bundle.
+
+set -e
+
+echo '=> Updating Ubuntu...'
+apt-get update
```
bskinner: This only updates the package definitions, and does not actually do any upgrades to the installed packages. In addition, this may fail when used non-interactively. If the intent is to upgrade to the latest version of the installed packages, this should be `apt update -y && apt upgrade -y`. If the intent is to only update the package definitions, the command should still use the `-y` flag.

cf-file-scanner/build-clamav-inside-docker.sh
```diff
@@ -0,0 +1,70 @@
...
+# running clamav scanner. This should be run inside a fresh image of Ubuntu
+# 18.04, which is the image that Cloud Functions use by default for the Java 11
+# runtime.
+#
+# The final tarball will be located in the `/build` directory, so you can run
+# docker with a mount point to grab the bundle.
+
+set -e
+
+echo '=> Updating Ubuntu...'
```
bskinner: Performing `apt` operations require `root` privileges. The `apt` call will fail if the user does not have an effective UID of `0`, but this may be a good spot to add a clearer error message that this script must be run as `root`.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,312 @@
...
+    private static final long refreshMillis;
+    private static final Storage storage;
+    private static final Publisher publisher;
+    private static Instant dbUpdatedAt = null;
+
+    // This is ran once on cold start.
+    static {
+        TimeUnit unit = TimeUnit.valueOf(getEnvOrThrow(ENV_REFRESH_UNIT));
+        if (unit == TimeUnit.MICROSECONDS || unit == TimeUnit.NANOSECONDS) {
+            throw new RuntimeException("Refresh time unit is too granular: " + unit);
```
bskinner: This seems like a very strange dance to be doing. Is there a reason we don't define `REFRESH_TIME` to be a fixed unit and leave it at that?

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
```diff
@@ -0,0 +1,312 @@
...
+        }
+        long refreshTime = Long.parseLong(getEnvOrThrow(ENV_REFRESH_TIME));
+        refreshMillis = unit.toMillis(refreshTime);
```
bskinner: Should we be doing a zero-check here?

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -711,6 +719,24 @@ void enrichWithDSMEventDates(Handle handle,
         });
     }
 
+    private void enrichWithFileRecords(Handle handle, FileUploadService fileService, long studyId, List<Participant> participants) {
+        var uploadDao = handle.attach(FileUploadDao.class);
+        Map<Long, List<FileRecord>> participantIdToFiles = new HashMap<>();
+        Set<Long> participantIds = participants.stream().map(p -> p.getUser().getId()).collect(Collectors.toSet());
+        try (var stream = uploadDao.findVerifiedAndAssignedUploadsForParticipants(studyId, participantIds)) {
```
MocanaAtBroad: Was trying to figure out what "assignedUploads" means...

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -453,6 +478,20 @@ AnswerSubmission.Date:
...
+    - type: object
+      required:
+        - value
+      properties:
+        value:
+          type: string
+          nullable: true
+          description: |
+            File upload guid provided by server when upload URL was requested.
+            A `null` value will unselect and unassign the uploaded file from
```
bskinner: What does `unselect` and `unassign` mean here? Will the file be deleted? This should be clarified in the docs.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -711,6 +719,24 @@ void enrichWithDSMEventDates(Handle handle,
...
+            stream.forEach(upload -> {
+                long key = upload.getParticipantUserId();
+                String bucket = fileService.getBucketForUpload(upload);
+                FileRecord record = new FileRecord(bucket, upload);
+                participantIdToFiles.computeIfAbsent(key, id -> new ArrayList<>()).add(record);
```
MocanaAtBroad, line 731: Also wonder if we should leave out the file location (bucket/blobname) from the export if the file has not yet been deemed safe.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+    - Activities
+  summary: create an authorized file upload
+  description: |
+    Request for a new URL for file upload. Link is accessible for limited
+    period of time (e.g. 5 mins). Resumable uploads are supported. With
+    information in response body, client will be able to upload the file
+    directly to a Google Bucket. If resumable=true is specified in request,
+    client will assume responsibility for creating the session URI with
+    provided URL in addition to follow up requests that contain the file
+    contents (see [documentation][doc]).
```
bskinner, line 18: Is there a reason the URL below isn't being linked to here?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+              type: string
+              description: The question identifier
+            fileName:
+              type: string
+              maxLength: 255
+              description: Name of file to be uploaded
+            fileSize:
+              type: integer
+              minimum: 1
+              description: Size of the file, in bytes (subject to a limit such as 2MB)
```
bskinner: Let's avoid mentioning a specific size here.

```suggestion
              description: Size of the file, in bytes. The server may limit the maximum size of uploaded files.
```

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+            mimeType:
+              type: string
+              maxLength: 255
+              default: application/octet-stream
+              description: MIME type, if not specified will default to `application/octet-stream`
+            resumable:
+              type: boolean
+              description: Whether URL should support resumable upload
```
bskinner:
```suggestion
              description: Whether URL should support resumable uploads
```

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
```diff
@@ -0,0 +1,12 @@
...
+public enum FileScanResult {
+    /**
+     * File upload should be sufficiently clean and safe. File is moved to safe storage and can be used.
+     */
+    CLEAN,
+    /**
+     * File upload was found to be malicious. File has been moved to quarantine and should not be used.
+     */
+    INFECTED,
+}
```
MocanaAtBroad, line 12: Could you confirm that result is truly binary, i.e., there is no "inconclusive" result?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+              - uploadUrl
+            properties:
+              uploadGuid:
+                type: string
+                description: Unique identifier for file upload
+              uploadUrl:
+                type: string
+                description: Signed URL which can be used to upload file contents
+    400:
+      description: invalid request or bad inputs
```
bskinner: This description needs to be expanded on. What constitutes an invalid request or bad input? Where should I start looking to debug the request?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+              description: Size of the file, in bytes (subject to a limit such as 2MB)
+            mimeType:
+              type: string
+              maxLength: 255
+              default: application/octet-stream
+              description: MIME type, if not specified will default to `application/octet-stream`
+            resumable:
+              type: boolean
+              description: Whether URL should support resumable upload
+  responses:
```
bskinner, line 61: I don't see any responses here indicating an error if the file size is too large.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateUserActivityUploadRoute.java
```diff
@@ -0,0 +1,125 @@
...
+            String prefix = String.format("%s/%s/%s", studyGuid, userGuid, instanceDto.getActivityCode());
+            return service.authorizeUpload(
+                    handle, instanceDto.getStudyId(), operatorUser.getId(), instanceDto.getParticipantId(), prefix,
+                    payload.getMimeType(), payload.getFileName(), payload.getFileSize(), payload.isResumable());
+        });
+
+        if (result.isExceededSize()) {
+            String msg = "File size exceeded maximum of " + service.getMaxFileSizeBytes() + " bytes";
+            LOG.warn(msg);
+            throw ResponseUtil.haltError(response, 422, new ApiError(ErrorCodes.NOT_SUPPORTED, msg));
```
MocanaAtBroad: Consider using HttpStatus.SC_UNPROCESSABLE_ENTITY , etc. where we have integers for status codes.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -0,0 +1,336 @@
...
+
+        public FileUpload getFileUpload() {
+            return fileUpload;
+        }
+
+        public URL getSignedUrl() {
+            return signedUrl;
+        }
+    }
+}
```
MocanaAtBroad, line 336: When I was asking about shared service code with A-T, this is the sort of thing I had in mind. Perhaps they way they ended up implementing things not amenable for sharing?

MocanaAtBroad (commented):
Looks good! Just a couple of comments.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -0,0 +1,106 @@
...
+            - fileName
+            - fileSize
+            - resumable
+          properties:
+            questionStableId:
+              type: string
+              description: The question identifier
+            fileName:
+              type: string
+              maxLength: 255
```
bskinner, line 47: What's going to happen if the length of the string is more than 255 bytes? Truncation, failed request, ...?

bskinner (commented):
Looks good after a first pass! Left a few comments to be addressed.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
yufengwng: Here's a screenshot from clamscan's man page, so yeah I think it's binary.
Screen Shot 2021-02-10 at 5 43 24 PM

cf-file-scanner/build-clamav-inside-docker.sh
yufengwng: Thanks for the tips! Intention is not "upgrade" -- we should keep whatever version of Ubuntu GCP is using for Cloud Functions. Will add the `-y` flag to the `update`!

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Wanted to provide easy way for us to specify time durations, e.g. "1 day", "12 hours", "20 minutes", etc.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: What will a refresh time of zero millis mean? In that case, I think it'll just refresh the db files on every function invocation, which I think is totally fine behavior.

cf-file-scanner/build-clamav-inside-docker.sh
yufengwng: The name of script indicates that this should be ran inside docker. And I believe docker uses root by default? In any case, will add more comments!

cf-file-scanner/build-clamav-inside-docker.sh
yufengwng: Pulled cf-file-scanner out into PR #636 and addressed this issue in https://github.com/broadinstitute/ddp-study-server/pull/636/commits/5e9c0f4c12e207d1c134d67bf8ded20968cbfdce.

cf-file-scanner/build-clamav-inside-docker.sh
yufengwng: Pulled cf-file-scanner out into PR #636 and addressed this issue in https://github.com/broadinstitute/ddp-study-server/pull/636/commits/5e9c0f4c12e207d1c134d67bf8ded20968cbfdce.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Please feel free to continue conversation in PR #636.

cf-file-scanner/src/main/java/org/broadinstitute/ddp/cf/FileScanner.java
yufengwng: Please feel free to continue conversation in PR #636.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
yufengwng: Also, this result is coming from the `cf-file-scanner` service, and that service only publishes these 2 result codes.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
yufengwng: "assigned uploads" is something I made up! I was looking for something to describe a "file upload that was associated with a file answer".

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
yufengwng: Given our current design, participant may answer a FileQuestion before we finish scanning the file. So I think we want to export their FileAnswer data along with the rest. So I'm inclined to export the data we have to Elasticsearch and give DSM the option to work with it.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
yufengwng: Yes, as far as I know, A-T went a different route.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Good question! I made up the terms `unselect` and `unassign` to describe the phenomenon of having a file that's uploaded but not actually "used" or "associated" with a FileAnswer. And yup, those files will get deleted eventually -- will add a note about in the docs! (Once again, I'm using quoted terms here, still not sure the right terminology to use!)

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: ~Right now, the response will have a general `FILE_ERROR` code, with the message having a more detailed description. Do we want more fine-grained error codes here? Like `FILE_SIZE_ERROR` or similar?~

Edit: My bad! I replied incorrectly. When file size exceeded, it returns a 422 with `NOT_SUPPORTED` error code.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: Link is showing up fine for me. Reminder that this is Markdown syntax.

Screen Shot 2021-02-11 at 12 26 52 PM

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateUserActivityUploadRoute.java
yufengwng: Done! d471b88524f8947ac322622378d83aa8fd0e9b5c

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Done! d471b88524f8947ac322622378d83aa8fd0e9b5c

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: Updated documentation to spell out that "invalid request or bad inputs" indicates issues with the payload properties. d471b88524f8947ac322622378d83aa8fd0e9b5c

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: Request will be rejected with a `400` status code. See updated description of `400` response.

yufengwng:
@pegahtah Do you have any comments about the Elasticsearch data for files?

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: I think `associated` may be a better fit than `assigned`. A file upload is associated with an answer (or an answer has an associated file upload). Thoughts?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
bskinner: That feels like a weird situation. Right now it seems like we have multiple disparate status codes for bad parameters. A file size too large is returning a `422` while an overly long string returns a `400`. Could we collapse these into a single error (`400` with a `BAD_PARAMETER` code, and explaining the error in the message)?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
bskinner: > Link is showing up fine for me. Reminder that this is Markdown syntax.

I am aware. I hadn't used the reference-style linking behavior before! That's a neat feature.

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: `associated` sounds good to me! Will switch to using that term unless other folks have objections.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
bskinner: Saw the descriptions added below- looks nice!

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: You're right, a 400 does seem to be a better fit! I'll switch this over to a `400` with the usual `BAD_PAYLOAD` error code.

bskinner: Sounds good! It looked like the included message clearly indicates the error, so I think this is set after the change.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
yufengwng: We converged on using the word "associated" instead. Hopefully that's clearer? e7eb94dd864889904a096f6f1da65ed1cb6bac12

pepper-apis/docs/specification/src/components/schemas.yml
yufengwng: Done! e7eb94dd864889904a096f6f1da65ed1cb6bac12

pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
yufengwng: Done! d52f3cabfa9bfd37262aea867e0b9f0f26a06dc0

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
MocanaAtBroad: "Some error(s) occurred" is a temporary condition or a terminal one?

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
MocanaAtBroad: IMO it defeats purpose  of scanning files if we allow access to contents before we scan them or after we find they are infected. And if we are going to include location but block access, then there is no point in including the location in first place.
Consider nulling out location unless there is a CLEAN scan result.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
yufengwng: I'm actually not sure. I haven't encountered an error yet. I can investigate.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
yufengwng: That sounds fine to me. @pegahtah please note that we'll null out the `bucket` and `blobName` properties if the file doesn't have a `CLEAN` scan result.

yufengwng: Done! 556de9b1f6d29e771ef672a3505e73b815b48dc6

pepper-apis/src/main/java/org/broadinstitute/ddp/model/files/FileScanResult.java
yufengwng: I've only done a cursory look at this, but my gut feeling is that `Some error(s) occurred` will most likely be "terminal" ones.

From [this link](https://unix.stackexchange.com/questions/510806/is-there-a-list-of-error-and-warning-messages-returned-by-clamscan), it appears that there's no official list of errors from `clamscan`, which sounds right as I couldn't find anything related to it on ClamAV's website. I also found [this link](https://www.drupal.org/project/clamav/issues/2441575#:~:text=2%20%3A%20Some%20error(s),records%20very%20little%20information%20e.g.) which lists some of the errors that `clamscan` _used_ to return. It's 5 years old, but looking at the list most of it seems like programmer/usage/operational errors.

The `cf-file-scanner` service (that runs the `clamscan` executable) does log all the output from `clamscan`, including any error messages, so we can investigate should an error pop up.

MocanaAtBroad: I am inclined to say let's go with what you have (which is damn good already), see what happens during operations, and adjust if we need to.

MocanaAtBroad (approved):
Looks nice! Look forward to start using it for real in a study.

PR #625, opened 2021-02-05
[DDP-3709] part 4: rework file upload status and upload verification

yufengwng:
## Context

Part of DDP-3709. This PR cleans up a few things in light of file scanning, in particular tracking file upload status and verifying the upload.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/endpoints/user.studies.activity.uploads.yml
```diff
@@ -44,13 +44,15 @@ post:
...
-              maxLength: 100
+              maxLength: 255
               description: Name of file to be uploaded
             fileSize:
               type: integer
+              minimum: 1
               description: Size of the file, in bytes (subject to a limit such as 2MB)
             mimeType:
               type: string
+              maxLength: 255
```
yufengwng, line 55: Note that I made some changes to the API docs to reflect updated implementation.

pepper-apis/src/main/resources/db-changes/schema/DDP-3709-file-upload.xml
```diff
@@ -23,16 +23,16 @@
             <column name="file_upload_guid" type="varchar(20)">
                 <constraints nullable="false" unique="true" uniqueConstraintName="file_upload_guid_uk"/>
             </column>
-            <column name="blob_name" type="varchar(300)">
+            <column name="blob_name" type="varchar(1024)">
```
yufengwng, line 26: According to [docs](https://cloud.google.com/storage/docs/naming-objects), blob object names can be up to 1024 bytes of UTF-8, which roughly translates to at most 1024 ASCII. So bumping this up.

pepper-apis/src/main/resources/db-changes/schema/DDP-3709-file-upload.xml
```diff
@@ -23,16 +23,16 @@
...
                 <constraints nullable="false"/>
             </column>
-            <column name="mime_type" type="varchar(50)">
+            <column name="mime_type" type="varchar(255)">
```
yufengwng, line 29: While digging around (see [here][1] and [here][2]), I found that mime type strings can be quite long, so bumping this up.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
[2]: https://stackoverflow.com/questions/643690/maximum-mimetype-length-when-storing-type-in-db

pepper-apis/src/main/resources/db-changes/schema/DDP-3709-file-upload.xml
```diff
@@ -23,16 +23,16 @@
...
                 <constraints nullable="false"/>
             </column>
-            <column name="file_name" type="varchar(100)">
+            <column name="file_name" type="varchar(255)">
```
yufengwng, line 32: While I'm in here, I'm bumping up `file_name` and `file_size` data types as well.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -133,51 +137,70 @@ URL generateSignedUrl(String blobName, String mimeType, HttpMethod method) {
...
+     * bucket. Since the file upload is locked, caller should commit transaction as soon as possible to unblock other
+     * transactions waiting on the file upload row.
      *
      * @param handle            the database handle
      * @param participantUserId the participant who this upload will be assigned to
      * @param upload            the file upload
      * @return check result
      */
-    public CheckResult checkAndSetUploadStatus(Handle handle, long participantUserId, FileUpload upload) {
+    public VerifyResult verifyUpload(Handle handle, long participantUserId, FileUpload upload) {
```
yufengwng, line 160: I struggled with implementing the verification here. I think it's mostly due to the asynchronous nature of the file operations. E.g. upload is done asynchronously by client, answering file question happens outside context of upload, files are scanned asynchronously somewhere else, file status might change or file gets moved to different bucket, etc.

I think the key here is to lock the file upload row in the database. So when user PATCHes answer, we lock the row and do the verification. In the "file scan result handler" which I'm implementing in upcoming PR, it'll try to lock the same file upload before updating scan status or moving the file to a bucket. So hopefully this helps avoid weird scenarios.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -133,51 +137,70 @@ URL generateSignedUrl(String blobName, String mimeType, HttpMethod method) {
...
         if (participantUserId != upload.getParticipantUserId()) {
-            return CheckResult.OWNER_MISMATCH;
+            return VerifyResult.OWNER_MISMATCH;
         }
 
-        if (upload.getStatus() == FileUploadStatus.AUTHORIZED) {
-            // File upload haven't been checked yet, do it now.
-            Blob blob = fetchBlob(upload.getBlobName());
+        if (!upload.isVerified()) {
```
yufengwng, line 165: While working through the "file scan result handler", I was struggling with how to deal with the different statuses and this "verification" of the upload. I think the key insight here is that there's ultimately two different concepts here: 1) whether file is "verified", and 2) file status can be separate from "verified". File needs to be "uploaded" to be marked "verified" of course, but outside of that we can update file status separately. So to make things more manageable, verification is only responsible for checking `is_verified`, while "file scan result handler" will be responsible for updating file status (e.g. uploaded timestamp and scan result).

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/FileUploadDao.java
```diff
@@ -19,40 +18,49 @@
...
             + "  from file_upload as f"
             + " where f.file_upload_guid = :guid")
     @RegisterConstructorMapper(FileUpload.class)
     Optional<FileUpload> findByGuid(@Bind("guid") String fileUploadGuid);
 
+    @SqlQuery("select f.*, (select file_scan_result_code from file_scan_result"
+            + "       where file_scan_result_id = f.scan_result_id) as scan_result"
+            + "  from file_upload as f"
+            + " where f.file_upload_id = :id for update")
+    @RegisterConstructorMapper(FileUpload.class)
```
MocanaAtBroad, line 61: Might we be locking `file_scan_result_code` unnecessarily?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
MocanaAtBroad: Is aim here to prevent scanner process from deleting file while this process is looking at its metadata? Guess on the other scanner side we lock when about to delete (2nd part of copy/delete original)? Am I understanding this correctly?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
MocanaAtBroad: Make sense to me that scan result is an independent status from user upload having been verified.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/FileUploadDao.java
yufengwng: According to mysql docs, `select for update` doesn't lock rows in subqueries so we should be good.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
yufengwng: Right. Imagine if the bucket file got moved/deleted right in-between the `select` and `fetchBlob()` call. We will not be able to find the file in the `uploads` bucket and will mistakenly think it's not uploaded, but in fact it was uploaded but got moved to the `scanned` bucket. On the file scan handler side, we'll first try to lock before moving the file.

PR #626, opened 2021-02-05
[DDP-3709] part 5: file scan result handler

yufengwng:
## Context

Part of DDP-3709. This PR implements a pubsub subscriber that listens for file scan results from the file scanner service. It'll manage moving the file upload's underlying bucket file to the appropriate destination bucket and updating status of file upload.

## FUD Score

## Testing

## Release

* Need to create a pubsub subscription to the `cf-file-scan-result` topic.
  * Add filter `attributes.bucketId="ddp-dev-file-uploads"`, use retry backoff, and set a long ack timeout.
* Create the scanned/quarantine buckets.
* Add new configs into Vault.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
@@ -9,17 +9,23 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
+import com.google.api.core.ApiService;
+import com.google.api.gax.core.ExecutorProvider;
 import com.google.api.gax.core.InstantiatingExecutorProvider;
 import com.google.api.gax.rpc.ApiException;
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Publisher;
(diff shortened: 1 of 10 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -505,6 +519,83 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
+        Subscriber newSubscriber = builder.build();
+
+        // Add handler that recursively calls back in here to renew subscriber.
+        newSubscriber.addListener(new ApiService.Listener() {
+            @Override
+            public void failed(ApiService.State from, Throwable failure) {
+                LOG.error("Subscriber to subscription {} encountered unrecoverable failure from {}"
+                        + ", rebuilding subscriber", subscription, from, failure);
+                if (!executorProvider.getExecutor().isShutdown()) {
+                    startNewSubscriberWithRecovery(subscription, builder, executorProvider, callbackExecutor, consumer);
```
yufengwng, line 584: @zyme @evgeniipr I hope I'm setting up this recovery correctly... Advice here will be appreciated!

evgeniipr: May be to add this line before `throw new DDPException`:
```
} catch (TimeoutException timeoutException) {
      subscriber.stopAsync();
}
```

evgeniipr: Looks like in this case will be called a signature `void error(String var1, Object... var2);`
```
LOG.error("Subscriber to subscription {} encountered unrecoverable failure from {}"
                        + ", rebuilding subscriber", subscription, from, failure);
```

yufengwng: @evgeniipr I'm curious why we need to call `stopAsync()`. If subscriber wasn't running already after waiting for the timeout, then do we need to call stop? It might be a good extra safeguard but just wasn't sure if we need it.

evgeniipr: I'm not sure if it really needs to call (may be really extra safeguard?). At least in this example it's called: https://cloud.google.com/pubsub/docs/samples/pubsub-subscriber-error-listener

yufengwng: In that code sample, they're demonstrating letting the subscriber run for 30 seconds and then shutting it down. The way they achieve the shutdown is using `awaitTerminated()` as a mechanism to throw a `TimeoutException` after 30 seconds, and then calling `stopAsync()` on timeout. (Note that subscriber won't be terminated since subscriber hasn't been stopped inside the `try` block, so that's why the timeout happens!) I'm not sure that has anything to do with starting up the subscriber though...

PR #633, opened 2021-02-10
atcp: refactor study contact cloud function

yufengwng:
## Context

This PR refactors the public Cloud Function for the A-T study's DAR feature. This used to be called `cf-study-contact`. I renamed it to `cf-atcp-contact` since this is specific to A-T study. See the README for more details. This is already deployed to `dev` environment so you can check it now.

## FUD Score

## Testing

I was unable to thoroughly test this since I was not able to get pass the recaptcha step (I don't know how to generate the client response!). I did deploy a custom version of the function that turned off recaptcha and I was able to verify everything else works, including permissions.

## Release

See `Setup and deployment` section in the README.

## Cloud Resources

This implementation requires certain paid resources:

* 1 secret in Secret Manager
* 1 service account
* 1 Google Bucket
* 1 Cloud Function


cf-atcp-contact/config/example.env
```diff
@@ -0,0 +1,20 @@
+# This is an example of the env vars needed by the function.
```
yufengwng, line 1: I switched to using env vars instead of a config file. Some advantages:

* Smaller size. No need to include the `typesafe config` dependency.
* Should be simple to use (just source the file in your shell).
* Works nicely with `gcloud functions deploy`.
* You can see the env vars in gcloud console website, so everything about a function is in one place.

cf-atcp-contact/deploy.sh
```diff
@@ -0,0 +1,32 @@
...
+  --entry-point org.broadinstitute.ddp.cf.ATCPStudyContact \
+  --trigger-http \
+  --allow-unauthenticated \
+  --service-account $SA_ACCT \
+  --env-vars-file env.yaml \
+  --region us-central1 \
+  --runtime java11 \
+  --memory 256MB \
+  --timeout 1m \
+  --max-instances 3
```
yufengwng, line 32: Since this is a publicly-facing HTTP endpoint, we should put limits on stuff. Looking at charts, I think these numbers should be good enough.

Screen Shot 2021-02-09 at 9 10 01 PM

Screen Shot 2021-02-09 at 9 10 16 PM

cf-atcp-contact/src/main/java/org/broadinstitute/ddp/cf/ATCPStudyContact.java
```diff
@@ -0,0 +1,172 @@
...
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.storage.HttpMethod;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import org.broadinstitute.ddp.cf.json.FileUploadUrl;
+import org.broadinstitute.ddp.cf.json.StudyContactRequest;
+
+public class ATCPStudyContact implements HttpFunction {
```
yufengwng, line 27: @YuriyPavlovBI I have refactored the implementation here, please take a close look. Some things I did:

* Refactored the recaptcha code out to a separate class to keep things clean. The `RecaptchaClient` is derived from work that @MocanaAtBroad had done in Pepper.
* Added more logging.
* Added check for `POST` HTTP method.
* Threw exceptions in appropriate places so they get picked up by GCP Error Reporting.
* Improved on request payload checking.

cf-atcp-contact/deploy.sh
YuriyPavlovBI: Sounds reasonable.

cf-atcp-contact/src/main/java/org/broadinstitute/ddp/cf/ATCPStudyContact.java
YuriyPavlovBI: LGTM

YuriyPavlovBI (approved):

PR #635, opened 2021-02-10
atcp: refactor emailer cloud function

yufengwng:
## Context

Second part of refactoring A-T study's Cloud Functions for DAR feature (see PR #633 for the first part). This PR updates the emailer to check file scan result.

## FUD Score

## Testing

This is running on `dev` environment already so you may give it a try. Note that currently the "to" email is pointing to an email `ddp_atcp_dar_emailer_dev@mailinator.com`. @YuriyPavlovBI introduced me to Mailinator, which is a free service where you can get an email inbox for testing email features. However, the email inbox is publicly-accessible and doesn't support viewing attachments in the free-tier, but at least there's something.

## Release

See notes in README.

## Cloud Resources

This implementation requires certain paid resources:

* 1 secret in Secret Manager
* 1 service account
* 1 pubsub subscription
* 1 Cloud Function


cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
```diff
@@ -0,0 +1,191 @@
...
+import com.sendgrid.Attachments;
+import com.sendgrid.Email;
+import com.sendgrid.Mail;
+import com.sendgrid.Method;
+import com.sendgrid.Personalization;
+import com.sendgrid.Request;
+import com.sendgrid.Response;
+import com.sendgrid.SendGrid;
+
+public class ATCPContactEmailer implements BackgroundFunction<ATCPContactEmailer.FileScanResult> {
```
yufengwng, line 26: @YuriyPavlovBI Once again, I made several changes:

* Function now listens for file scan pubsub result instead of bucket notification.
* Use env vars instead of config file.
* More logging and error handling.

cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
```diff
@@ -0,0 +1,191 @@
...
+        try {
+            result = ScanResult.valueOf(scanResult);
+        } catch (Exception e) {
+            throw new RuntimeException("Unexpected file scan result: " + scanResult, e);
+        }
+
+        if (result == ScanResult.INFECTED) {
+            throw new RuntimeException(String.format(
+                    "File is infected! Will not send ATCP DAR email! (bucket=%s file=%s)", bucketName, filename));
+        }
```
yufengwng: This is the most significant change -- we're now inspecting the file scan result! I don't know what we want to do if we get an infected file. For now, the function will halt and yell loudly.

cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
YuriyPavlovBI: How do we know if a file was uploaded for DAR form and not as an answer to the file upload question?

cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
YuriyPavlovBI: I think it's fine, there's nothing we can do in this case. Maybe we should just remove the file?

cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
yufengwng: DAR form uses a separate bucket. The bucket name is passed in as an env var and we match on that when processing the result. We also have that `FILENAME_PATTERN` regex.

cf-atcp-emailer/src/main/java/org/broadinstitute/ddp/cf/ATCPContactEmailer.java
yufengwng: Yeah, you know what, I think we should remove the file but keep the metadata file, so we have some trace of what it is.

yufengwng: Done! 2dbe19f097b841736422308d06ccf6d858527402

YuriyPavlovBI (approved):

PR #640, opened 2021-02-12
DDP-5235 - fix exceptions processing: replace ERROR level with WARN

evgeniipr:
## Context

Fix exceptions processing: replace ERROR level with WARN level for cases when errors came from auth0 or from data validation. This is needs in order to prevent excessive appending of error messages to Slack alerts channels. We don't need to publish messages as ERROR which related to auth0 checking failures or data validation failures.

## Checklist

## FUD Score

## Testing

## Release


evgeniipr:
@zyme, @yufengwng , this is a fix for pussub api (profile update): it replaces log level from ERROR to WARN for cases when error caused by auth0 user checks, invalid parameters in pubsub message, or data validation errors.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateEmailHandler.java
```diff
@@ -30,7 +32,7 @@ private void updateEmail(Handle handle, String userGuid, Properties payload) {
             String email = payload.getProperty(FIELD_EMAIL);
             var userDto = handle.attach(JdbiUser.class).findByUserGuid(userGuid);
             if (userDto == null) {
-                throw new PubSubTaskException("User profile is not found for guid=" + userGuid);
+                throw new PubSubTaskException("User profile is not found for guid=" + userGuid, WARN);
```
yufengwng, line 35: Curious question: why couldn't we just log the message right here instead of passing it back up through an exception?

evgeniipr: I had an idea to move to common class that switch() where analyzed result from auth0.. And in that code exceptions are thrown (I took it from UpdateUserEmailRoute.java and do not want to change it's logic). But I agree that better to replace with warn log messages as you advice since Zim suggested not to parse the auth0 result but just return it's code.. I will change it..

evgeniipr: Ah, no. There is yet another reason to throw exceptions: I need to catch it in one place (in PubSubTaskReceiver.java) and send to receive topic the error message..  So I'd better to stay with this approach (passing WARN/ERROR type in exeption) - in order to log the appropriate log level inside catch method.
Otherwise I need to do something like this in every place:
```
String msg = "Error message...";
LOG.warn(msg);
throw new PubSubTaskException(msg);
```

yufengwng (approved):

PR #641, opened 2021-02-12
DDP-5573 2/3: support child activities and child instances

yufengwng:
## Context

Part of DDP-5573. This PR implements:

* study-builder changes to specify parent/child nested activities.
* DAO and database changes for creating nested activities.
* Support new `createOnParentCreation` activity configuration property.
* Updated `ACTIVITY_INSTANCE_CREATION` action to require `ACTIVITY_STATUS` for creating child activities.
* Updated "list activity instance summaries" route to filter out child instances.
* Ensured that get/patch/put routes works with child instances (see updated test suite).

NOT IMPLEMENTED:

* New status logic for parent instances that takes child instances into account. This is tackled in a separate PR.

Part 1: #644
Part 3: #648 

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/FormActivityDao.java
```diff
@@ -52,24 +57,72 @@
...
+     * activities. Changing content should be done with the appropriate add/disable methods.
      *
-     * @param activity   the activity to create, without generated things like ids
-     * @param revisionId the revision to use, will be shared for all created data
+     * @param activity         the activity to create, without generated things like ids
+     * @param nestedActivities the list of nested activities to created, that the activity references.
+     * @param revisionId       the revision to use, will be shared for all created data
      */
-    default void insertActivity(FormActivityDef activity, long revisionId) {
+    default void insertActivity(FormActivityDef activity, List<FormActivityDef> nestedActivities, long revisionId) {
```
yufengwng, line 79: I have the nested activities as a separate list here because if I'm understanding correctly, we want parent/child to be more-or-less standalone activities, and just have them "reference" each other. E.g. child references parent with just a `parentActivityCode` property.

However, I've pretty much made it a requirement that nested activities are created alongside the parent (so you have to pass the child activities along in this list). This is because I think we have a "chicken-and-egg" problem. Parent and child activities reference each other, so we need to create them within the same "transaction".

pepper-apis/src/main/resources/db-changes/schema/DDP-5573-activity-nesting.xml
```diff
@@ -0,0 +1,28 @@
...
+            <column name="nested_activity_id" type="bigint">
+                <constraints nullable="false"
+                             references="study_activity(study_activity_id)"
+                             foreignKeyName="activity_nesting_nested_activity_fk"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint tableName="activity_nesting"
+                             columnNames="parent_activity_id, nested_activity_id"
+                             constraintName="activity_nesting_parent_nested_uk"/>
+    </changeSet>
```
MocanaAtBroad, line 28: Did you consider just adding a parent_activity_id column to study_activity?

MocanaAtBroad (commented):
Not seeing handling of child activity instances (i.e., where we link an child instance to a parent instance). Will it be part of this P.R.?

yufengwng:
@MocanaAtBroad 
> where we link an child instance to a parent instance

The linking is done in `FormActivityDao`. After we finish creating both the parent and nested activities, we add a link from parent to child.

> Not seeing handling of child activity instances

I don't think proper handling of child activity instances is necessary for this PR -- again, this PR is meant to provide enough to unblock other tickets. For example, with this PR merged, we can start working on new `ActivityBlock` type. The other requirements of handling child activities will come in separate PR.

pepper-apis/src/main/resources/db-changes/schema/DDP-5573-activity-nesting.xml
yufengwng: I did. However, I chose to go this route because:

1. It's more consistent with how we do "nesting" in other aspects. See tables for "nested blocks" and "composite child questions".
2. It gives us room to grow. For example, composite question nesting table has `display_order` attribute. We don't need a "display order" for nested activities, but should we need to add more properties to describe parent/child activities, we have a place to do this with this `activity_nesting` table.

MocanaAtBroad: I can't say I see the need for room for growth. Order does not appear to be relevant as we getting that from the containing blocks (sort of). The table does facilitate a many-to-many relationship, but the constraint added explicitly precludes it.
Not going to push the issue here. If you feel this is more consistent with the rest of the code and schema I will accept your judgement.

MocanaAtBroad:
> @MocanaAtBroad
> 
> > where we link an child instance to a parent instance
> 
> The linking is done in `FormActivityDao`. After we finish creating both the parent and nested activities, we add a link from parent to child.
> 
>>> I believe `FormActivityDao` does not concern itself with actvity instances, only instances correct?

> @MocanaAtBroad
> 
> > where we link an child instance to a parent instance
> 
> The linking is done in `FormActivityDao`. After we finish creating both the parent and nested activities, we add a link from parent to child.
> 
> > Not seeing handling of child activity instances
> 
> I don't think proper handling of child activity instances is necessary for this PR -- again, this PR is meant to provide enough to unblock other tickets. For example, with this PR merged, we can start working on new `ActivityBlock` type. The other requirements of handling child activities will come in separate PR.
OK.

yufengwng:
@MocanaAtBroad I couldn't figure out what your last comment was about, but I think your question is this?
> I believe FormActivityDao does not concern itself with actvity instances, only instances correct?

My answer: `FormActivityDao` is concerned with activity definitions and not instances.

yufengwng:
~I just realized that we also need to establish the parent/child relationship for _activity INSTANCES_, which I did not implement yet. I'm addressing that in the upcoming PR I'm working on. Other than that, this PR is ready-to-go.~

pepper-apis/src/main/resources/db-changes/schema/DDP-5573-activity-nesting.xml
yufengwng: After working with some real sql queries that deals with parent/child activities, I changed my mind and went the simpler solution of just adding new column. This simplifies the queries, and we can always pull it out into a new join table if needed.

yufengwng:
Sorry for the confusion! I have implemented support for both child activities and child instances, and rolled it into this one PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
```diff
@@ -68,8 +67,9 @@ public Object handle(Request request, Response response, ActivityInstanceCreatio
                 activityInstanceDao.bulkUpdateIsHiddenByActivityIds(participantId, true, Set.of(studyActivityId));
             }
 
+            Long parentInstanceId = null;
             String instanceGuid = activityInstanceDao
-                    .insertInstance(studyActivityId, operatorGuid, participantGuid, InstanceStatusType.CREATED, null)
+                    .insertInstance(studyActivityId, operatorGuid, participantGuid, parentInstanceId)
```
MocanaAtBroad, line 72: `parentInstanceId` is always null. I am assuming handling children is part of DDP-5574/separate PR. Yes?

MocanaAtBroad (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
yufengwng: Yup, that's right!

PR #642, opened 2021-02-15
DDP-5235 - fix PubSub task result generation

evgeniipr:
## Context

In this PR fixed some logical problems, improved code structure, written comments for most unclear parts of code.
The following improvements and fixes are done:
1) **PubSubTaskResult and it's creation code** is refactored: now it is possible to define any set of attributes and payload for certain task types - it is possible by overriding PubSubTaskProcessorAbstract.addCustomDataToResult(). By default during PubSubTaskResult creation it is copied all attributes from PubSubTask, added attr "taskMessageId" (with task message ID), payload {'resultType':'ERROR/SUCCESS', 'errorMessage':'..error msg (in case of ERROR)..'}.
2) Prevent logging of errors which can be logged with WARN level (so it makes less verbose error log).
3) Limit number of retries (in case of retriable errors) up to 4 (use AtomicLongMap to store counters for each message Id).
4) Adjust format of result payload: don't add 'errorMessage' in case of resultType=SUCCESS.
5) Create more clear comments.
6) Fix log messages.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskReceiver.java
```diff
@@ -31,6 +33,10 @@
 
     private static final Logger LOG = getLogger(PubSubTaskReceiver.class);
 
+    private static final int RETRY_MESSAGE_MAX_COUNT = 5;
```
yufengwng: Pubsub supports limiting retries by tweaking the settings of the subscription. Have we considered that?

evgeniipr: I tried to find some standard approach which could be good but without any luck..
What happens when limit (configured in subscription) is gained?
Some kind of exception thrown?

evgeniipr: Ah, I understood: it  is published to a special topic...

yufengwng: Per Slack conversation, we'll want to send back an error result to DSM when retry is exceeded, so pubsub's native retry support of dead-lettering topic won't help us here. So I'm okay with this implementation of dealing with retries.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskReceiver.java
```diff
@@ -76,18 +74,63 @@ private PubSubTask parseMessage(PubsubMessage message) {
...
+        if (e instanceof PubSubTaskException && ((PubSubTaskException)e).isShouldRetry()) {
+            handleRetriableErrors(consumer, pubSubTask, e);
+        } else {
+            handleNonRetriableErrors(consumer, pubSubTask, e);
+        }
+    }
+
+    private void handleRetriableErrors(AckReplyConsumer consumer, PubSubTask pubSubTask, Exception e) {
+        long count = retryMessageCounters.incrementAndGet(pubSubTask.getMessageId());
+        if (count <= RETRY_MESSAGE_MAX_COUNT) {
```
yufengwng: Shouldn't this be `<` (less than)? The `count` variable (and the `retryMessageCounters`, for that matter) is really tracking the _number of attempts_ instead of number of retries, right? For example, the first time we see a message and fails, `count` will be 1. So when `count` is 5, it means we have tried 5 attempts already, so at this point we should stop. This means when `count == RETRY_MESSAGE_MAX_COUNT` we should call `handleNonRetrieableErrors()` instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/util/GsonCreateUtil.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class GsonCreateUtil {
+
+    public static String createJsonIgnoreNulls(String name1, Object value1) {
+        Map<String, Object> map = new HashMap<>();
+        map.put(name1, value1);
+        return  GsonUtil.standardGson().toJson(excludeNullValues(map));
+    }
+
+    public static String createJsonIgnoreNulls(String name1, Object value1, String name2, Object value2) {
```
yufengwng, line 12: The `Map.of()` static method lets you create maps with arbitrary number of pairs, as long as you pass in even number of arguments. E.g.
```
Map.of(key1, val1)
Map.of(key1, val1, key2, val2)
```
I recommend just having one version of `createJsonIgnoreNulls()` method and have that take in a Map as parameter. Callers can use `Map.of()` to create the map they need.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskReceiver.java
evgeniipr: I thought that it limits number of retries. But may be you are right - it needs to count num of total attempts.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/GsonCreateUtil.java
evgeniipr: The problem with Map.of() that it not allows pairs with null values - it throws NPE. I also thought to use Map.of initially.
But I want to exclude fields with null values in result payload: for example, to exclude errorMessage=null in case if resultType='SUCCESS' (so that the returned payload will be {'resultType': 'SUCCESS'}.

yufengwng: Ah I see. Didn't know about issue with nulls!

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskReceiver.java
yufengwng: Wait, maybe I have it wrong here. If we're saying we allow 5 retries, that means total number of attempts will be 6. In that case, the logic here looks correct. My mistake!

evgeniipr: Fixed

yufengwng (approved):
Looks fine to me.

PR #652, opened 2021-02-22
DDP-5574 update create instance endpoint to require parent instance

yufengwng:
## Context

See DDP-5574. This PR updates the [CreateActivityInstance](https://pepper-dev.datadonationplatform.org/docs#operation/createActivity) API endpoint to check for parent instance when creating child instance.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/endpoints/user.studies.activities.yml
+++ b/pepper-apis/docs/specification/src/endpoints/user.studies.activities.yml
@@ -31,7 +31,16 @@ post:
   tags:
     - Activities
   summary: create a new instance of an activity
-  description: Given an `activityCode`, the server will create a new instance of the activity for the specified user if the configured conditions for the activity is met.
+  description: |
+    Given an `activityCode`, the server will create a new instance of the activity
+    for the specified user if the configured conditions for the activity is met.
+    When creating instances of child nested activities, request must provide the
+    corresponding parent activity instance.
+
+    **Study Admins**
+
+    Creating child instance is not supported if parent instance is hidden or
+    read-only, unless operator is a study admin.
   requestBody:
     required: true
     content:
@@ -47,7 +56,7 @@ post:
               example: ABCDEFGH12
             parentInstanceGuid:
               description: |
-                in the case the activity instance being created is a child activity, need to specify the parent instance with their guid.
+                In the case the activity instance being created is a child activity, need to specify the parent instance guid.
               type: string
               example: ABCDEFGH12
   responses:
@@ -76,10 +85,10 @@ post:
             schema:
               $ref: '../pepper.yml#/components/schemas/Error.UserStudyActivityNotFound'
       422:
-        description: activity instance failed to be created
+        description: activity instance failed to be created, e.g. too many instances or parent is read-only
         content:
           application/json:
             schema:
-              $ref: '../pepper.yml#/components/schemas/Error.TooManyInstances'
+              $ref: '../pepper.yml#/components/schemas/Error.CreateActivityInstance'
       default:
         $ref: '../pepper.yml#/components/responses/ErrorResponse'
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
@@ -12,15 +12,18 @@ import org.broadinstitute.ddp.db.dao.ActivityInstanceDao;
 import org.broadinstitute.ddp.db.dao.DataExportDao;
 import org.broadinstitute.ddp.db.dto.ActivityInstanceCreationValidation;
 import org.broadinstitute.ddp.db.dto.ActivityInstanceDto;
+import org.broadinstitute.ddp.db.dto.StudyDto;
 import org.broadinstitute.ddp.json.ActivityInstanceCreationPayload;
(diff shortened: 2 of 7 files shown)


pepper-apis/docs/specification/src/endpoints/user.studies.activities.yml
```diff
@@ -31,7 +31,16 @@ post:
   tags:
     - Activities
   summary: create a new instance of an activity
-  description: Given an `activityCode`, the server will create a new instance of the activity for the specified user if the configured conditions for the activity is met.
```
yufengwng, line 34: @alextanlan @bskinner I've updated documentation descriptions to better describe parent/child dependency when creating child instances, particularly about read-only behavior.

bskinner: Thank you for the heads up

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
```diff
@@ -52,22 +64,54 @@ public Object handle(Request request, Response response, ActivityInstanceCreatio
...
+            }
+
+            Long parentInstanceId = null;
+            if (validation.getParentActivityCode() != null) {
+                if (StringUtils.isBlank(parentInstanceGuid)) {
+                    String msg = "Creating child nested activity instance requires parent instance guid";
+                    LOG.warn(msg);
+                    throw ResponseUtil.haltError(response, HttpStatus.SC_BAD_REQUEST,
+                            new ApiError(ErrorCodes.BAD_PAYLOAD, msg));
+                }
```
evgeniipr: This logging/exception throwing lines could be extracted to method (repeated 3 times in this class):
```
private void haltError(int status, String code, String msg) {
        LOG.warn(msg);
        throw ResponseUtil.haltError(status, new ApiError(code, msg));
}       
```

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
```diff
@@ -52,22 +64,54 @@ public Object handle(Request request, Response response, ActivityInstanceCreatio
                 var err = new ApiError(ErrorCodes.ACTIVITY_NOT_FOUND,
                         "Could not find creation validation information for activity " + activityCode);
                 LOG.warn(err.getMessage());
-                throw ResponseUtil.haltError(response, 404, err);
+                throw ResponseUtil.haltError(response, HttpStatus.SC_NOT_FOUND, err);
+            }
+
+            Long parentInstanceId = null;
+            if (validation.getParentActivityCode() != null) {
```
evgeniipr, line 73: I'd suggest to extract code under this `if` to method `validateParentActivity(...)`.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceDao.sql.stg
```diff
@@ -2,6 +2,11 @@ group ActivityInstanceDao;
 
 queryActivityInstanceCreationValidation() ::= <<
 select act.study_activity_id,
+       act.parent_activity_id,
+       (select study_activity_code
+          from study_activity
+         where study_activity_id = act.parent_activity_id
+       ) as parent_activity_code,
```
evgeniipr, line 9: Not sure about performance (needs to compare execution plans) but seems it's possible to replace nested select with
left join study_activity sa on sa.study_activity_id = act.parent_activity_id.
In this post there are opinions that `left join` is better by performance that `subquery`: https://stackoverflow.com/questions/2577174/join-vs-sub-query
For example, one guy said:
`As per my observation like two cases, if a table has less then 100,000 records then the join will work fast.
But in the case that a table has more than 100,000 records then a subquery is best result.`

MocanaAtBroad (dismissed):

evgeniipr (dismissed):

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceDao.sql.stg
yufengwng: Most activities will not have a parent activity, so most of the time the subquery will be querying with `null` as `act.parent_activity_id`. I'm hoping the DB can see that it's null and optimize out the subquery. I took a brief look of the query plan as I was writing this query, and I didn't see any immediate red flags, so I think I'll leave this query as-is for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
yufengwng: All set! 6238d092369add6c34bf28cd07d5aff6ce0d9be3

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
yufengwng: Done! 6238d092369add6c34bf28cd07d5aff6ce0d9be3

MocanaAtBroad (approved):

PR #654, opened 2021-02-23
DDP-5576 include canDelete in instance/summary API responses

yufengwng:
## Context

See DDP-5576. This PR adds the new `canDelete` property to the "instance summary" and "full instance" JSON response objects returned by the activity API endpoints.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -115,6 +115,7 @@ Activity:
     - sections
     - lastUpdatedText
     - lastUpdated
+    - canDelete
     - isFollowup
     - isHidden
     - sectionIndex
@@ -177,6 +178,9 @@ Activity:
       format: date-time
       nullable: true
       description: The date time the activity definition (NOT the instance) was last updated
+    canDelete:
+      type: boolean
+      description: whether the activity instance can be deleted
     isFollowup:
       type: boolean
       description: whether this activity is considered a "follow-up"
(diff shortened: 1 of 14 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -115,6 +115,7 @@ Activity:
     - sections
     - lastUpdatedText
     - lastUpdated
+    - canDelete
```
yufengwng, line 118: @alextanlan @bskinner I noticed the documentation was missing the new `canDelete` property on the full `Activity` object, so I added it.

evgeniipr (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/ActivityInstance.java
```diff
@@ -146,4 +157,12 @@ public boolean isInstanceHidden() {
     public boolean isHidden() {
         return isHidden;
     }
+
+    public String getParentInstanceGuid() {
+        return parentInstanceGuid;
+    }
+
+    public void setParentInstanceGuid(String parentInstanceGuid) {
+        this.parentInstanceGuid = parentInstanceGuid;
```
MocanaAtBroad: Do we have to worry that this is only property in object whose value is not set in constructor? Would be more consistent if if treated like others.

(In ActivityInstanceSummary is almost same deal, but looks to me like code for setters here was added before?)

evgeniipr (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/ActivityInstance.java
yufengwng: I think this was from a merge conflict. I have resolved conflicts and don't see this change in this PR.

MocanaAtBroad (approved):

PR #656, opened 2021-02-23
DDP-5573 fix creation event handling of child activities

yufengwng:
## Context

Noticed a bug where if we use the `ACTIVITY_INSTANCE_CREATION` event, activities will get created as "child instances". Bug was introduced as part of DDP-5573. This PR fixes handling of child activities in that event.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/ActivityInstanceCreationEventAction.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/ActivityInstanceCreationEventAction.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.ddp.model.event;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -101,7 +102,7 @@ public class ActivityInstanceCreationEventAction extends EventAction {
         }
 
         Long parentInstanceId = null;
-        if (signal.getEventTriggerType() == EventTriggerType.ACTIVITY_STATUS) {
+        if (activityDto.getParentActivityId() != null && signal.getEventTriggerType() == EventTriggerType.ACTIVITY_STATUS) {
             parentInstanceId = ((ActivityInstanceStatusChangeSignal) signal).getActivityInstanceIdThatChanged();
         }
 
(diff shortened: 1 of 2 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/ActivityInstanceCreationEventAction.java
```diff
@@ -101,7 +102,7 @@ public void doAction(Handle handle, EventSignal signal) {
         }
 
         Long parentInstanceId = null;
-        if (signal.getEventTriggerType() == EventTriggerType.ACTIVITY_STATUS) {
+        if (activityDto.getParentActivityId() != null && signal.getEventTriggerType() == EventTriggerType.ACTIVITY_STATUS) {
```
yufengwng, line 105: I was missing the check for parent activity. Without this check, we're always setting the `parentInstanceId`, which means any instances we create from this event will be effectively a "child" instance and won't show up on the dashboard.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/ActivityInstanceCreationEventAction.java
```diff
@@ -167,6 +168,41 @@ public void doAction(Handle handle, EventSignal signal) {
...
+            EventService.getInstance().processAllActionsForEventSignal(handle, new ActivityInstanceStatusChangeSignal(
+                    signal.getOperatorId(),
+                    signal.getParticipantId(),
+                    signal.getParticipantGuid(),
+                    signal.getOperatorGuid(),
+                    newChildInstanceId,
+                    childActivityId,
+                    signal.getStudyId(),
+                    InstanceStatusType.CREATED));
+        }
```
MocanaAtBroad, line 205: I could have **sworn** that I had seen the child instance creation logic in your original PRs!

MocanaAtBroad (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/ActivityInstanceCreationEventAction.java
yufengwng: I thought I had covered it too. But realized that the creation event uses "lower-level" facilities to create instances so it missed out on handling child activities.

PR #657, opened 2021-02-25
DDP-5402, DDP-5657 - use JdbiQuestionCached (using Redis-cache) 

evgeniipr:
## Context

Use **JdbiQuestionCached** (using Redis-cache) instead of JdbiQuestion  when reading Def-data from DB.
This class unitilized also in **GetActivityInstanceRoute**.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/SectionBlockDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/SectionBlockDao.java
@@ -7,13 +7,14 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+
 import org.broadinstitute.ddp.db.dao.ComponentDao;
 import org.broadinstitute.ddp.db.dao.JdbiBlock;
 import org.broadinstitute.ddp.db.dao.JdbiBlockContent;
 import org.broadinstitute.ddp.db.dao.JdbiBlockGroupHeader;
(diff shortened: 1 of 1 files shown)


yufengwng (commented):
Nice! Do you have any measurements of response times to give us a sense of how much this improves the performance of the GET route?

evgeniipr:
> Nice! Do you have any measurements of response times to give us a sense of how much this improves the performance of the GET route?

Hi @yufengwng ! Yes, I did measurements in the code (using StopWatch class): I added start/stop lines before/after that part of GetActivityInstanceRoute code which calls SectionBlockDao.
I made 7 runs with cached version (QuestionCachedDao) and 7 runs with non-cached (QuestionDao). Before each run I restarted backend and restarted MySQL (I tested with local server). In average I got:
- with cached version ~ 0.554,857 sec
- with non-cached version ~ 0.690,143 sec.

So, the cached version gives quite noticeable performance boost even with local version of MySQL (and needs to add that there are quite small number of rows in tables in the local DB). So, with dev and prod the boost should be much more noticeable.
I thought about writing a performance checking unit test but then decided that it will not be useful: it not possible to realistically imitate DB SQL query delays in unit test..

yufengwng (approved):
Thanks! Maybe we need to get this up to `dev` environment where we have more data and real Redis running to get a better sense of performance. Let's get this merged!

MocanaAtBroad (approved):

PR #659, opened 2021-02-25
DDP-5659 fix model notification in JdbiUmbrellaStudyCached

yufengwng:
## Context

See DDP-5659. Disclaimer: I don't actually know what I'm doing in this PR. I'm not too familiar with how the "model notification" works for the cache, but the code looks wrong so I'm trying to fix it. I discovered this while working on DDP-5583, where I had to add new sql statements to `JdbiUmbrellaStudy` and thus `JdbiUmbrellaStudyCached`. Please advice.

## Checklist

## FUD Score

## Testing

I did NOT test this.

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyCached.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyCached.java
@@ -139,7 +139,7 @@ public class JdbiUmbrellaStudyCached extends SQLObjectWrapper<JdbiUmbrellaStudy>
         if (val > 0) {
             StudyDto dto = findByStudyGuid(guid);
             if (dto != null) {
-                notifyModelUpdated(ModelChangeType.UMBRELLA, dto.getId());
+                notifyModelUpdated(ModelChangeType.STUDY, dto.getId());
                 removeFromCache(dto);
             }
         }
@@ -164,7 +164,7 @@ public class JdbiUmbrellaStudyCached extends SQLObjectWrapper<JdbiUmbrellaStudy>
         if (val > 0) {
             StudyDto dto = findByStudyGuid(guid);
             if (dto != null) {
-                notifyModelUpdated(ModelChangeType.UMBRELLA, dto.getUmbrellaId());
+                notifyModelUpdated(ModelChangeType.STUDY, dto.getId());
                 removeFromCache(dto);
             }
         }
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyCached.java
```diff
@@ -139,7 +139,7 @@ public int updateShareLocationForStudy(boolean shareLocation, String guid) {
         if (val > 0) {
             StudyDto dto = findByStudyGuid(guid);
             if (dto != null) {
-                notifyModelUpdated(ModelChangeType.UMBRELLA, dto.getId());
+                notifyModelUpdated(ModelChangeType.STUDY, dto.getId());
```
yufengwng, line 142: The "thing" that's changing here is the study. STUDY and UMBRELLA are totally different things. I don't know what "notify model updated" really means, but I think this should say STUDY instead.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyCached.java
```diff
@@ -190,7 +190,7 @@ public long insert(String studyName, String studyGuid, long umbrellaId, String w
                        Long precisionId, boolean shareLocation, String studyEmail, String recaptchaSiteKey) {
         long studyId = delegate.insert(studyName, studyGuid, umbrellaId, webBaseUrl, auth0TenantId, irbPassword, precisionId, shareLocation,
                 studyEmail, recaptchaSiteKey);
-        notifyModelUpdated(ModelChangeType.UMBRELLA, umbrellaId);
+        notifyModelUpdated(ModelChangeType.STUDY, studyId);
```
yufengwng, line 193: If we're inserting a new study, is the "model" that changed a STUDY or UMBRELLA? Again, not familiar with this.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiUmbrellaStudyCached.java
MocanaAtBroad: I think you are right to change it to STUDY.

Idea with the notifyModelUpdated is to alert other caches that a change has occurred that might have made their cached data invalid.

We had currently defined 3 scopes for model changes: user, study, and umbrella. As names suggest, they refer to data that is just user-specific, or study specific or for a whole umbrella of studies.
If our data was perfectly factored/normalized then each cache would have data that is not replicated anywhere else. But our data model has data that is replicated in multiple spots (e.g., activity instances have some data that is also in activities and activity summaries, etc).

So invention of the change scopes was let other caches that share that scope that they should invalidate themselves.

In this case, the a change to the study should only affect stuff in the study scope (despite having "umbrella" as part of this name which is probably why messed up here).

MocanaAtBroad (approved):

PR #664, opened 2021-03-03
DDP-5402 - fix question validation rules caching

evgeniipr:
## Context

Fix question validation rules caching.
Before the fix validation saved with the redis cache with not correct key (the key should be an activity ID whereas was used question ID). As a consequence the stored validation rules not were not found in the cache and the query repeated every time.
This fix improved the performance by ~ 10% with local DB. Needs to check how it works on dev.

The cause of the problem was that validations stored in Redis cache map with key=questionId.
And it causes multiple SQL query of validation rules for same activityId. Because quite many questions are without validation rules. Therefore when querying Redis cache it returns null and SQL query executed again..
Now the cache key = activityId and value Map of lists of rules (key=questionId).

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiQuestionValidationCached.java
```diff
@@ -57,15 +57,15 @@ public long insert(long questionId, long validationId) {
         } else {
             List<RuleDto> validations = null;
             try {
-                validations = questionIdToValidationsCache.get(questionDto.getId());
+                validations = questionIdToValidationsCache.get(questionDto.getActivityId());
```
yufengwng: Hmmm, I'm getting a bit confused -- the variable name is `questionIdToValidationsCache` but we're looking up by activity id?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiQuestionValidation.java
```diff
@@ -61,7 +61,7 @@
         Map<Long, List<RuleDto>> questionIdToRuleDtos = new HashMap<>();
         try (var stream = getAllActiveValidationDtosForActivity(activityId)) {
             stream.forEach(ruleDto -> questionIdToRuleDtos
-                    .computeIfAbsent(ruleDto.getQuestionId(), k -> new ArrayList<>())
+                    .computeIfAbsent(activityId, k -> new ArrayList<>())
```
yufengwng: I think the current code is correct. I think this `getAllActiveValidationsForActivity()` method is meant to find all `RuleDto`s for the given activity id, but grouped by question id. So the resulting Map is `(question_id -> list)`.

This Map is then added to the `questionIdToValidationsCache`, which will accumulate all these `RuleDto`s for all the questions since questions across activities have unique question ids.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ValidationDao.java
```diff
@@ -140,15 +140,9 @@ default void insertValidations(QuestionDef question, long revisionId) {
         I18nContentRenderer i18nContentRenderer = new I18nContentRenderer();
 
         for (RuleDto ruleDto : ruleDtos) {
-            String correctionHint = null;
-            if (ruleDto.getHintTemplateId() != null) {
-                correctionHint = i18nContentRenderer.renderContent(getHandle(), ruleDto.getHintTemplateId(), langCodeId);
+            if (ruleDto.getQuestionId() == questionDto.getId()) {
```
yufengwng: This `if` condition here means the list of `RuleDto`s might not belong to the question. I think this is a red flag that something is wrong here... `getAllActiveValidations(questionDto)` should only return rules for the question!

yufengwng (changes requested):
This seems like a nice find, but taking a closer look I got confused and not sure what the problem is...

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiQuestionValidation.java
evgeniipr: The problem was that for different questions the result was not found in the cache (in the cache the key should be activity ID). But it was searched in cache by question ID. 
Seems you are right, the code was correct but not quite correct. I'll refix it.

evgeniipr: @yufengwng I remembered what was a problem: it is not actually correct to store in Redis cache the question validations by a question ID. Because there are questions without validation. And if we query a cache by such question ID then it's validation are not found (which is quite correct - it not defined for this question) but the next step: the SQL query 'getAllActiveValidationsForActivity' is executed again.
The correct way - to store a list of question validations for whole activity with key = activity ID.
If you ok with it then I will try to fix the implementation (rename 'questionIdToValidationsCache' etc)..

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiQuestionValidationCached.java
evgeniipr: yes, agree. It needs to be renamed.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ValidationDao.java
evgeniipr: The idea was that it's flat list with all rules of all questions of whole activity.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiQuestionValidation.java
yufengwng: Okay, I see. So I'll say in this case the code is correct -- it was doing what it was designed for. But this design is not helping us because it results in cache misses for scenarios where question doesn't actually have validations. Going up a level and caching by activity id sounds reasonable to me. Go for it!

evgeniipr: @youfeng, I reimplemented the more proper way. Now the only class changed is JdbiQuestionValidationCached.java:
key = activityId, value = Map<Long, List<RuleDto>> (where key = questionId).
This solved the problem with repeated SQL query.

MocanaAtBroad:
I like the change! Not only reduces the number of missed cache queries, but also there only one query per activity instead of one query per question.

yufengwng (approved):
LGTM

PR #671, opened 2021-03-09
DDP-5045 fix sendgrid events verification

yufengwng:
## Context

While testing the sendgrid webhook feature in `dev` environment, I noticed we didn't turn on "Signed Event Webhooks". After turning it on and setting the appropriate `sendgrid.eventsVerificationKey` in the config file, I noticed there was an issue in the logs:

```
2021-03-09 17:49:38.092 EST POST 401 1.72 KiB 15 ms SendGrid Event API /pepper/v1/sendgrid-event
2021-03-09 17:49:38.105 EST 22:49:38.105 c7c0dcd6-5168-451b-a7f8-d1e4eb9ed7e1 167.89.117.70, 169.254.1.1 C: S: [qtp2061423649-157] WARN o.b.d.f.SendGridEventVerificationFilter Error during SendGrid event signature verification: java.security.NoSuchProviderException: no such provider: BC
```

After doing some digging, I found this [GitHub issue][issue] and [this code example][example], which shows how to add the security provider. I manually deployed this fix to `dev` and verified that it fixes the problem. While I'm in here, I also fixed a few things.

[issue]: https://github.com/sendgrid/sendgrid-java/issues/624
[example]: https://github.com/sendgrid/sendgrid-java/blob/main/examples/helpers/eventwebhook/Example.java

## Checklist

## FUD Score

- [x] Go through some study workflows that sends emails or trigger example email events from SendGrid.

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/SendGridEventVerificationFilter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/filter/SendGridEventVerificationFilter.java
@@ -10,12 +10,13 @@ import java.io.IOException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
+import java.security.Security;
 import java.security.SignatureException;
 import java.security.spec.InvalidKeySpecException;
 
-
 import com.sendgrid.helpers.eventwebhook.EventWebhook;
 import com.sendgrid.helpers.eventwebhook.EventWebhookHeader;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.broadinstitute.ddp.json.errors.ApiError;
 import org.broadinstitute.ddp.util.ResponseUtil;
 import org.slf4j.Logger;
@@ -43,6 +44,20 @@ public class SendGridEventVerificationFilter implements Filter {
 
     public SendGridEventVerificationFilter(String cfgParamSendGridEventsVerificationKey) {
         this.cfgParamSendGridEventsVerificationKey = cfgParamSendGridEventsVerificationKey;
+        registerSecurityProvider();
+    }
+
+    private void registerSecurityProvider() {
+        boolean alreadyRegistered = false;
+        for (var provider : Security.getProviders()) {
+            if (BouncyCastleProvider.PROVIDER_NAME.equals(provider.getName())) {
+                alreadyRegistered = true;
+                break;
+            }
+        }
+        if (!alreadyRegistered) {
+            Security.addProvider(new BouncyCastleProvider());
+        }
     }
 
     @Override
--- a/pepper-apis/src/main/resources/db-changes/schema/DDP-5045-sendgrid-event-message-id-nullable.xml
+++ b/pepper-apis/src/main/resources/db-changes/schema/DDP-5045-sendgrid-event-message-id-nullable.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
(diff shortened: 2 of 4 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/filter/SendGridEventVerificationFilter.java
```diff
@@ -43,6 +44,20 @@
...
+    private void registerSecurityProvider() {
+        boolean alreadyRegistered = false;
+        for (var provider : Security.getProviders()) {
+            if (BouncyCastleProvider.PROVIDER_NAME.equals(provider.getName())) {
+                alreadyRegistered = true;
+                break;
+            }
+        }
+        if (!alreadyRegistered) {
+            Security.addProvider(new BouncyCastleProvider());
```
yufengwng, line 59: We already have Bouncy Castle (BC) as a dependency, so no changes to POM files needed. I admit I don't know much about BC, so let me know if you have advice here!

pepper-apis/src/main/resources/db-changes/schema/DDP-5045-sendgrid-event-message-id-nullable.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="yufeng" id="20210309-sendgrid-event-sg-message-id-column-nullable">
+        <dropNotNullConstraint tableName="sendgrid_event" columnName="sg_message_id" columnDataType="varchar(100)"/>
```
yufengwng, line 9: Noticed an issue while looking through the logs:

Screen Shot 2021-03-09 at 6 34 06 PM

Looking through the [docs](https://sendgrid.com/docs/for-developers/tracking-events/event/#json-objects), it appears that `sg_message_id` can be null, so I'm dropping the null constraint here.
> sg_message_id - a unique, internal SendGrid ID for the message. The first half of this ID is pulled from the smtp-id. The message ID will be included in most cases. In the event of an asynchronous bounce, the message ID will not be available.

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/SendGridEventVerificationFilter.java
evgeniipr: I looked some articles about BC usage as a provider in production - I haven't found any problems described.
The only concern: once BC is set it this filter then it will be used in DDP backend in all subsequent code. But as it  supports standard algorithms then it should not be a problem?

pepper-apis/src/main/resources/db-changes/schema/DDP-5045-sendgrid-event-message-id-nullable.xml
evgeniipr: Thanks.
And I need to complete SendGrid doc yet..

evgeniipr (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/SendGridEventVerificationFilter.java
yufengwng: Interesting. I had not considered the fact that BC will be used for the whole backend once this is set. I don't have any insight into whether this is a good thing or if there's other side-effects. Unfortunately, I don't know of a way to restrict this security provider to _just_ this sendgrid filter, so I think I will have to proceed with this as-is.

yufengwng: I think the way it works is that each provider is registered under a unique name. For BouncyCastleProvider, its name is `BC`. And then whoever wants a provider, they get it by using its name. For example, SendGrid code for verification has something like this:

```
KeyFactory.getInstance("ECDSA", "BC");
```

Currently, none of our stuff in the backend uses the `BC` provider, since we know that the `BC` provider is not registered yet (otherwise I wouldn't have found this issue in the first place!). So I think adding the `BC` provider here is safe and won't affect any existing code. For future code that wants a `BC` provider, I think the BouncyCastleProvider implementation should suffice. So I think this is good to go.

PR #672, opened 2021-03-10
DDP-5675 Add First Name to User's profile

YuriyPavlovBI:


charlottedye:
This looks good to me but would you mind linking me to the ticket so I can double check?  Thanks

YuriyPavlovBI:
> This looks good to me but would you mind linking me to the ticket so I can double check? Thanks

https://broadinstitute.atlassian.net/browse/DDP-5675

charlottedye:
I think we should probably review this with Jeff that this is what we want.  I actually find the ticket description for this confusing (what does he mean by adding them to ElasticSearch specifically?) so let's make sure we are on the same page.

YuriyPavlovBI:
> I think we should probably review this with Jeff that this is what we want. I actually find the ticket description for this confusing (what does he mean by adding them to ElasticSearch specifically?) so let's make sure we are on the same page.

The problem was that firstName and lastName in profile were always null.

charlottedye (approved):

charlottedye:
The commit log is broken, though

YuriyPavlovBI:
> The commit log is broken, though

This was merged together with another PR

PR #673, opened 2021-03-10
Modify study builder to load labeled Event Configurations

MocanaAtBroad:
## Context

In the interest of avoiding creating java patch code to update configuration, we are adding a command line options to study builder to load specific event configuration entries.
Part of implementation involves adding a "label" property to Event Configurations.

## FUD Score

Can try to add one or more event configuration entries into a study.conf file and specify them in the command line.

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
@@ -12,6 +12,7 @@ import org.broadinstitute.ddp.db.dto.NotificationTemplateSubstitutionDto;
 import org.broadinstitute.ddp.db.dto.QueuedEventDto;
 import org.broadinstitute.ddp.db.dto.QueuedNotificationDto;
 import org.broadinstitute.ddp.db.dto.QueuedPdfGenerationDto;
+import org.broadinstitute.ddp.exception.DDPException;
 import org.broadinstitute.ddp.model.activity.types.EventActionType;
 import org.broadinstitute.ddp.model.activity.types.EventTriggerType;
 import org.broadinstitute.ddp.model.event.EventConfiguration;
@@ -37,6 +38,16 @@ public interface EventDao extends SqlObject {
         return getEventConfigurationDtosByStudyId(studyId).stream()
                 .map(dto -> new EventConfiguration(dto))
                 .collect(Collectors.toList());
+    }
+
+    default Optional<EventConfiguration> getAllEventConfigurationsByStudyIdAndLabel(long studyId, String label) {
+        if (label == null) {
+            throw new DDPException("label argument cannot be null");
+        }
+        return getEventConfigurationDtosByStudyId(studyId).stream()
+                .map(dto -> new EventConfiguration(dto))
+                .filter(cfg -> label.equals(cfg.getLabel()))
+                .findFirst();
     }
 
     default List<EventConfiguration> getAllEventConfigurationsByStudyIdAndTriggerType(long studyId,
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
@@ -40,7 +40,7 @@ public interface EventDao extends SqlObject {
                 .collect(Collectors.toList());
     }
 
-    default Optional<EventConfiguration> getAllEventConfigurationsByStudyIdAndLabel(long studyId, String label) {
+    default Optional<EventConfiguration> getEventConfigurationByStudyIdAndLabel(long studyId, String label) {
         if (label == null) {
             throw new DDPException("label argument cannot be null");
         }
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
@@ -45,8 +45,8 @@ public interface EventDao extends SqlObject {
             throw new DDPException("label argument cannot be null");
         }
         return getEventConfigurationDtosByStudyId(studyId).stream()
-                .map(dto -> new EventConfiguration(dto))
                 .filter(cfg -> label.equals(cfg.getLabel()))
+                .map(dto -> new EventConfiguration(dto))
                 .findFirst();
     }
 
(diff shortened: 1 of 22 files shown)


evgeniipr:
What is the purpose of the new field "label"? Just interesting to know.

MocanaAtBroad:
@evgeniipr 
> What is the purpose of the new field "label"? Just interesting to know.

Purpose is to identify within the study.conf file the specific event configurations to load into system via Study Builder. Before this code change, you have to either load the whole study (which we can't) or load all the events (which we can't either). 
For family history, we will be adding a few new event configurations (e.g., when consent is completed, create a family history activity instance). 
With this change we'll be able to add labels to the event configurations in the HOCON file and specify them from command line in StudyBuilder using the "--only-labeled-events" command line option.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
```diff
@@ -39,6 +40,16 @@
                 .collect(Collectors.toList());
     }
 
+    default Optional<EventConfiguration> getAllEventConfigurationsByStudyIdAndLabel(long studyId, String label) {
```
yufengwng: Method name of `getAllEventConfigurations...` might be confusing since we're only returning an `Optional`.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
```diff
@@ -39,6 +40,16 @@
...
+        if (label == null) {
+            throw new DDPException("label argument cannot be null");
+        }
+        return getEventConfigurationDtosByStudyId(studyId).stream()
+                .map(dto -> new EventConfiguration(dto))
+                .filter(cfg -> label.equals(cfg.getLabel()))
```
yufengwng: A minor optimization is to do the `filter()` before the `map()`, so we're just mapping over zero or one object instead of the whole list.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
```diff
@@ -150,6 +152,13 @@ private void run(String[] args) throws Exception {
             execute(builder::runEvents, isDryRun);
             log("done");
             return;
+        } else if (cmd.hasOption("only-labeled-events")) {
+            log("executing events setup...");
```
yufengwng: Maybe we can be a bit more specific?
```suggestion
            log("executing labeled events setup...");
```

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -79,9 +89,45 @@ private void insertEvents(Handle handle) {
...
+                .filter(eventCfg -> eventCfg.hasPath("label") && eventCfg.getString("label") != null)
+                .collect(toList());
+       List<String> allCfgLabels = allLabeledCfgs.stream().map(cfg -> cfg.getString("label")).collect(toList());
+       Set<String> notFoundEventLabels = eventLabelsToLoad.stream().filter(eventName -> !allCfgLabels.contains(eventName)).collect(toSet());
+       if (!notFoundEventLabels.isEmpty()) {
+           throw new DDPException("Could not find events " + String.join(", ", notFoundEventLabels));
+       }
+       if (allCfgLabels.size() > eventLabelsToLoad.size()) {
+           throw new DDPException("Found duplicate names in event configuration entries");
+       }
```
yufengwng: Hmmm I'm not too sure about this scenario. Not seeing how the size of the lists/sets will determine duplicates? Also, could it be possible that configuration files has a lot of labeled events (so `allCfgLabels` size is big) but on command line we only specify a few labels we want in that run (so `eventLabelsToLoad` size is small), therefore the sizes won't match up?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -79,9 +89,45 @@ private void insertEvents(Handle handle) {
...
+       List<Config> eventCfgsToLoad = allLabeledCfgs.stream()
+                   .filter(cfg -> allCfgLabels.contains(cfg.getString("label")))
```
yufengwng: Did we mean to check against the `eventLabelsToLoad` set?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -79,9 +89,45 @@ private void insertEvents(Handle handle) {
...
+                   .collect(toList());
+
+       List<String> existingCfgLabelsInDb = eventCfgsToLoad.stream()
+                .map(eventCfg -> handle.attach(EventDao.class)
+                        .getAllEventConfigurationsByStudyIdAndLabel(studyDto.getId(), eventCfg.getString("label")))
```
yufengwng: Okay, I see how this query is being used. In this case, how about we make the query take in a set of labels to query and return a list/stream of those that are found?

yufengwng (commented):
Have a few comments, but overall looks good.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
MocanaAtBroad: That was a dumb name. Renaming.
c72355d9

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventDao.java
MocanaAtBroad: Dude!
03120ac5

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
MocanaAtBroad: Thought of doing query, then decided this was easier to do. For seldom executed code iterating over a handful of items ought to be fine.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
MocanaAtBroad: Right. This was wrong.
dbea8bc3b184e2e39ed61bc53e39c4bef2aa58bb

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
MocanaAtBroad: Wrong too.
dbea8bc3b184e2e39ed61bc53e39c4bef2aa58bb

yufengwng (dismissed):
LGTM

yufengwng (approved):

PR #676, opened 2021-03-11
DDP-5587 support rendering answers into title/subtitle

yufengwng:
## Context

See DDP-5587. This PR adds support for rendering answer values into an activity instance (or summary) title and subtitle.

The syntax is straightforward:
```
$ddp.answer("<question_stable_id>", "<fallback_value>")
```

Example usage:
```
You have selected: $ddp.answer("SOME_PICKLIST_QUESTION", "no option selected yet")
```

Behavior: this feature will render the answer value for the given question. `COMPOSITE` and `FILE` questions are currently not supported, `PICKLIST` will render the option text rather than option stable id since the former is more common use-case, and the other answer types will render a string representation of the answer value. The main change is in `RenderValueProvider::answer()`.

We currently only support answer rendering for metadata properties like `title` and `subtitle`. For instance summaries, we additionally support answer rendering for `name` and `summary` (which already supports other kinds of substitutions). There are two main scenarios when we'll be doing answer rendering/substitutions:

1. Fetching activity instance summaries. See changes in `ActivityInstanceService::renderInstanceSummaries()` for this.
2. Fetching whole activity instance object. See changes in `FormInstance::renderContent()` for this.

_Note: I intentionally did not implement support for answer rendering within templates inside sections/blocks. This can be implemented, however, a major issue here is that the answer substitutions WILL NOT BE DYNAMIC. This is due to the fact that rendering/substitutions are handled by the server, and the fact that our current interaction model is that client only fetches the activity once. Imagine if we have an answer substitution in a question prompt template that references a previous question. The client only fetches the activity once, so when this is called, server renders the activity, and since there's no answer to previous question yet, substitution is not done. User answers the previous question, but client doesn't reload the activity or make another API request to the server, so server has no way of telling client that a specific question prompt changed. I think we need to address this issue before we support answer rendering in sections/blocks._

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
@@ -213,7 +213,7 @@ public class RenderValueProvider {
 
     private String renderAnswerUsingFormResponse(String questionStableId, String fallbackValue) {
         Answer answer = formResponse.getAnswer(questionStableId);
-        if (answer == null) {
+        if (answer == null || answer.isEmpty()) {
             // No answer response for this question yet, so use fallback.
             return fallbackValue;
         }
@@ -242,7 +242,7 @@ public class RenderValueProvider {
         Question question = formInstance.getQuestionByStableId(questionStableId);
         Answer answer = question != null && question.isAnswered()
                 ? (Answer) question.getAnswers().get(0) : null;
-        if (answer == null) {
+        if (answer == null || answer.isEmpty()) {
             return fallbackValue;
         }
         switch (answer.getQuestionType()) {
@@ -359,6 +359,9 @@ public class RenderValueProvider {
             return this;
         }
 
+        /**
+         * If caller has a FormResponse object available, then use this.
+         */
         public Builder withFormResponse(FormResponse formResponse, FormActivityDef formActivity, String isoLangCode) {
             provider.formResponse = formResponse;
             provider.formActivity = formActivity;
@@ -366,6 +369,9 @@ public class RenderValueProvider {
             return this;
         }
 
+        /**
+         * If caller has a FormInstance object available, then use this.
+         */
         public Builder withFormInstance(FormInstance formInstance) {
             provider.formInstance = formInstance;
             return this;
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
@@ -7,8 +7,22 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.ddp.exception.DDPException;
+import org.broadinstitute.ddp.model.activity.definition.FormActivityDef;
+import org.broadinstitute.ddp.model.activity.definition.question.PicklistOptionDef;
+import org.broadinstitute.ddp.model.activity.definition.question.PicklistQuestionDef;
+import org.broadinstitute.ddp.model.activity.definition.question.QuestionDef;
+import org.broadinstitute.ddp.model.activity.instance.FormInstance;
+import org.broadinstitute.ddp.model.activity.instance.FormResponse;
+import org.broadinstitute.ddp.model.activity.instance.answer.Answer;
+import org.broadinstitute.ddp.model.activity.instance.answer.PicklistAnswer;
+import org.broadinstitute.ddp.model.activity.instance.question.PicklistOption;
+import org.broadinstitute.ddp.model.activity.instance.question.PicklistQuestion;
+import org.broadinstitute.ddp.model.activity.instance.question.Question;
 import org.broadinstitute.ddp.model.dsm.KitReasonType;
 import org.broadinstitute.ddp.model.dsm.TestResult;
 import org.slf4j.Logger;
@@ -32,6 +46,13 @@ public class RenderValueProvider {
     private String testResultCode;
     private Instant testResultTimeCompleted;
     private Integer activityInstanceNumber;
+
+    // To minimize database round-trips, we lookup answers using existing objects that should have the answer objects.
+    // Depending on what's available for the provider, we'll use either response + activity or the instance object.
+    private FormResponse formResponse;
+    private FormActivityDef formActivity;
+    private String isoLangCode;
+    private FormInstance formInstance;
 
     private RenderValueProvider() {
         // Use builder.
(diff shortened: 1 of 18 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -166,6 +187,72 @@ public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCuto
...
+     *
+     * <p>Because common case is displaying the answer within content, this substitution will by default renders the
+     * "user-friendly" view of the answer. For example, if answer is a picklist, then the picklist option text will be
+     * rendered rather than the option stable id.
+     *
+     * @param questionStableId the question stable id
+     * @param fallbackValue    the fallback value
+     * @return answer string representation if available, otherwise null
+     */
+    public String answer(String questionStableId, String fallbackValue) {
```
evgeniipr, line 202: May be to extract this method to separate class (for example RenderAnswerProvider.java) because RenderValueProvider is quite large now.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -166,6 +187,72 @@ public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCuto
...
+                                    .getOptionLabelTemplate()
+                                    .render(isoLangCode))
+                            .collect(Collectors.joining(","));
+                case COMPOSITE: // Fall-through
+                case FILE:
+                    // Have not decided what composite or file answers will look like yet.
+                    throw new DDPException("Rendering answer type " + answer.getQuestionType() + " is currently not supported");
+                default:
+                    // Everything else will get turned into a string.
+                    return answer.getValue().toString();
```
evgeniipr: Is it possible NPE here (seems getValue() can be empty or null) ?

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -166,6 +187,72 @@ public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCuto
...
+            if (answer == null) {
+                return fallbackValue;
+            }
+            switch (answer.getQuestionType()) {
+                case PICKLIST:
+                    Map<String, String> options = ((PicklistQuestion) question)
+                            .streamAllPicklistOptions()
+                            .collect(Collectors.toMap(PicklistOption::getStableId, PicklistOption::getOptionLabel));
+                    return ((PicklistAnswer) answer).getValue().stream()
+                            .map(selected -> options.get(selected.getStableId()))
```
evgeniipr: Why getOptionLabelTemplate().render(isoLangCode) not called here?

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -264,6 +351,18 @@ public Builder setActivityInstanceNumber(Integer activityInstanceNumber) {
             return this;
         }
 
+        public Builder withFormResponse(FormResponse formResponse, FormActivityDef formActivity, String isoLangCode) {
```
evgeniipr, line 365: This a bit unclear why in some cases built withResponse() and in other - withInstance(). Small comments would help.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/GroupBlock.java
```diff
@@ -65,6 +67,11 @@ public Long getTitleTemplateId() {
         return nested;
     }
 
+    @Override
+    public Stream<Question> streamQuestions() {
+        return nested.stream().flatMap(FormBlock::streamQuestions);
```
evgeniipr, line 72: Is it possible that `nested` is null?

yufengwng:
@MocanaAtBroad @evgeniipr Please take a look at this PR when you get a chance!

MocanaAtBroad:
> @MocanaAtBroad @evgeniipr Please take a look at this PR when you get a chance!
Doing the PR rounds. Getting here soon!

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -166,6 +187,72 @@ public String activityInstanceNumberDisplay(int offsetToSubtract, int numberCuto
...
+     * <p>Because common case is displaying the answer within content, this substitution will by default renders the
+     * "user-friendly" view of the answer. For example, if answer is a picklist, then the picklist option text will be
+     * rendered rather than the option stable id.
+     *
+     * @param questionStableId the question stable id
+     * @param fallbackValue    the fallback value
+     * @return answer string representation if available, otherwise null
+     */
+    public String answer(String questionStableId, String fallbackValue) {
+        if (formResponse != null) {
```
MocanaAtBroad, line 203: Consider splitting code in method into 2: one for when formResponse is available and another for when only formInstance is available.

MocanaAtBroad (commented):
OK. Formula can be used within a translation, so we would have one per language, yes?
~~The fallback value cannot be internationalized/has to be a static string, am I correct?
In family history, when we initially create and display/before user has had chance of select which type of relation (e.g.. before they select "brother" vs "sister"), we will have to show some other title, like "Sibling" or "Sibling1" if working in English or "Hermano/Hermana" in Spanish. This something we can do?~~

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: Done! https://github.com/broadinstitute/ddp-study-server/pull/676/commits/7478060ee110b36d1bc630d1fee793002f2382e0

yufengwng:
@MocanaAtBroad Regarding translations. Yes, it'll work with the fact that we have different `Translation`s for the title/subtitle, so the `text` for those translations will have the same piece of syntax but with a different fallback string.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -457,6 +447,27 @@ public void countQuestionsAndAnswers(Handle handle, String userGuid, String oper
...
+
+    private FormActivityDef getDefinitionForSummary(Handle handle, ActivityDefStore activityDefStore,
+                                                    String studyGuid, ActivityInstanceSummary summary) {
+        String activityCode = summary.getActivityCode();
+        String versionTag = summary.getVersionTag();
+        long versionId = summary.getVersionId();
+        long revisionStart = summary.getRevisionStart();
+
+        FormActivityDef formActivityDef = activityDefStore.getActivityDef(studyGuid, activityCode, versionTag);
+        if (formActivityDef == null) {
```
evgeniipr, line 462: Same block of code exists in `ActivityInstanceDao.java` - could be extracted to method and reused in both places:
```
                       Optional<ActivityDto> activityDto = handle.attach(JdbiActivity.class)
                                .findActivityByStudyGuidAndCode(studyGuid, activityCode);
                        formActivityDef = handle.attach(FormActivityDao.class).findDefByDtoAndVersion(
                                activityDto.get(), versionTag, versionId, revisionStart);
                        activityDefStore.setActivityDef(studyGuid, activityCode, versionTag, formActivityDef);
 ```

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/GroupBlock.java
yufengwng: `nested` is always initialized to `new ArrayList<>()` so it won't be `null`.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: Good catch! I'll double-check.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: `FormInstance` objects and the picklist options inside a FormInstance does not have `Template` objects. Instead, they have `optionLabel` which is a string that is already translated/rendered.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: I think I'll keep it for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
yufengwng: I think I'll keep it for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: I have other comments that mentions we'll use either the `FormResponse` or `FormInstance` depending on what's available.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: Added a check using answer's `isEmpty()` method, which checks for nulls, empty text strings, etc. c42ede19

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
yufengwng: Added small comments. c42ede19

evgeniipr (approved):

PR #681, opened 2021-03-16
DDP-5707 brain: fix copy event for Post Consent

yufengwng:
## Context

See DDP-5707.

The root issue is that we removed old sex/gender questions and added new ones as part of Brain Tumor rename. And the POST_CONSENT copy answer event was referencing the old questions. The fix is to check that questions actually exist in `CopyExecutor`.

Another issue is that in order to do proper copying for the sex/gender questions, the picklist options need to match up. So I added code to fix up those option stable ids.

## Checklist

## FUD Score

- [x] Getting dev into a state where this is user-visible requires some tech fiddling
    - Need to run `BrainRenameFixPostConsentCopy` patch.

## Testing

## Release

    - Need to run `BrainRenameFixPostConsentCopy` patch.


MocanaAtBroad (approved):
Looks OK to me. Do we have to deal with migrating existing answers?

yufengwng:
Re: migrating existing answers. Answers internally reference the option numeric id and not the string stable id, so I think current answers should be fine. As for copying answers, there are no answers that need to be copied. If there was, it will have ran into the bug and resulted in a prod Slack alert, but we haven't seen any yet.

We'll need to refresh the cache and update activity definitions in Elasticsearch, but no data migration needed.

PR #683, opened 2021-03-22
DDP-5402 - implemented an API ActivityInstanceFromDefinitionBuilder for ActivityInstance creation

evgeniipr:
## Context

It is implemented an API for `ActivityInstance` creation from a study definition stored in `ActivityDefStore`. 
The instance data (like activity instance, answers) are fetched from DB. Validation messages also fetched from DB but once it's fetched it is cached in `ActivityDefStore` cache and next time is taken from the cache.

The main class of this API is `ActivityInstanceFromDefinitionBuilder`. It has one method:
```
public Optional<ActivityInstance> buildActivityInstance(
            Handle handle,
            String userGuid,
            String operatorGuid,
            String studyGuid,
            String instanceGuid,
            ContentStyle style,
            String isoLangCode
    );
```

This API currently used in `GetActivityInstanceRoute` instead of old `ActivityInstanceService.getTranslatedActivity()`:
```
 Optional<ActivityInstance> inst = actInstService.buildInstanceFromDefinition(
                handle, userGuid, operatorGuid, studyGuid, instanceGuid, style, isoLangCode
 );
```
The performance tests indicated about 3.5 times of speed increase of `ActivityInstance` reading in endpoint  `GetActivityInstanceRoute`.

The next step will be: remove old implementation `ActivityInstanceService.getTranslatedActivity()` from the code (subtask https://broadinstitute.atlassian.net/browse/DDP-5758).

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/FormInstanceCreator.java
```diff
@@ -0,0 +1,91 @@
...
+                context.getHandle(),
+                context.getInterpreter(),
+                context.getUserGuid(),
+                context.getOperatorGuid(),
+                context.getActivityInstanceDto().getGuid(),
+                null);
+    }
+
+    private void render(FormInstance formInstance, FormActivityDef formActivityDef) {
+        // TODO
```
evgeniipr: @yufengwng could you advise about rendering of `FormInstance` templates?
There is a method `FormInstance.renderContent()` where rendered `readonlyHint`, `activityDefinitionLastUpdatedText`.  What is the best way to render this strings then without excessive calls to DB?
I believe it is not possible to avoid DB calls in this case. What do you think?
Possibly it needs to update `ActivityDefStore`  
And another question about rendering: some of `..Def` classes contains only templateID-s (and `Template` objects are not assigned - i.e. nulls): does it mean that it needs to query DB to get `Template` data? I didn't find any cache (map) with templates from where it could be fetched by IDs.

yufengwng: I'm not sure I have all the answers, but here's my recommendation:

* Try updating the various DAOs (such as `ContentBlockDao::colleckBlockDefs()`) so that they query out all the `Template` objects. This way, we'll have what we need from the `ActivityDef` objects. It does add cost but this is only done once and then it'll be cached in `ActivityDefStore`.

* Try adding a new overloaded `render()` method to `Template` class that takes in a ` Map<String, Object> context`. This way, we can pass in a `RenderValueProvider` object, just like how things are done in `FormInstance::renderContent()`.

* Lastly, instead of calling `FormInstance::renderContent()` or rendering templates with `bulkRender()` and template ids, take advantage of the fact that we have `Template` objects that we can render from. For things like `readonlyHint` and `lastUpdatedText`, maybe try to add new constructor that takes in a string instead of the template id.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ContentBlockDao.java
```diff
@@ -89,11 +90,22 @@ default void disableContentBlock(long blockId, RevisionMetadata meta) {
...
 
+        Set<Long> templateIds = new HashSet<>();
+        contentDtos.values().forEach(v -> {
+            templateIds.add(v.getBodyTemplateId());
+            if (v.getTitleTemplateId() != null) {
+                templateIds.add(v.getTitleTemplateId());
+            }
+        });
+        Map<Long, Template> templates = getTemplateDao().collectTemplatesByIdsAndTimestamp(templateIds, timestamp);
+
```
evgeniipr, line 101: @yufengwng please check this change: here is added a ContentBlock templates collecting, and then the collected templates are passed to a ContentBlock constructor (see below).

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/FormInstanceCreator.java
```diff
@@ -0,0 +1,106 @@
...
+        formInstance.updateBlockStatuses(
+                context.getHandle(),
+                context.getInterpreter(),
+                context.getUserGuid(),
+                context.getOperatorGuid(),
+                context.getActivityInstanceDto().getGuid(),
+                null);
+    }
+
+    private void render(FormInstance formInstance) {
```
evgeniipr: @yufengwng in this method rendered `readonlyHint` and `activityDefinitionLastUpdatedText`. Currently the result is the same as during ActivityInstance normal reading (the texts are identical). But yet I'm not sure if it's correct from the versions perspective etc.. I need to check still. And as you see here I just call template.render(0 and then HtmlConverter. So it looks quite similar to formInstance.renderContent() logic but I think I need more deeper understanding of `activityDefinitionLastUpdated` rendering (because it depends on `activityDefinitionLastUpdated'). So, this logic I definitely need to fix in my code still.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/ActivityInstanceAssembleService.java
```diff
@@ -0,0 +1,188 @@
...
+ *       {@link FormBlockCreator}
+ *         {@link SectionIconCreator}
+ *         {@link QuestionCreator}
+ *           {@link ValidationRuleCreator}
+ * </pre>
+ *
+ * <p>NOTE: it is defined a class {@link Context} used to pass the basic parameters to each element creator (so it no
+ * need to pass multiple parameter in each creator - only one parameter.
+ */
+public class ActivityInstanceAssembleService {
```
yufengwng: Can we just put this in `ActivityInstanceService` instead of creating a new class?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/FormInstanceCreator.java
```diff
@@ -0,0 +1,106 @@
...
+        FormActivityDef formActivityDef = context.getFormActivityDef();
+
+        FormInstance formInstance = new FormInstance(
+                instanceDto.getParticipantId(),
+                instanceDto.getId(),
+                instanceDto.getActivityId(),
+                instanceDto.getActivityCode(),
+                formActivityDef.getFormType(),
+                instanceDto.getGuid(),
+                activityI18nDetail != null ? activityI18nDetail.getTitle() : null,
```
yufengwng: The `FormActivityDef` object should already have things like title and subtitle. I think we should reuse those instead of querying for an `ActivityI18nDetail`. Take a look at `FormActivityDao::findDefByDtoAndVersion()` method -- it already queries for those `ActivityI18nDetail`. I think what you need to do here is iterate through those lists and pick out the right one using the language code.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ContentBlockDao.java
yufengwng: Looks about right.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/FormInstanceCreator.java
yufengwng: Take a look at `FormActivityDao::findDefByDtoAndVersion()`. When we build a `FormActivityDef` object, we are already querying for the appropriate version. So the read-only hint and last updated text you get from the `FormActivityDef` object is what you should use.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstanceassembler/ActivityInstanceAssembleService.java
evgeniipr: I renamed `ActivityInstanceAssembleService` to `ActivityInstanceFromActivityDefStoreBuilder` and added to `ActivityInstanceService` a method:
```
public Optional<ActivityInstance> buildInstanceFromActivityDefStore(
            Handle handle,
            String isoLangCode,
            ContentStyle style,
            String studyGuid,
            String userGuid,
            String operatorGuid,
            ActivityInstanceDto instanceDto) {
        return new ActivityInstanceFromActivityDefStoreBuilder().buildActivityInstance(
                handle, isoLangCode, style, studyGuid, userGuid, operatorGuid, instanceDto
        );
}
```

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ComponentBlockDef.java
```diff
@@ -26,6 +26,9 @@
     @SerializedName("hideNumber")
     private boolean hideNumber;
 
+    @SerializedName("revisionId")
+    private long revisionId;
```
yufengwng: Note that these definition objects are what we use to deserialize study-builder activity configurations in order to insert them into the database (check out `ActivityBuilder` in study-builder codebase). `revisionId` is not something we'll deserialize from the activity configuration JSON. How about we change the following so it doesn't play a part in JSON?

```
private transient long revisionId;
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -1,19 +1,5 @@
...
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
```
yufengwng: Order of imports here is not consistent with Java files in this repo. Can you fix your code editor settings so that the order of imports is:
```
import static

import java.*
import javax.*

<third part imports>
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -563,4 +531,21 @@ public void deleteInstance(Handle handle, ActivityInstanceDto instanceDto) {
         int numDeleted = instanceDao.deleteAllByIds(Set.of(instanceDto.getId()));
         DBUtils.checkDelete(1, numDeleted);
     }
+
+    /**
+     * Build {@link ActivityInstance} from data cached stored in {@link ActivityDefStore}.
+     * Some of data (answers, rule messages) are queried from DB.
+     */
+    public Optional<ActivityInstance> buildInstanceFromActivityDefStore(
```
yufengwng: Consider renaming the method. The focus here is that we're rendering the activity instance from its definition. It just so happens that the definition is stored in a class called ActivityDefStore. The definition could very well be stored somewhere else (like Redis).

```
buildInstanceFromDefinition()

ActivityInstanceFromDefinitionBuilder()
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormInstanceCreator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private FormInstance constructFormInstance() {
+        ActivityInstanceDto instanceDto = context.getActivityInstanceDto();
+        FormActivityDef formActivityDef = context.getFormActivityDef();
+
+        String title = extractOptionalActivityTranslation(formActivityDef.getTranslatedTitles(), context.getIsoLangCode());
+        String subtitle = extractOptionalActivityTranslation(formActivityDef.getTranslatedSubtitles(), context.getIsoLangCode());
+
+        boolean readonly = ActivityInstanceUtil.isReadonly(
+                context.getActivityInstanceDto().getEditTimeoutSec(),
```
yufengwng: `editTimeoutSed` is a property of definition. Use `ActivityDef::getEditTimeoutSec()`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormInstanceCreator.java
```diff
@@ -0,0 +1,108 @@
...
+        FormInstance formInstance = constructFormInstance();
+        addChildren(formInstance);
+        updateBlockStatuses(formInstance);
+        formInstance.setDisplayNumbers();
+        render(formInstance);
+        return formInstance;
+    }
+
+    private FormInstance constructFormInstance() {
+        ActivityInstanceDto instanceDto = context.getActivityInstanceDto();
```
yufengwng: Have you considered passing in the whole `FormResponse` object instead of both a `ActivityInstanceDto` and list of answers? FormResponse is missing `isHidden` and `sectionIndex`, but once we add those, it should give you everything you need from ActivityInstanceDto.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormInstanceCreator.java
```diff
@@ -0,0 +1,108 @@
...
+public class FormInstanceCreator extends ElementCreator {
+
+    public FormInstanceCreator(ActivityInstanceFromActivityDefStoreBuilder.Context context) {
+        super(context);
+    }
+
+    public FormInstance createFormInstance() {
+        FormInstance formInstance = constructFormInstance();
+        addChildren(formInstance);
+        updateBlockStatuses(formInstance);
```
yufengwng: Tracing through the current code,  looks like`updateBlockStatuses()` is the last thing we call.
```
1. new instance
2. load sections
3. render
4. set display numbers
5. update block statuses
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormInstanceCreator.java
```diff
@@ -0,0 +1,108 @@
...
+        formInstance.updateBlockStatuses(
+                context.getHandle(),
+                context.getInterpreter(),
+                context.getUserGuid(),
+                context.getOperatorGuid(),
+                context.getActivityInstanceDto().getGuid(),
+                null);
+    }
+
+    private void render(FormInstance formInstance) {
```
yufengwng: As you're aware, this is not feature-parity with existing render method. We need to create a `RenderValueProvider` and pass that into the `Template` so it can use it as rendering context. For last updated text, we also need to pass `activityDefinitionLastUpdated` into the rendering context map. Please see the original method for more details.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormSectionCreator.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public FormSectionCreator(ActivityInstanceFromActivityDefStoreBuilder.Context context) {
+        super(context);
+    }
+
+    public FormSection createSection(FormSectionDef formSectionDef) {
+        if (formSectionDef != null) {
+            FormSection formSection = constructFormSection(formSectionDef);
+            addChildren(formSection, formSectionDef);
+            applyRenderedTemplates(formSection);
```
yufengwng: Can we use the name template object rather than calling `applyRenderedTemplates()`?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormInstanceCreator.java
```diff
@@ -0,0 +1,108 @@
...
+                instanceDto.getActivityId(),
+                instanceDto.getActivityCode(),
+                formActivityDef.getFormType(),
+                instanceDto.getGuid(),
+                title,
+                subtitle,
+                instanceDto.getStatusType() != null ? instanceDto.getStatusType().name() : null,
+                readonly,
+                formActivityDef.getListStyleHint(),
+                getTemplateId(formActivityDef.getReadonlyHintTemplate()),
```
yufengwng: Hmmm, I'm not sure about calling `getTemplateId()` -- it's doing extra work. The implementation of `getTemplateId()` calls `template.render()`, which means we're rendering the read-only hint twice.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/SectionIconCreator.java
```diff
@@ -0,0 +1,30 @@
...
+        super(context);
+    }
+
+    public SectionIcon createSectionIcon(SectionIcon sectionIconDef) {
+        SectionIcon sectionIcon = constructSectionIcon(sectionIconDef);
+        return sectionIcon;
+    }
+
+    private SectionIcon constructSectionIcon(SectionIcon sectionIconDef) {
+        SectionIcon sectionIcon = new SectionIcon(
```
yufengwng, line 12: Hmmm, they're the same thing..?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/impl/ContentBlockCreator.java
```diff
@@ -0,0 +1,24 @@
...
+public class ContentBlockCreator extends ElementCreator {
+
+    public ContentBlockCreator(ActivityInstanceFromActivityDefStoreBuilder.Context context) {
+        super(context);
+    }
+
+    public ContentBlock createContentBlock(ContentBlockDef contentBlockDef) {
+        return new ContentBlock(
+                getTemplateId(contentBlockDef.getTitleTemplate()),
+                getTemplateId(contentBlockDef.getBodyTemplate())
```
yufengwng: Note that the whole approach with template ids and `registerTemplateIds`/`applyRenderedTemplates` was a workaround to optimize rendering templates when we didn't have the definition objects. But now that we're rendering directly from activity definition objects, I recommend we take full advantage of those template objects.

For example, right here we can create a new `ContentBlock` constructor that takes in the rendered title/body strings rather than ids. If you take a look at the code, no one is calling `getTitleTemplateId`/`getBodyTemplateId`, so we can probably just get rid of those entirely.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/impl/GroupBlockCreator.java
```diff
@@ -0,0 +1,40 @@
...
+                groupBlockDef.getListStyleHint(),
+                groupBlockDef.getPresentationHint(),
+                getTemplateId(groupBlockDef.getTitleTemplate())
+        );
+    }
+
+    private void addChildren(GroupBlock groupBlock, GroupBlockDef groupBlockDef) {
+        if (groupBlockDef.getNested() != null) {
+            groupBlockDef.getNested().forEach(b ->
+                    groupBlock.getNested().add(new FormBlockCreator(context).createBlock(b)));
```
yufengwng: Move the `new FormBlockCreator()` outside the loop so we're not new-ing up a bunch of objects.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/ValidationRuleCreator.java
```diff
@@ -0,0 +1,170 @@
...
+                getHintTitle(ruleDef),
+                ruleDef.getAllowSave(),
+                ruleDef.getMin(),
+                ruleDef.getMax()
+        );
+    }
+
+    private String resolveRuleMessage(RuleDef ruleDef) {
+        ValidationDao validationDao = context.getHandle().attach(ValidationDao.class);
+        return validationDao.getJdbiI18nValidationMsgTrans().getValidationMessage(
```
yufengwng: I think we can cache these default validation message texts. This is probably a good time to do it.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/QuestionUtil.java
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.ddp.service.actvityinstancebuilder.block.question;
+
+import org.broadinstitute.ddp.model.activity.definition.question.QuestionDef;
+import org.broadinstitute.ddp.service.actvityinstancebuilder.ActivityInstanceFromActivityDefStoreBuilder;
+
+public class QuestionUtil {
+
+    public static boolean isReadOnly(ActivityInstanceFromActivityDefStoreBuilder.Context context, QuestionDef questionDef) {
```
yufengwng: This is not feature-parity with existing calculation of question read-only. Please see `ddp.util.QuestionUtil` class. Also, I recommend removing this class here and add a new method in the other util class -- having two classes with same name is a bit confusing.
```
isReadOnly(Handle handle, QuestionDef questionDef, InstanceStatusType instanceStatus, long instanceId)
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/impl/BoolQuestionCreator.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Creates {@link BoolQuestion}
+ */
+public class BoolQuestionCreator extends ElementCreator {
+
+    public BoolQuestionCreator(ActivityInstanceFromActivityDefStoreBuilder.Context context) {
+        super(context);
+    }
+
+    public BoolQuestion createBoolQuestion(QuestionCreator questionCreator, BoolQuestionDef questionDef) {
```
yufengwng: For simple question types like these, perhaps we can move these construction methods into `QuestionCreator` itself? I'll rather have a bigger class file than a ton of tiny files.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/impl/CompositeQuestionCreator.java
```diff
@@ -0,0 +1,54 @@
...
+                getTemplateId(questionDef.getAdditionalItemTemplate()),
+                getChildQuestions(questionDef),
+                questionDef.getChildOrientation(),
+                questionCreator.getAnswers(CompositeAnswer.class, questionDef.getStableId())
+        );
+    }
+
+    private List<Question> getChildQuestions(CompositeQuestionDef questionDef) {
+        List<Question> childQuestions = new ArrayList<>();
+        QuestionCreator questionCreator = new QuestionCreator(context);
```
yufengwng: Is it important to create a new object? We already have a `QuestionCreator` we can use.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/ElementCreator.java
```diff
@@ -0,0 +1,29 @@
+package org.broadinstitute.ddp.service.actvityinstancebuilder;
+
+
+import org.broadinstitute.ddp.content.Renderable;
+import org.broadinstitute.ddp.model.activity.definition.template.Template;
+
+/**
+ * Abstract creator class
+ */
+public class ElementCreator {
```
yufengwng: Is there advantage to having this `ElementCreator`? I don't see how having this parent class helps us. If we really need these, I think `getTemplateId()` and `applyRenderedTemplates()` could be implemented as methods on the `Context` class itself.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetActivityInstanceRoute.java
```diff
@@ -127,6 +125,27 @@ public Object handle(Request request, Response response) {
...
+            String isoLangCode) {
+
+        Optional<ActivityInstance> inst = actInstService.buildInstanceFromActivityDefStore(
+                handle, isoLangCode, style, studyGuid, userGuid, operatorGuid, instanceDto
+        );
+        if (inst.isEmpty()) {
+            inst = actInstService.getTranslatedActivity(
+                    handle, userGuid, operatorGuid, instanceDto.getActivityType(), instanceGuid, isoLangCode, style
+            );
+        }
```
yufengwng: I don't think we should fallback to the old approach here. Once we switch to new approach, we should deprecate this. I look forward to deleting code for `getTranslatedActivity()` and related methods!

evgeniipr:
@yufengwng I pushed next commit: 
- removed small question/block creation classes and created 2 bigger classes instead;
- did some optimization (all Creators' construction - at the beginning of the process); 
- fixed Question.isReadOnly detection.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormResponse.java
```diff
@@ -36,9 +36,12 @@ public FormResponse(@ColumnName("instance_id") long id,
                         @ColumnName("activity_id") long activityId,
                         @ColumnName("activity_code") String activityCode,
                         @ColumnName("activity_version_tag") String activityVersionTag,
+                        @ColumnName("is_hidden") Boolean isHidden,
+                        @ColumnName("sectionIndex") Integer sectionIndex,
```
yufengwng: JDBI is smart enough to work with both snake_case and camelCase, but let's be consistent here.
```suggestion
                        @ColumnName("section_index") Integer sectionIndex,
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/AbstractCreator.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public abstract class AbstractCreator {
+
+    protected final ActivityInstanceFromDefinitionBuilder.Context context;
+
+    public AbstractCreator(ActivityInstanceFromDefinitionBuilder.Context context) {
+        this.context = context;
+    }
+
+    protected Long renderTemplateIfDefined(Template template) {
```
yufengwng: I still think this can be a util method on the `context` object itself, rather than needing to share an abstract class. Or maybe put it in `TemplateRenderUtil` if you want to keep methods out of `context`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/QuestionCreator.java
```diff
@@ -0,0 +1,80 @@
...
+
+    <T extends Answer> List<Rule<T>> getValidationRules(QuestionDef questionDef) {
+        return CollectionMiscUtil.createListFromAnotherList(questionDef.getValidations(),
+                (ruleDef) -> validationRuleCreator.createRule(ruleDef));
+    }
+
+    <T extends Answer> List<T> getAnswers(Class<T> type, String questionStableId) {
+        List<T> answers = new ArrayList<>();
+        if (context.getFormResponse().getAnswers() != null) {
+            answers = context.getFormResponse().getAnswers().stream()
```
yufengwng: You can lookup the answer by questionStableId instead of streaming it. The choice of using a `List` of answers was a legacy design decision -- there should only be one answer per question.

evgeniipr:
> I still think this can be a util method on the context object itself, rather than needing to share an abstract class.

I removed the abstract class.

evgeniipr:
> You can lookup the answer by questionStableId instead of streaming it. 

Fixed. I also marked the `FormResponse.getAnswers()` as `deprecated`.

evgeniipr:
> JDBI is smart enough to work with both snake_case and camelCase, but let's be consistent here.

Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityDefStore.java
```diff
@@ -161,6 +166,15 @@ public void setActivityDef(String studyGuid, String activityCode, String version
         }
     }
 
+    public String findValidationRuleMessageMap(Handle handle, RuleType ruleType, long langCodeId) {
```
yufengwng: Can you rename this method? We're not finding a "Map" but rather just the rule message.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/FormResponse.java
```diff
@@ -36,15 +36,28 @@ public FormResponse(@ColumnName("instance_id") long id,
...
+                latestStatus);
     }
 
+    /**
+     * Get list of answers.
+     * @deprecated An answer can be looked up by questionStableId instead of streaming it.
+     *       The choice of using a List of answers was a legacy design decision -
+     *       there should only be one answer per question.
+     */
+    @Deprecated
```
yufengwng: I think there's a misunderstanding here. This method should not be deprecated. When I said "legacy design decision", I was not talking about this method -- what I meant was that if you looked at the `Question` class it has a `List<Answer>`. Question should have one answer, so this `List<Answer>` was an old design decision (and no, we do not need to refactor that right now!).

And note that this `getAnswers()` method does NOT return a list of answers for a single question, rather it returns all the answers for all questions in the activity instance. Which is why I suggested not to use this one -- since we want the answer for a single question, we should use `getAnswers(String questionStableId)` instead.

Sorry if I've caused any confusions! Can you please remove the `@Deprecated` annotation?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/ActivityInstanceFromDefinitionBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+    public Optional<ActivityInstance> buildActivityInstance(
+            Handle handle,
+            String userGuid,
+            String operatorGuid,
+            String studyGuid,
+            String instanceGuid,
+            ContentStyle style,
+            String isoLangCode
+    ) {
+        LOG.info("Start ActivityInstance building from definition (ActivityDefStore). StudyGuid={}, instanceGuid={}",
```
yufengwng: Might want to change these to `debug()` logs -- there's going to be alot of activity instance fetching, so we don't want to clog up production logs! Likewise with the `SUCCESSFULLY` log below.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/ActivityInstanceFromDefinitionBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+                studyGuid, instanceGuid);
+        var formResponse = ActivityInstanceUtil.getFormResponse(handle, instanceGuid);
+        if (formResponse.isEmpty()) {
+            LOG.warn("Error reading form activity by guid=" + instanceGuid);
```
yufengwng, line 92: There's already logging for the `isEmpty()` case in `GetActivityInstanceRoute` so this warning here is a bit redundant. In general, probably good idea to let caller deal with logging/reporting.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/ActivityDefStore.java
```diff
@@ -161,6 +182,37 @@ public void setActivityDef(String studyGuid, String activityCode, String version
...
+    }
+
+    private String detectValidationRuleMessage(
+            Handle handle, RuleType ruleType, Long hintTemplateId, long langCodeId, long timestamp) {
+        String correctionHint = null;
+        if (hintTemplateId != null) {
+            correctionHint = i18nContentRenderer.renderContent(handle, hintTemplateId, langCodeId, timestamp);
+        }
+        if (correctionHint != null) {
+            return correctionHint;
```
yufengwng, line 208: I don't think `correctionHint` belongs here. ActivityDefStore should only deal with caching the default validation message that corresponds to each type of validation rule.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/ValidationRuleCreator.java
```diff
@@ -0,0 +1,166 @@
...
+                ruleDef.getRuleId(),
+                resolveRuleMessage(ctx, ruleDef),
+                getHintTitle(ctx, ruleDef),
+                ruleDef.getAllowSave(),
+                ruleDef.getMin(),
+                ruleDef.getMax()
+        );
+    }
+
+    private String resolveRuleMessage(AIBuilderContext ctx, RuleDef ruleDef) {
```
yufengwng: I don't think we need to do any "resolving" here. If you look in the abstract parent `Rule` class, it will decide whether to use the `correctionHint` or the `defaultMessage`. So we just need to grab the default one from ActivityDefStore and pass it in.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/block/question/PicklistCreatorHelper.java
```diff
@@ -0,0 +1,61 @@
...
+                    CollectionMiscUtil.createListFromAnotherList(picklistOptionDef.getNestedOptions(),
+                        (nestedPicklistOptionDef) -> createPicklistOption(ctx, nestedPicklistOptionDef,  picklistGroupDefs, true))
+        );
+
+        setGroupStableIdIfOptionBelongsToGroup(picklistGroupDefs, picklistOption);
+
+        return picklistOption;
+    }
+
+    private void setGroupStableIdIfOptionBelongsToGroup(List<PicklistGroupDef> picklistGroupDefs, PicklistOption picklistOption) {
```
yufengwng, line 50: Might be better to have some sort of `Map<option_stable_id, group_def>` here instead of a `List` so we can turn this into `O(n)` (lookup in map is O(1) and we're doing N options) instead of `O(n^2)` (each lookup is O(n) and we're doing N options).

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
```diff
@@ -28,6 +28,7 @@
     private Long firstCompletedAt;
     private boolean allowUnauthenticated;
     private int sectionIndex;
+    private Long editTimeoutSec;
```
yufengwng: I think we can remove this now?

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiActivityInstance.sql.stg
```diff
@@ -14,6 +14,7 @@ select ai.activity_instance_id,
        act.allow_unauthenticated,
        act.study_activity_code,
        act_type.activity_type_code as activity_type,
+       act.edit_timeout_sec,
```
yufengwng: Don't need this anymore since we're grabbing this from the activity definition.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ComponentBlockDef.java
```diff
@@ -26,6 +26,8 @@
     @SerializedName("hideNumber")
     private boolean hideNumber;
 
+    private transient long revisionId;
```
yufengwng: I got a bit confused on why we need this, and then I saw that this is being used in `FormBlockCreatorHelper`. Two things:
* Maybe we should rename this to `componentRevisionId`. I got confused on thinking that this is the "revision" of the block itself...
* If we inspect the code, we see that component revision id is not really used anywhere. Maybe there's a way to not require on this.

yufengwng (dismissed):
Have a few small comments. This looks good already, so let's get this merged and fix the small issues in a new PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/ActivityInstanceDto.java
```diff
@@ -27,7 +27,6 @@
     private String sessionId;
     private Long firstCompletedAt;
     private boolean allowUnauthenticated;
-    private int sectionIndex;
```
yufengwng: Wait... why did we remove this one? Can we put section index back in?

yufengwng (approved):

PR #684, opened 2021-03-23
Adds an Application flag to enable skipping Pepper registration [DDP-5719]

bskinner:
[DDP-5719](https://broadinstitute.atlassian.net/browse/DDP-5719)

## Context

When handling Auth0 login attempts, all the logins result in a callback to the [registerUser](https://pepper.datadonationplatform.org/docs#operation/registerUser) operation. While necessary for Pepper, this is undesirable for clients which do not need to be registered. This PR adds `skipPepperRegistration` as an optional flag which can be added to an Application's client metadata in order to bypass this process.

In the case where:
* `skipPepperRegistration` is undefined or null
  * The script will use the legacy behavior and check to see if the client's name is in the `m2mClients` array
* `skipPepperRegistration` is set to the literal string `true` in the client metadata
  * The registration process with Pepper is skipped during authentication attempts
* `skipPepperRegistration` is set to any other string
  * Assumes the client has opted-in to the new behavior, and has declined to skip the registration process. Registration should continue as normal.

## Checklist

## FUD Score

This ticket does not make any explicitly user-visible changes.

## Testing

## Release

### DIFF
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -16,9 +16,15 @@ function (user, context, callback) {
     // to the value of 'true' in their client metadata if the Pepper registration process
     // is not required.
     var m2mClients = ['dsm', 'Count Me In (Salt CMS)'];
-    var skipPepperRegistration = context.clientMetadata.skipPepperRegistration || 'false'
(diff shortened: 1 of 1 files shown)


Rinchan20 (dismissed):
lgtm - thanks for answering my questions on it

Rinchan20 (approved):
lgtm

PR #687, opened 2021-03-25
DDP-5738 more error logging in user registration

yufengwng:
## Context

DDP-5738. We had an issue in production and we were not aware of it because it was logged as a warning. Changing it to an error instead.

## Checklist

## FUD Score

## Testing

- [x] I have written zero automated tests

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -382,7 +382,7 @@ private User handleGovernancePolicy(Response response, UserRegistrationPayload p
             shouldCreateGoverned = policy.shouldCreateGovernedUser(handle, interpreter, operatorUser.getGuid());
         } catch (Exception e) {
             String msg = "Error while evaluating study governance policy for study " + policy.getStudyGuid();
-            LOG.warn(msg, e);
+            LOG.error(msg, e);
```
yufengwng, line 385: These are the only two places in this route where we're returning a server 500 response. In these two cases, we should log an error so we can get alerted about it.

MocanaAtBroad (approved):

evgeniipr (approved):

PR #688, opened 2021-03-25
Ddp 5096 rgp seqr export

charlottedye:
## Context

See DDP-5096.  The Rare Genomes Project staff needs the data for the main (enrollment) survey exported on a weekly basis so it can be imported into another system.  They want each survey to be exported once after it is completed.  They are interested in a subset of fields and of those fields, the two most likely to contain PHI that would need to be removed should be toward the beginning of the spreadsheet to make reviewing them easier.  Additionally, study staff like to be notified when these exports are done.  

This will allow study staff to easily add this data to Seqr for further analysis.  This is a feature that exists in Gen2 but this has been modified to make use of the existing Pepper infrastructure.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/config/application.conf.ctmpl
```diff
@@ -84,7 +84,37 @@
         "cancerLoader": "{{$conf.Data.schedules.cancerLoader}}",
         "studyExport": "{{$conf.Data.schedules.studyExport}}",
         "fileUploadCleanup": "{{$conf.Data.schedules.fileUploadCleanup}}",
-        "tempUserCleanup": "{{$conf.Data.schedules.tempUserCleanup}}"
+        "tempUserCleanup": "{{$conf.Data.schedules.tempUserCleanup}}",
+        "rgpExport": "{{$conf.Data.schedules.rgpExport}}"
+    },
+    "rgpExport": {
```
yufengwng: I want to make a note: Pepper is meant to be a general _platform_ for different studies. This configuration is the config file for the Pepper platform. If you look through this file, you won't find mentions of _study-specific_ values. There's no mention of A-T, Angio, Osteo, TestBoston, etc. It concerns me that we have study-specific values (in this case "RGP") in what is meant to be a platform config file. What should have happened is that we come up with a more general feature (e.g. "incremental backup to a bucket") and then these become study-level configuration.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExportCoordinator.java
```diff
@@ -31,13 +39,17 @@
 
     public static final int DEFAULT_BATCH_SIZE = 100;
     public static final int READER_BUFFER_SIZE_IN_BYTES = 10 * 1024;
+    public static final String RGP_GUID = "RGP";
+    public static final String RGP_ACTIVITY = "ENROLLMENT";
```
yufengwng: To me, this is a red flag. Seeing study-specific IDs hardcoded in source code doesn't seem right to me. We're building special code for a specific study. Remember that the Pepper codebase is open-source. Imagine if someone wants to setup their own instance of Pepper -- they're not going to have a study called "RGP".

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-study-rgp-export-last-completed.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="charlotte" id="20210315-study-rgp-export-last-completed">
+        <createTable tableName="rgp_export">
```
yufengwng: Did you speak with other folks (e.g. Zim) about this approach? Did they give their blessing on having study-specific code/tables?

yufengwng (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExportCoordinator.java
yufengwng: We do have code that mentions some studies, but those are mostly utility tools / command-line tools that we developers run, or some unit tests. For example, there might be some code to help with data migration. In other words, code that are not actual production code.

pepper-apis/config/application.conf.ctmpl
charlottedye: This was my initial approach but @zyme suggested I implement this as a one-off.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExportCoordinator.java
charlottedye: I think the above mostly addresses this, but I'm going to at least move this stuff to the config file.  It's still ugly, though.

charlottedye: I'll clean this piece up and then add Zim as a reviewer to make sure that this is the approach he suggested.

pepper-apis/config/application.conf.ctmpl
zyme: How about if we make a separate config file for the RGP export?

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-study-rgp-export-last-completed.xml
zyme: I would prefer we not name tables with study identifiers.  Can we changes this to `export_operation` or some such, and remove `rgp` prefixes in general?

pepper-apis/config/application.conf.ctmpl
zyme: How about making the RGP specific configurations live in secret manager?  No reason to keep going with vault...

charlottedye: Good idea on both counts.  I keep forgetting about secret manager

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-study-rgp-export-last-completed.xml
charlottedye: That makes sense.  Plus this also makes it easier to make this available to other studies if at some point we decide to do that.

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-custom-export-last-completed.xml
```diff
@@ -0,0 +1,23 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="charlotte" id="20210315-custom-export-last-completed">
+        <createTable tableName="custom_export">
+            <column name="custom_export_id" type="bigint" autoIncrement="true" startWith="1">
+                <constraints primaryKey="true" primaryKeyName="custom_export_pk"/>
+            </column>
+            <column name="umbrella_study_id" type="bigint">
```
yufengwng: study_id column should be unique right? So that we only have one row per study? Maybe we should just make the study_id the PK as well.

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-custom-export-last-completed.xml
```diff
@@ -0,0 +1,23 @@
...
+                <constraints nullable="false"
+                             references="umbrella_study(umbrella_study_id)"
+                             foreignKeyName="custom_export_umbrella_study_id_fk"/>
+            </column>
+            <column name="custom_export_last_completed" type="bigint" defaultValueNumeric="0">
```
yufengwng: For new tables, I think we should use data type `datetime(6)` (a timestamp with microsecond precision). Nicer to work with when querying stuff manually, and can be mapped to an `Instant` in Java.

pepper-apis/parent-pom.xml
```diff
@@ -513,6 +519,13 @@
                 <artifactId>commons-beanutils</artifactId>
                 <version>1.9.4</version>
             </dependency>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>libraries-bom</artifactId>
```
yufengwng, line 522: Is the bom here necessary?

pepper-apis/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
```diff
@@ -49,6 +50,15 @@ public Bucket getBucket(String bucketName) {
         return storage.get(bucketName);
     }
 
+    public Bucket getBucketFromBlob(String bucketName) {
```
yufengwng: Why is this implementation necessary? Can we use the `getBucket(String)` method above?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/constants/CustomExportConfigFile.java
```diff
@@ -0,0 +1,28 @@
+package org.broadinstitute.ddp.customexport.constants;
+
+public class CustomExportConfigFile {
+
+    // Custom Export fields
+    public static final String BUCKET_NAME = "bucketName";
+    public static final String FILE_PATH = "filePath";
+    public static final String BASE_FILE_NAME = "baseFileName"; // The main file name, not including the time stamp or the extension
+    public static final String GUID = "guid";
```
yufengwng: A number of things have guids. How about we call this one `STUDY_GUID`?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,94 @@
...
+
+public interface CustomExportDao extends SqlObject {
+
+    default long getLastCompleted(long studyId) {
+        Optional<Long> optionalLastCompleted = findExportLastCompletedByStudyId(studyId);
+        if (optionalLastCompleted.isPresent()) {
+            return optionalLastCompleted.get();
+        } else {
+            // Return 0 but make sure to add this to the database
+            long exportId = insertExport(studyId, 0);
```
yufengwng: If we use `datetime(6)`, we can probably insert/return `Instant.EPOCH` as the "minimum" value.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,94 @@
...
+    Optional<Long> findExportLastCompletedByStudyId(@Bind("id") long id);
+
+    @SqlUpdate("update custom_export set custom_export_last_completed = :lastCompleted where umbrella_study_id = :studyId")
+    int updateLastCompletedByStudyId(@Bind("lastCompleted") long lastCompleted, @Bind("studyId") long studyId);
+
+    @SqlQuery("SELECT CASE WHEN ( "
+            + "SELECT COUNT(user_id) FROM ( "
+            + "SELECT user_id FROM user_study_enrollment AS usen  "
+            + "JOIN study_activity AS sa ON sa.study_id = usen.study_id  "
+            + "JOIN activity_instance AS aci ON usen.user_id = aci.participant_id  "
```
yufengwng: I think we should add another join clause here so we match things up properly
```
AND aci.study_activity_id = sa.study_activity_id
```

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @SqlQuery("select custom_export_last_completed from custom_export where umbrella_study_id = :id")
+    Optional<Long> findExportLastCompletedByStudyId(@Bind("id") long id);
+
+    @SqlUpdate("update custom_export set custom_export_last_completed = :lastCompleted where umbrella_study_id = :studyId")
+    int updateLastCompletedByStudyId(@Bind("lastCompleted") long lastCompleted, @Bind("studyId") long studyId);
+
+    @SqlQuery("SELECT CASE WHEN ( "
+            + "SELECT COUNT(user_id) FROM ( "
+            + "SELECT user_id FROM user_study_enrollment AS usen  "
```
yufengwng: Hmmm, I don't think we need to join over to `user_study_enrollment`. It feels to me that we're mostly just looking things by activityCode. An activity belongs in one study. So in a way, querying for RGP's activity is already only querying for users in RGP.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/housekeeping/schedule/CustomExportJob.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(CustomExportJob.class);
+
+    private static JobKey getKey() {
+        return Keys.Export.CustomExportJob;
+    }
+
+    public static void register(Scheduler scheduler, Config cfg) throws SchedulerException {
+        // Test that we can initialize the credentials.
+        GoogleCredentialUtil.initCredentials(true);
```
yufengwng: What if I'm running things locally and I want to turn off this job? I don't want to have to pass in credentials... Can you put this after we have checked that the `schedule` is present and is not set to `off`?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/housekeeping/schedule/CustomExportJob.java
```diff
@@ -0,0 +1,138 @@
...
+            LOG.info("Finished job {}. Took {} ms", getKey(), elapsed);
+        } catch (Exception e) {
+            LOG.error("Error while executing job {}", getKey(), e);
+            throw new JobExecutionException(e, false);
+        }
+    }
+
+    private void runExport() {
+        Config cfg = ConfigManager.getInstance().getConfig();
+        Config exportCfg = SecretUtil.getConfigFromSecret(cfg, ConfigFile.CUSTOM_EXPORT_CONFIG_SECRET);
```
yufengwng: Should we just do this once and cache it? So we don't constantly make API requests to Secret Manager? Although, one benefit of this setup is that we can change the config "on-the-fly" sort of without rebooting.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/housekeeping/schedule/CustomExportJob.java
```diff
@@ -0,0 +1,138 @@
...
+        Bucket bucket = bucketClient.getBucketFromBlob(bucketName);
+        if (bucket == null) {
+            LOG.error("Could not find google bucket {}, skipping job {}", bucketName, getKey());
+            return;
+        }
+
+        String studyGuid = exportCfg.getString(CustomExportConfigFile.GUID);
+
+        StudyDto customDto = TransactionWrapper.withTxn(TransactionWrapper.DB.APIS,
+                handle -> handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid));
```
yufengwng: Can use `JdbiUmbrellaStudyCached` here for a bit more efficiency.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+        //Find out whether there is anything to export
+        boolean needExport = withAPIsTxn(handle -> {
+            CustomExportDao customExportDao = handle.attach(CustomExportDao.class);
+            long lastCompleted = customExportDao.getLastCompleted(customDto.getId());
+            return customExportDao.needCustomExport(customDto.getId(), customExportStatus, lastCompleted, customActivity);
+        });
+
+        if (!needExport) {
+            LOG.info("Skipping custom export: nothing to export");
+            ApiResult res = client.sendMail(createNotificationEmail(false, true));
```
yufengwng: Woohoo! Nice to see that these wrappers around 3rd party APIs are getting some reuse!

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+            if (res.hasThrown()) {
```
yufengwng: Aside from an exception, we should also check if the API result have an error payload as well (e.g. perhaps there was no exceptions but SendGrid returned a 4xx or 5xx with some error response).

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+        }
+
+        // Proceed with the export
+        List<ActivityExtract> activityExtracts = withAPIsTxn(handle -> exporter.extractCustomActivity(handle));
+
+        withAPIsTxn(handle -> {
+            ExportUtil.computeMaxInstancesSeen(handle, activityExtracts);
+            ExportUtil.computeActivityAttributesSeen(handle, activityExtracts);
+            return null;
+        });
```
yufengwng: I think we can do these in one transaction.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+        return createMail(subject, fromEmail, toEmail, content);
+    }
+
+    private void addSuccessContent(List<Content> content, String fullContent, String fileUrl) {
+        Content linkContent = new Content("text/html", "<a href=\"" + fileUrl + "\">" + fileUrl + "</a>");
+        if (!fullContent.contains("{{bucketLink}}")) {
+            content.add(new Content("text/plain", fullContent));
+            content.add(linkContent);
+        } else {
+            int startLinkIndex = fullContent.indexOf("{{bucketLink}}");
```
yufengwng: Is this how things are done in Gen2 version of this export? Can we perhaps pass in the whole content to SendGrid, pass in data as part of the `Personalization` object, and leverage SendGrid's dynamic templating?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+                        studyDto.getId(), customExportStatus, customActivity);
+                long lastCompletionTime = optionalLastCompletionTime.orElse((long)0);
+                customExportDao.updateLastCompleted(lastCompletionTime, studyDto.getId());
+
+                return true;
+            });
+
+            if (!success) {
+                LOG.error("Error while running csv export for study {}", studyGuid);
+            }
```
yufengwng: Hmmm, `success` is never false?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -0,0 +1,224 @@
...
+
+        Optional<ActivityDto> activityDtoOptional = jdbiActivity.findActivityByStudyGuidAndCode(customGuid, customActivity);
+        if (activityDtoOptional.isEmpty()) {
+            LOG.error("Custom enrollment activity DTO not found");
+            return null;
+        }
+        ActivityDto activityDto = activityDtoOptional.get();
+        List<ActivityExtract> activities = extractVersionsOfActivity(handle, activityDto, customGuid);
+
+        LOG.info("Custom export found {} versions of enrollment activity", activities.size());
```
yufengwng:
```suggestion
        LOG.info("Custom export found {} versions of activity {}", activities.size(), customActivity);
```

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -0,0 +1,224 @@
...
+
+        List<String> excludedFields = new ArrayList<>(cfg.getStringList(CustomExportConfigFile.EXCLUDED_PARTICIPANT_FIELDS));
+        participantMetaFmt = new ParticipantMetadataFormatter(excludedFields);
+
+
+        List<String> headers = new LinkedList<>(participantMetaFmt.headers());
+
+        Map<String, Integer> activityTagToNormalizedMaxInstanceCounts = new HashMap<>();
+        Map<String, ActivityResponseCollector> responseCollectors = new HashMap<>();
+        Map<String, ActivityAttributesCollector> attributesCollectors = new HashMap<>();
```
yufengwng, line 137: Since this is custom code, you can probably get rid of the whole business with `activityTagToNormalizedMaxInstanceCounts` and `attributesCollectors`. Those aren't really applicable to RGP.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -0,0 +1,224 @@
...
+            try {
+                row = new LinkedList<>(participantMetaFmt.format(pt.getStatus(), pt.getUser()));
+                ComponentDataSupplier supplier = new ComponentDataSupplier(pt.getUser().getAddress(), pt.getProviders());
+                for (ActivityExtract activity : activities) {
+                    String activityTag = activity.getTag();
+                    int maxInstances = activityTagToNormalizedMaxInstanceCounts.get(activityTag);
+                    ActivityResponseCollector responseCollector = responseCollectors.get(activityTag);
+                    ActivityAttributesCollector attributesCollector = attributesCollectors.get(activityTag);
+
+                    List<ActivityResponse> instances = pt.getResponses(activityTag).stream()
```
yufengwng: Do we need those 4 checkboxes? If so, those are in the Prequal, so we'll get to grab it from there. Do we need family member data? Those are represented as "child activities", so we'll need to query for the activity definition for those and grab the data from those activity instances.

Also note that I've updated the question stable ids in RGP so they match closer to what we have in DatStat (in the hope that things will look more familiar for study staff). You'll need to update the configuration in Secret Manager.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/ExportUtil.java
```diff
@@ -0,0 +1,146 @@
...
+    public static final int DEFAULT_BATCH_SIZE = 100;
+    public static final int READER_BUFFER_SIZE_IN_BYTES = 10 * 1024;
+
+    public static <R, X extends Exception> R withAPIsTxn(HandleCallback<R, X> callback) throws X {
+        return TransactionWrapper.withTxn(TransactionWrapper.DB.APIS, callback);
+    }
+
+    public static String makeExportCSVFilename(String baseFileName, Instant timestamp) {
+        DateTimeFormatter fmt = DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmssX").withZone(ZoneOffset.UTC);
+        return String.format("%s%s.csv", fmt.format(timestamp), baseFileName);
```
yufengwng: This format is different from what we have before. We're missing an underscore. And the argument order is swapped.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/SecretUtil.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.gson.JsonParser;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import org.broadinstitute.ddp.constants.ConfigFile;
+import org.broadinstitute.ddp.exception.DDPException;
+
+public class SecretUtil {
+    public static Config getConfigFromSecret(Config mainCfg, String secretName) {
+        try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
+            SecretVersionName versionName = SecretVersionName.of(mainCfg.getString(ConfigFile.GOOGLE_PROJECT_ID), secretName, "latest");
```
yufengwng: Looks like GCP project id is the only thing we need here. Can we just pass that in instead of a whole config object?

pepper-apis/src/main/resources/db-changes/schema/DDP-5096-custom-export-last-completed.xml
charlottedye: Good point.  I think I did it this way out of habit

pepper-apis/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
charlottedye: So the reason we need this is that in production, the bucket that gets used is actually a Terra bucket, so there are some weird access constraints on it.  The Java API in particularly handles this badly, so this is a workaround to handle that.

yufengwng: Ah okay. So if I'm understanding things correctly, I see that the `BlobListOption.prefix("")` part means to list all the top-level folders in the bucket. But I think `getBlobId().getBucket()` just returns a string, which is exactly what `bucketName` is. So I don't see what the difference is here... Is `storage.list()` doing something else important for working around access constraints? (is it caching some bucket object or something while doing the listing?)

Also, will there be a case where there's nothing in the bucket? In this case, there's nothing to iterate from the `Page`, so this will throw an exception. Is that what we're expecting?

pepper-apis/parent-pom.xml
charlottedye: It's for secret manager.

pepper-apis/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
charlottedye: You're right; this doesn't make any sense.  Plus then I looked at the RGP backend code and I can't even find the code that does this.  I could have sworn I got it from there, too.  Let me switch this to use the existing getBucket method and test it against the Terra bucket and see what happens.  Maybe I was looking at an old version of the code?

pepper-apis/parent-pom.xml
yufengwng: I don't know all the details about how maven does dependency resolution, but I don't think we need to include the BOM if we pin the dependency to a specific version. I see we have already spelled out `<version>1.4.2</version>` for the secret manager dependency, so how about we remove this BOM?

charlottedye: I'll look into this more but I was getting errors when I tried removing it.  Not totally sure what the bom is, actually

pepper-apis/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
charlottedye: I switched it to use the existing getBucket method and everything was fine.  I must have been looking at an old version of the RGP code or something.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
charlottedye: We don't need the 4 checkboxes but we will need the child activities.  Are the nested activities and updated question IDs in develop?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
charlottedye: Good idea

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
yufengwng: Just merged in my PR, so nested activities and updated question IDs are in `develop` now!

charlottedye: Nice!

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
```diff
@@ -55,13 +55,17 @@
     // whether to use local filesystem for pdf archiving or not
     public static final String PDF_ARCHIVE_USE_FILESYSTEM = "pdfArchiveUseFilesystem";
 
+    // config file for custom export job
+    public static final String CUSTOM_EXPORT_CONFIG_SECRET = "custom-export";
```
yufengwng: Small nitpick: this config file has traditionally use camelCase. Also, I'll recommend renaming it to something like `customExportSecretId` or similar.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/SecretUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class SecretUtil {
+    public static Config getConfigFromSecret(String googleProjectId, String secretName) {
+        try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
+            SecretVersionName versionName = SecretVersionName.of(googleProjectId, secretName, "latest");
+            AccessSecretVersionResponse response = client.accessSecretVersion(versionName);
+            String secret = response.getPayload().getData().toStringUtf8();
+            secret = secret.replaceAll("\n", "");
+            JsonElement elem = new JsonParser().parse(secret);
+            return ConfigFactory.parseString(new Gson().toJson(elem));
```
yufengwng: Curious why we need to transform it before passing it into the parse method. The content of the secret is just a `.conf` file right? I feel like `ConfigFactory` should be able to handle it.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
```diff
@@ -334,6 +342,14 @@ private void collectComponentMappings(Map<String, Object> props, ComponentBlockD
             if (definition.getActivityType() == ActivityType.FORMS) {
                 flattenHeadersByOrderedDepthTraversal((FormActivityDef) definition);
             }
+            if (excludedFields != null) {
+                responseHeaders.removeAll(excludedFields);
```
yufengwng, line 346: Should we remove from `deprecatedHeaders` too? Will there be a chance an exclude-field is in there? On the other hand, I'm not too sure how we'll handle first-fields for `deprecatedHeaders`... maybe it's safe to punt on first-fields?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/housekeeping/schedule/CustomExportJob.java
```diff
@@ -0,0 +1,144 @@
...
+        String schedule = ConfigUtil.getStrIfPresent(cfg, ConfigFile.CUSTOM_EXPORT_SCHEDULE);
+        if (schedule == null || schedule.equalsIgnoreCase("off")) {
+            LOG.warn("Job {} is set to be turned off: not scheduled", getKey());
+            return;
+        }
+
+        GoogleCredentialUtil.initCredentials(true);
+        CustomExportJob.cfg = cfg;
+        exportCfg = SecretUtil.getConfigFromSecret(cfg.getString(ConfigFile.GOOGLE_PROJECT_ID),
+                ConfigFile.CUSTOM_EXPORT_CONFIG_SECRET);
```
yufengwng:
```suggestion
                cfg.getString(ConfigFile.CUSTOM_EXPORT_CONFIG_SECRET));
```

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,346 @@
...
+            if (res.hasThrown()) {
+                LOG.error("Sending notification email for study {} with skipped custom export failed", customDto.getGuid(),
+                        res.getThrown());
+            } else if (res.hasError()) {
+                LOG.error("Sending notification email for study {} with skipped custom export failed", customDto.getGuid(),
+                        res.getError());
+            } else if (res.hasFailure()) {
+                LOG.error("Sending notification email for study {} with skipped custom export failed with status code {}",
+                        customDto.getGuid(), res.getStatusCode());
+            }
```
yufengwng: I don't think we need a `hasFailure()` branch. We should log the status code in the `hasError()` branch.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    private Mail createMail(String subject, Email fromEmail, Email toEmail, List<Content> content, String fileUrl) {
+        Mail mail = new Mail();
+        mail.setFrom(fromEmail);
+        mail.setSubject(subject);
+        Personalization p = new Personalization();
+        p.addTo(toEmail);
+        mail.addPersonalization(p);
+        p.addSubstitution("{{bucketLink}}", "<a href=\"" + fileUrl + "\">" + fileUrl + "</a>");
```
yufengwng: `addSubstitution` are for the legacy template syntax (e.g. `-ddp.baseWebUrl-`). I think what you're looking for is `addDynamicTemplateData`. Also, it'll be good to cleanup the `addSuccessContent()` if we're going with dynamic templating approach.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,93 @@
...
+            @Bind("studyId") long studyId,
+            @Bind("lastCompleted") Instant lastCompleted
+    );
+
+    @SqlQuery("select * from custom_export where study_id = :id")
+    @RegisterConstructorMapper(CustomExportDto.class)
+    Optional<CustomExportDto> findExportDtoByStudyId(@Bind("id") long id);
+
+    @SqlQuery("select custom_export_last_completed from custom_export where study_id = :id")
+    Optional<Timestamp> findExportLastCompletedByStudyId(@Bind("id") long id);
```
yufengwng: We should it directly to `Instant` instead of the sql data type. JDBI is able to handle that conversion.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @SqlUpdate("update custom_export set custom_export_last_completed = :lastCompleted where study_id = :studyId")
+    int updateLastCompletedByStudyId(@Bind("lastCompleted") Instant lastCompleted, @Bind("studyId") long studyId);
+
+    @SqlQuery("SELECT CASE WHEN ( "
+            + "SELECT COUNT(study_activity_id) FROM ( "
+            + "SELECT sa.study_activity_id FROM study_activity AS sa  "
```
yufengwng: Shouldn't this query look for _activity instance ids_ instead of `study_activity_id`? `study_activity` are the activity definitions, which doesn't change, whereas `activity_instance` holds instances for each user. Also beware that each activity instance will have multiple `activity_instance_status` rows (especially if they're completed), so we're going to end up with a cross-product here, but I guess that's okay since we're just trying to figure out if there's at least one new instance.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,93 @@
...
+            + ") > "
+            + "0 THEN true ELSE false END")
+    boolean needCustomExport(@Bind("studyId") long studyId,
+                             @Bind("statusType") String statusType,
+                             @Bind("lastCompletion") long lastCompletion,
+                             @Bind("activityCode") String activityCode);
+
+    @SqlQuery("SELECT usen.user_id FROM user_study_enrollment AS usen "
+            + "JOIN study_activity AS sa ON sa.study_id = usen.study_id "
+            + "JOIN activity_instance AS aci ON usen.user_id = aci.participant_id "
```
yufengwng, line 70: I think we should add another join clause like
```
AND aci.study_activity_id = sa.study_activity_id
```
otherwise I suspect we might get some weird combination of activities to instances.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,93 @@
...
+                                        @Bind("offset") long offset);
+
+    @SqlQuery("SELECT MAX(aci.first_completed_at) FROM activity_instance AS aci "
+            + "JOIN user_study_enrollment AS usen ON aci.participant_id = usen.user_id "
+            + "JOIN study_activity AS sa ON aci.study_activity_id = sa.study_activity_id "
+            + "JOIN activity_instance_status AS ais ON aci.activity_instance_id = ais.activity_instance_id "
+            + "JOIN activity_instance_status_type AS aist ON aist.activity_instance_status_type_id = ais.activity_instance_status_type_id "
+            + "WHERE  sa.study_id = :studyId AND aist.activity_instance_status_type_code = :statusType  "
+            + "AND sa.study_activity_code=:activityCode AND usen.valid_to IS NULL")
+    Optional<Long> getLastCustomCompletionDate(@Bind("studyId") long studyId,
```
yufengwng: Hmmm, I feel like we should update `findCustomUserIdsToExport()` to return both the user id as well as the `first_completed_at` timestamp, so we can keep track of the max as we iterate through the data set. The scenario I'm worried about is, what happens if we finish iterating, and right before we call this get method another user completes their survey?

yufengwng (commented):
Looking really good! I have a few more comments.

pepper-apis/parent-pom.xml
yufengwng: Any update on removing the BOM?

charlottedye: I keep getting this error:
`Exception in thread "main" java.util.ServiceConfigurationError: io.grpc.NameResolverProvider: io.grpc.grpclb.SecretGrpclbNameResolverProvider$Provider Unable to get public no-arg constructor
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:581)
	at java.base/java.util.ServiceLoader.getConstructor(ServiceLoader.java:672)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1232)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1264)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1299)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1384)
	at io.grpc.ServiceProviders.loadAll(ServiceProviders.java:67)
	at io.grpc.NameResolverProvider.<clinit>(NameResolverProvider.java:53)
	at io.grpc.internal.AbstractManagedChannelImplBuilder.<clinit>(AbstractManagedChannelImplBuilder.java:86)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelProvider.builderForAddress(NettyChannelProvider.java:37)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelProvider.builderForAddress(NettyChannelProvider.java:23)
	at io.grpc.ManagedChannelBuilder.forAddress(ManagedChannelBuilder.java:37)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createSingleChannel(InstantiatingGrpcChannelProvider.java:293)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.access$1700(InstantiatingGrpcChannelProvider.java:74)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider$1.createSingleChannel(InstantiatingGrpcChannelProvider.java:220)
	at com.google.api.gax.grpc.ChannelPool.create(ChannelPool.java:72)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createChannel(InstantiatingGrpcChannelProvider.java:227)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.getTransportChannel(InstantiatingGrpcChannelProvider.java:210)
	at com.google.api.gax.rpc.ClientContext.create(ClientContext.java:169)
	at com.google.cloud.secretmanager.v1.stub.GrpcSecretManagerServiceStub.create(GrpcSecretManagerServiceStub.java:250)
	at com.google.cloud.secretmanager.v1.stub.SecretManagerServiceStubSettings.createStub(SecretManagerServiceStubSettings.java:342)
	at com.google.cloud.secretmanager.v1.SecretManagerServiceClient.<init>(SecretManagerServiceClient.java:153)
	at com.google.cloud.secretmanager.v1.SecretManagerServiceClient.create(SecretManagerServiceClient.java:134)
	at com.google.cloud.secretmanager.v1.SecretManagerServiceClient.create(SecretManagerServiceClient.java:125)
	at org.broadinstitute.ddp.util.SecretUtil.getConfigFromSecret(SecretUtil.java:18)
	at org.broadinstitute.ddp.customexport.housekeeping.schedule.CustomExportJob.register(CustomExportJob.java:65)
	at org.broadinstitute.ddp.Housekeeping.setupScheduler(Housekeeping.java:516)
	at org.broadinstitute.ddp.Housekeeping.start(Housekeeping.java:237)
	at org.broadinstitute.ddp.Housekeeping.main(Housekeeping.java:204)
Caused by: java.lang.VerifyError: Cannot inherit from final class
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:802)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:700)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:623)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3138)
	at java.base/java.lang.Class.getConstructor0(Class.java:3343)
	at java.base/java.lang.Class.getConstructor(Class.java:2152)
	at java.base/java.util.ServiceLoader$1.run(ServiceLoader.java:659)
	at java.base/java.util.ServiceLoader$1.run(ServiceLoader.java:656)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.util.ServiceLoader.getConstructor(ServiceLoader.java:667)
	... 27 more`

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @SqlQuery("SELECT usen.user_id as user_id, aci.first_completed_at as completed_time FROM user_study_enrollment AS usen "
+            + "JOIN study_activity AS sa ON sa.study_id = usen.study_id "
+            + "JOIN activity_instance AS aci ON usen.user_id = aci.participant_id "
+            + "AND aci.study_activity_id = sa.study_activity_id "
+            + "JOIN activity_instance_status AS ais ON aci.activity_instance_id = ais.activity_instance_id "
+            + "JOIN activity_instance_status_type AS aist ON aist.activity_instance_status_type_id = ais.activity_instance_status_type_id "
+            + "WHERE usen.study_id = :studyId AND aist.activity_instance_status_type_code=:statusType "
+            + "AND aci.first_completed_at>:lastCompletion AND usen.valid_to IS NULL AND sa.study_activity_code=:activityCode "
+            + "GROUP BY usen.user_id, aci.first_completed_at ORDER BY MAX(aci.first_completed_at) DESC LIMIT :limit OFFSET :offset")
```
yufengwng: Have we considered sorting in `ASC` order instead? I think it will be safer that way. If we're in the middle of paging through the result set, and a couple more participants finish their survey, won't that throw off our iteration since we're effectively starting at the back of the iteration? (I think this is called "phantom" paging or something, not sure...) If we sort by ASC, then any newly finished ones will be at the end of the iteration and the results we have paged through so far won't be affected.

yufengwng (commented):
Looks good! Just one more comment to consider.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/db/dao/CustomExportDao.java
charlottedye: Good point!

yufengwng (approved):
Looks ready-to-go as far as I can tell!

PR #689, opened 2021-03-29
DDP-5708 rgp: refactor questions and add version 2

yufengwng:
## Context

DDP-5708. For data migration purposes, I refactored the configurations so it's easier to know what questions / picklist options are. Added missing old questions/options. Added patch code to create version 2 of surveys.

## Checklist

## FUD Score

- [x] Need to redeploy RGP in dev

## Testing

## Release

- [x] Part of RGP migration/release.


study-builder/studies/rgp/i18n/es.conf
```diff
@@ -1,11 +1,12 @@
 {
   //Prequal
   "prequal_name": "[es]Tell us your story",
-  "prompt_prequal_question": "[es]I am, or am applying on behalf of, a patient who: ",
-  "prequal_understands_english": "[es]Understands English",
-  "prequal_lives_us": "[es]Lives in the United States",
-  "prequal_has_condition": "[es]Has a rare and genetically undiagnosed condition",
-  "prequal_has_doctor": """[es]Is currently under the care of a doctor that is helping to
+  "prompt_general_info": "[es]I am, or am applying on behalf of, a patient who: ",
+  "general_info_english": "[es]Understands English",
```
charlottedye: I lost where you mentioned this, but you're right: this should be "Understands English or Spanish"

study-builder/studies/rgp/i18n/en.conf
```diff
@@ -1,11 +1,12 @@
 {
   //Prequal
   "prequal_name": "Tell us your story",
-  "prompt_prequal_question": "I am, or am applying on behalf of, a patient who: ",
-  "prequal_understands_english": "Understands English",
-  "prequal_lives_us": "Lives in the United States",
-  "prequal_has_condition": "Has a rare and genetically undiagnosed condition",
-  "prequal_has_doctor": """Is currently under the care of a doctor that is helping to
+  "prompt_general_info": "I am, or am applying on behalf of, a patient who: ",
+  "general_info_english": "Understands English",   // or Spanish?
```
charlottedye: You're right: this should be "Understands English or Spanish"

study-builder/studies/rgp/enrollment-questions.conf
```diff
@@ -0,0 +1,4245 @@
+{
+  // We use `include`s to reduce duplication and file sizes. However, HOCON parser has a hard time
```
charlottedye, line 2: Thanks for this: I ran into this stumbling block with the HOCON parser, too

study-builder/studies/rgp/i18n/en.conf
yufengwng: Fixed!

study-builder/studies/rgp/i18n/en.conf
```diff
@@ -116,127 +121,88 @@
   condition?""",
   "website_prompt_optional": "(optional)",
   "website_placeholder": "If yes, provide the link here",
```
charlottedye: Missed this in the previous version: there should be a period after "provide the link here"

study-builder/studies/rgp/i18n/es.conf
```diff
@@ -116,127 +121,88 @@
   condition?""",
   "website_prompt_optional": "[es](optional)",
   "website_placeholder": "[es]If yes, provide the link here",
```
charlottedye: Missed this in the previous version: there should be a period after "provide the link here"

study-builder/studies/rgp/enrollment.conf
```diff
@@ -8535,133 +746,26 @@
...
-                  ]
-                }
-              }
-            ]
-          }
+          "question": ${_questions.agreement}
+        },
+
+        //Questions no longer used
+        {
```
charlottedye, line 758: I would consider moving this question above or below the current ethnicity question.  It could also be reasonable to leave it here or put it just above the agreement checkboxes, but I think this was just an earlier version of the ethnicity question.

study-builder/studies/rgp/enrollment.conf
```diff
@@ -8535,133 +746,26 @@
...
+          "blockType": "QUESTION",
+          "shownExpr": null,
+          "question": ${_questions.ethnicity_categories}
+        },
+        {
+          "blockType": "QUESTION",
```
charlottedye, line 764: I'm not sure where this question went, but similar to the above, I would consider putting this at the top of the third page (I believe the third page questions were added when this one was remove, but I couldn't swear to it) or putting it above the agreement checkboxes (just because logically I think it makes sense to have them last).  I could be convinced to agree with any of these locations, though; there isn't really a right answer given that this version of the survey never really existed.

charlottedye (changes requested):
Just a few picky comments: this seems good to me!

study-builder/studies/rgp/enrollment.conf
yufengwng: > given that this version of the survey never really existed

Interesting note...

---

So the way I went about it with these two questions is by marking them `isDeprecated`. What this flag does is to indicate that this question is no longer used. It exists in the question definition, but won't be displayed in any way to the user. This is nice so that there is a place we can migrate data to for these two questions, but users won't see them at all, and we don't really need to fiddle with versioning here since they're already hidden. We have actually used this feature when migrating one of the other studies, if I recall correctly.

As you have pointed out, looking at the nature of these two questions, it appears to me that they're a variation of the `ethnicity` question and the set of "biological family" questions we currently have. So I figured we can just hide these, even for participants with v1 of the survey. Does this sound good, or should I approach this differently?

study-builder/studies/rgp/enrollment.conf
yufengwng: This question is not visible to users, so I think it's fine wherever we put it. Leaving it here for now.

study-builder/studies/rgp/enrollment.conf
yufengwng: Leaving this question here for now, unless you have a strong opinion here.

study-builder/studies/rgp/enrollment.conf
charlottedye: Ah!  Okay

study-builder/studies/rgp/enrollment.conf
charlottedye: That seems reasonable to me

charlottedye (approved):

PR #695, opened 2021-04-08
DDP-5571 support canDeleteFistInstance flag for nested activities

yufengwng:
## Context

This adds a new flag, `canDeleteFirstInstance`, in backend activity definition that will control whether an instance of a nested activity can be deleted or not.

## Checklist

## FUD Score

Need a study that uses this configuration, but you can try it out locally by tweaking existing activity definitions (such as Osteo's Family History).

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/DeleteActivityInstanceRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/DeleteActivityInstanceRoute.java
@@ -60,10 +60,12 @@ public class DeleteActivityInstanceRoute implements Route {
                     activityDef.getCanDeleteFirstInstance(),
                     isFirstInstance);
 
-            if (!activityDef.canDeleteInstances()) {
+            if (!canDelete && !activityDef.canDeleteInstances()) {
                 throwNotAllowed(response, "Activity does not allow deleting instances");
-            } else if (!canDelete) {
+            } else if (!canDelete && isFirstInstance) {
                 throwNotAllowed(response, "Activity does not allow deleting the first instance");
+            } else if (!canDelete) {
+                throwNotAllowed(response, "Activity instance is not allowed to be deleted");
             } else if (instanceDto.getParentActivityCode() == null) {
                 throwNotAllowed(response, "Deleting non-nested activity instances is not supported");
             }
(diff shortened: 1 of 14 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/DeleteActivityInstanceRoute.java
```diff
@@ -53,16 +53,19 @@ public ActivityInstanceSummary handle(Request request, Response response) {
...
+                    activityDef.getCanDeleteFirstInstance(),
+                    isFirstInstance);
+
             if (!activityDef.canDeleteInstances()) {
-                String msg = "Activity does not allow deleting instances";
-                LOG.warn(msg);
-                throw ResponseUtil.haltError(response, HttpStatus.SC_UNPROCESSABLE_ENTITY,
-                        new ApiError(ErrorCodes.OPERATION_NOT_ALLOWED, msg));
+                throwNotAllowed(response, "Activity does not allow deleting instances");
+            } else if (!canDelete) {
```
evgeniipr: `canDelete` means can delete an instance (even if it's not a first one). But error text says that first instance cannot be deleted..
What if `computeCanDelete` logic will be changed (for example, added a check of deleting a last instance): in that case the message could be incorrect.
May be to change it like:
```
boolean canDeleteFirstInstance = isFirstInstance + activityDef.getCanDeleteFirstInstance();
"Activity deletion not possible." + (!canDeleteFirstInstance ?  " First instance cannot be deleted." : "")
```
Or at least for more clear logic I'd suggest to set if-s like:
```
if (activityDef.canDeleteInstances()) {
   if (!canDelete) {
         throwNotAllowed(response, "Activity does not allow deleting the first instance");
   }
} else {
      throwNotAllowed(response, "Activity does not allow deleting instances");
}
```

yufengwng: Good point! I'll change it.

yufengwng: Should be better now!

evgeniipr (approved):
Looks ok

PR #698, opened 2021-04-12
DDP-5822 . Multi Language support for Pdf generation.

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-5822
Added platform level support to handle multi language pdf generation
Added StudyBuilder support to handle multi language pdf templates in pdf configuration

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfTemplate.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfTemplate.java
@@ -12,16 +12,24 @@ public abstract class PdfTemplate {
     private long id;
     private PdfTemplateType type;
     private byte[] rawBytes;
+    private Long languageCodeId;
 
-    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes) {
+    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes, Long languageCodeId) {
         this.id = id;
         this.type = type;
         this.rawBytes = rawBytes;
+        this.languageCodeId = languageCodeId;
     }
 
     PdfTemplate(PdfTemplateType type, byte[] rawBytes) {
         this.type = type;
         this.rawBytes = rawBytes;
+    }
+
+    PdfTemplate(PdfTemplateType type, byte[] rawBytes, Long languageCodeId) {
+        this.type = type;
+        this.rawBytes = rawBytes;
+        this.languageCodeId = languageCodeId;
     }
 
     public long getId() {
@@ -38,6 +46,10 @@ public abstract class PdfTemplate {
 
     public byte[] getRawBytes() {
         return rawBytes;
+    }
+
+    public Long getLanguageCodeId() {
+        return languageCodeId;
     }
 
     /**
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfTemplate.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfTemplate.java
@@ -12,21 +12,16 @@ public abstract class PdfTemplate {
     private long id;
     private PdfTemplateType type;
     private byte[] rawBytes;
-    private Long languageCodeId;
+    private long languageCodeId;
 
-    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes, Long languageCodeId) {
+    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes, long languageCodeId) {
         this.id = id;
         this.type = type;
         this.rawBytes = rawBytes;
         this.languageCodeId = languageCodeId;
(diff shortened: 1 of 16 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/model/pdf/PdfTemplate.java
```diff
@@ -12,18 +12,26 @@
     private long id;
     private PdfTemplateType type;
     private byte[] rawBytes;
+    private Long languageCodeId;
 
-    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes) {
+    PdfTemplate(long id, PdfTemplateType type, byte[] rawBytes, Long languageCodeId) {
```
yufengwng: Can we have the subclasses default to `en` language code id, and then we can make this non-null?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/PdfBuilder.java
```diff
@@ -268,23 +270,33 @@ private PdfTemplate buildPdfTemplate(Handle handle, Config fileCfg) {
...
         } else if (InstitutionType.INSTITUTION.name().equals(type)) {
-            return buildProviderTemplate(fileCfg, InstitutionType.INSTITUTION, rawBytes);
+            return buildProviderTemplate(fileCfg, InstitutionType.INSTITUTION, rawBytes, languageCodeId);
         } else {
             throw new DDPException("Unsupported pdf template file type " + type);
         }
     }
 
-    private PdfTemplate buildMailingAddressTemplate(Config fileCfg, byte[] rawBytes) {
+    private PdfTemplate buildMailingAddressTemplate(Config fileCfg, byte[] rawBytes, Long languageCodeId) {
```
yufengwng: We should make these methods take a non-null language code id.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/PdfBuilder.java
```diff
@@ -268,23 +270,33 @@ private PdfTemplate buildPdfTemplate(Handle handle, Config fileCfg) {
             throw new DDPException(e);
         }
 
+        Long languageCodeId = LanguageStore.getDefault().getId();
```
yufengwng: Will it be better if we try figuring out which language the study has set as the "default" and using that instead?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,16 +139,28 @@ public InputStream generateFlattenedPdfForConfiguration(PdfConfiguration configu
                                        Handle handle) throws IOException {
         Participant participant = loadParticipantData(handle, configuration, userGuid);
         Map<Long, ActivityResponse> instances = loadActivityInstanceData(handle, configuration, participant);
+        Long languageId = null;
+        if (participant.getUser().getProfile() != null) {
+            languageId = participant.getUser().getProfile().getPreferredLangId();
+        }
+        if (languageId == null) {
+            //use default 'en'
+            languageId = LanguageStore.getDefault().getId();
```
yufengwng: Could we perhaps figure the study's "default" language and use that? Might avoid situations where our platform default is English but the study doesn't have English and their default is instead French or something.

yufengwng (commented):
Looks good! Just a few small questions.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -138,16 +140,35 @@ public InputStream generateFlattenedPdfForConfiguration(PdfConfiguration configu
...
         Map<Long, ActivityResponse> instances = loadActivityInstanceData(handle, configuration, participant);
+        Long languageId = null;
+        if (participant.getUser().getProfile() != null) {
+            languageId = participant.getUser().getProfile().getPreferredLangId();
+        }
+        if (languageId == null) {
+            //use study default language if exists
+            StudyLanguageDao studyLanguageDao = handle.attach(StudyLanguageDao.class);
+            List<Long> defaultLanguages = studyLanguageDao.getStudyLanguageSql().selectDefaultLanguageCodeId(configuration.getStudyId());
+            if (org.apache.commons.collections4.CollectionUtils.isNotEmpty(defaultLanguages)) {
```
yufengwng: Ah, I see we have already imported `org.apache.commons.collections.CollectionUtils`. Maybe we should update code to use the `collections4` one? We can also just drop the usage of `isNotEmpty()` and simply check `!defaultLanguages.isEmpty()`. JDBI won't return nulls if return type is a collection.

yufengwng (dismissed):
LGTM

yufengwng (approved):

PR #699, opened 2021-04-12
Add gitignore rules to ignore more config files

yufengwng:
We had a hiccup where we checked in secret `.conf` files. Tweaking our gitignore rules so we don't run into this again.


MocanaAtBroad (approved):

bskinner (approved):
👍

PR #700, opened 2021-04-12
Better handling of config loading when using CLI tools

yufengwng:
## Context

When we're running CLI tools, we get this annoying message about failing to initialize SlackAppender. This PR addresses that issue by fixing handling of config loading. Essentially, the idea is to handle nulls in our static blocks. See comments for more details.

_What about production config loading in study-server and housekeeping?_
* These use the config extensively, so we will have gotten NullPointerException very early in the process if config is not provided.

_What about the study-builder CLI tool?_
* I tried it out locally and study-builder CLI works as normal.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
@@ -36,7 +36,10 @@ public class TransactionWrapper {
     // https://bugs.mysql.com/bug.php?id=91112
     // Note 03/10/2020: Release of fixed Java connector is imminent and this might no longer be necessary
     static {
-        TimeZone.setDefault(TimeZone.getTimeZone(ConfigManager.getInstance().getConfig().getString(ConfigFile.DEFAULT_TIMEZONE)));
+        ConfigManager manager = ConfigManager.getInstance();
+        if (manager != null && manager.getConfig() != null) {
+            TimeZone.setDefault(TimeZone.getTimeZone(manager.getConfig().getString(ConfigFile.DEFAULT_TIMEZONE)));
+        }
     }
 
     private static final Logger LOG = LoggerFactory.getLogger(TransactionWrapper.class);
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
@@ -67,26 +67,6 @@ public class SlackAppender<E> extends AppenderBase<ILoggingEvent> {
                       String slackChannel,
                       Integer queueSize,
                       Integer intervalInMillis) {
-        httpClient = HttpClient.newHttpClient();
-        try {
-            this.slackHookUrl = new URI(slackHookUrl);
-        } catch (URISyntaxException e) {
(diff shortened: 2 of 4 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
```diff
@@ -54,7 +54,7 @@ private ConfigManager(Config cfg) {
      * @return the config for the app
      */
     public static synchronized ConfigManager getInstance() {
-        if (configManager == null) {
+        if (configManager == null && TYPESAFE_CONFIG_FILE != null) {
```
yufengwng, line 57: When we're running CLI tools (like `AesUtil`), the first thing that runs is the SlackAppender constructor, which calls this `getInstance()` method to grab the config object. Somehow, the `static` block in this file doesn't get run, which means `TYPESAFE_CONFIG_FILE` variable is null. So we add null handling here. Idea here is: if there's no config file, we won't initialize the config manager, so callers need to handle null config managers.

pepper-apis/src/main/java/org/broadinstitute/ddp/log/SlackAppender.java
```diff
@@ -67,26 +67,6 @@ private void init(String slackHookUrl,
                       String slackChannel,
                       Integer queueSize,
                       Integer intervalInMillis) {
-        httpClient = HttpClient.newHttpClient();
```
yufengwng, line 70: We might have a config object, but that doesn't mean it'll have slack configuration properties. So we're moving these inside the `if` block below.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -36,7 +36,10 @@
     // https://bugs.mysql.com/bug.php?id=91112
     // Note 03/10/2020: Release of fixed Java connector is imminent and this might no longer be necessary
     static {
-        TimeZone.setDefault(TimeZone.getTimeZone(ConfigManager.getInstance().getConfig().getString(ConfigFile.DEFAULT_TIMEZONE)));
+        ConfigManager manager = ConfigManager.getInstance();
+        if (manager != null && manager.getConfig() != null) {
```
yufengwng, line 40: Another `static` block that uses the config manager, so we handle null here.

pepper-apis/src/main/java/org/broadinstitute/ddp/security/AesUtil.java
```diff
@@ -159,10 +159,12 @@ public static void main(String[] args) throws Exception {
 
         if (cmd.hasOption('d')) {
             System.out.println(AesUtil.decrypt(cmd.getOptionValue('s'), encryptionSecret));
+            return;
```
yufengwng, line 162: Each flag in AesUtil does one thing, so we return after finishing that one thing.

yufengwng:
Some screenshots to illustrate (CC @bskinner):

---

Before:
Screen Shot 2021-04-12 at 7 27 49 PM

---

After:
Screen Shot 2021-04-12 at 7 25 18 PM

ssettipalli (approved):
looks good to me

PR #702, opened 2021-04-13
DDP-5938 prion: fix syntax issue in password reset page

yufengwng:
Another syntax fix. `password_complexity_options` is an object, so when templating is done, we end up with something like this:

```
password_complexity_options:  "{"min_length": 8}",
```
which is an error due to the unescaped quotes. The fix is to remove the outer set of quotes since we want it to be substituted as a whole object.

See here for more: https://community.auth0.com/t/hosted-password-reset-page-with-custom-length-password-at-connection-level-not-working/21308.

### DIFF
--- a/study-builder/tenants/prion/tenant.yaml
+++ b/study-builder/tenants/prion/tenant.yaml
@@ -1,5 +1,6 @@
 tenant:
   default_directory: Username-Password-Authentication
+  picture_url: '##BASE_URL##/assets/images/project-logo-dark.svg'
 
 rules:
   - name: Register User in Pepper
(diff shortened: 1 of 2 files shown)


study-builder/tenants/prion/tenant.yaml
```diff
@@ -1,5 +1,6 @@
 tenant:
   default_directory: Username-Password-Authentication
+  picture_url: '##BASE_URL##/assets/images/project-logo-dark.svg'
```
yufengwng, line 3: The password reset page has this:
```
    theme: {
        icon: "{{tenant.picture_url | default: '//cdn.auth0.com/styleguide/1.0.0/img/badge.png'}}",
            primaryColor: "{{tenant.colors.primary | default: '#ea5323'}}"
    }
```
I noticed on dev `tenant.picture_url` was pointing to some old `dev.prionregistry.org` URL. And on other environments, test/staging and even production does not have anything set for this. So I updated this to point to prion's project logo.

yufengwng:
Going to move forward with this now!

PR #703, opened 2021-04-13
DDP-5802 - use new AI builder in PutFormAnswersRoute

evgeniipr:
## Context
In this code improvement the `ActivityInstanceFromDefinitionBuilder` is used it in `PutFormAnswerRoute` 
instead of the following code:
```
        FormInstance form = formInstanceDao.getBaseFormByGuid(handle, instanceGuid, isoLangCode);
        formInstanceDao.loadAllSectionsForForm(handle, form, langCodeId);
        form.updateBlockStatuses(handle, interpreter, userGuid, operatorGuid, instanceGuid, instanceSummary);
```
Short info about `ActivityInstanceFromDefinitionBuilder` (shortly "AI builder"): initially it was implemented as a replacement of existing code providing `ActivityInstance` object creation in `GetAcrtivityInstanceRoute`: the old code always get study static and instance data from DB; the new AI builder gets static data from `ActivityDefStore` cache - and this approach improved performance of an activity loading more than 2 times.
In `PutFormAnswerRoute` it was needed to use a "cutted" version of AI builder, therefore the builder is refactored in order to be possible to execute only some **steps** of building: so the AI builder called not as a single method but with using of a **builder pattern** when we can specify which steps to apply and which to skip. 
Example (here is a version of AI builder used in `PutFormAnswerRoute`: steps `renderContent()` and `setDisplayNumbers()` are not applied):
```
var formInstance = new ActivityInstanceFromDefinitionBuilder(handle,
                createParams(userGuid, studyGuid, instanceGuid)
                        .setOperatorGuid(operatorGuid)
                        .setIsoLangCode(isoLangCode)
                        .setInstanceSummary(instanceSummary))
                .checkParams()
                    .readFormInstanceData()
                    .readActivityDef()
                .startBuild()
                    .buildFormInstance()
                    .buildFormChildren()
                    .renderFormTitles()
                    .updateBlockStatuses()
                .endBuild()
                    .getContext()
                    .getFormInstance();
```
NOTE: in the ticket https://broadinstitute.atlassian.net/browse/DDP-5758 it will need to do the following change (and new builder approach will suit fine):  
- create AI builder instance at the beginning of `GetActivityInstanceRoute.handle()`;
- call `readFromInstanceData()` (instead of `RouteUtil.findAccessibleInstanceOrHalt()`);
- the rest of AI building steps called aftewards.

## Checklist

## FUD Score

I fill ok, but it required an approvement about new design of `ActivityInstanceFromDefinitionBuilder`.

It can be detected a performance improvement of a form submit.

## Testing
Test will be implemented after approving the design of `ActivityInstanceFromDefinitionBuilder`.

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormSectionCreator.java
```diff
@@ -5,6 +5,7 @@
 
 import org.broadinstitute.ddp.model.activity.definition.FormSectionDef;
 import org.broadinstitute.ddp.model.activity.instance.FormSection;
+import org.broadinstitute.ddp.service.actvityinstancebuilder.context.AIBuilderContext;
```
yufengwng, line 8: Unused import?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -81,7 +87,7 @@ public ActivityInstanceService(ActivityInstanceDao actInstanceDao, PexInterprete
      * @return activity instance, if found
      * @throws DDPException if pex evaluation error
      * @deprecated Thus method should be removed as soon as
-     * {@link #buildInstanceFromDefinition(Handle, String, String, String, String, ContentStyle, String)} is carefully tested
+     *     @link #buildInstanceFromDefinition(Handle, String, String, String, String, ContentStyle, String)} is carefully tested
```
yufengwng: Missing an opening curly brace in the javadoc here?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -551,8 +566,71 @@ public void deleteInstance(Handle handle, ActivityInstanceDto instanceDto) {
...
+     */
+    public Optional<ActivityInstance> buildInstanceFromDefinition(
+            Handle handle,
+            String userGuid,
+            String operatorGuid,
+            String studyGuid,
+            String instanceGuid,
+            String isoLangCode,
+            UserActivityInstanceSummary instanceSummary) {
+
```
yufengwng, line 614: This looks nice, but I don't think this buys us much. What we wanted in the PUT route was to avoid rendering templates. We only need to load the form structure, load the answer data, and the block statuses -- that's all we need to know to figure out whether the user has completed the form or not; no need to render the templates since we're not displaying anything.

However, the current implementation of AIBuilder is always rendering templates, even with this "builder pattern" refactor. That's because when building the "form instance" and the "form children", every object we create as part of the structure will call `addAndRenderTemplate()` which will basically render each template we encounter during traversal. It's not exactly "rendering the form" since they are stashed away in the context and is not actually put into the form structure objects. But still, that's work we don't want to pay for.

Anyways, what I'm trying to say here is, this refactor doesn't actually help us much. I think if we skipped this "builder pattern" refactor and just used AIBuilder as-is in the PUT route, the performance will probably be very similar to what this refactor gives us.

But I could be wrong! Maybe I've misunderstood, and that this is part of a bigger upcoming refactor? Regardless, this "builder pattern" style usage of AIBuilder might be useful in other cases, so I'll let you decide how you want to proceed here.

yufengwng (commented):
Looks fine.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormSectionCreator.java
evgeniipr: Import is used: AIBuilderContext is moved into subpackage therefore it is imported now.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
evgeniipr: Fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
evgeniipr: Ok, I understood the idea with templates. I will add a parameter to turn off templates rendering.
Another place where this builder could be used (or at least a possibility to call building steps separately) - in `GetActivityInstanceRoute`: to replace `RouteUtil.findAccessibleInstanceOrHalt()` with the 1st step of AIBuilder: readFormInstanceData() which reads all data which reads `RouteUtil.findAccessibleInstanceOrHalt()` (or actually reads `.getByActivityInstanceGuid(instanceGuid)` which called from `RouteUtil.findAccessibleInstanceOrHalt()` ).
So we can avoid yet one extra query in `GetActivityInstanceRoute`.

evgeniipr: Added the flag to disable templates rendering and used it for `PutFormAnswerRoute`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/FormSectionCreator.java
yufengwng: Ah right! 👍

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/context/AIBuilderContext.java
```diff
@@ -23,79 +24,120 @@
...
-        this.userGuid = userGuid;
-        this.operatorGuid = operatorGuid;
-        this.isoLangCode = isoLangCode;
-        this.style = style;
-        this.langCodeId = LanguageStore.get(isoLangCode).getId();
-        this.formActivityDef = formActivityDef;
-        this.formResponse = formResponse;
+        this.params = params;
+        LanguageDto languageDto = LanguageStore.get(params.getIsoLangCode());
+        this.langCodeId = languageDto != null ? languageDto.getId() : 0;
```
yufengwng: We should probably throw an error (or somehow return `Optional.empty()`) if it's an invalid language instead of defaulting to zero?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
yufengwng: Looks good!

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/context/AIBuilderContext.java
evgeniipr: It is checked at the beginning of the process (when parameters checked) and eventually Optional.empty() is returned:
```
if (languageDto == null) {
            context.setFailedMessage("Unknown language code: " + context.getParams().getIsoLangCode());
            context.setFailedStep(CHECK_PARAMS);
 } ...
```
And in old implementation it doesn't throw an exception when language code is unknown, therefore I did the same processing.
Or do you think that better to throw an exception in such case? (will need to change couple of tests then)

evgeniipr: Anyway I added check and exception throwing in AIBuilderContext constructor (double-checking)

yufengwng: I think the old implementation assumes that this is non-null because the language should had been handled in `StudyLanguageResolutionFilter`. I think it'll be good to do similar thing here -- assume it's non-null instead of trying to fallback to something like zero. In my view, getting a NullPointerException will be better than having this silently go along with zero.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PutFormAnswersRouteStandalone2Test.java
```diff
@@ -0,0 +1,47 @@
...
+import org.broadinstitute.ddp.model.activity.definition.QuestionBlockDef;
+import org.broadinstitute.ddp.model.activity.definition.validation.RequiredRuleDef;
+import org.broadinstitute.ddp.model.activity.revision.RevisionMetadata;
+import org.junit.Test;
+
+/**
+ * The test moved to a separate class because result of other tests affected on it and
+ * it's behaviour was incorrect
+ */
+public class PutFormAnswersRouteStandalone2Test extends PutFormAnswersRouteStandaloneTestAbstract {
```
yufengwng, line 23: Do we need to add these new route tests to `ServerInSameProcessIntegrationTestSuite`?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/context/AIBuilderContext.java
evgeniipr: Fixed

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PutFormAnswersRouteStandalone2Test.java
evgeniipr: Yes, thanks. Fixed.

yufengwng (approved):
LGTM

PR #718, opened 2021-04-23
Use study default language if no pdf templates for user preferred lang templates

ssettipalli:
Handle a very rare edge case where users preferred language pdf templates does not exist
Use study default language if no pdf templates for user preferred lang templates


pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -141,10 +142,21 @@ public InputStream generateFlattenedPdfForConfiguration(PdfConfiguration configu
...
+                if (templateIds.isEmpty()) {
+                    LOG.warn("User : {} has language {} as preferred language but NO PDF templates found for the language. " 
+                            + "Using study default language ", userGuid, participant.getUser().getProfile().getPreferredLangCode());
+                }
+            }
         }
-        if (languageId == null) {
+
+        if (languageId == null || templateIds.isEmpty()) {
             //use study default language if exists
```
ssettipalli, line 160: Expectation here is that if  a study has NON 'english' as default language, pdf templates exist for this language. 
We might want to have this check in study builder if we are concerned about this expectation.

yufengwng (approved):

PR #719, opened 2021-04-26
DDP-5924 api docs for admin participants lookup endpoint

yufengwng:
Here is the initial API documentation for the new admin participants-lookup endpoint being added as part of DDP-5924.


yufengwng:
CC @MocanaAtBroad @bskinner @alextanlan @aglushko-broad

evgeniipr:
@yufengwng looks like `proxy` always should be specified (required)?

evgeniipr:
@yufengwng is it correct that `totalCount` always should be specified (even if found rows count <= MAX_COUNT) ?

yufengwng:
@evgeniipr:

> @yufengwng looks like `proxy` always should be specified (required)?

Good catch! Let's make it nullable!

> @yufengwng is it correct that `totalCount` always should be specified (even if found rows count <= MAX_COUNT) ?

Yeah, I think it should always be specified.

evgeniipr:
@yufengwng is it possible that a participant could have more than 1 proxy?

yufengwng:
@evgeniipr yes, it's possible given our current model. However, so far, we've had only one proxy per participant. I think we'll need to work out more things related to access-control before we're ready to see "one participant multiple proxies" in real usage.

evgeniipr (dismissed):
Looks ok

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2384,3 +2384,91 @@ Statistics:
...
+    - EXITED_AFTER_ENROLLMENT
+    - CONSENT_SUSPENDED
+    - COMPLETED
+ParticipantsLookupResultItem:
+  type: object
+  required:
+    - guid
+    - hruid
+    - legacyAltPid
+    - legacyShortId
```
yufengwng, line 2420: @evgeniipr @aglushko-broad @MocanaAtBroad Added legacy identifiers.

evgeniipr (approved):

PR #723, opened 2021-04-27
DDP-5676 Add workflow statuses to /status endpoint

YuriyPavlovBI:
## Context

Participant status now is being taken from ES, without querying DSM
*The response has changed so it could break existing functionality. Also, It's still not clear why we were taking MIN dates here: org.broadinstitute.dsm.db.ParticipantStatus#SQL_SELECT_MR_STATUS , so I just passed all the records through*
See
https://broadinstitute.atlassian.net/browse/DDP-5676
https://broadinstitute.atlassian.net/browse/DDP-5735
https://broadinstitute.slack.com/archives/G01S430UNES

## Checklist

## FUD Score

- [x] I have no idea how to demo this. Please help me!

## Testing

## Release


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2178,9 +2178,15 @@ ParticipantStatusTrackingInfo:
...
+      type: array
+      minItems: 0
+      items:
+        $ref: '../pepper.yml#/components/schemas/ParticipantStatusTrackingInfo.Record'
     tissueRecord:
-      $ref: '../pepper.yml#/components/schemas/ParticipantStatusTrackingInfo.Record'
+      type: array
+      minItems: 0
+      items:
+        $ref: '../pepper.yml#/components/schemas/ParticipantStatusTrackingInfo.Record'
```
bskinner: If this is just adding workflows to the status, why are these changes here? This breaks multiple clients.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2214,9 +2224,11 @@ ParticipantStatusTrackingInfo.Record:
         - RECEIVED
         - UNKNOWN
     requestedAt:
-      type: integer
+      type: string
+      description: ISO-8601 date
     receivedBackAt:
-      type: integer
+      type: string
+      description: ISO-8601 date
```
bskinner: What's the technical reason for this change?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2233,9 +2245,14 @@ ParticipantStatusTrackingInfo.Kit:
         - RECEIVED
         - UNKNOWN
     sentAt:
-      type: integer
+      type: string
+      description: ISO-8601 date
     receivedBackAt:
-      type: integer
+      type: string
+      description: ISO-8601 date
```
bskinner: Again, what's the technical reason for breaking the API here?

bskinner (changes requested):

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: It's not just adding workflows. As we discussed in the slack channel I mentioned in the description, we are going to switch to ES entirely for participant status. Previously there was strange aggregation of multiple records on DSM side (see SQL query I put the reference to), now we're getting all records from ES. Now we have two ways:
1) pass all the records to clients (that breaks clients)
2) do the same strange aggregation on DSS side
Let's discuss

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: That was proposed in the slack channel discussion

bskinner: Can you share a link to the relevant discussion?

YuriyPavlovBI: https://broadinstitute.slack.com/archives/G01S430UNES/p1616680027055800

bskinner: Thank you! How about a different proposal for this- create a new pair of properties: `requestedAtDate`, and `receivedBackAtDate` in the Pepper API only. These will be the new ISO-8601 date fields. The older properties `requestedAt` and `receivedBackAt` should remain as the unix timestamp formatted version of this information, but be marked as deprecated. New or updated applications will be able to opt-in to using the new field, and the old one can be deprecated without immediately breaking clients.

YuriyPavlovBI: Sounds good!

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: @YuriyPavlovBI Let's use the same approach here as in the [previous comment](https://github.com/broadinstitute/ddp-study-server/pull/723/files/e0f3762bed10f10ac8ae7274f9f5573bee528fa5#r622223630). Leave `sentAt` and `receivedBackAt` as `integer` type fields, mark them as deprecated in OpenAPI, and add the `sentAtDate` and `receivedBackAtDate` as `string` type fields, with the `format` `ISO-8601 Date`

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: How is the data expected to change? If a client is getting a single record now, what does the new response look like now? What I'm thinking here is leave the original response, but add `tissueRecords` and `medicalRecords` fields using the pluralized name for the lists, and mark the `medicalRecord` and `tissueRecord` fields as deprecated for future removal.

yufengwng: This is not what I expected. Yes, we agreed to have DSM write out data into ES and that DSS should read data from ES instead of calling DSM's API. However, I think it is best that we keep the API the same. This means internally DSS should "massage" the data into the same format that we have now. In the Slack conversation, you'll see that I've noted this and have asked Simone about the expected behavior there. Overall, I think the only thing that should be changing in the API is a new `workflows` list.

YuriyPavlovBI: @yufengwng Can you check aggregation in DSM SQL queries? It seems meaningless. Could you confirm I should aggregate data the same way?
For example: org.broadinstitute.dsm.db.ParticipantStatus#SQL_SELECT_MR_STATUS

yufengwng: @YuriyPavlovBI Based on my brief reading of that SQL, it looks like the query is aggregating medical records by participant id, and taking the minimum of the `requested`/`received` dates. This is consistent with what Simone described in the Slack conversation:
* Basically, a participant has a list of medical/tissue records.
* Each record is for a different institution.
* The way DSM builds the response of a single record is by taking the minimum date, _regardless_ of which institution it is.

I think we can just do the same in DSS -- read the `medicalRecord` or `tissueRecord` list and reduce it to a single object by taking the minimum dates.

yufengwng: This is the example that Simone gave in the Slack conversation:
```
institution a: requested 01/01/2020
                received: ---
institution b: requested 02/01/2020
                received: 05/10/2020
would result in requested 01/01/2020 and received 05/10/2020
```

bskinner: It's sounding like none of the API beyond the `workflows` is expected to change- is this an accurate understanding, @yufengwng?

yufengwng: Yes, my understanding is that the API should stay the same except for the `workflows` piece. In fact, I think we already have the `workflows` in there already. The original issue we were trying to solve was to workaround the authentication issue between DSS and DSM. So idea was to have DSM put data into ES so we don't have to do the auth dance. But folks got a little too enthusiastic about ES so we have to do some more refactors on DSS side when reading data out of ES.

bskinner: Good to know- thank you!

YuriyPavlovBI: Ok, thanks for clarification. I rolled back the interface part and added data post-processing to keep the same output.

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Rolled back

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Rolled back

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: image

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
```diff
@@ -12,29 +14,28 @@
...
         INELIGIBLE, PENDING, SENT, RECEIVED, DELIVERED, UNKNOWN
     }
 
-    private static final Logger LOG = LoggerFactory.getLogger(ParticipantStatus.class);
-
-    @SerializedName("medicalRecords")
+    @SerializedName("medicalRecord")
     private MedicalRecord medicalRecord;
-    @SerializedName("tissue")
+    @SerializedName("tissueRecord")
```
yufengwng, line 25: This is an agreed-upon API change to rename the property names to `medicalRecord` and `tissueRecord`. We also have these two names in the OpenAPI docs, so making the code to match.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
```diff
@@ -12,29 +14,28 @@
...
+    private final transient String userGuid;
 
     private RecordStatus figureOutMedicalRecordStatus(
-            String userGuid,
             String entityName,
             EnrollmentStatusType enrollmentStatusType,
-            Long requested,
-            Long received
+            LocalDate requested,
+            LocalDate received
```
yufengwng, line 38: The timestamps in ES are now dates instead of epoch seconds.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
```diff
@@ -48,70 +49,85 @@ private RecordStatus figureOutMedicalRecordStatus(
...
+        LocalDate minRequested = null;
+        if (statusES.getMedicalRecords() != null) {
+            for (ParticipantStatusES.MedicalRecord medicalRecord : statusES.getMedicalRecords()) {
+                if (medicalRecord.getReceived() != null) {
+                    if (minReceived == null || minReceived.compareTo(medicalRecord.getReceived()) > 0) {
+                        minReceived = medicalRecord.getReceived();
+                    }
+                }
+                if (medicalRecord.getRequested() != null) {
+                    if (minRequested == null || minRequested.compareTo(medicalRecord.getRequested()) > 0) {
```
yufengwng, line 72: We always return a medical or tissue record object, even if there's no timestamps. Before, DSM will compute the timestamps to use. Now, we're querying the data directly from ES, so we have to do this computation of finding the minimum timestamp to use.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/GetDsmParticipantStatusRoute.java
```diff
@@ -57,15 +55,14 @@ public GetDsmParticipantStatusRoute(DsmClient dsmClient, RestHighLevelClient esC
...
 
         LOG.info("Attempting to fetch DSM participant status for {} in study {}", userGuid, studyGuid);
 
         response.type(ContentType.APPLICATION_JSON.getMimeType());
-        return process(studyGuid, userGuid, token);
+        return process(studyGuid, userGuid);
     }
 
-    ParticipantStatusTrackingInfo process(String studyGuid, String userGuid, String token) {
+    ParticipantStatusTrackingInfo process(String studyGuid, String userGuid) {
```
yufengwng, line 65: We now get rid of needing a token, solving our original authentication issue.

pepper-apis/docs/specification/src/endpoints/user.studies.status.yml
```diff
@@ -21,5 +21,11 @@ get:
   responses:
     200:
       $ref: '../pepper.yml#/components/responses/ParticipantStatusResponse'
+    404:
```
yufengwng, line 24: API is already returning a 404, so while I'm in here, just updating the docs to reflect reality.

yufengwng (commented):
Yuriy is going on vacation, so I'm picking up and wrapping up this PR. Original issue we had was that user's authentication token for RGP study was not being processed properly by DSM. Approach we agreed on was to have DSM put data into ES, then DSS will query from ES and build the necessary medical/tissue/kit/workflow objects. So the goal of this PR is to switch DSS's `GetDsmParticipantStatusRoute` to query ES.

@ssettipalli can I get a second pair of eyes on this?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
bskinner: Thanks for cleaning these up! I'll get iOS ready for the change.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
ssettipalli: Variable names something like latestReceivedDate , latestRequestedDate will be more clear !

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
```diff
@@ -212,43 +225,47 @@ public static RecordStatus figureOutStatus(
             } else if (delivered == null && received != null) {
                 return RecordStatus.RECEIVED;
             } else {
-                LOG.error("Something completely unexpected happened with ", entityName);
+                LOG.error("Something completely unexpected happened with kit");
```
bskinner, line 228: Any more information we can give here? This message could use more actionable information if we have it.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/DateTimeUtils.java
```diff
@@ -39,4 +40,8 @@ public static ZoneId parseTimeZone(String givenTimeZone) {
             return null;
         }
     }
+
+    public static Long localDateToEpochSeconds(LocalDate date) {
+        return date == null ? null : date.atStartOfDay(ZoneId.of("UTC")).toEpochSecond();
```
bskinner, line 45: Is assuming `UTC` instead of the system's timezone here accurate?

bskinner (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
yufengwng: Actually, we're looking for the minimum date, so the oldest date and not latest.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/dsm/ParticipantStatusTrackingInfo.java
yufengwng: Actually, I don't think we'll ever execute this branch. The other `if` branch conditions cover all possible cases. I don't actually know why we have this `else` branch. My purpose here is not to refactor/change implementation here so I had kept it the way it is.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/DateTimeUtils.java
yufengwng: Yes, accurate.

ssettipalli (approved):
looks good to me

bskinner (approved):
👍

PR #729, opened 2021-04-28
MBC es email fixes

ssettipalli:
minor formatting and cleanup. 
Updated StudyBuilder patch: UpdateEmailEventTemplates to handle emails with same "execution_order" that gets triggered on same event.


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateEmailEventTemplates.java
```diff
@@ -78,15 +78,23 @@ public void run(Handle handle) {
     }
 
     private String hashEvent(Config eventCfg) {
-        return String.format("%s-%d",
+        //use delaySeconds in key to handle emails with same "order" triggered in same event
```
yufengwng, line 81: Nice! Using delay makes sense to me.

yufengwng (approved):

PR #734, opened 2021-04-30
DDP-5924 participants lookup endpoint

evgeniipr:
## Context

It is implemented a REST endpoint `POST ../admin/participants-lookup` which allows to search study participants by a specified fragment of string.                                         
The searching delegated to interface `ParticipantsLookupService` which has implementation                    
`ESParticipantsLookupService` - it searches for participants in Pepper **ElasticSearch** database.               
                                                                                                                   
Request payload `ParticipantsLookupPayload` contains parameter 'query' with substring by which to search  
for participants.                                                                                                  
                                                                                                                   
Response object `ParticipantsLookupResponse` contains result of the lookup:                               
<ul>                                                                                                               
    <li>`totalCount` - total number of found participants (note: it can be higher than fetched number of participants
      - in case if specified limit (parameter 'resultsMaxCount') less than the found number of participants);      
      </li>                                                                                                        
    <li>`participants` - list of fetched participants (it's size could be equal to 'totalCount' or could be less -   
    in case if number of found participants exceeds 'resultsMaxCount' - in such case real size of this list ==     
    'resultsMaxCount' (and totalCount contains real found count).</li>                                             
</ul>                                                                                                              


## Checklist

## FUD Score

It may be needs to discuss more about ElasticSearch Query DSL - how to use it better.

As soon as a related frontend ticket DDP-5923 will be implemented it will be possible to test this ticket. 

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/util/ElasticsearchServiceUtil.java
```diff
@@ -31,6 +32,13 @@
     private static final Logger LOG = LoggerFactory.getLogger(ElasticsearchServiceUtil.class);
     private static final Map<Integer, RestHighLevelClient> ES_CLIENTS = new HashMap<>();
 
+    public static String detectEsIndex(String studyGuid, ElasticSearchIndexType elasticSearchIndexType) {
+        return TransactionWrapper.withTxn(handle -> {
+            StudyDto studyDto = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
```
yufengwng: Let's use `JdbiUmbrellaStudyCached` here.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -207,6 +209,9 @@
     private static final AtomicBoolean isReady = new AtomicBoolean(false);
     private static final int DEFAULT_BOOT_WAIT_SECS = 30;
 
+    private static final int DEFAULT_PARTICIPANTS_LOOKUP_RESULT_MAX_COUNT = 100;
```
yufengwng: Eventually, this might come from the request payload. For now, maybe move this into the route and remove the constructor parameter?

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
```diff
@@ -44,6 +44,8 @@
                 fmt(ADMIN_BASE + "/studies/%s/invitation-details", PathParam.STUDY_GUID);
         public static final String ADMIN_STUDY_USER_LOGIN_ACCOUNT =
                 fmt(ADMIN_BASE + "/studies/%s/user/%s/login-account", PathParam.STUDY_GUID, PathParam.USER_GUID);
+        public static final String ADMIN_STUDY_PARTICIPANTS_LOOKUP =
```
yufengwng: Can you move this up to be right below `ADMIN_STUDY_PARTICIPANTS`?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ParticipantsLookupRoute.java
```diff
@@ -0,0 +1,80 @@
...
+        var studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
+        var query = payload.getQuery();
+
+        response.type(ContentType.APPLICATION_JSON.getMimeType());
+
+        LOG.info("Started participants lookup (query=\"{}\")", query);
+
+        var result = new ParticipantsLookupResponse(0, emptyList());
+
+        if (StringUtils.isNotBlank(query)) {
```
yufengwng: I think it'll be better (and safer) if we move this logic into the service itself -- if query is empty/blank, `lookupParticipants()` call will return result.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ParticipantsLookupRoute.java
```diff
@@ -0,0 +1,80 @@
...
+            var lookupResult = participantsLookupService.lookupParticipants(studyGuid, query, resultsMaxCount);
+            result = new ParticipantsLookupResponse(
+                    Long.valueOf(lookupResult.getTotalCount()).intValue(),
```
yufengwng: We should probably stick with one data type. I think `int` will be a safe bet here?

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupPayload.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.validation.constraints.NotEmpty;
+
+import com.google.gson.annotations.SerializedName;
+
+public class ParticipantsLookupPayload {
+
+    @SerializedName("query")
+    private String query;
+
+    public ParticipantsLookupPayload(@NotEmpty String query) {
```
yufengwng: I don't think we need the annotation?
```suggestion
    public ParticipantsLookupPayload(String query) {
```

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupResponse.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Result of {@link ParticipantsLookupRoute}.
+ */
+public class ParticipantsLookupResponse {
+
+    @SerializedName("totalCount")
+    private int totalCount;
+
+    @SerializedName("participants")
```
yufengwng: I think we called this property `results` in the API docs.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
```diff
@@ -0,0 +1,109 @@
...
+
+
+    public static Map<String, ESUsersIndexResultRow> searchInUsersIndex(
+            RestHighLevelClient esClient,
+            String studyGuid,
+            String mainQuery,
+            Map<String, String> governedUserToProxy,
+            int resultMaxCount
+    ) throws IOException {
+        var esIndex = ElasticsearchServiceUtil.detectEsIndex(studyGuid, ElasticSearchIndexType.USERS);
```
yufengwng: Can we make a single query for the `StudyDto` object (perhaps in the route) and then pass that around instead of `studyGuid`?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
```diff
@@ -0,0 +1,109 @@
...
+        var queryBuilder = createESQueryForUsersIndex(mainQuery);
+        var searchRequest = prepareSearching(
+                esIndex, USERS__INDEX__SOURCE, queryBuilder, doubleResultMaxCount(resultMaxCount));
+        var response = esClient.search(searchRequest, RequestOptions.DEFAULT);
+        return readUsersIndexResult(response, governedUserToProxy);
+    }
+
+    private static AbstractQueryBuilder createESQueryForUsersIndex(String query) {
+        var builder = QueryBuilders.queryStringQuery(addWildcards(query)).defaultOperator(Operator.OR);
```
yufengwng: I think using the `query_string` query is fine. I tried a few searches with it in Kibana `dev` and it seems to work. On a related note, should we escape potential reserved characters in the raw user-provided query? ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-query-string-query.html#_reserved_characters))

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.broadinstitute.ddp.elastic.participantslookup.model.ESParticipantsLookupIndexType.USERS;
+
+/**
+ * Names of fields in ElasticSearch indices ("users", "participants_structured")
+ * by which to search results.<br>
+ */
+public enum ESParticipantsLookupField {
+
+    PROFILE__GUID("profile.guid", ALL_TYPES),
+    PROFILE__HRUID("profile.hruid", ALL_TYPES),
```
yufengwng, line 14: Can we add fields `profile.legacyAltPid` and `profile.legacyShortId`? I think these 2 are for `ALL_TYPES`.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
```diff
@@ -0,0 +1,41 @@
...
+    PROFILE__FIRST_NAME("profile.firstName", ALL_TYPES),
+    PROFILE__LAST_NAME("profile.lastName", ALL_TYPES),
+    PROFILE__EMAIL("profile.email.text", ALL_TYPES),
```
yufengwng: Should we do both this and the keyword `profile.email` field?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
```diff
@@ -0,0 +1,109 @@
...
+     * For example: MAX = 10
+     * Query: "john"
+     * In index "users" found 5 of proxy users having in lastName substring "john" and
+     * all their children (governed users) have same lastName therefore they found also (5 users). Total found = 10.
+     * When searching in index "participants_structured" it is found governed users only (5 participants)
+     * therefore final list is 5 (which twice less than specified limit 10).
+     * </pre>
+     * That's why it needs to double MAX limit when searching in index "users".
+     */
+    private static int doubleResultMaxCount(int resultMaxCount) {
```
yufengwng: Maybe an alternative solution here is to restrict our `users` index search request to only proxy users? I think the way to do this is to add a clause that says "`governedUsers` field should exist and is not empty". If this is a regular JSON query request, I think we can tack on the `exist` clause, but I'm not too sure how to do it for the `query_string` query. I think we might have to generate the query string ourselves with all the fields instead of building a list of fields with `builder.field()` and then tack on the `_exists_` clause at the end. Something like this:

```
{
  "query": {
    "query_string": {
      "query": "(profile.guid:*yufeng* OR profile.hruid:*yufeng* OR ...) AND _exists_:governedUsers"
    }
  }
}
```

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
```diff
@@ -0,0 +1,109 @@
...
+        return builder;
+    }
+
+    public static Map<String, ESUsersIndexResultRow> readUsersIndexResult(
+            SearchResponse response,
+            Map<String, String> governedUserToProxy) {
+        Map<String, ESUsersIndexResultRow> resultData = new HashMap<>();
+        for (var hit : response.getHits()) {
+            var participantResultData = gson.fromJson(getJsonElement(hit, SOURCE__PROFILE), ESUsersIndexResultRow.class);
+            JsonElement governedUsersJson = getJsonElement(hit, SOURCE__GOVERNED_USERS);
```
yufengwng: I think it'll be more efficient if we do `hit.getSourceAsString()` once and deserialize that into a single POJO, then we can grab the `profile` and `governedUsers` out of it, instead of doing parsing 3 times.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsStructuredIndexSearchHelper.java
```diff
@@ -0,0 +1,131 @@
...
+        participantsLookupResult.setTotalCount(response.getHits().getTotalHits());
+        return readParticipantsStructuredIndexResult(response);
+    }
+
+    private static AbstractQueryBuilder createESQueryForParticipantsStructuredIndex(String query, Map<String, String> governedUserToProxy) {
+        var aggregatedQuery = getAggregatedQuery(query, governedUserToProxy);
+        var builder = QueryBuilders.queryStringQuery(aggregatedQuery.toString()).defaultOperator(Operator.OR);
+        for (ESParticipantsLookupField field : ESParticipantsLookupField.values()) {
+            if (isQueryFieldForIndex(field, PARTICIPANTS_STRUCTURED)) {
+                builder.field(field.getEsField());
```
yufengwng: Hmmm, I think we should switch to the "field names" syntax instead of using list of fields. This way, we can specify the exact search we want for invitation code (without affect other fields), as well as governed users. For example:
```
profile.guid:*foo-bar* OR profile.hruid:*foo-bar* ... OR invitation:*foo-bar* OR invitation:*foobar* OR profile.guid:(GOVERNED_GUID1 GOVERNED_GUID2 ...)
```

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsStructuredIndexSearchHelper.java
```diff
@@ -0,0 +1,131 @@
...
+    public static Map<String, ESParticipantsStructuredIndexResultRow> searchInParticipantsStructuredIndex(
+            RestHighLevelClient esClient,
+            String studyGuid,
+            String mainQuery,
+            Map<String, String> governedUserToProxy,
+            int resultMaxCount,
+            ParticipantsLookupResult participantsLookupResult) throws IOException {
+        var esIndex = ElasticsearchServiceUtil.detectEsIndex(studyGuid, ElasticSearchIndexType.PARTICIPANTS_STRUCTURED);
+        var queryBuilder = createESQueryForParticipantsStructuredIndex(mainQuery, governedUserToProxy);
+        var searchRequest = prepareSearching(esIndex, PARTICIPANTS_STRUCTURED__INDEX__SOURCE, queryBuilder, resultMaxCount);
```
yufengwng: Now that I'm looking at this more closely, I think there's some flaws with our approach.

One concern is this second query and the `resultMaxCount` limit. What if there's more than 100 matches, and ES doesn't return all the governed users that we're looking for? Then we won't be able to build the correct result set.

Another concern is also related to governed users. With our approach, it's possible that this second query returns a governed participant. It's possible that we don't have a proxy for them in our list from our first query since our first query only focused on matching on proxy data.

I was hoping we can avoid doing multiple search queries. But maybe it's unavoidable. Perhaps we need 4 queries:

* First, search proxies in `users` index. Second, lookup their corresponding governed users in `participants_structured` index (this second request is not a search but rather just looking up by `guid`).
* Third, search participants in `participants_structured` index. Fourth, lookup their proxy in the `users` index (again, lookup by `guid` instead of a search).

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ElasticsearchServiceUtil.java
evgeniipr: Tried to do it but it gives exception. Will try to check what is wrong..

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
evgeniipr: Moved to route

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ParticipantsLookupRoute.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/route/ParticipantsLookupRoute.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupPayload.java
evgeniipr: Removed annotation

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupResponse.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
evgeniipr: Added the fields.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
evgeniipr: Did a redesign according to https://github.com/broadinstitute/ddp-study-server/pull/734#discussion_r625338695

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsStructuredIndexSearchHelper.java
evgeniipr: I agree with it.
Redesigned according to the suggested logic.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsStructuredIndexSearchHelper.java
evgeniipr: Yes, fixed this (invitation code queried separately)

pepper-apis/src/test/java/org/broadinstitute/ddp/route/ParticipantsLookupRouteTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private static String urlTemplate;
+    private static TestDataSetupUtil.GeneratedTestData testData;
+
+    @BeforeClass
+    public static void setupServer() {
+        var jsonSerializer = new NullableJsonTransformer();
+
+        int port = RouteTestUtil.findOpenPortOrDefault(5559);
+        Spark.port(port);
```
yufengwng: This is great!! I think we should do more of these unit tests that spin up their own Spark server so things are more isolated.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public String getEsField() {
+        return esField;
+    }
+
+    public ESParticipantsLookupIndexType getIndexType() {
+        return indexType;
+    }
+
+    public static boolean isQueryFieldForIndex(ESParticipantsLookupField field, ESParticipantsLookupIndexType index) {
```
yufengwng: This can just be a method on the `enum` rather than a static method.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
```diff
@@ -0,0 +1,148 @@
...
+                .setQuery(query)
+                .search();
+        // add to proxy result found extra proxies (which were not found during search step 1)
+        for (var proxyGuid : proxiesExtraResults.keySet()) {
+            proxiesResults.put(proxyGuid, proxiesExtraResults.get(proxyGuid));
+        }
+        // copy to map governedUser=proxy the found extra proxies
+        governedUserToProxy.putAll(governedUserToProxyExtraSearch);
+        // repeated merge of results
+        esResultRows = mergeResults(
```
yufengwng: This will create twice the amount of objects... is there any way to reuse `esResultRows` from the first merge and just fill in what's missing?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
```diff
@@ -0,0 +1,148 @@
...
+            Map<String, String> governedUserToProxyExtraSearch,
+            String usersEsIndex,
+            Map proxiesResults,
+            Map participantsStructuredResults) throws IOException {
+        List esResultRows;
+        Map<String, ESUsersIndexResultRow> proxiesExtraResults = new ESUsersProxiesExtraSearch(esClient, governedUserToProxyExtraSearch)
+                .setResultMaxCount(resultsMaxCount)
+                .setEsIndex(usersEsIndex)
+                .setFetchSource(USERS__INDEX__SOURCE)
+                .setQuery(query)
```
yufengwng: Setting the query here is confusing -- this extra search is looking up proxies by guid and not user-provided `query`.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
```diff
@@ -0,0 +1,148 @@
...
+     * Adds to results of "participants_structured" index the 'proxies' data results (taken from
+     * "users" index results).
+     */
+    private List<ParticipantsLookupResultRow> mergeResults(
+            Map<String, ESUsersIndexResultRow> usersResults,
+            Map<String, ESParticipantsStructuredIndexResultRow> participantsStructuredResults,
+            Map<String, String> governedUserToProxy,
+            Map<String, String> governedUserToProxyExtraSearch) {
+        List<ParticipantsLookupResultRow> resultList = new ArrayList<>();
+        for (var participant : participantsStructuredResults.values()) {
```
yufengwng: I don't see anywhere that addresses this scenario: what if there are matches in the first search for proxies, but then the second search in the participants index does not return the governed user from the first search? It might be possible that the `query` matched a bunch of non-governed users, and because of `resultsMaxCount` some of the governed users from the first search is not returned.

But I think the approach you have here is slightly different and I think it'll be fine. If I'm understanding, the main result set of final response will come from this second search on the participants index. If the search returned a governed user, then great it'll be included in final response along with the proxy that we found in the first search. If not, then the proxies we found in the first search will simply be left out. I think this is fine.

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupResultRowBase.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * Base set of data to be fetched during participants lookup.
+ */
+public class ParticipantsLookupResultRowBase {
+
+    @SerializedName("guid")
+    protected String guid;
```
yufengwng: Can we also return `legacyAltPid` and `legacyShortId` in the final response? Some studies (like Angio and MBC) have legacy users and it'll be good to return those two pieces of data. These 2 properties should only be set on the participant object, not the proxy. I'll update the API docs to reflect this.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
yufengwng: Can you double-check this comment?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
yufengwng: Have you tried doing  a search with special characters? Do we need to do some escaping?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: But user-provided query is executed in 1st step in "users" index (when searched for proxies). 
Is it correct that "participants" contains all except proxies? In that case the searching in "participants" index is enough to find everyone except proxies ( + proxies found in "users")?

yufengwng: It is correct that the `participants_structured` index contains only participants -- no proxies. But that is not what my comment is about. I'm saying this line here is confusing because the "proxy extra search" does not use the user-provided `query` as search term. Rather, it is looking for proxies by guid. So having `.setQuery(query)` is confusing.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: > and because of resultsMaxCount some of the governed users from the first search is not returned

Yes, it's possible. I can split search in "participants" into 2 runs:
1) search for participants by "query" (coming from frontend), limited by `resultsMaxCount`.
This search result consider as higher priority (than proxies found in "users"). Therefore take the results of this search (+ proxy users of the governedUsers existing in this list). If this result >= `resultsMaxCount` then take only this list as a final result. 
2) If it < `resultsMaxCount` then find in proxies list (found in `users`) the proxies which not exist in the 1st result, and find governed users for them (also checking that the extra amount together with 1st list will be <= `resultsMaxCount`).

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: Ah, understood)
Yes, I thought about the same, but forgot to remove this setQuery().. Will remove

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: So, that priority of finding among participants data (first,last name, email, guid, hruid) will be higher than among proxies (first, last name, email, guid, hruid).

yufengwng: I don't see how the 2 runs you described is any better than what you have now. For case 2 of `< resultsMaxCount`, if this is the case, then that means we have room to search for the governed users of proxies. But isn't that exactly what you get back now in your current approach? If the `query` doesn't find enough matches to fill `resultsMaxCount`, then the rest of the matches you get back should be the governed users that we're looking for?

yufengwng: I think the way you have it now is fine.

evgeniipr: Let me double-check..
Currently it works like this:
1) Search in "users" (in proxy users only) by specified "query" (in name/email/guid/hruid/legacyId...).
2) Search in "participants" by compound search: governed users (by guids detected from 1st seaech in "users") + by "query" (in name/email/guid/hruid/invitations/legacyId...). The total result contains governed users detected from proxies (found in 1st search) + governed users AND normal users found by "query". 
And here can happen that the total list exceeds `resultsMaxCount` and some of governed users/normal users (found by "query") excluded from result.. 
But ok, I think it is not so big issue probably..

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/model/ESParticipantsLookupField.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/json/admin/participantslookup/ParticipantsLookupResultRowBase.java
evgeniipr: Added

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESParticipantsLookupService.java
evgeniipr: I agree. Redesigning..

evgeniipr: Changed (and simplified) this code

yufengwng:
I have a few unresolved comments above from May 3rd. Can you take another look?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
evgeniipr: Ah, yes, sorry. I made a single query.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
evgeniipr: I tried to search by query having special characters: for example by email containing '+' and '@' and it finds it without escaping.

evgeniipr:
> I have a few unresolved comments above from May 3rd. Can you take another look?

Yes, sure. Answered/fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ElasticsearchServiceUtil.java
evgeniipr: Eventually changed to JdbiUmbrellaStudyCached.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/participantslookup/ESUsersIndexSearchHelper.java
yufengwng: Nice!

yufengwng (dismissed):
Looks good-to-go.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/participantslookup/ParticipantsLookupService.java
```diff
@@ -0,0 +1,79 @@
...
+     * in {@link ParticipantsLookupResult#getTotalCount()}.
+     *
+     * @param studyGuid GUID of a study where to search participants
+     * @param query string containing a fragment by which to lookup the participants (do a full-text search).
+     * @return ParticipantsLookupResult - an object containing the participants lookup results.
+     */
+    public ParticipantsLookupResult lookupParticipants(String studyGuid, String query, int resultsMaxCount) {
+
+        if (studyGuid == null) {
+            throw new ParticipantsLookupException(STUDY_GUID_UNKNOWN, "studyGuid cannot be null");
```
yufengwng: How about we query for the study in the route and:
* If good, then pass in the `StudyDto`, so then you don't have to query for it in `detectEsIndices()`;
* If missing, halt with `404 Not Found`.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/participantslookup/ParticipantsLookupService.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        if (query != null && query.length() > QUERY_MAX_LENGTH) {
+            throw new ParticipantsLookupException(INVALID_QUERY, "query length should be <= 100");
```
yufengwng: We can put a constraint on the request payload itself, for example:
```
    @Size(max = 255)
    @SerializedName("fileName")
    private String fileName;
```
And then I recommend overriding `getValidationErrorStatus()` in the route so it returns `400` when the constraint is not met. (Make sure to update the OpenAPI docs!)

pepper-apis/src/main/java/org/broadinstitute/ddp/service/participantslookup/ParticipantsLookupService.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/service/participantslookup/ParticipantsLookupService.java
evgeniipr: Fixed

yufengwng: One last thing: can you override `getValidationErrorStatus()` and make it return `400`?

evgeniipr: Done (getValidationErrorStatus() is overridden).

yufengwng (approved):

PR #737, opened 2021-05-03
Add language support for Join Mailing List 

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-6032
Add language support for Join Mailing List 
PR includes backward compatible API change and we do have Blakes blessings.

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/json/mailinglist/GetMailingListResponse.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/json/mailinglist/GetMailingListResponse.java
@@ -16,6 +16,8 @@ public class GetMailingListResponse {
     private String info;
     @SerializedName("dateCreated")
     private long dateCreated;
+    @SerializedName("isoLanguageCode")
+    private String languageCode;
 
     public GetMailingListResponse(MailingListEntryDto dto) {
         this.firstName = dto.getFirstName();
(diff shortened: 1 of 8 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/json/mailinglist/GetMailingListResponse.java
```diff
@@ -16,13 +16,16 @@
     private String info;
     @SerializedName("dateCreated")
     private long dateCreated;
+    @SerializedName("isoLanguageCode")
```
ssettipalli, line 19: @SimoneMaiwald : Simone adding "isoLanguageCode" to MailingList Response as part of adding platform level  language support for MailingList

pepper-apis/src/main/java/org/broadinstitute/ddp/route/JoinMailingListRoute.java
```diff
@@ -50,6 +56,34 @@ public Object handle(Request request, Response response, JoinMailingListPayload
...
+            Long languageCodeId = null;
+            if (StringUtils.isEmpty(languageCode)) {
+                LanguageDto preferredUserLangDto = RouteUtil.getUserLanguage(request);
+                if (preferredUserLangDto != null) {
+                    languageCode = preferredUserLangDto.getIsoCode();
+                }
+            }
+
+            if (StringUtils.isNotEmpty(languageCode)) {
+                languageCodeId = LanguageStore.get(languageCode).getId();
```
yufengwng, line 74: Might want to wrap this in null-check in case client sends request with an unknown language code, in which case I suggest we return a `400`.

yufengwng (commented):
Overall looks good!

yufengwng (approved):

PR #751, opened 2021-05-11
DDP-6097 Remove US/Canada check from Prequal in rare-x

MocanaAtBroad:
### **KEEPING THIS PR OPEN TO HANDLE AT END EVERYTHING REGARDING COUNTRY CHECKS**


yufengwng (dismissed):

study-builder/studies/rarex/study.conf
```diff
@@ -149,7 +123,7 @@
               user.studies["rarex"].forms["PREQUAL"].questions["SELF_COUNTRY"].answers.hasOption("PR")
               && user.studies["rarex"].forms["PREQUAL"].questions["SELF_CURRENT_AGE"].answers.value() < 21
             ) || (
-              user.studies["rarex"].forms["PREQUAL"].questions["SELF_COUNTRY"].answers.hasAnyOption("GU", "VI", "MP", "AS")
+              !user.studies["rarex"].forms["PREQUAL"].questions["SELF_COUNTRY"].answers.hasAnyOption("US", "CA", "PR")
```
yufengwng, line 126: This is the key change:
* If we're in US/CA/PR, then we should follow whatever those rules are.
* Only if it's not US/CA/PR, then we do check for 18.

This ensures we handle the case where, e.g. you're in US+Alabama but you're not 19.

study-builder/studies/rarex/study.conf
```diff
@@ -816,11 +766,12 @@
         "type": "ACTIVITY_INSTANCE_CREATION",
         "activityCode": "PARENTAL_CONSENT"
```
yufengwng, line 783: Took me a while to wrap my head around the parental thing. Here's the [relevant ticket](https://broadinstitute.atlassian.net/browse/DDP-5696) that I used as reference. Essentially, in English:
```
if self:
  give self-consent

if child_age < 7:
  give consent-parental

if (7 <= child_age < majority) and child_not_capable:
  give consent-parental

if (7 <= child_age < majority) and child_is_capable:
  give consent-assent

if child_age >= majority and child_not_capable:
  give lar-consent

if child_age >= majority and child_is_capable:
  give lar-consent-assent
```

study-builder/studies/rarex/study.conf
MocanaAtBroad: Are we sure Guam, Virgin Islands and other American territories should be treated differently than the territory of Puerto Rico?

yufengwng: [See Slack comment](https://broadinstitute.slack.com/archives/C01A6HWK3PX/p1620679623109700?thread_ts=1620671795.091400&cid=C01A6HWK3PX).

Screen Shot 2021-05-12 at 3 49 04 PM

MocanaAtBroad: OK. Never mind. Makes more sense when you see all the rules.

MocanaAtBroad (commented):
@yufengwng github won't allow me to approve the PR (because I created it), but here you go ✅

yufengwng (approved):
🤦  'doh! Oh right. I guess I'll give myself a ✅  ... :)

PR #752, opened 2021-05-11
DDP-6102 rarex: proxy emails

yufengwng:
## Context

PR to add "proxy" versions of some of the emails. Honestly, I'm not too confident about the changes because I don't actually know what the emails are supposed to say. However, that's not the goal here. My focus is to just add new proxy versions where the salutation says "Dear proxy". 

## Release

1. Delete all rarex emails in dev sendgrid.
2. Use study-builder to create brand new email templates.
3. Put new email template ids into Vault.

### DIFF
--- a/study-builder/studies/rarex/emails/data-sharing-complete-reminder-proxy.html
+++ b/study-builder/studies/rarex/emails/data-sharing-complete-reminder-proxy.html
@@ -89,22 +89,29 @@
                     </p>
 
                     <p style="font-family: 'Roboto', sans-serif; text-align: left; font-weight: 400; font-size: 16px; line-height: 24px;">
-                        We are excited for you to begin contributing your information to the RARE-X Data Collection
-                        Program (DCP). Although you created an account you have not completed the Consent Assent Form, so
-                        you are not yet enrolled in the DCP and cannot share your information with us.
+                        Thank you for taking the time to sign up for the Data Collection Program (DCP). Your participation is vital to advancing rare disease research.
                     </p>
 
                     <p style="font-family: 'Roboto', sans-serif; text-align: left; font-weight: 400; font-size: 16px; line-height: 24px;">
-                        If you wish to fully enroll please review, sign, and submit the Consent Assent Form.
+                        It looks like you have started to answer the Data Sharing Survey but did not complete it.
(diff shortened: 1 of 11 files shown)


study-builder/studies/rarex/emails/data-sharing-complete-reminder-proxy.html
```diff
@@ -0,0 +1,160 @@
+<html>
```
yufengwng, line 1: All the new `*-proxy.html` files are basically copy-paste of existing files, where the only change is the salutation `Dear -ddp.proxy.firstName- -ddp.proxy.lastName-,`.

study-builder/studies/rarex/study.conf
```diff
@@ -1191,22 +1191,19 @@
...
-            "emailTemplate": ${emails.finishConsent1wkReminder},
+            "emailTemplate": ${emails.consent1wkReminder},
             "language": "en",
             "isDynamic": false
           }
         ],
         "linkedActivityCode": "CONSENT",
         "pdfAttachments": []
       },
-      "cancelExpr": """user.studies["rarex"].forms["CONSENT"].hasInstance()
```
yufengwng, line 1202: This event is triggered by the creation of the consent, so there's no need to check that we have an instance of it...

study-builder/studies/rarex/study.conf
```diff
@@ -1566,14 +1549,12 @@
         "linkedActivityCode": "FAMILY_MEMBER",
         "pdfAttachments": []
       },
-      "cancelExpr": """user.studies["rarex"].forms["FAMILY_MEMBER"].hasInstance()
-        && user.studies["rarex"].forms["FAMILY_MEMBER"].isStatus("COMPLETE")""",
-      "delaySeconds": 172800,
+      "cancelExpr": """user.studies["rarex"].isGovernedParticipant()
```
yufengwng, line 1552: Some activities are implemented as a single activity that swaps out self/parental text within the activity itself, so we're linking to just one activity. But we still need separate email event configuration for the proxy email. So we need to `isGovernedParticipant()` PEX to help us out.

MocanaAtBroad (dismissed):

PR #760, opened 2021-05-12
DDP-6074: RareX Shorten prompt so placeholder and label can be full-visible

MocanaAtBroad:
Per Slack discussion:
https://broadinstitute.slack.com/archives/C01A6HWK3PX/p1620843518231900


yufengwng (approved):

yufengwng:
Ah! So the new text is here. Let's get this merged.

PR #764, opened 2021-05-13
rarex: more edits

yufengwng:
Few more edits:
* Make General Information form "Other" details required.
* Legacy form: switch to month-day-year.
* H&D form: add asterisk for required questions.


MocanaAtBroad (approved):
Two thoughts: we should really tie together the order of fields in date to Language.
Marking questions required: not sure why we keep putting asterisks everywhere. It ought to be easier to do it using client logic. We do it on our studies, but not sure why.
At any rate, had to get that out. Tried it locally and looks good.

PR #774, opened 2021-05-20
DDP-5808 - file upload API extension

evgeniipr:
## Context

File upload API extension: add `maxFileSize` and `mimeTypes` to FileQuestionDef.
Add to DB the tables for storing file question and mime types data:
- `file_question` - to store file-upload specific data;
- `mime_type` - to store mime types;
- `file_question__mime_type` - relation table between file question and mime type.
It can be defined more than 1 mime-types for a certain file question.
It can be specified an empty list of mime types (in that case any mime type is allowed). 
Updated OpenAPI doc: added Question.FILE properties: maxFileSize (integer, int64), mimeTypes (list of strings).

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1491,12 +1491,22 @@ Question.File:
     - type: object
       required:
         - answers
+        - maxFileSize
```
yufengwng, line 1494: Let's mark `mimeTypes` property as required too.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1491,12 +1491,22 @@ Question.File:
...
         questionType:
           type: string
           default: FILE
           enum:
             - FILE
+        maxFileSize:
+          type: integer
+          format: int64
+          description: a max size of uploaded file, in bytes
+        mimeTypes:
```
yufengwng, line 1506: Should we make it so we always return an array and minimize on `null`s? Let's add this property:
```
minItems: 0
```

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1491,12 +1491,22 @@ Question.File:
...
+          type: array
+          items:
+            type: string
+          description: a list of allowed mime-types (if the list if empty then any type allowed)
```
yufengwng:
```suggestion
          description: a list of allowed mime-types (if the list is empty then any type is allowed)
```

pepper-apis/src/main/resources/db-changes/schema/DDP-5808-create-file-question-mime-type-tables.xml
```diff
@@ -0,0 +1,47 @@
...
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="evgeniipr" id="20210517-create-mime-type-table">
+        <createTable tableName="mime_type">
+            <column name="mime_type_id" type="bigint(20)" autoIncrement="true" startWith="1">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="mime_type_pk"/>
+            </column>
+            <column name="mime_type" type="varchar(255)">
```
yufengwng: To follow existing pattern for our other "enumeration" tables, I think column should be named `mime_type_code`.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/FileQuestionDef.java
```diff
@@ -1,12 +1,27 @@
...
+import org.broadinstitute.ddp.util.ConfigManager;
+
+public final class FileQuestionDef extends QuestionDef implements FileUploadSettings {
+
+    @Positive
+    @SerializedName("maxFileSize")
+    private long maxFileSize;
+
+    @SerializedName("mimeTypes")
+    private List<String> mimeTypes;
```
yufengwng: Should we make this a set? What happens if someone specifies the same mime type twice?
```suggestion
    private Set<String> mimeTypes = new HashSet<>();
```

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/FileQuestionDef.java
```diff
@@ -20,14 +35,45 @@ public static Builder builder(String stableId, Template prompt) {
...
+    public long getMaxFileSize() {
+        return maxFileSize;
+    }
+
+    @Override
+    public List<String> getMimeTypes() {
+        return mimeTypes;
+    }
+
+    private static String validateFileMaxSize(long maxFileSize) {
```
yufengwng: How about we check the size in `QuestionDao` instead, when we go insert the definition?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.stringtemplate4.UseStringTemplateSqlLocator;
+
+
+@UseStringTemplateSqlLocator
+public interface JdbiFileQuestion extends SqlObject {
```
yufengwng: Since these sql are pretty simple, how about we put these into `JdbiQuestion` and minimize having so many "jdbi" files?

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.sql.stg
```diff
@@ -0,0 +1,13 @@
+group JdbiFileQuestion;
+
+insert() ::= <<
+insert into file_question (question_id, max_file_size) values (:questionId, :maxFileSize)
```
yufengwng: I recommend just putting these into the annotation itself. They're just one or two lines. And we'll have one less file to navigate to.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @SqlUpdate("insert")
+    int insert(@Bind("questionId") long questionId, @Bind("maxFileSize") long maxFileSize);
+
+    @SqlUpdate("insertMimeType")
+    @GetGeneratedKeys
+    long insertMimeType(@Bind("mimeType") String mimeType);
```
yufengwng: Should this be `findMimeTypeIdOrInsert()`? How should we deal with duplicate mime types? Should be be unique in the `mime_type` table?

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiQuestion.sql.stg
```diff
@@ -89,6 +89,13 @@ select qt.question_type_code as question_type,
          where dqfo.date_question_id = dq.question_id
        ) as date_fields,
 
+       (select group_concat(mt.mime_type separator ',')
```
yufengwng: Nice! But I think [mysql server has a limit](https://dev.mysql.com/doc/refman/5.7/en/aggregate-functions.html#function_group-concat) on the string length returned from `group_concat`.

> The result is truncated to the maximum length that is given by the `group_concat_max_len` system variable, which has a default value of 1024.

`group_concat` works well for something like the date fields since there's only 3 of those (month, day, year), but the mime types are bigger strings in general. Let's assume an average of 20 characters per mime type, we can concat about 50 of those, so maybe this should be fine?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/FileQuestion.java
```diff
@@ -2,20 +2,37 @@
...
 import org.broadinstitute.ddp.model.activity.instance.validation.Rule;
 import org.broadinstitute.ddp.model.activity.types.QuestionType;
 
 public final class FileQuestion extends Question<FileAnswer> {
 
+    @SerializedName("maxFileSize")
+    private long maxFileSize;
+
+    @SerializedName("mimeTypes")
+    private List<String> mimeTypes;
```
yufengwng: Can we ensure that this is not `null` so we don't return null for this in API response?

yufengwng (commented):
Looking good!

yufengwng:
When this is merged, we should update the `fileUploads.maxFileSizeBytes` in Vault. What do folks think we should set the max platform limit to be? 10MB (10485760)?

pepper-apis/docs/specification/src/components/schemas.yml
evgeniipr: But we need to allow null or empty mime types - which means "all mime types are allowed".

pepper-apis/docs/specification/src/components/schemas.yml
evgeniipr: Fixed

pepper-apis/src/main/resources/db-changes/schema/DDP-5808-create-file-question-mime-type-tables.xml
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/FileQuestionDef.java
evgeniipr: Fixed

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiQuestion.sql.stg
evgeniipr: Is it really possible a lot of mime-types in one File Question?
Actually it would be much more convenient to group mime-types (may be to define enum fo it) and then assign to FileQiestions the groups instead of mime types. 
Like the following groups: IMAGE ("image/gif","image/jpeg"...), VIDEO ("video/mpeg", "video/ogg",...), etc.
I mean that in the study builder it would be much more convenient to operate with such groups of mime types than each time to remember which mime types possible for image or video or text formats..
And when returning to frontend a group would be translated to a real list of mime types..

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/FileQuestionDef.java
evgeniipr: May be to do it in both places? Just move out this method to a separate validation class.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/FileQuestionDef.java
evgeniipr: Btw, may be to validate the uniqueness in the FileQuestionDef constructor and throw an exception in case of duplicates? So, it could be a list in such case..

pepper-apis/docs/specification/src/components/schemas.yml
evgeniipr: > When this is merged, we should update the `fileUploads.maxFileSizeBytes` in Vault. What do folks think we should set the max platform limit to be? 10MB (10485760)?

Could it be possible that a user want to upload a video? And 10MB may be not quite enough?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
evgeniipr: Fixed

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.sql.stg
evgeniipr: Fixed

pepper-apis/docs/specification/src/components/schemas.yml
evgeniipr: Made it "required".

pepper-apis/docs/specification/src/components/schemas.yml
evgeniipr: added minItems: 0

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/FileQuestion.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
evgeniipr: Fixed.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiQuestion.sql.stg
evgeniipr: > Let's assume an average of 20 characters per mime type, we can concat about 50 of those, so maybe this should be fine?

I think it could be enough.

But I can provide an alternative way to read mime types during questions data reading: read it by a separate query
in this method (in class JdbiQuestion.java):
```
    Stream<QuestionDto> findQuestionDtosByIds(
            @BindList(value = "questionIds", onEmpty = EmptyHandling.NULL) Iterable<Long> questionIds);
```
If it is ok then I'll do this change.

.circleci/config.yml
```diff
@@ -122,6 +122,9 @@ commands:
             develop)
               DEPLOY_ENV=dev
             ;;
+            DDP-5808-extend-file-upload-api)
```
ssettipalli: May be we can remove this before merge to develop

pepper-apis/docs/specification/src/components/schemas.yml
ssettipalli: 10MB should be a good start. maxFileSizeBytes is a configurable value any way which can be changed any time.

.circleci/config.yml
evgeniipr: Removed FB deploy settings from Circle CI config.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1717,4 +1735,14 @@ private CompositeQuestionDef buildCompositeQuestionDef(CompositeQuestionDto dto,
 
         return builder.build();
     }
+
+    private long findMimeTypeIdOrInsert(String mimeType) {
```
yufengwng: You can put this in `JdbiFileQuestion` as a `default` method.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/FileUploadValidator.java
```diff
@@ -0,0 +1,28 @@
...
+     * @throws IllegalArgumentException the exception is thrown in case of invalid value
+     */
+    public static void validateFileMaxSize(long maxFileSize) throws IllegalArgumentException {
+        Long maxFileSizeConf = ConfigManager.getInstance().getConfig().getLong(MAX_FILE_SIZE_BYTES);
+        String errorMessage = null;
+        final String errorMessagePrefix = "Invalid value of maxFileSize=" + maxFileSize + ". ";
+        if (maxFileSize <= 0) {
+            errorMessage = errorMessagePrefix + "It should be greater than 0.";
+        } else if (maxFileSizeConf != null && maxFileSize > maxFileSizeConf) {
+            errorMessage = errorMessagePrefix + "It should not exceed config value=" + maxFileSizeConf + ".";
```
yufengwng:
```suggestion
            errorMessage = errorMessagePrefix + "It should not exceed max value=" + maxFileSizeConf + ".";
```

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiQuestion.sql.stg
yufengwng: I think typical use-case we'll see is allowing PDFs, or images, so just a handful of mime types. So I think we should be fine for now.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
yufengwng: I still recommend merging this into `JdbiQuestion` so we have fewer files to go through.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/FileQuestion.java
yufengwng: I don't see this addressed yet. How are we ensuring that it's not null? Also, let's be consistent and make this a `Set<String>` like in `FileQuestionDef`.

evgeniipr: It was added @NotNull in one of previous commits.

> Set<String> like in FileQuestionDef.

Done

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
evgeniipr: But I see we have separate classes: JdbiBooleanQuestion, JdbiDateQuestion, JdbiPickListQuestion, JdbiTextQuestion..
So, I added a separate class according to this pattern.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
evgeniipr: I was trying to make it default initially, but it not allows combination "private default". But I wasn't going to make this method available to other classes.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/FileUploadValidator.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
yufengwng: Why not make it available? If someone wants to "find or insert" a mime type (e.g. in a unit test) then they'll have to write their own?

evgeniipr: ok, did it 'default'

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFileQuestion.java
yufengwng: We actually want to get rid of those other `Jdbi*Question` files. For example, the `JdbiBooleanQuestion` file only has one `insert` method... no need for that to be a separate thing.

evgeniipr: Removed. Moved methods to JdbiQuestion.java

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1450,6 +1462,16 @@ default void disableRequiredRule(long questionId, RevisionMetadata meta) {
         return questionDefs;
     }
 
+    default long findMimeTypeIdOrInsert(String mimeType) {
```
yufengwng: Please move this into `JdbiQuestion`.

evgeniipr: Done

yufengwng (approved):

PR #775, opened 2021-05-25
Updates the Count Me In study names [DDP-6117]

bskinner:
[DDP-6117](https://broadinstitute.atlassian.net/browse/DDP-6117)

## Context
The iOS and Android apps use the names of the study to present a list of all available studies on signup and login. This PR replaces the initial names in the `ANGIO`, `cmi-brain`, `CMI-OSTEO`, `cmi-mbc`, `cmi-mpc`, `cmi-esc` projects with the full name of the project.

These changes can be demoed by setting up a new server instance, and initializing one (or all) of the updated studies using `study-builder`. The names can then be checked by calling the [getStudies](https://pepper.datadonationplatform.org/docs#operation/getStudies) operation- with the `umbrella` parameter set to `cmi`- and checking the response JSON object. The updated value should match the `name` property for each study guid.

For example, with a server running on `127.0.0.1`, the url would be ``http://127.0.0.1:5555/pepper/v1/studies?umbrella=cmi`

## Release
Further steps are needed in order for these modifications to be visible in the public environments. This will require a manual database update. See the associated ticket for more details about what is needed.

### DIFF
--- a/study-builder/studies/angio/study.conf
+++ b/study-builder/studies/angio/study.conf
@@ -9,7 +9,7 @@
         "guid": "cmi"
     },
     "study": {
-        "name": "ascproject",
+        "name": "Angiosarcoma Project",
         "guid": "ANGIO",
         "studyEmail": "info@ascproject.org",
         "baseWebUrl": ${baseWebUrl},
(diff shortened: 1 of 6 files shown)


yufengwng (approved):
LGTM

PR #777, opened 2021-05-25
Auth0 migrate v2 -> v3

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-5130?oldIssueView=true
1. Use v3 search engine for user list API calls
2. Add pagination for connections API calls. Didn't find any other API calls like clients that need pagination.


pepper-apis/src/main/java/org/broadinstitute/ddp/client/Auth0ManagementClient.java
```diff
@@ -210,18 +211,30 @@ private boolean shouldRequestToken() {
...
+                try {
+                    mgmtApi.setApiToken(getToken());
+                    ConnectionsPage page = mgmtApi.connections().listAll(filter).execute();
+                    return ApiResult.ok(200, page.getItems());
+                } catch (APIException e) {
+                    return ApiResult.err(e.getStatusCode(), e);
+                } catch (Exception e) {
+                    return ApiResult.thrown(e);
+                }
+            });
```
yufengwng, line 233: What should we do if we get an error for a page result while iterating? Should we return early from the function? A simple test to perform might be:

```java
ApiResult<> apiResult = withRetries(...);
if (apiResult.hasFailure()) {
  return apiResult;
}
```

yufengwng (commented):
Looks good, just a small comment.

pepper-apis/src/main/java/org/broadinstitute/ddp/client/Auth0ManagementClient.java
ssettipalli: Makes sense to return early on failures during iteration.

yufengwng (approved):

PR #779, opened 2021-05-26
Ddp 5924 refactoring to define API for ElasticSearch data query

evgeniipr:
## Context

It is refactored the code in order to define some sort of API for ElasticSearch data query.
This change is initiated for the following reasons:
- in response to a request from @charlottedye to find some example of ElasticSearch data query;
- earlier it was decided that it needs to define some API providing more easy and convenient way of making query to ElasticSearch (considering the logic and data structures of Pepper project) which could be used in all Pepper (backend) projects, like DSS, DSM and others.

After this refactoring the following **Pepper ElasticSearch query API** classes can be defined:
- `ElasticSearchQueryExecutor` - main class for  ElasticSearch data query execution.
- `ElasticSearchQueryUtil` - contains static methods used for query execution, for query string pre-processing and result data parsing.
- `ElasticSearchQueryBuilderUtil` - contains static methods used for building a QueryBuilder which passed as a parameter for searching in ElasticSearch.
- `ElasticSearchIndexType` - enum with list of Pepper ElasticSearch indices' names (not real names but short synonyms: the real names can be resolved by method `ElasticsearchServiceUtil.detectEsIndices()`. 
- `ElasticsearchServiceUtil` - contains different methods including a method `detectEsIndices()` used to detect real indices names.

In addition to these classes there are classes which used for participants lookup but also could be used as samples of ElasticSearch data query:
- `ESParticipantsLookupService`,
- `ESSearchSteps`,
- `ESUsersProxiesResultReader`,
- `ESParticipantsResultReader`,
- `ESParticipantsLookupData`.

Below is an example of querying data by participant GUID:
```
var indices = ElasticsearchServiceUtil.detectEsIndices(studyDto, of(ElasticSearchIndexType.PARTICIPANTS_STRUCTURED));
var participantsEsIndex = indices.get(ElasticSearchIndexType.PARTICIPANTS_STRUCTURED);
var elasticSearchQueryExecutor = new ElasticSearchQueryExecutor(esClient);
elasticSearchQueryExecutor
     .setResultMaxCount(resultsMaxCount)
     .setEsIndex(participantsEsIndex)
     .setFetchSource(new String[] {"profile"})
     .search(
         () -> QueryBuilders.matchQuery("profile.guid", participantGuid),
         (response) -> {
             Map<String, ESUsersIndexResultRow> resultData = new HashMap<>();
             for (var hit : response.getHits()) {
                var hitAsJson = ElasticSearchQueryUtil.getJsonObject(hit);
                var userData = gson.fromJson(hitAsJson.get("profile"), ESUsersIndexResultRow.class);
                resultData.put(userData.getGuid(), userData);
             }
             return resultData;
         }
     );
```
The code defining a query and result processing are called as Lambda-functions.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/ElasticSearchQueryExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.client.RestHighLevelClient;
+import org.elasticsearch.index.query.QueryBuilder;
+
+
+/**
+ * Executes query to ElasticSearch.
+ */
+public class ElasticSearchQueryExecutor<T> {
```
charlottedye, line 39: I like the name of this class, which makes it easy to tell what exactly its role is (and to indicate that other classes do not have this particular role)

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/ElasticSearchQueryExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+
+    protected final RestHighLevelClient esClient;
+
+    protected String query;
+    protected String esIndex;
+    protected String[] fetchSource;
```
charlottedye, line 44: You might consider adding a comment here to tell people to look at ESSearchUtil for the list of options.  Alternately, you might consider creating an enum type for the different sources so that you could have something like `protected Source[] fetchSources;` here, which would naturally lead people to look at the Source type.  Just an option to consider.

charlottedye (dismissed):
I added one (picky) comment to consider, but I generally like this!  Even just changing the name of the class makes things much easier to read.

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/ElasticSearchQueryExecutor.java
evgeniipr: Thanks @charlottedye! I decided to make more better design of this class (so that it could return not only map with results but also list with result rows (added another method for this purpose). And added comments with code sample how to use this class (ElasticSearchQueryExecutor).

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/ElasticSearchQueryExecutor.java
evgeniipr: Yes, I added the comment with code example to the class-level comment. And in the comment I replaced reference to ESSearchUtil to string constants (in order to be clear what these constants mean). 
And I created enum class `IndexSource` where defined all sources and implemented a method to get list of sources for each of used indices.
Hopefully the API now became a bit more clear.

evgeniipr:
@yufengwng do you think I can merge this code to dev after the local testing (and then to test participant lookup functionality on dev - verify that noting is broken)? Or better to create a new ticket for this refactoring?

pepper-apis/src/main/java/org/broadinstitute/ddp/elastic/ElasticSearchQueryExecutor.java
charlottedye: Nice!

charlottedye (approved):
This is great!

PR #785, opened 2021-06-01
DDP-4925 replace app-engine-dispatch with load-balancer

yufengwng:
## Context

Here are the scripts that will help us setup the initial load-balancer to replace app engine `dispatch.yaml`. Start by reading the `init-lb.sh` script. The `add-study-to-lb.sh` script can be used down the line to more easily add LB configs for a new study. See this [wiki page](https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/2635497487/Dispatch+Load+Balancer) for more details.

CC @bskinner since I saw we have a LB for Salt so I could probably use some advice here.

## Checklist

## FUD Score

I have already setup the LB in `dev` environment so that it can be demo-ed there.

## Testing

Initial testing of the LB in `dev` did not show any issues. I was able to reach the various apps and services just fine.

## Release

TLDR: There will be downtime!

First, we need to run `init-lb.sh` to setup the load-balancer. Then, we can run `add-study-to-lb.sh` to add in all the various studies. We might need to do some custom/additional steps depending on the study.

The main source of downtime is switching DNS and provisioning SSL certs. From what I can tell in `dev`, updated DNS records seem to propagate pretty quickly, so downtime was mostly dominated by SSL cert provisioning which took anywhere from 8 - 12 mins. So I think a rough estimate of downtime is about 15 mins.

We should be able to apply this without having to create a new code release.

### DIFF
--- a/scripts/load-balancer/add-study-to-lb.sh
+++ b/scripts/load-balancer/add-study-to-lb.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+#
+# Setup the frontend, backend, and url mappings for a new study to the LB.
+
+set -e
+
+NAME="$(basename "$0")"
+if (( $# < 5 )); then
+  echo "usage: $NAME <GCP_PROJECT> <LB_NAME> <GAE_SERVICE> <STUDY_DOMAIN> <STUDY_IDENTIFIER>"
+  echo ""
+  echo "For STUDY_IDENTIFIER, choose a short concise name for the study."
+  echo "This identifier will be used to name various LB resources for the study."
+  exit 1
+fi
+
+PROJECT_ID="$1"
+REGION='us-central1'
+LB_NAME="$2"
+GAE_SERVICE="$3"
+STUDY_DOMAIN="$4"
+STUDY="$5"
+
+echo "using gcp project: $PROJECT_ID"
+echo "using region: $REGION"
+echo "using lb name: $LB_NAME"
+echo "using gae service: $GAE_SERVICE"
+echo "using study domain: $STUDY_DOMAIN"
+echo "using study identifier: $STUDY"
+echo ""
+
+echo_run() {
+  echo "+ $@"
+  "$@"
+  echo ""
+}
+
+echo "=> setting up backend..."
+
+neg_name="$STUDY-gae"
+service_name="$STUDY-service"
+
+echo_run gcloud --project="$PROJECT_ID" \
+  compute network-endpoint-groups create "$neg_name" \
+  --network-endpoint-type=serverless \
+  --app-engine-service="$GAE_SERVICE" \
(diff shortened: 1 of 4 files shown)


scripts/load-balancer/add-study-to-lb.sh
```diff
@@ -0,0 +1,106 @@
...
+echo "using lb name: $LB_NAME"
+echo "using gae service: $GAE_SERVICE"
+echo "using study domain: $STUDY_DOMAIN"
+echo "using study identifier: $STUDY"
+echo ""
+
+echo_run() {
+  echo "+ $@"
+  "$@"
+  echo ""
```
bskinner, line 34: Would using `set -v` or `set -x` before line 33 and disabling the option (`set +<opt>`) be an option here instead of `echo`, or does the output need to be on `stdout`?

scripts/load-balancer/init-lb.sh
```diff
@@ -0,0 +1,185 @@
...
+echo "=> setting up second LB for http-to-https redirects..."
+
+# Note: we use another LB to handle http-to-https redirects. The second LB
+# mostly just contains frontends with port 80 and same IPs, and no backends.
+# There doesn't seem to be a way to create this using cli interface, so we have
+# to import a YAML file.
+#
+# We're only going to use this http LB for redirecting the study site and not
+# our DSS and DSM backends. HTTP requests might contain credentials, so we want
+# to reject those requests.
```
bskinner, line 117: That's a nice comment 👍

bskinner (approved):
Had a question regarding the `echo_run` function, but looks good otherwise!

bskinner:
Did some digging into using Cloud Armor for blocking the `http` routes explicitly instead of implicitly `404`ing them. Your approach here works great, and achieves the same end. I think we may get some benefit out of explicitly blocking those routes when `http` is used, but that's something to look at later.

scripts/load-balancer/add-study-to-lb.sh
yufengwng: Doesn't have to echo to stdout. I can use `set -x` here.

yufengwng: Hmm nevermind, `set -x` is not what I want here. I think I'll stick with `echo`. Using `set -x` will echo every command, so it means it will also echo the `set +x` after executing the command, resulting in something like:
```
+ gcloud compute ...
+ set +x

+ gcloud compute ...
+ set +x
```

yufengwng: Also, I believe [the docs](https://tldp.org/LDP/abs/html/options.html) is saying both `-v` and `-x` prints to stdout. So I guess my approach with `echo` is pretty much the same in that regard.

PR #789, opened 2021-06-03
DSS API specification cleanup & update

bskinner:
## Context

* Cleanup of discriminator use
* Cleanup of some date/time format descriptions
* Adds more `format` and `description` elements
* Removes `speccy` and switches to [redocly/openapi-cli](https://github.com/Redocly/openapi-cli)
* Updates summaries to fix linter errors, and prevent operation ids from appearing in the index column
* Removed unreferenced types & schemas

Mailing Address Before & After

Cc. @yufengwng @ssettipalli 

## FUD Score

- [x] :relaxed: All good, business as usual

in `docs/specification` run `npm run start`. The docs can then be accessed at `http://127.0.0.1:8080`

## Testing

## Release

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -132,9 +132,9 @@ Activity:
     - sectionIndex
   properties:
     activityType:
-      $ref: '../pepper.yml#/components/schemas/ActivityType'
+      $ref: '#/ActivityType'
     formType:
-      $ref: '../pepper.yml#/components/schemas/FormType'
+      $ref: '#/FormType'
     activityCode:
       type: string
       format: guid
@@ -148,7 +148,7 @@ Activity:
       nullable: true
       description: guid for the parent activity instance in case this is a child activity instance
     statusCode:
-      $ref: "../pepper.yml#/components/schemas/ActivityInstanceStatusType"
+      $ref: "#/ActivityInstanceStatusType"
     title:
       type: string
       description: title of the activity to be displayed
@@ -171,15 +171,15 @@ Activity:
         `readonly` flag is `true`). If the `ddp-Content-Style` header is set to
         `BASIC`, this field will contain unstructured text.
     listStyleHint:
-      $ref: '../pepper.yml#/components/schemas/ListStyle'
+      $ref: '#/ListStyle'
     introduction:
-      $ref: '../pepper.yml#/components/schemas/Section'
+      $ref: '#/Section'
     sections:
       type: array
       items:
-        $ref: '../pepper.yml#/components/schemas/Section'
+        $ref: '#/Section'
     closing:
-      $ref: '../pepper.yml#/components/schemas/Section'
+      $ref: '#/Section'
     lastUpdatedText:
       type: string
       nullable: true
@@ -235,9 +235,9 @@ Activity.Summary:
       type: string
       format: guid
     activityType:
-      $ref: "../pepper.yml#/components/schemas/ActivityType"
+      $ref: "#/ActivityType"
     activitySubtype:
-      $ref: "../pepper.yml#/components/schemas/FormType"
+      $ref: "#/FormType"
     activityName:
       type: string
       description: name of the activity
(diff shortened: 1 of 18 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -57,15 +59,17 @@ JoinMailingListPayload:
       description: The umbrella id of the mailing list the user is interested in (one of umbrellaId OR studyGuid but not both is required)
       example: cmi
       type: string
+      format: guid
     studyGuid:
       description: The studyGuid of the mailing list the user is interested in (one of umbrellaId OR studyGuid but not both is required)
       example: ANGIO
       type: string
+      format: guid
```
MocanaAtBroad, line 60: Is this `guid` format something that is defined somewhere (in this doc or is a well-known thing)?

MocanaAtBroad (commented):
I see why you wanted to upgrade! A bunch of stuff got cleaned up, particularly those refs. discriminators are also a nice addition.
Looks good overall and redoc and related scripts seem to handle new spec just fine.
Have one question.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Nope, it's not- the list of supported formats are defined [here](https://swagger.io/specification/#data-type-format). If the value of the `format` property isn't one of the known values, tools are supposed to treat it as a open-string valued property that can have any value.

From my understanding of the platform, the definition of `GUID` is pulling double-duty- as a specific set of types (Study GUID, Answer GUID, etc), and also as a general way of identifying a unique identifier type. This is using the latter definition, and my intent here is to more visibly call out the identifier types in the API outside of just their name.

bskinner:
It feels quite nice 😂- those `$refs` and the `discriminator` weirdness have been bugging me from the beginning.

> I see why you wanted to upgrade! A bunch of stuff got cleaned up, particularly those refs. discriminators are also a nice addition.
> Looks good overall and redoc and related scripts seem to handle new spec just fine.
> Have one question.

bskinner:
This is ready for review!

pepper-apis/docs/specification/src/endpoints/user.profile.address.temp.yml
```diff
@@ -39,7 +39,7 @@ put:
 
 delete:
   operationId: deleteTemporaryMailingAddress
-  description: removed the temporary mailing address for a participant
+  summary: Delete a patient's temporary mailing address
```
yufengwng:
```suggestion
  summary: Delete a participant's temporary mailing address
```

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2117,24 +2171,27 @@ Invitation:
...
       nullable: true
       example: 2020-05-13T18:58:33.887Z
+      format: YYYY-MM-DDThhmmss.sssZ (ISO-8601)
     acceptedAt:
-      description: timestamp of when invitation was accepted, in ISO-8601 UTC format
+      description: timestamp of when invitation was accepted.
       type: string
       nullable: true
       example: 2020-05-13T19:58:33.887Z
+      format: YYYY-MM-DDThhmmss.sssZ (ISO-8601)
```
yufengwng: Do we need colon separator in the time portion of the format?

yufengwng (commented):
Looks great!

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Ooo, good catch. Yes, we do- we're using the extended format, not the basic. Thank you!

bskinner: Fixed in 829b72f48950963b66cc03e82b6b03020ae5fe80

yufengwng (dismissed):
👍

bskinner:
Brought the branch up to date with `develop`.

yufengwng (approved):

PR #790, opened 2021-06-04
Ddp 6239 rgp export family

charlottedye:
## Context

See DDP-6239.  In Gen2, the code that exports the RGP survey data for use in Seqr includes the family ID.  This was not present in the initial code because the family ID is stored in DSM rather than in Pepper.  This code will read the family ID from ElasticSearch and add it to the export.  If the family ID has not yet been populated, the export will fail.  This is a draft pull request because the family ID is not yet being exported to ElasticSearch.  When that gets added, I plan to test this against the dev ElasticSearch one more time.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +56,67 @@
...
+    public CustomExporter(Config mainConfig, Config exportConfig) {
+        this.mainConfig = mainConfig;
+        this.exportConfig = exportConfig;
+        this.customActivity = exportConfig.getString(CustomExportConfigFile.ACTIVITY);
+        this.customGuid = exportConfig.getString(CustomExportConfigFile.STUDY_GUID);
     }
 
+    List<CustomExportParticipant> extractParticipantDataSetByIds(Handle handle, StudyDto studyDto, Set<Long> userIds)
+            throws IOException {
+        RestHighLevelClient elasticsearchClient = ElasticsearchServiceUtil.getElasticsearchClient(this.mainConfig);
```
yufengwng: Probably makes more sense to initialize this in the constructor.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +56,67 @@
...
-        this.customGuid = cfg.getString(CustomExportConfigFile.STUDY_GUID);
+    public CustomExporter(Config mainConfig, Config exportConfig) {
+        this.mainConfig = mainConfig;
+        this.exportConfig = exportConfig;
+        this.customActivity = exportConfig.getString(CustomExportConfigFile.ACTIVITY);
+        this.customGuid = exportConfig.getString(CustomExportConfigFile.STUDY_GUID);
     }
 
+    List<CustomExportParticipant> extractParticipantDataSetByIds(Handle handle, StudyDto studyDto, Set<Long> userIds)
+            throws IOException {
```
yufengwng: I don't see where `IOException` is coming from. It looks like we're wrapping it in an unchecked exception already.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExportCoordinator.java
```diff
@@ -230,18 +230,22 @@ public boolean hasNext() {
...
+                        if (!userIds.isEmpty()) {
+                            exportLastCompleted = userIds.get(userIds.size() - 1).getCompletedTime();
+                        }
+                        List<CustomExportParticipant> extract = exporter.extractParticipantDataSetByIds(handle, studyDto,
+                                userIds.stream().map(CompletedUserDto::getUserId).collect(Collectors.toSet()));
+                        return new ArrayDeque<>(extract);
+                    });
+                } catch (IOException e) {
+                    throw new DDPException("Error retrieving next participant", e);
+                }
```
yufengwng: Maybe remove the catch of `IOException`? I don't see where it's coming from..

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +56,67 @@
...
+        return createCustomParticipantsFrom(ExportUtil.extractParticipantDataSetByIds(handle, studyDto, userIds, emailStore), handle,
+                studyDto, elasticsearchClient);
+    }
+
+    private static List<CustomExportParticipant> createCustomParticipantsFrom(List<Participant> participants, Handle handle,
+                                                                              StudyDto studyDto,
+                                                                              RestHighLevelClient esClient) {
+        List<CustomExportParticipant> customExportParticipants = new ArrayList<>();
+        for (Participant p : participants) {
+            String familyId = getFamilyId(handle, studyDto, p.getUser().getGuid(), esClient);
```
yufengwng, line 90: I was going to suggest something more efficient, like a bulk query or something, but thinking about the use-case here, I see that most of the time this is an incremental export with a handful of new participants, so I guess this is okay.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +56,67 @@
...
+    private static String getFamilyId(Handle handle, StudyDto studyDto, String userGuid, RestHighLevelClient esClient) throws DDPException {
+        String esIndex = ElasticsearchServiceUtil.getIndexForStudy(handle, studyDto, ElasticSearchIndexType.PARTICIPANTS_STRUCTURED);
+        GetRequest getRequest = new GetRequest(esIndex, "_doc", userGuid);
+        String[] includes = {"dsm"};
+        getRequest.fetchSourceContext(new FetchSourceContext(true, includes, null));
+        GetResponse esResponse;
+        try {
+            esResponse = esClient.get(getRequest, RequestOptions.DEFAULT);
+        } catch (IOException e) {
+            throw new DDPException("Error getting ElasticSearch client to use to get family ID for user guid " + userGuid, e);
```
yufengwng: `Error getting ElasticSearch client` doesn't seem quite right here?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -270,6 +327,7 @@ public int exportDataSetAsCsv(StudyDto studyDto, List<CustomActivityExtract> act
             for (var i = 1; i <= maxInstances; i++) {
                 // Add the headers
                 headers.addAll(attributesCollector.headers());
+                headers.add(familyIdHeader);
```
yufengwng: Are we sure we want to add a column for family_id for every instance? Shouldn't this be added just once? On L359 I see that we add family_id right after the participant columns, so I think this should go right after L274?

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +56,67 @@
...
+        Object sourceElement = source.get("dsm");
+        if (sourceElement == null) {
+            throw new DDPException("Error reading ElasticSearch source: DSM element is null for user guid" + userGuid);
+        } else if (!(sourceElement instanceof Map)) {
+            throw new DDPException("Error reading ElasticSearch source: DSM element is not a map for user guid " + userGuid);
+        }
+
+        Object familyId = ((Map) sourceElement).get(familyIdHeader);
+        if (familyId == null) {
+            throw new DDPException("Error retrieving family ID from ElasticSearch: familyId is not present for user guid " + userGuid);
```
yufengwng: I'm a bit worried about the timing of this. How often will we encounter situation where we need to do a seqr export but DSM hasn't populated the family_id in ES yet? If it's often, will we see a lot of alerts in Slack about this? Will that be too much noise?

charlottedye: I typed up a whole set of options and then I realized probably the first thing I should do is find out how often DSM is populating the family_ids.  I'll check that!

charlottedye: I talked to Simone and this shouldn't be an issue.  The family ID will be generated when the participant signs up and then immediately added to ElasticSearch.  This export only runs on participants who have completed their survey, at which point they should have family IDs.

yufengwng: Great!

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
charlottedye: I tried doing this with a boolean search query and ElasticSearch complained about having to sort the results so I think I'll keep it this way for now.  Good point, though.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -47,20 +57,66 @@
 
     // A cache for user auth0 emails, storing (auth0UserId -> email).
     private static final Map<String, String> emailStore = new HashMap<>();
+    private static final String familyIdHeader = "familyId";
```
yufengwng, line 60: I think all the column names in CSV export use snake_case, so you might want to be consistent and name it `family_id`.

yufengwng (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
charlottedye: I agree with this point and noticed it myself but only after I told the Seqr people that it would be written this way so I think I'm stuck with it...

PR #795, opened 2021-06-08
Ddp 6278 implement event UPDATE_CUSTOM_WORKFLOW

evgeniipr:
## Context

Add new study event configuration in DSS that allows updating the workflow status (means the workflow defined in DSM). 
The event name = `UPDATE_CUSTOM_WORKFLOW`.
Configuration (in study builder):
```
{
  "trigger": { ... },
  "action": {
    "type": "UPDATE_CUSTOM_WORKFLOW",
    "workflow": "string",  # name of the workflow to update
    "status": "string"  # new status value to use
  },
```
The event extra data stored in a separate DB table `update_custom_workflow_event_action` (to store workflow and status).

This event handled by `UpdateCustomWorkflowEventAction.java` which publishes PubSub message to a specified topic (should be specified in config by parameter `pubsub.eventActionTopic`).
The format of PubSub message:
```
**Attributes:**
taskType="UPDATE_CUSTOM_WORKFLOW"
studyGuid="study guid"
participantGuid="participant guid"
**Payload:**
{
  "workflow": "string",
  "status": "string"
}
```
NOTE: it is possible to publish any other types of events with the same API (may be required in future).
NOTE: most of changes in files are because it is added a new parameter `studyGuid` which needs to pass to the event handler.

The document describing this event configuration is in progress (coming soon)...

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/test/java/org/broadinstitute/ddp/model/event/UpdateCustomWorkflowEventActionTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+            handle.rollback();
+        });
+    }
+
+    static class TestEventActionPublisher implements EventActionPublisher {
+
+        @Override
+        public void publishEventAction(EventActionType eventActionType, String eventPayload, String studyGuid, String participantGuid) {
+            assertEquals(expectedEventActionType, eventActionType);
```
yufengwng: I think we could have used `mock`s so that we can run asserts in the test itself and not outside in a separate place like this.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/types/EventActionType.java
```diff
@@ -11,7 +11,8 @@
     PDF_GENERATION(false),
     REVOKE_PROXIES(true),
     UPDATE_USER_STATUS(false),
-    USER_ENROLLED(true);
+    USER_ENROLLED(true),
+    UPDATE_CUSTOM_WORKFLOW(false);
```
yufengwng: Looks like enum is sorted alphabetically? Can we move the new enum variant up?

study-builder/studies/study-example.conf
```diff
@@ -893,5 +893,18 @@
       "dispatchToHousekeeping": "false",  # Action handled synchronously.
       # ...
     },
+
+    # Update DSM workflow status.
+    #
+    # This action can define custom workflow status which will be exported to DSM
+    {
+      "action": {
+        "type": "UPDATE_CUSTOM_WORKFLOW",
```
yufengwng: Looks like things are sorted here. Can you move this block up?

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
```diff
@@ -105,6 +105,7 @@
     public static final String PUBSUB_TASKS_SUB = "pubsub.pubSubTasksSubscription";
     public static final String PUBSUB_TASKS_RESULT_TOPIC = "pubsub.pubSubTasksResultTopic";
     public static final String PUBSUB_TASKS_SUBSCRIBER_AWAIT_RUNNING_TIMEOUT = "pubsub.pubSubTaskSubscriberAwaitRunningTimeout";
+    public static final String PUBSUB_EVENT_ACTION_TOPIC = "pubsub.pubSubEventActionTopic";
```
yufengwng: Why do we need another topic? Can we just use the current pubsub topic that we use to communicate with DSM?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+ * an event of a specified type to Google PubSub topic.<br>
+ * The parameters for PubSub publishing are taken from config.
+ * <b>Event publishing config parameters</b>:
+ * <ul>
+ *     <li>`usePubSubEmulator` - (true or false)</li>
+ *     <li>`googleProjectId` - project ID (for a pubsub topic)</li>
+ *     <li>`pubsub.eventActionTopic` - name of PubSub topic where to publish event</li>
+ * </ul>
+ */
+public class EventActionPubSubPublisher implements EventActionPublisher {
```
yufengwng: Why do we have to create a special "event action" publisher? Isn't there a general "pubsub task" publisher we can use?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/EventConfiguration.java
```diff
@@ -77,46 +85,51 @@ public EventConfiguration(EventConfigurationDto dto) {
...
+                    break;
+                case UPDATE_CUSTOM_WORKFLOW:
+                    eventAction = new UpdateCustomWorkflowEventAction(this, dto, new EventActionPubSubPublisher());
+                    break;
+                default:
+                    throw new DDPException("Event action type: " + eventActionType.name() + " is not properly configured in "
+                            + "the EventConfiguration ctor");
+            }
+        } catch (Exception e) {
+            LOG.error(format("Error handle event %s", eventActionType), e);
```
yufengwng: This will swallow the exception... meaning the constructor will finish as if nothing has happened. Should we forego catching the exception so it propagates up and halts the execution of event processing? Should we rethrow the exception here?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/UpdateCustomWorkflowEventAction.java
```diff
@@ -0,0 +1,54 @@
...
+        super(eventConfiguration, dto);
+        this.workflow = dto.getCustomWorkflowName();
+        this.status = dto.getCustomWorkflowStatus();
+        this.eventActionPublisher = eventActionPublisher;
+    }
+
+    @Override
+    public void doAction(PexInterpreter pexInterpreter, Handle handle, EventSignal signal) {
+        eventActionPublisher.publishEventAction(
+                EVENT,
```
yufengwng: I don't like how the "pubsub task" we're pulishing is an "event action" type. We shouldn't tie EventActionType into this. What we should be publishing is a "pubsub task" type. For example, we currently have `UPDATE_PROFILE` type. We should have another type such as `UPDATE_CUSTOM_WORKFLOW`. Although they have the same name, we shouldn't reuse EventActionType like this.... imagine someone passing in an EventActionType that DSM don't even support!

Also, any chance we can refactor the "pubsub task" thing? I had a hard time digging through the code to figure out what "pubsub task" types we support... Will be great if we had an enum somewhere.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+     * Constructor. Creates PubSub topic for publishing events.
+     */
+    public EventActionPubSubPublisher() throws IOException {
+        this.cfg = ConfigManager.getInstance().getConfig();
+        final boolean usePubSubEmulator = cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
+        this.pubsubConnectionManager = new PubSubConnectionManager(usePubSubEmulator);
+        this.pubSubTopicName = ProjectTopicName.of(
+                cfg.getString(ConfigFile.GOOGLE_PROJECT_ID),
+                cfg.getString(ConfigFile.PUBSUB_EVENT_ACTION_TOPIC));
+        this.eventPublisher = pubsubConnectionManager.getOrCreatePublisher(pubSubTopicName);
```
yufengwng: These feel like pretty heavy operations... should we have a singleton "pubsub task" publisher that we can share?

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
evgeniipr: The existing PubSubTask service works a different way: PubSubTasks service used to listen tasks from DSM (or any other external app) for DSS. So it is task to be executed by DSS. There are 2 special topics used for it:
- "pubsub.pubSubTasksSubscription" - the scubscription listening by DSS for tasks from DSM (or any other external app);
- "pubsub.pubSubTasksResultTopic" - topic where DSS publishes result of a task execution.
Therefore we can not use this service in this event publishing service (where we have an opposite purpose: to send a task from DSS to DSM). So I suggest to use a separate topic for this.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
evgeniipr: The existing PubSubTask service - for opposite purpose (it used to consume tasks from DSM to be executed on DSS side). And the current ticket: to send a task from DSS for DSM (to be consumed by DSM).
In the "PubSub Task service" the publisher used to publish Result of PubSub task execution by DSS (assigned to DSS by outside app like DSM via incoming topic): the result published to a dedicated topic specified by conf parameter "pubsub.pubSubTasksResultTopic". So I'm not sure that it makes sense to reuse the same topic for sending tasks to DSM (as the semantic of these actions: "send results of PubSub task execution by DSS" and "send a task to be executed by DSM" are different and in can confuse developers. If it not very resource/cost expensive to have yet another PubSub topic for tasks assigned for DSM I'd better suggest to add such separate topic..
But I agree that we can/should reuse Java code implemented PubSub Publisher. I'll try to do it.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/UpdateCustomWorkflowEventAction.java
evgeniipr: I don't see a possibility to reuse "pubsub task" because it is used for opposite purpose: to consume tasks created by DSM for DSS (actually it can be not only tasks from DSM but from any other external application which can put tasks to the "pubSubTaskTopic", but anyway this PubSubTask service intended to consume and process tasks assigned for DSS by outside applications).
In this current ticket the similar task consuming service should be implemented on DSM side and on DSS side we implement a client side publishing task for DSM. So the currently implementing service can be named "DSM tasks client".
But I agree that needs to design this DSM tasks client a more generic way: replace `EventActionType` with a more generic parameter: String dsmTask.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
evgeniipr: I agree about singleton. But I'd better not touch "pubsub task" for a while. As anyway we can not reuse this service in this (having opposite purpose) task.
May be we could refactor "pubsub task" later when implement such publisher singleton and reuse it in PubSub task service. The matter is that PubSubTask service is initialized in the Housekeeping class where it is already initialized some of PubSub classes.. But in this current ticket (to process events and send them to DSM via PubSub) we have a problem to pass parameters from main Housekeeping class because `UpdateCustomWorkflowEventAction.doAction()` is called from `EventConfiguration.doAction()` which eventually called by `EventService.processAllActionsForEventSignal` which triggered in different parts of DSS code.. Therefore we need somehow to read configuration/initialize pubsub publisher, we can not reuse classes already instantiated in class `Housekeeping`. But anyway I agree about the moving PubSub initialization code to a separate singleton.
Doing this task we need somehow to keep in mind that similar code we have in `Housekeeping` and it needs to create common solution for entire DSS..

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/EventConfiguration.java
evgeniipr: But if an exception not swallowed and re-thrown then it can interrupt the execution of other logic where events thrown. Is it good that due to some PubSub communication problem a process which triggers an event will be failed? May be for reliable delivering of an event to DSM via PubSub it needs to invent some special approach: some flag in table `update_custom_workflow_event_action` which should be checked by an async job publishing events to PubSub: and if flag is false then try to publish event and set it on successful  publishing ...

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/types/EventActionType.java
evgeniipr: Fixed

study-builder/studies/study-example.conf
evgeniipr: Fixed

pepper-apis/src/test/java/org/broadinstitute/ddp/model/event/UpdateCustomWorkflowEventActionTest.java
evgeniipr: Fixed (run asserts in test)

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/EventConfiguration.java
evgeniipr: I fixed: DDPException() is re-thrown now, but I'm not sure that it is a good solution. In future I suggest to implement asynchronous job for such events publishing.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
evgeniipr: PubSub publisher initialization code is moved out from the `EventActionPubSubPublisher.java` to class `PubSubPublisherInitializer.java`.

pepper-apis/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
yufengwng: Can you connect with @Irakli-Khaladze or whoever is working on the DSM-side for this feature? And make sure we're on the same page and agree on the topics and such?

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
yufengwng: The way I had been thinking about it is that we just have a "communication channel" between DSS and DSM, and we're passing messages around. The message could be a "result for update_profile" or it could be "please update_custom_workflow". Maybe I'm misunderstanding things here.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/PubSubPublisherInitializer.java
```diff
@@ -0,0 +1,55 @@
...
+            } catch (IOException e) {
+                throw new DDPException(format("Error during creation of PubSub publisher for topic %s", key), e);
+            }
+        });
+    }
+
+    public static class PubSubPublisherData {
+
+        private final Publisher publisher;
+        private final ProjectTopicName pubSubTopicName;
```
yufengwng: The `Publisher` actually know what GCP project and topic it's for (e.g. `publisher.getTopicName()`). So you don't have to bundle this into another class.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventPubSubPublisher.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public void publishEvent(
+            String eventType,
+            String payload,
+            String studyGuid,
+            String participantGuid) {
+
+        var publisherData = PubSubPublisherInitializer.getOrCreatePubSubPublisherData(
+                ConfigFile.PUBSUB_EVENT_ACTION_TOPIC);
```
yufengwng: What happens if we want to send a message to DSM outside of an event action? I recommend renaming this to something else.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/PubSubPublisherInitializer.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventPubSubPublisher.java
evgeniipr: @yufengwng you mean to rename ConfigFile parameter?

yufengwng: Not just ConfigFile parameter, but in general.

evgeniipr: Fixed. Refactored

yufengwng (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/EventActionPubSubPublisher.java
evgeniipr: > "communication channel" between DSS and DSM

@yufengwng I strongly agree it is a good idea to have a single communication channel: may be we could discuss it with DSM team and redesign it such way that 2 topics are used for communication between DSM and DSS, but it require some redesign on DSS and DSM sides. Unfortunately we not discussed common design (some common approach - I mentioned this problem during some of "Retro" meetings): currently we design features separately. But I think that it could be reused the same code for both apps (DSM and DSS)..

Yet another understanding of PubSubTasks exists: it is not tasks for DSS from DSM only but from any system from outside world. In theory a task for DSS could be published to "incoming PubSub topic" by any external application (and this external app can process response posted by DSS to "result topic"). So, PubSubTask - an instrument to command by DSS from outside.
In current task we do similar thing but for DSM (to command it from outside - and such commands could be assigned not only by DSS but any app having access to dedicated PubSub topics)...
So, it requires another pair of topics to command DSM and process its results (though in the current events processing task we do not used result topic - to check results from DSM)...
Of course we can reuse the same 2 topics for both apps (for commands for DSS and for commands for DSM) but it needs some additional attribute in it.. May be we could define more general design in some next iteration?
And it is definitely should be re-used same code in both apps as it does same things. Now it is completely separate implementations..

evgeniipr:
@yufengwng please re-approve: I made some changes. 
The basic ones: 
- add topic param to config templates;
- rename topic config parameter to make it more conforming to already existing pubsub params;
- done small refactoring of PubSub publisher classes.

yufengwng (approved):

PR #802, opened 2021-06-10
DDP-5710 rgp: add dashboard summaries

yufengwng:
Add dashboard summaries per DDP-5710.

Example of what it looks like now:

Screen Shot 2021-06-10 at 5 13 19 PM


charlottedye (approved):
Looks good!

PR #806, opened 2021-06-11
DDP-6351 rgp: add other option to title question

yufengwng:
Make the "title" question optional and add an "Other" option as per DDP-6351.

Note: This change should not affect our data migration efforts, since 1) in Gen2 this question is required so folks should have an answer for it, 2) it's a new option so older participants shouldn't have it, and 3) migration only cares about the existing options and not this new Other option.


charlottedye (approved):
Looks good

PR #807, opened 2021-06-11
load-balancer: update scripts to handle already reserved IP

yufengwng:
## Context

For some studies, we don't manage the DNS records. Therefore, we'll need to reserve an IP for the study ahead of time so we can work with study staff to plan DNS switch. I updated the load-balancer scripts to handle this case.

## Checklist

## FUD Score

## Testing

I tested this by running it for the basil study:

```
$ ./reserve-ip.sh broad-ddp-dev pepper-lb basil
$ ./add-study-to-lb.sh broad-ddp-dev pepper-lb basil basil.dev.datadonationplatform.org basil
```

### DIFF
--- a/scripts/load-balancer/add-study-to-lb.sh
+++ b/scripts/load-balancer/add-study-to-lb.sh
@@ -60,11 +60,25 @@ cert_name="$STUDY-cert"
 proxy_name="$LB_NAME-$STUDY-proxy"
 frontend_name="$LB_NAME-$STUDY-frontend"
 
-echo_run gcloud --project="$PROJECT_ID" \
-  compute addresses create "$ip_name" --global
+# Check if IP address is already reserved or not. Since this can fail if it
+# doesn't exit, we temporary turn off `-e` and redirect fd 2 so we minimize
+# noise on script output.
+set +e
 ip_addr="$(gcloud --project="$PROJECT_ID" \
   compute addresses describe "$ip_name" \
-  --global --format='get(address)')"
+  --global --format='get(address)' 2>/dev/null)"
+set -e
+
+if [[ -z "$ip_addr" ]]; then
+  echo_run gcloud --project="$PROJECT_ID" \
+    compute addresses create "$ip_name" --global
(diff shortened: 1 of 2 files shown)


scripts/load-balancer/add-study-to-lb.sh
```diff
@@ -60,11 +60,25 @@ cert_name="$STUDY-cert"
...
-echo_run gcloud --project="$PROJECT_ID" \
-  compute addresses create "$ip_name" --global
+# Check if IP address is already reserved or not. Since this can fail if it
+# doesn't exit, we temporary turn off `-e` and redirect fd 2 so we minimize
+# noise on script output.
+set +e
 ip_addr="$(gcloud --project="$PROJECT_ID" \
   compute addresses describe "$ip_name" \
-  --global --format='get(address)')"
+  --global --format='get(address)' 2>/dev/null)"
```
bskinner, line 69: What are the upsides of using a `global` address instead of a `regional` one?

yufengwng: I'm actually not too sure. It appears to me that that's what we've been using, so I just followed suite.

bskinner: How difficult would this be to rework to take a region parameter (either via the CLI, the environment, or google's config, etc)? From my understanding, most, if not all, of our infrastructure is within the `us-central1-a` region, and having a `global` IP could hit us with additional cross-region fees for traffic. We currently don't have a setup in which Google's routing for the global IPs would be a boon for us.

yufengwng: So far, the load-balancer is setup as a global LB, because I think that was how we did previously. So I think we need a global IP here?

bskinner: Sounds like switching from global IPs would be an intensive change. Let's keep it as-is.

bskinner (approved):
Looks good 👍

PR #820, opened 2021-06-17
Ddp 6281 autocomplete render mode

evgeniipr:
## Context
This new feature is being added to support the entry of type of cancers in Pan Cancer study.
We currently support 3 render modes for pick list questions
LIST
DROPDOWN
CHECKBOX_LIST
We want to support a fourth kind: AUTOCOMPLETE.

**The following tasks are done:**
- added picklist render mode of type AUTOCOMPLETE;
- created a Liquibase script to add this new type to DB;
- Updated Open API; 
- added demo list of type AUTOCOMPLETE to **basil** project. 
NOTE: The last task still in progress because in order to generate a study builder conf with a list of more than 100 options (for example knowns cancers), it needs to implement a small tool for it. I'm working on this tool now.
I just started to create a list (I wanted to create a list of all countries as a demo of a big list) but then understood than to do it manually - not a most optimal solution.. So the list is just started but not completed. Will complete it as soon as create that generation tool.


yufengwng:
> I wanted to create a list of all countries as a demo of a big list

Using a list of countries as an example is a great idea! A bunch of our studies have a "country" question, so I guess we could have just copy-pasted from there if we want to call it a day...

study-builder/studies/basil/i18n/en.conf
```diff
@@ -7,4 +7,16 @@
...
   },
+  "autocomplete": {
+    "name": "Picklist Autocomplete",
+    "title": "Title: Picklist Autocomplete feature",
+  },
+  "country": {
+    "afghanistan": "Afghanistan",
+    "albania": "Albania",
+    "algeria": "Algeria",
+    "american_samoa": "American Samoa",
```
yufengwng, line 18: Since there's a lot of countries, and since it's just the `basil` demo study, I think we can skip translations. Just use the English for the country names and be done with it, e.g. just hardcode
```
{
   ... [
    { "language": "en", "text": "Canada" },
    { "language": "es", "text": "Canada" },
   ]

   ... [
    { "language": "en", "text": "..." },
    { "language": "es", "text": "..." },
   ]
}
```

study-builder/studies/basil/picklist-autocomplete-activity.conf
```diff
@@ -0,0 +1,31 @@
+{
+  include required("../snippets/activity-general-form.conf"),
+  "studyGuid": ${id.study},
+  "activityCode": ${id.act.autocomplete},
+  "versionTag": "v1",
+  "displayOrder": 2,
+  "writeOnce": false,
+  "editTimeoutSec": null,
+  "maxInstancesPerUser": 5,
```
yufengwng: Let's remove `maxInstancesPerUser` so we can create as many of these as we want for testing.

study-builder/studies/basil/i18n/en.conf
evgeniipr: I generated this conf with StringTemplate, therefore not a big problem to generate 2 languages.

study-builder/studies/basil/picklist-autocomplete-activity.conf
evgeniipr: Fixed

yufengwng (dismissed):

yufengwng (approved):

PR #822, opened 2021-06-17
Config updates

ssettipalli:
Get assets bucket name from config 
we need this to make config compatible for prod bucket name / config

### DIFF
--- a/study-builder/studies/atcp/assent.conf
+++ b/study-builder/studies/atcp/assent.conf
@@ -302,13 +302,13 @@
           "state": "INCOMPLETE",
           "height": 25,
           "width": 25,
-          "1x": "https://storage.googleapis.com/${assetsBucketName}/icons/Assent_itemDot_purple.svg"
+          "1x": "https://storage.googleapis.com/"${assetsBucketName}"/icons/Assent_itemDot_purple.svg"
         },
         {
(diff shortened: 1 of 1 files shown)


study-builder/studies/atcp/assent.conf
```diff
@@ -302,13 +302,13 @@
           "state": "INCOMPLETE",
           "height": 25,
           "width": 25,
-          "1x": "https://storage.googleapis.com/dev-at-assets/icons/Assent_itemDot_purple.svg"
+          "1x": "https://storage.googleapis.com/${assetsBucketName}/icons/Assent_itemDot_purple.svg"
```
yufengwng: I don't think HOCON handles substitutions inside strings. I think you need to do _string concatenation_
```suggestion
          "1x": "https://storage.googleapis.com/"${assetsBucketName}"/icons/Assent_itemDot_purple.svg"
```

bskinner: @yufengwng's correct- it does not. From the [Concatenation](https://github.com/lightbend/config#concatenation) section of the HOCON [docs](https://github.com/lightbend/config)
```
Note: the ${} syntax must be outside the quotes!
```

ssettipalli:
will do a diff PR because repo changed

PR #834, opened 2021-06-23
DDP-6334 - implemented templating syntax for picklist answer substitution

evgeniipr:
## Context

DDP-6334 - implemented templating syntax for picklist answer substitution: $ddp.answer(stableId, fallbackText, useDetailTextForPickList) - it overloads existing $ddp.answer() adding 3rd boolean parameter: if it's true, then get 'detailText' (for picklist only);
fixed a problem with $ddp.answer() calculation: run templates rendering and templates apply 2 times - in order to precalculate values in referred questions (referred by 1st stableId parameter).


pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -248,7 +268,27 @@ private String renderAnswerUsingFormResponse(String questionStableId, String fal
...
 
-    private String renderAnswerUsingFormInstance(String questionStableId, String fallbackValue) {
+    private String optionLabelRender(Map<String, PicklistOptionDef> options, SelectedPicklistOption selected, String fallbackValue) {
+        PicklistOptionDef option = options.get(selected.getStableId());
+        return option.getOptionLabelTemplate() == null ? fallbackValue : option.getOptionLabelTemplate().render(isoLangCode);
+    }
+
+    private String detailLabelRender(Map<String, PicklistOptionDef> options, SelectedPicklistOption selected, String fallbackValue) {
+        PicklistOptionDef option = options.get(selected.getStableId());
+        Template template = option.getDetailLabelTemplate() == null ? option.getOptionLabelTemplate() : option.getDetailLabelTemplate();
```
yufengwng: Sorry, I think I've caused a misunderstanding here. What I mean by "detail text" in the ticket is that, we want what the user typed in for the "please provide more details" input. To explain:

Screen Shot 2021-07-01 at 5 48 09 PM

Ultimately, what we want is to substitute the user's answer. What does that mean for picklist question? If they selected an option, then we want to substitute what they have selected. If they selected a regular picklist option, we substitute the "option label" text, because that's what the user sees, so the substitute will make sense to them. What about an option that have an "detail" input? We have two choices: 1) substitute the "option label" text, 2) substitute the user's answer to the "detail" input.

This new 3rd argument is to control which one we substitute. When it's `true` we want to do (2) and substitute what they have typed.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreator.java
```diff
@@ -92,12 +94,35 @@ public void updateBlockStatuses(AIBuilderContext ctx) {
...
+    /**
+     * The rendering done in 2 runs - in order to correctly calculate templates depending on results of other templates
+     * (for example with using of Velocity method $ddp.answer())
+     */
+    public void renderContent(AIBuilderContext ctx) {
+        Map<Long, String> renderedTemplates = new TreeMap<>();
+
+        // 1st run
+        renderTemplates(ctx, renderedTemplates);
+        applyRenderedTemplates(ctx, renderedTemplates);
```
yufengwng: Just so I understand: Is this first call to "apply" needed because we need to associate the option label text, so that on the second run we'll see it in the `$ddp.answer()` substitutions?

yufengwng (commented):
Nice! I'm glad there was support for overloading and we've implemented it that way. However, I think I've caused a misunderstanding -- see my comment below. I wasn't expecting the refactor to how we're rendering templates so still wrapping my head around what that is about.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreator.java
evgeniipr: Yes, for this purpose.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
evgeniipr: Yes. Sorry, my fault: misunderstood what meant "detail text". Now I fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreator.java
evgeniipr: In the design doc there is a requirement: "Another limitation is that we can substitute answers into things like activity name and title, but not contents/questions.".
I think that 2 runs of templates applying should help to solve this problem.

yufengwng: Right. That problem is captured in another ticket, so I wasn't expecting it here. But that's fine.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
```diff
@@ -0,0 +1,67 @@
...
+ * Utility methods for rendering selected picklist options and picklist option detail text.
+ */
+public class SelectedPickListOptionRenderUtil {
+
+    public static String selectedOptionLabelsRender(QuestionDef questionDef, Answer answer, String fallbackValue, String isoLangCode) {
+        Map<String, PicklistOptionDef> options = ((PicklistQuestionDef) questionDef)
+                .getAllPicklistOptions().stream()
+                .collect(Collectors.toMap(PicklistOptionDef::getStableId, Function.identity()));
+        return ((PicklistAnswer) answer).getValue().stream()
+                .map(selected -> renderOptionLabelOrFallback(options, selected, fallbackValue, isoLangCode))
```
yufengwng: Rendering option label OR fallback is different from current behavior. Currently, we don't use the fallback in this way. Notice that in `RenderValueProvider`, the fallback is used if there is no answer for the question (specifically we check `answer.isEmpty()`). So by the time we get to this rendering here, we're guaranteed that there's at least one selected option. Also, the option_stable_id should match up, otherwise, it means we've selected an option that's not part of the question, and that will mean there's something else wrong here. So my point is that, we shouldn't need to deal with the fallback here. In fact, if we used fallback here, it might be hiding a potential bug.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
```diff
@@ -0,0 +1,67 @@
...
+                .collect(Collectors.joining(","));
+    }
+
+    public static String selectedOptionLabelsRender(Question question, Answer answer, String fallbackValue) {
+        Map<String, String> options = ((PicklistQuestion) question)
+                .streamAllPicklistOptions()
+                .collect(Collectors.toMap(PicklistOption::getStableId,
+                        (p) -> p.getOptionLabel() == null ? fallbackValue : p.getOptionLabel()));
```
yufengwng: Likewise, let's reconsider using fallback here.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
```diff
@@ -0,0 +1,67 @@
...
+        return ((PicklistAnswer) answer).getValue().stream()
+                .map(selected -> options.get(selected.getStableId()))
+                .collect(Collectors.joining(","));
+    }
+
+    public static String detailTextRender(QuestionDef questionDef, Answer answer, String fallbackValue, String isoLangCode) {
+        String detailText = getDetailText(answer);
```
yufengwng: Hmmm, recall that EACH picklist option MAY allow inputting additional details. In our use-case for PanCan, there's only one option with detail text, but that might not be the case for others. Notice how in the methods above we're doing "map" and "collect" and joining with a semicolon. I think what we should do here, is to iterate through all the selected options (like what we're doing in `selectedOptionLabelsRender()`) and if `useDetails` is true AND the selected option has detail text, then we use that instead of the option label. So I think it might be better off passing `useDetails` into `selectedOptionLabelsRender()` and handling it there.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/ActivityInstanceFromDefinitionBuilder.java
```diff
@@ -204,6 +204,9 @@ public ActivityInstanceFromDefinitionBuilder buildFormInstance() {
                 var formInstance = context.getAIBuilderFactory().getFormInstanceCreator().createFormInstance(context);
                 context.setFormInstance(formInstance);
 
+                context.getAIBuilderFactory().getTemplateRenderHelper().addInstanceToRendererInitialContext(
+                        context, context.getFormInstance());
```
yufengwng: I was thinking it'll be better to pass in the "form response" object instead. If we have the "form response", then I think we'll have less of an issue with render (and not needing to do 2 rendering runs) since the rendering will pull from the activity definition and template instead of the "form instance" label strings.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreator.java
yufengwng: Now that I think about it, maybe we can avoid this need for 2 runs if we tried passing in the "form response" object and use the "render from form response" approach in `RenderValueProvider` instead? I suspect we're having this issue because we're looking for option label in the "form instance", which is not fully created yet because that's the thing we're trying to build in the "activity instance builder"! With "form response", we shouldn't have this problem because we're pulling the label right from the template and activity definition.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/ActivityInstanceFromDefinitionBuilder.java
evgeniipr: Fixed: put form response and activity definition object to renderer context. 
Removed 2nd pass of templates apply.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
evgeniipr: Fixed.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreator.java
evgeniipr: Fixed: removed 2 runs and replaced "form instance" with "form response".

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
```diff
@@ -0,0 +1,59 @@
...
+                .collect(Collectors.toMap(PicklistOption::getStableId, (p) -> p.getOptionLabel()));
+        return ((PicklistAnswer) answer).getValue().stream()
+                .map(selected -> optionRender(selected, options, useDetailTextForPickList))
+                .collect(Collectors.joining(","));
+    }
+
+    private static String optionRender(SelectedPicklistOption selected, Map<String, PicklistOptionDef> options,
+                                       String isoLangCode, boolean useDetailTextForPickList) {
+        if (useDetailTextForPickList) {
+            return selected.getDetailText() != null ? selected.getDetailText() :
```
yufengwng: Should we check that it's not-null and not-blank?

yufengwng (commented):
Tried this locally and looks good to me! Just one small comment to consider.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/SelectedPickListOptionRenderUtil.java
evgeniipr: You mean to check if not null in this expression?
`options.get(selected.getStableId()).getOptionLabelTemplate()`
I hope that `stableId` and `optionLabelTemplate` can not be null: at least in DB it is `NOT NULL`.

yufengwng: I mean:
```
selected.getDetailText() != null && !selected.getDetailText().isBlank()
```

evgeniipr: Added isNotBlank() check.

yufengwng (approved):

PR #860, opened 2021-07-07
DDP-5910 Picklist-boolean PDF fields

YuriyPavlovBI:
I extended abilities of PDF form filling: now we can set checkboxes based on some picklist answers. For example:

This:
image

Could be used to generate PDF like this:
image


yufengwng:
Have we considered expanding the "picklist" pdf substitution type to do this? I think it will feel more natural to specify a picklist substitution and specify a mapping of form fields to picklist options.

YuriyPavlovBI:
> Have we considered expanding the "picklist" pdf substitution type to do this? I think it will feel more natural to specify a picklist substitution and specify a mapping of form fields to picklist options.

I was thinking about this too but it turned out to be more "natural" and simple the way it is in this PR. Check org/broadinstitute/ddp/service/PdfGenerationService.java:718, there we have one PDF form field per substitution which makes mapping **picklist->set of fields** not possible. Also, boolean substitution already has extension DB table so we just add one field and keep it simple.

yufengwng:
> Check PdfGenerationService.java:718

I don't see why we couldn't make changes here to accommodate new picklist substitution feature where we have a mapping of options->fields? In L720, I see:

```
if (field == null && substitution.getQuestionType() != QuestionType.COMPOSITE) {
```

Could we not add a new condition in the `if` for picklist?

YuriyPavlovBI:
> > Check PdfGenerationService.java:718
> 
> I don't see why we couldn't make changes here to accommodate new picklist substitution feature where we have a mapping of options->fields? In L720, I see:
> 
> ```
> if (field == null && substitution.getQuestionType() != QuestionType.COMPOSITE) {
> ```
> 
> Could we not add a new condition in the `if` for picklist?

So, "placeholder" will contain a prefix and not an actual PDF field. If it is so (no field found with that name), we will iterate through picklist selected options trying to find a field named _prefix_optionstableid_ and set it as checked.
Does this sound ok?

yufengwng:
How about this?

(1) For the `PICKLIST` pdf substitution type, we'll make the `field` property optional, and allow setting a `fields` mapping of field names to option stable ids:
```
# study.conf
...
"substitutions": [
  {
    "type": "PICKLIST",
    "activityCode": "...",
    "questionStableId": "...the picklist question stable id...",
    "fields": {
      "field1": "option1",
      "field2": "option2",
      ...
    }
  },
]
```

(2) In `PdfBuilder`, handle the new configuration for `PICKLIST`. If `field` is not there, just set an empty string or something.

(3) Add new tables to store new configurations.

(4) Update DAOs and DTOs to handle new configurations.

(5) In `PdfGenerationService`, update `substitutePicklist()` to handle possibility of `fields` mapping by iterating through and finding form fields to check off.

YuriyPavlovBI:
> How about this?
> 
> (1) For the `PICKLIST` pdf substitution type, we'll make the `field` property optional, and allow setting a `fields` mapping of field names to option stable ids:
> 
> ```
> # study.conf
> ...
> "substitutions": [
>   {
>     "type": "PICKLIST",
>     "activityCode": "...",
>     "questionStableId": "...the picklist question stable id...",
>     "fields": {
>       "field1": "option1",
>       "field2": "option2",
>       ...
>     }
>   },
> ]
> ```
> 
> (2) In `PdfBuilder`, handle the new configuration for `PICKLIST`. If `field` is not there, just set an empty string or something.
> 
> (3) Add new tables to store new configurations.
> 
> (4) Update DAOs and DTOs to handle new configurations.
> 
> (5) In `PdfGenerationService`, update `substitutePicklist()` to handle possibility of `fields` mapping by iterating through and finding form fields to check off.

OK, sounds reasonable.
Please check the implementation

pepper-apis/src/main/resources/db-changes/schema/DDP-5910-pdf-picklist-fields-mapping.xml
```diff
@@ -0,0 +1,28 @@
...
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet id="20181003-pdf-picklist-fields-mapping" author="ypavlov">
+        <createTable tableName="pdf_picklist_field_mapping">
+            <column name="pdf_substitution_id" type="bigint">
+                <constraints nullable="false"
```
yufengwng, line 11: Should we set this as the primary key for this table as well?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
```diff
@@ -510,27 +522,37 @@ public void accumulate(Map<Long, PdfTemplate> container, RowView view) {
...
-                    QuestionType qtype = QuestionType.valueOf(view.getColumn("a_question_type", String.class));
-                    if (qtype == QuestionType.BOOLEAN) {
-                        sub = view.getRow(BooleanAnswerSubstitution.class);
-                    } else {
-                        sub = view.getRow(AnswerSubstitution.class);
-                    }
-                    break;
-                default:
-                    throw new DaoException("unhandled pdf substitution type " + type);
+            long substId = view.getColumn("p_pdf_substitution_id", Long.class);
```
yufengwng: They're the same thing, but for consistency I think we should use column `a_pdf_substitution_id` here.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
```diff
@@ -845,6 +859,25 @@ private void substitutePicklist(Answer answer, PdfFormField field) {
...
+        if (answer != null) {
+            for (SelectedPicklistOption option : answer.getValue()) {
+                selectedOptions.add(option.getStableId());
+            }
+        }
+
+        for (String fieldName : mapping.keySet()) {
+            PdfFormField field = form.getField(fieldName);
+            if (field == null) {
+                LOG.warn(String.format("Could not find PDFFormField field with name: %s", fieldName));
```
yufengwng: Should this be an error instead? Maybe we should pass in `List<String> errors` and append to it?

yufengwng (commented):
Looks good! Just a few comments.

pepper-apis/src/main/resources/db-changes/schema/DDP-5910-pdf-picklist-fields-mapping.xml
YuriyPavlovBI: It's not a primary key and not unique

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
YuriyPavlovBI: Agree

YuriyPavlovBI: Hm, `a_pdf_substitution_id` is null for substitutions without `fields`

pepper-apis/src/main/java/org/broadinstitute/ddp/service/PdfGenerationService.java
YuriyPavlovBI: Sure

pepper-apis/src/main/resources/db-changes/schema/DDP-5910-pdf-picklist-fields-mapping.xml
yufengwng: Oh right, a substitution can have many mappings.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PdfDao.java
YuriyPavlovBI: @yufengwng What do you think?

yufengwng: Wait, I'm not understanding why it's null when it's coming from the same column as `p_pdf_substitution_id`?

YuriyPavlovBI: My bad, I misunderstood the point. Fixed

yufengwng (approved):
LGTM

PR #862, opened 2021-07-08
Ddp 6306 extend activity instance creation

evgeniipr:
## Context
(draft PR: still in progress): in needs some feedback about the implementation.

Extend activity instance creation event action to allow creating multiple instances and copying values from picklist option answers.
The basic idea of this new feature: a parent activity instance contains a composite component where can be selected (or entered as detail text) N answers inside picklists (for example a picklist with cancers). And after submitting of this parent activity it should be created a set of new activities (of a specified type) - number of activities equals to selected answers. And each of answers option (or detail text) should be copied to a hidden picklist (containing same elements as source picklist): each answer into it's own newly created activity.
The most of the changes re done in class `ActivityInstanceCreationEventAction`.
The logic of activity instances creation is moved to 2 new classes:
`ActivityInstanceCreatorDefault`,
`ActivityInstanceCreatorFromAnswers`.
It is also created a class containing methods used during activity instance creation: `ActivityInstanceCreationService`.

The study builder configuration which enables this new feature:
```
{
  ...
  “action”: {
    “type”: “ACTIVITY_INSTANCE_CREATION”,
    “activityCode”: “THE_ACT_TO_CREATE”,
    “createFromAnswer”: true,
    “sourceQuestionStableId”: “SOURCE_CHILD_PICKLIST_INSIDE_COMPOSITE”,
    “targetQuestionStableId”: “TARGET_PICKLIST”
  }
}
```

## FUD Score

These changes could be tested in PanCan project: when cancer data are selected in prequal activity and this activity is submitted.

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
```diff
@@ -277,6 +286,18 @@ public Long getActivityInstanceCreationStudyActivityId() {
         return activityInstanceCreationStudyActivityId;
     }
 
+    public Boolean isCreateFromAnswer() {
```
yufengwng: Maybe call this `getCreateFromAnswer()` since it's a nullable boolean?

study-builder/studies/study-example.conf
```diff
@@ -684,10 +684,18 @@
...
+    # will be copied to component "targetQuestionStableId" (in each new instance). So, 1st answer copied to 1st
+    # instance, 2nd answer - to 2nd instance, etc...
     {
       "action": {
         "type": "ACTIVITY_INSTANCE_CREATION",
-        "activityCode": "string"
+        "activityCode": "string",
+        "createFromAnswer": "bool",           # if true then create instances from "sourceQuestionStableId" answers
+                                              # (by default it is false)
+        "sourceQuestionStableId": "string",   # required, where to grab the source answer
```
yufengwng: "required if createFromAnswer is true"

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -257,7 +257,10 @@ private long insertEventAction(Handle handle, Config actionCfg, Config triggerCf
                     .orElseThrow(() -> new DDPException("Could not find pdf configuration with name " + pdfName));
             return actionDao.insertPdfGenerationAction(pdfId);
         } else if (EventActionType.ACTIVITY_INSTANCE_CREATION.name().equals(type)) {
-            String activityCode = actionCfg.getString(ACTIVITY_CODE_FIELD);
+            String activityCode = actionCfg.getString("ACTIVITY_CODE_FIELD");
+            Boolean createFromAnswer = ConfigUtil.getBoolIfPresent(actionCfg, "createFromAnswer");
+            String sourceQuestionStableId = actionCfg.getString("sourceQuestionStableId");
+            String targetQuestionStableId = actionCfg.getString("targetQuestionStableId");
```
yufengwng: `getString()` requires that the key is present in the config object. Prefer to use `ConfigUtil.getStrIfPresent()` since these are optional.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventActionDao.java
```diff
@@ -67,9 +67,11 @@ private long insertNotificationAction(SendgridEmailEventActionDto eventAction, N
         return actionId;
     }
 
-    default long insertInstanceCreationAction(long targetActivityId) {
+    default long insertInstanceCreationAction(long targetActivityId,
+                      Boolean createFromAnswer, String sourceQuestionStableId, String targetQuestionStableId) {
```
yufengwng: If `createFromAnswer` is true, then let's check that `sourceQuestionStableId` is not blank.

pepper-apis/src/main/resources/db-changes/schema/DDP-6306-add-activity-instance-creation-properties.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="evgeniipr" id="20210629-add-activity-instance-creation-properties">
+        <addColumn tableName="activity_instance_creation_action">
+            <column name="create_from_answer" type="boolean">
+                <constraints nullable="true"/>
```
yufengwng: One thing we can do is mark this as non-null, set the default as `false`, and backfill existing rows with `false`. This way, we have one less null to worry about.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
```diff
@@ -0,0 +1,165 @@
...
+            ActivityInstanceCreationService creationService) {
+        super(studyActivityId, creationService);
+        this.sourceQuestionStableId = sourceQuestionStableId;
+        this.targetQuestionStableId = targetQuestionStableId;
+    }
+
+    /**
+     * Create N activity instances.
+     * where N = number of answers in composite question `sourceQuestionStableId`.
+     * IF N > not_used_activities_count THEN N = not_used_activities_count,
```
yufengwng: I think for now, it might be better if we throw an exception when it's `N > not_used` ? If we're using this feature, then I'm pretty sure we're expecting to create all N, so if we can't then let's give an error so we can further investigate.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
```diff
@@ -0,0 +1,165 @@
...
+        JdbiActivityInstanceStatus jdbiActivityInstanceStatus = handle.attach(JdbiActivityInstanceStatus.class);
+        JdbiActivity jdbiActivity = handle.attach(JdbiActivity.class);
+
+        ActivityDto activityDto = jdbiActivity.queryActivityById(studyActivityId);
+        creationService.checkSignalIfNestedTargetActivity(activityDto);
+
+        // detect parent instance id
+        Long parentInstanceId = creationService.detectParentInstanceId(activityDto);
+        if (parentInstanceId == null) {
+            throw new DDPException("Failed to detect parent instance ID");
```
yufengwng: Wait, why are we requiring that there is a parent instance? We're not creating child activities of a parent activity. They're totally separate activities. We should get the "source" activity instance from the `signal` without requiring that there is a parent activity.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        // check if number of non-used activities is enough (it can be limited if a special limit parameter specified)
+        Integer numberOfActivitiesLeft = creationService.detectNumberOfActivitiesLeft(studyActivityId, activityDto, jdbiActivityInstance);
+        if (numberOfActivitiesLeft == null || numberOfActivitiesLeft > 0) {
+
+            List<Answer> sourceAnswers = detectSelectedAnswers(handle, parentInstanceId);
+
+            // detect number of new activities to be created
+            int newInstancesCount = sourceAnswers.size();
```
yufengwng: Probably safer to turn it into a `List<SelectedPicklistOption>` so the count is more accurate?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
```diff
@@ -0,0 +1,165 @@
...
+     * We expect than answer - PickListAnswer (i.e. currently we deal only with picklist-inside-composite).
+     *
+     * @param targetInstanceId ID of target activity instance
+     * @param sourceAnswer     source answer (in question `sourceQuestionStableId`) to be copied to question
+     *                         `targetQuestionStableId`
+     */
+    private void copyAnswerToNewActivityInstance(Handle handle, EventSignal signal, long targetInstanceId, Answer sourceAnswer) {
+        var targetAnswer = cloneAnswer(sourceAnswer, sourceQuestionStableId, targetQuestionStableId);
+        var questionDto = getQuestionDto(handle, signal.getStudyId(), targetQuestionStableId);
+        handle.attach(AnswerDao.class).createAnswer(signal.getOperatorId(), targetInstanceId, questionDto.getId(), targetAnswer);
```
yufengwng: I think you can call `createAnswer()` without getting `questionDto`, in fact it's probably the way to go here.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
```diff
@@ -0,0 +1,165 @@
...
+        Long parentInstanceId = creationService.detectParentInstanceId(activityDto);
+        if (parentInstanceId == null) {
+            throw new DDPException("Failed to detect parent instance ID");
+        }
+
+        // check if number of non-used activities is enough (it can be limited if a special limit parameter specified)
+        Integer numberOfActivitiesLeft = creationService.detectNumberOfActivitiesLeft(studyActivityId, activityDto, jdbiActivityInstance);
+        if (numberOfActivitiesLeft == null || numberOfActivitiesLeft > 0) {
+
+            List<Answer> sourceAnswers = detectSelectedAnswers(handle, parentInstanceId);
```
yufengwng: I recommend grabbing the answer here, then match on the answer type and call the appropriate implementation. We might want to support other types in the future, and the number of target instances to create will vary depending on type. For example, if we have a `numeric` answer then the number to create is the answer itself. Matching on the type will hopefully make our code be a bit more general and it'll give us room to grow.
```
var answer = ...
switch (answer.getQuestionType()) {
  case COMPOSITE:
    ...
  default:
    ... unsupported ...
}

**Review by yufengwng — COMMENTED** [review]
Overall looks reasonable to me. Main issue I saw was the assumption about needing a "parent" activity.

File: pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java (hunk 1 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java]: Fixed

File: study-builder/studies/study-example.conf (hunk 2 above)
Comment (evgeniipr) [reply study-builder/studies/study-example.conf]: Fixed

File: study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java (hunk 3 above)
Comment (evgeniipr) [reply study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java]: Fixed

File: pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventActionDao.java (hunk 4 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/EventActionDao.java]: Fixed

File: pepper-apis/src/main/resources/db-changes/schema/DDP-6306-add-activity-instance-creation-properties.xml (hunk 5 above)
Comment (evgeniipr) [reply pepper-apis/src/main/resources/db-changes/schema/DDP-6306-add-activity-instance-creation-properties.xml]: Fixed

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 6 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: Fixed

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 8 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: I looked through Pepper code and didn't found any samples of processing CompositeAnswers and producing SelectedPicklistOptions from it..
I found some examples of collecting composite child answers. Seems here the purpose is the same - to get all answers selected or entered by a user: 
```
return Optional.of(answer).stream()
                .flatMap(parent -> parent.getValue().stream())
                .flatMap(row -> row.getValues().stream())
                .filter(Objects::nonNull)
```
May be more proper code would be like this (but seems it's the same):
```
List<Answer> answersToCheck = new ArrayList<>();
        answersToCheck.add(answer);
        if (answer.getQuestionType() == QuestionType.COMPOSITE) {
            ((CompositeAnswer) answer).getValue().stream()
                    .map(answerRow -> answerRow.getValues())
                    .flatMap(Collection::stream)
                    .filter(each -> each != null)
                    .forEach(answersToCheck::add);
        }
```

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 7 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: Initially I thought that it detects a current activity. Now I see that I was wrong.. Fixed

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 9 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: But how to detect question Id which is required?

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 10 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: Done. Refactored the code.

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 8 above)
Comment (yufengwng) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: What I'm saying is that, right now our only use-case is having a "single-select picklist inside a composite". So it's safe to assume that "one child answer == one picklist option". But imagine if we have a "multi-select picklist inside a composite". Then, one child answer can have multiple selected options. I'm just recommending that we drill down a little more into the list of selected options for a more accurate count.

File: pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java (hunk 9 above)
Comment (yufengwng) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: Each `Answer` object has the `questionStableId`, so `AnswerDao` will figure out the right question id for that activity instance.

Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java]: You mean that it's ok to use this `AnswerDao` method?
` default Answer createAnswer(long operatorId, long instanceId, Answer answer);`

File: study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java [hunk 11]
```diff
@@ -258,7 +258,10 @@ private long insertEventAction(Handle handle, Config actionCfg, Config triggerCf
                     .orElseThrow(() -> new DDPException("Could not find pdf configuration with name " + pdfName));
             return actionDao.insertPdfGenerationAction(pdfId);
         } else if (EventActionType.ACTIVITY_INSTANCE_CREATION.name().equals(type)) {
-            String activityCode = actionCfg.getString(ACTIVITY_CODE_FIELD);
+            String activityCode = actionCfg.getString("ACTIVITY_CODE_FIELD");
```
yufengwng: Typo
```suggestion
            String activityCode = actionCfg.getString(ACTIVITY_CODE_FIELD);
```

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
```diff
@@ -258,7 +258,10 @@ private long insertEventAction(Handle handle, Config actionCfg, Config triggerCf
...
+            Boolean createFromAnswer = ConfigUtil.getBoolIfPresent(actionCfg, "createFromAnswer");
```
yufengwng, line 263: Just to be extra careful, can we add a check that, if `createFromAnswer` is true then `triggerCfg` type needs to be `ACTIVITY_STATUS`?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
yufengwng: Yes.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
evgeniipr: But to copy an answer to target it is copied like this:
`new PicklistAnswer(null, targetQuestionStableId, null, picklistAnswer.getValue());`
And I believe this new answer (to be copied) contains all set of SelectedPicklistOption's (in getValue()).
So in case of multiple selected options all should be copied ok.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
evgeniipr: Thanks! Fixed.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/EventBuilder.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
yufengwng: Remember that the idea is, _for each_ selected option create a new instance. So if an child answer has multiple answers, I think the right thing to do is create multiples instances from that child answer. Also, you can easily create a new picklist answer, like `new PicklistAnswer(..., selectedOption)`.

Anyways, we're not going to see that use-case come up for PanCan so I guess it's fine.

yufengwng (approved):
I think this does what it needs to do, so should be ready-to-go.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/event/activityinstancecreation/ActivityInstanceCreatorFromAnswers.java
evgeniipr: Ah, yes! I totally forgot that each of selection in multiple selection should create a separate instance...
Ok, may be to create a separate ticket for such case? Which I could implement after PanCan

PR #868, opened 2021-07-09
DDP-6546 fix validating child activities during parent activity submission

yufengwng:
## Context

See DDP-6546.

## Checklist

## FUD Score

Try out RGP study.

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
@@ -284,18 +284,15 @@ public class PutFormAnswersRoute implements Route {
                 }
 
                 for (var childInstanceDto : childInstanceDtos) {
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
```diff
@@ -284,18 +284,15 @@ private FormInstance loadFormInstance(Response response,
                 }
 
                 for (var childInstanceDto : childInstanceDtos) {
-                    if (childInstanceDto.getStatusType() != InstanceStatusType.COMPLETE) {
-                        // Child instance is not finished but it might not have required questions, so check it.
```
yufengwng, line 288: Before, we only checked the child activity instance if it's not `COMPLETE`. This is based on assumption that child instance won't be changed after submission, and if it was, then frontend should submit the child before the parent and so we'll be able to catch any missing required questions. However, those assumptions turn out to be wrong. So now we're always checking the child instances to ensure they're completed.

ssettipalli (approved):

PR #870, opened 2021-07-09
v1.16.1 hotfix 1

yufengwng:
Includes two things:

* DDP-6523: use sendgrid proxy for RGP seqr export so we can send emails properly
* DDP-6546: fix validation of child activities when parent is submitted


charlottedye:
Yep, those are the right changes

PR #875, opened 2021-07-13
Ddp 6306 extend activity instance creation fix

evgeniipr:
## Context

NPE fix: provide default value 'false' for `createFromAnswer`


YuriyPavlovBI (approved):

evgeniipr:
DDP-6306 - NPE fix: provide default value 'false' for `createFromAnswer`

PR #883, opened 2021-07-15
DDP-6321 . Add user profile during temp user creation.

ssettipalli:
Add user profile during temp user creation.


pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
```diff
@@ -33,6 +38,19 @@ protected int getValidationErrorStatus() {
     public Object handle(Request request, Response response, CreateTemporaryUserPayload payload) {
         String auth0ClientId = payload.getAuth0ClientId();
         String auth0Domain = payload.getAuth0Domain();
+        String languageCode = payload.getLanguageCode();
+        if (languageCode == null) {
+            //try from request header
+            LanguageDto preferredLanguage = RouteUtil.getUserLanguage(request);
```
ssettipalli: @yufengwng :
Is this good enough to get languageCode from request header (with client changes) OR do I need to add filter to temp-user route in DataDonationPlatform like:
before(API.BASE + "/temporary-users/*", new StudyLanguageResolutionFilter());

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
```diff
@@ -13,16 +13,29 @@
...
 
+    @SerializedName("languageCode")
+    private String languageCode;
+
     public CreateTemporaryUserPayload(String auth0ClientId, String auth0Domain) {
         this.auth0ClientId = auth0ClientId;
         this.auth0Domain = auth0Domain;
     }
 
+    public CreateTemporaryUserPayload(@NotBlank String auth0ClientId, String auth0Domain, String languageCode) {
```
yufengwng: I think you can remove the parameter annotation, don't think it does much for us.
```suggestion
    public CreateTemporaryUserPayload(String auth0ClientId, String auth0Domain, String languageCode) {
```

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
yufengwng: If you take a look at implementation of `RouteUtil. getUserLanguage()` you see that it's grabbing something set by `StudyLanguageResolutionFilter`, so if you want to use this, then you should add the `before()` filter.

However, I don't think it is ideal here because `StudyLanguageResolutionFilter` is meant for "study-related" URL paths, and create-temp-user is not one of them. You might be better off just parsing out the language header (or use some utils for it).

pepper-apis/src/main/java/org/broadinstitute/ddp/json/CreateTemporaryUserPayload.java
yufengwng: Sorry, I meant you can remove annotation from constructor. Let's keep the one on the member field on L8...

yufengwng (dismissed):

yufengwng (approved):

PR #886, opened 2021-07-16
testboston: backfill for completed status

yufengwng:
## Context

There is a prod issue for TestBoston where a large number of participants are not transitioning to `completed` status. After investigating, it turns out that we did not do backfill properly when we rolled out feature for TB. Adding properly backfill now.

## Testing

Will run this in non-prod envs to make sure it work.s

## Release

1. Create a hotfix branch off of `production`.
2. Cherry-pick this commit over.
3. Compile study-builder
4. Run:

```
$ java -Dconfig.file=... -jar target/StudyBuilder.jar --vars ... --substitutions ... testboston --run-tasks TestBostonBackfillCompleted
```

### DIFF
--- a/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/TestBostonBackfillCompleted.java
+++ b/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/TestBostonBackfillCompleted.java
@@ -1,0 +1,180 @@
+package org.broadinstitute.ddp.studybuilder.task;
+
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+import com.typesafe.config.Config;
+import org.apache.commons.collections4.ListUtils;
+import org.broadinstitute.ddp.db.dao.EventDao;
+import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
+import org.broadinstitute.ddp.db.dao.QueuedEventDao;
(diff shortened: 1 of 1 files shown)


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/TestBostonBackfillCompleted.java
```diff
@@ -0,0 +1,180 @@
...
+                    candidate.userId, candidate.userId, delaySecondsLeft);
+            LOG.info("Queued status change event for participant {} with queued_event_id={} after delay_seconds={}",
+                    candidate.userId, queuedId, delaySecondsLeft);
+        }
+
+        LOG.info("Handling eligible participants now...");
+        int batchSize = 100;
+        int bufferSecs = 10 * 60; // 10 mins
+        int delaySecsToUse = 0;
+        // Since there's likely a large number of these, we partition into batches and make each batch 10 mins apart.
```
yufengwng: There's roughly 4500+ that need this, so roughly 45 batches. So it'll take 450 mins to run through, roughly 7 hrs. Is this okay, or should we beef it up?

MocanaAtBroad (dismissed):

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/TestBostonBackfillCompleted.java
yufengwng: Bumping this up to 5 mins apart.

PR #887, opened 2021-07-16
Pancan redirect pex updates

ssettipalli:
Pex updates to handle redirects correctly depending on self/child option selection.


study-builder/studies/pancan/prequal.conf
```diff
@@ -143,83 +143,15 @@
           "blockType": "QUESTION",
           "shownExpr": null
         },
-        {
```
ssettipalli, line 146: apparently advanced-breast/prostate question is needed  only if self_only option is selected.
hence removed body template prompts for self , child and added self prompt to question.

yufengwng: Ah nice, thanks for catching that! No wonder I didn't see mention of a "child" variant of the text in the spreadsheet...

study-builder/studies/pancan/prequal.conf
```diff
@@ -143,83 +143,15 @@
...
-          },
-          "blockType": "CONTENT",
-          "shownExpr": ${_pex.is_child_only}"&&"${_pex.has_prostate}
+          "shownExpr": ${_pex.is_self_only}"&&"${_pex.has_breast}
         },
         {
           "question": ${_includes.question_advanced_prostate},
           "blockType": "QUESTION",
-          "shownExpr": ${_pex.has_prostate}
+          "shownExpr": ${_pex.is_self_only}"&&"${_pex.has_prostate}
```
yufengwng, line 154: Wait... when they select both "self" and "child" diagnosed options, the behavior is that we treat it as if it's just "self". That's why I had put in checks like `is_self` and `is_child_only`. If they selected both "self" and "child", I think we should still show advanced breast/prostate questions, since it's technically a "self" case?

ssettipalli: will update to is_self

study-builder/studies/pancan/prequal.conf
yufengwng: Wait... when they select both "self" and "child" diagnosed options, the behavior is that we treat it as if it's just "self". That's why I had put in checks like `is_self` and `is_child_only`. If they selected both "self" and "child", I think we should still show advanced breast/prostate questions, since it's technically a "self" case?

ssettipalli: If so: They can never enter child stuff right, because its redirected to existing study

yufengwng: I think so? In the "self only" and "self+child" cases, we're only displaying self questions, so don't think they'll answer any child questions.

study-builder/studies/pancan/study-workflows.conf
```diff
@@ -28,25 +28,25 @@
           "type": "STUDY_REDIRECT",
           "studyGuid": "cmi-mbc",
           "redirectUrl": ${mbcRedirectUrl},
-          "expression": ${_pex.has_breast}"&&"${_pex.is_advanced_breast}"&&"${_pex.is_lang_en},
+          "expression": ${_pex.is_self_only}"&&"${_pex.has_breast}"&&"${_pex.is_advanced_breast}"&&"${_pex.is_lang_en},
```
yufengwng: Need to update these?

study-builder/studies/pancan/prequal.conf
ssettipalli: we can change it later if needed. flow chart has  a 3rd diff flow if self + child are chosen. that doesn't have any mention of study-redirects

study-builder/studies/pancan/study-workflows.conf
ssettipalli: flow chart has a 3rd diff flow if self + child are chosen and that doesn't have any mention of study-redirects.
we can update once we get clarified.

study-builder/studies/pancan/prequal.conf
ssettipalli: Here either way advanced question is only asked for "self" , we should be fine here.

study-builder/studies/pancan/study-workflows.conf
ssettipalli: will update here

yufengwng (approved):

PR #891, opened 2021-07-19
Pancan prequal updates 2

ssettipalli:
Updates to handle pancan external site study redirects.
Add ESC to study redirects.

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -2382,7 +2382,6 @@ WorkflowNextSuggestion.StudyRedirect:
     - type: object
       required:
         - studyName
-        - studyGuid
         - redirectUrl
       properties:
         studyName:
--- a/pepper-apis/src/main/resources/db-changes/schema/DDP-6313-add-study-redirect-workflow.xml
+++ b/pepper-apis/src/main/resources/db-changes/schema/DDP-6313-add-study-redirect-workflow.xml
@@ -12,7 +12,7 @@
             </column>
             <column name="study_guid" type="varchar(10)">
             </column>
-            <column name="study_name" type="varchar(100))">
+            <column name="study_name" type="varchar(100)">
             </column>
             <column name="redirect_url" type="varchar(200)">
                 <constraints nullable="false"/>
--- a/pepper-apis/src/main/resources/db-changes/schema/DDP-6313-add-study-redirect-workflow.xml
+++ b/pepper-apis/src/main/resources/db-changes/schema/DDP-6313-add-study-redirect-workflow.xml
@@ -12,7 +12,7 @@
             </column>
             <column name="study_guid" type="varchar(10)">
             </column>
-            <column name="study_name" type="varchar2(100))">
+            <column name="study_name" type="varchar(100))">
             </column>
             <column name="redirect_url" type="varchar(200)">
                 <constraints nullable="false"/>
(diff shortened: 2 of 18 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2382,7 +2382,6 @@ WorkflowNextSuggestion.StudyRedirect:
     - type: object
       required:
         - studyName
-        - studyGuid
```
yufengwng, line 2385: Another way to do this is keep this here in `required` but mark it as `nullable: true` in `properties`. And then you can update `description` to mention that it might not be available if it's not a pepper study.

pepper-apis/src/main/resources/db-changes/schema/DDP-6313-add-study-redirect-workflow.xml
```diff
@@ -11,7 +11,8 @@
                              references="workflow_state(workflow_state_id)" foreignKeyName="workflow_study_redirect_state_workflow_state_fk"/>
             </column>
             <column name="study_guid" type="varchar(10)">
-                <constraints nullable="false"/>
+            </column>
+            <column name="study_name" type="varchar(100))">
```
yufengwng: There's an extra parenthesis. Also, since we have ran this liquibase file in `dev` environment already, I think liquibase migration is going to fail. I think merging this is fine since it's not in the other environments yet, but we'll need to manually fix things up in `dev`.
```suggestion
            <column name="study_name" type="varchar(100)">
```

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/WorkflowBuilder.java
```diff
@@ -94,8 +99,9 @@ private String stateAsStr(Config stateCfg) {
             return String.format("%s/%s", type, activityCode);
         } else if (StateType.STUDY_REDIRECT.name().equals(type)) {
             String studyGuid = stateCfg.getString("studyGuid");
+            String studyName = stateCfg.getString("studyName");
```
yufengwng: Should these be `ConfigUtil.getStrIfPresent()`?

study-builder/studies/pancan/study-workflows.conf
```diff
@@ -27,39 +27,66 @@
         {
           "type": "STUDY_REDIRECT",
           "studyGuid": "cmi-mbc",
+          "studyName": "Metastatic Breast Cancer"
```
yufengwng: For these that are Pepper studies, I recommend not setting `studyName` so we use the other configurations (and also let us know the code for looking it up is working!). Also, `Metastatic Breast Cancer` is not exactly what we want. I think we should add `Project` at the end there. Again, I recommend not setting it and use our other configurations.

study-builder/studies/study-example.conf
```diff
@@ -484,6 +484,7 @@
         {
           "type": "STUDY_REDIRECT",
           "studyGuid": "string",
+          "studyName": "string",
           "redirectUrl": "string",
           "expression": "string",
         }
```
yufengwng: Can you add a quick explanation?

study-builder/studies/pancan/study-workflows.conf
ssettipalli: I thought now that we have more flexibility, we can pass exactly what we need for front-end to show. Avoid cases where like "ascproject" is passed. 
Any way will remove and we can revisit if needed.

yufengwng: We have ticket to clean up those names: https://broadinstitute.atlassian.net/browse/DDP-6329.

study-builder/studies/pancan/study-workflows.conf
```diff
@@ -48,6 +48,12 @@
           "redirectUrl": ${angioRedirectUrl},
           "expression": ${_pex.is_self}"&&"${_pex.has_angio},
         },
+        {
+          "type": "STUDY_REDIRECT",
+          "studyGuid": "CMI-ESC",
```
yufengwng: ESC study guid is lowercase `cmi-esc`.

yufengwng (dismissed):

yufengwng (approved):

PR #893, opened 2021-07-19
Adds iOS client to the pancan study

bskinner:
## Context

Adds two new opt-in keys for including the iOS client ID and client secret when deploying the Pancan study. This is necessary to prevent access to the study from being revoked after a deploy. If a study does not have the `iosClientId` key in their Vault config, the additional keys will not appear in the generated vars file.

If further changes are necessary, let me know and I'll incorporate them into this PR.

## FUD Score

On initial deployment or update, the `pancan` study should have a new client added, beginning with the string `1q8ZS...`.

## Testing

## Release

Currently deployed Pancan stud\y instances will not pick up the changes until they have been re-deployed.


yufengwng (approved):
Nice!

PR #897, opened 2021-07-20
Ddp 6490 freeze address in form

evgeniipr:
## Context

Implements DDP-6490 (Allow consent to become read-only and freeze address in the form).

**The problem is**: in PanCan the address entered in the consent activity and after submit all entered data should be _frozen_ (become read-only) including the address. But address associated with the user’s profile that is used to sent kits should  be editable.

**The solution is**: it is provided a possibility (can be enabled/disabled) to create a default address snapshot (a copy of a created default address) and `addressGuid` (new guid of address copy) to save to activity instance substitutions (to table `activity_instance_substitutions`) with key `ADDRESS_GUID`. 
And in case if an address snapshot exist (in activity instance substitutions) then it is used instead of a default address.

**In more details**:
- In `FormActivity` added a flag `snapshotAddressOnSubmit` (true/false). By default it is `false`. If it is `true` then during form submit a default email address is cloned (with new `id` and `guid`, of course) and inserted to DB (with default flag = false, so it is a `non-default address`). The snapshotted address creation is done in service PUT /answers (`PutFormAnswersRoute`) which called during submit.
- It needs to use a `snapshotted address` (if exists) during data export (to ElasticSearch and CSV): when Participants data is fetched (before starting export) it is detected if `snapshotted address` exist and if it is then it is used, otherwise a default address will be used.
(Important: the address object that we export to Elasticsearch should always be the default address because kit shipment logic depends on this being the latest address!)

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
```diff
@@ -168,6 +169,13 @@ public String isGovernedParticipant(String ifTrueString, String ifFalseString) {
         return governedParticipant != null && governedParticipant ? ifTrueString : ifFalseString;
     }
 
+    /**
+     * Get GUID of a snapshotted address
+     */
+    public String addressGuid() {
+        return addressGuid;
+    }
```
yufengwng: Public methods in `RenderValueProvider` are exposed as `$ddp.<method_name>()` method calls in templates. We don't need to expose this in templates right now. Let's remove this method.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
```diff
@@ -557,10 +566,14 @@ private void collectQuestionIntoRecord(Map<String, String> record, QuestionDef q
         }
     }
 
-    private void collectComponentIntoRecord(Map<String, String> record, ComponentBlockDef component, ComponentDataSupplier supplier) {
+    private void collectComponentIntoRecord(FormResponse formResponse, Map<String, String> record,
+                                            ComponentBlockDef component, ComponentDataSupplier supplier) {
         switch (component.getComponentType()) {
             case MAILING_ADDRESS:
-                record.putAll(addressFmt.collect(supplier.getAddress()));
+                MailAddress address = withAPIsTxn(handle -> {
```
yufengwng: At this point in the code, we're in the middle of transforming the data to be exported, and we're very likely inside a transaction already, so opening another transaction is not good. We should query the data beforehand and pass it in. I think `ComponentDataSupplier` is the ideal place to store and pass along the activity snapshot data and the snapshotted address.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -969,12 +973,13 @@ private String formatParticipantToStructuredJSON(
...
 
-    private QuestionRecord createRecordForComponent(String component, Participant participant) {
+    private QuestionRecord createRecordForComponent(Handle handle, FormResponse formResponse, String component, Participant participant) {
         ComponentQuestionRecord record = null;
         List<List<String>> values;
         switch (component) {
             case "MAILING_ADDRESS":
-                MailAddress address = participant.getUser().getAddress();
+                MailAddress address =
+                        getSnapshottedAddress(handle, formResponse.getId(), addressService, participant.getUser().getAddress());
```
yufengwng: I think we have already queried for the activity substitutions data so doing another query (for each user and activity) is not as efficient. If you look at the `Participant` class, you'll see this field:
```
private Map<Long, Map<String, String>> activityInstanceSubstitutions;
```
And this is populated in `ExportUtil.extractParticipantsFromResultSet()`. So we have already bulk-queried for that data. Now, if only we can figure out a way to bulk query for the snapshotted addresses beforehand and pass it in, that'll be nice.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
```diff
@@ -226,6 +229,48 @@ public Object handle(Request request, Response response) {
...
+    private void snapshotAddress(Handle handle, String participantGuid, String operatorGuid, FormInstance form,
+                                 boolean shouldSnapshotAddressOnSubmit) {
+        if (shouldSnapshotAddressOnSubmit) {
+            // find default address
+            Optional<MailAddress> defaultAddress = addressService.findDefaultAddressForParticipant(handle, participantGuid);
+            if (defaultAddress.isPresent()) {
+                MailAddress mailAddress = defaultAddress.get();
+                mailAddress.setDefault(false);
+                // create a copy of the default address (but setting it as non-default)
+                mailAddress = addressService.addAddress(handle, mailAddress, participantGuid, operatorGuid);
```
yufengwng: If you look at the implementation of `AddressService.addAddress()`, under the hood it's doing a lot of things: calculating plus codes, calling EasyPost API to verify address, etc. Those are nice things, but we don't need them in this case. The default address should already be in a valid state, so we should just "copy" it instead of doing all this extra work. Is there something else we can use here?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreatorHelper.java
```diff
@@ -86,6 +89,18 @@ public int setDisplayNumbers(FormInstance formInstance) {
...
     }
 
+    /**
+     * If snapshotted address GUID is preserved in activity instance substitutions (with key ADDRESS_GUID)
+     * then find MAILING_ADDRESS component in the created formInstance and set addressGuid with this found
+     * snapshotted address GUID
+     */
+    public void populateSnapshottedAddress(MailingAddressComponent mailingAddressComponent, Map<String, String> activitySnapshots) {
+        String addressGuid = activitySnapshots.get(I18nTemplateConstants.Snapshot.ADDRESS_GUID);
+        if (addressGuid != null) {
```
yufengwng: Should we check mailing-address component is not null?

pepper-apis/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
evgeniipr: Done

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
evgeniipr: I tried to check where this code called from. There are 3 methods (actually 4):
```
DataExporter.exportToElasticsearch()
   formatParticipantToStructuredJSON()
   formatParticipantToFlatJSON
DataExporter.exportDataSetAsCsv()
CustomExporter.exportDataSetAsCsv()
```
`ComponentDataSupplier` used in 3 of them except of `formatParticipantToStructuredJSON()`. Therefore I can not rely on this class 100%.
Another problem of calculating snapshot address before creating `ComponentDataSupplier` is that it needs to know `addressGuid` which corresponds to a certain `instanceId`. Therefore it needs to know activity instance data.
You advised to get from `Participant#activityInstanceSubstitutions` but we can use this substitution data only after it set to Participants. Possible way to do it: pass a callback to `Participant#putActivityInstanceSubstitutions` and in this callback to detect snapshotted address and save it to `ComponentDataSupplier` (but unfortunately not for a case of `formatParticipantToStructuredJSON()`)...

pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
evgeniipr: > And this is populated in ExportUtil.extractParticipantsFromResultSet()

Is this method called for all 4 cases?
 ```
 DataExporter.exportToElasticsearch()
   formatParticipantToStructuredJSON()
   formatParticipantToFlatJSON
DataExporter.exportDataSetAsCsv()
CustomExporter.exportDataSetAsCsv()
```

File: pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java (hunk 4 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java]: Done

File: pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreatorHelper.java (hunk 5 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/FormInstanceCreatorHelper.java]: Sure, thanks!
Fixed.

File: pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java (hunk 2 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java]: Actually I did some refactoring: trying to get Participant#activityInstanceSubstitutions and detecting snapshotted address during creation of ComponentDataSupplier.

File: pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java (hunk 3 above)
Comment (evgeniipr) [reply pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java]: Reused Participant#activityInstanceSubstitutions when it is possible

File: pepper-apis/src/main/java/org/broadinstitute/ddp/export/ExportUtil.java [hunk 6]
```diff
@@ -115,4 +119,66 @@ public static void computeActivityAttributesSeen(ActivityInstanceDao instanceDao
...
+            MailAddress defaultAddress) {
+        String addressGuid = null;
+        if (activityInstanceSubstitutions != null) {
+            addressGuid =
+                    activityInstanceSubstitutions.values().stream()
+                            .flatMap(subs -> subs.entrySet().stream())
+                            .filter(subs -> subs.getKey().equals(I18nTemplateConstants.Snapshot.ADDRESS_GUID))
+                            .map(subs -> subs.getValue())
+                            .findFirst()
+                            .orElse(null);
```
yufengwng: So we're looking through all the substitutions for all activity instances to grab the snapshotted address? This doesn't look right... What happens if we have mailing address component in two different activities? What happens if we have multiple activity instances of the same activity? Each individual activity instance should have its own snapshotted address, and we shouldn't use another instance's snapshotted address for this instance!

I had an idea while reading through your PR. How does this sound?

(1) We bulk query for "snapshotted, non-default addresses" upfront. We can do this by adding bulk query to our "extract participant dataset" pipeline. It looks like all the various exports should go through `ExportUtil.extractParticipantsFromResultSet()` so that's probably a good place to put it. Try to come up with a bulk query that looks for non-default addresses for a set of participant ids.

* The default address should have already been queried as part of building the `Stream<Participant> resultset` that is passed in by caller. The default address should have been stored in the `Participant` object.
* There should only be one default address, so the rest are non-default addresses. And we know that non-default addresses are what we use for "snapshotting".
* We can filter out the default address in our bulk query by looking in the `default_mailing_address` and excluding that `address_id` from our result set.
* You have the `Stream<Participant> resultset` and we're already iterating through it, so it is ideal to collect all the participant ids from the `resultset` into a `Set`, so we can do a bulk query for all these participants in one go.
* After we done the bulk query, we can iterate through the queried rows, group them by participant, and assign them to the appropriate participant (notice that we have `Map<String, Participant> participants` which should help with this! But you might need to change how we map things from participant guid `String` to participant id `Long` for an easier implementation).
* There's probably more efficient ways to do this, but the end result is that we have bulk  queried all non-default addresses for all participants in our dataset.

(2) In `Participant` class, we have the activity instance substitutions. We are getting passed the list of snapshotted addresses from (1). So at this point, we can do a bit of preprocessing to figure out which activity instance has which snapshotted address.
* Perhaps we add a method called `associateInstanceWithSnapshottedAddresses(Map<String, MailAddress> nonDefaultAddresses)` (need a better name here...) which step (1) will call.
* The end result is that in `Participant` we store a mapping of `Map<Long, MailAddress> instanceIdToNonDefaultAddress` (any better name?)
* Now, the rest of the export code don't need to grab a `Handle` object and don't need to run any more queries! It's all there in the extracted participant dataset!

(3) For exports that use `ComponentDataSupplier`, we can use the supplier as an "interface" to get the snapshotted address.
* Pass in the `instanceIdToAddress` mapping  to `ComponentDataSupplier` constructor.
* Either add a new method or update existing `getAddress()` method to take a `long activityInstanceId` argument.
* Now, getting an address means looking in the map. If there is an entry, use that, otherwise fallback to the default address.
* Tip: in `ActivityResponseCollector` you can pass the `instance` to `collectComponentIntoRecord()`.

(4) For exports that don't use `ComponentDataSupplier` but has access to the `Participant` object, we can do similar thing and add method to `Participant` to fetch an address for an activity instance (which will look it up in the map or use default).

By doing a bulk query, things are bit more efficient. And by collecting all the data upfront, now we don't need to open up a transaction while exporting data and holding the transaction open for so long.

evgeniipr: Implemented this logic

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
```diff
@@ -63,11 +65,14 @@
...
     private final String customActivity;
     private final RestHighLevelClient elasticsearchClient;
+    private final AddressService addressService;
+
 
     CustomExporter(Config mainConfig, Config exportConfig) {
         this.exportConfig = exportConfig;
         this.customActivity = exportConfig.getString(CustomExportConfigFile.ACTIVITY);
         this.customGuid = exportConfig.getString(CustomExportConfigFile.STUDY_GUID);
+        addressService = new AddressService();
```
yufengwng: Doesn't look like we use `AddressService` here?

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
```diff
@@ -63,14 +66,18 @@
     private MailingAddressFormatter addressFmt = new MailingAddressFormatter();
     private MedicalProviderFormatter providerFmt = new MedicalProviderFormatter();
 
+    private final AddressService addressService;
+
     public ActivityResponseCollector(ActivityDef definition) {
         this.definition = definition;
+        this.addressService = new AddressService();
```
yufengwng: `AddressService` is not used in this collector.

study-builder/studies/snippets/activity-general-form.conf
```diff
@@ -21,6 +21,11 @@
   # Optional, defaults to false. This is typically used with write-once activities.
   "snapshotSubstitutionsOnSubmit": false,
 
+  # Should a snapshot of addressGuid be saved to activity substitutions on the first submit?
```
yufengwng:
```suggestion
  # Should a snapshot of the default address be saved for the activity instance on the first submit?
```

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiMailAddress.sql.stg
```diff
@@ -95,6 +95,47 @@ findAllAddressesByParticipantGuid() ::= <<
...
+               LEFT JOIN
+            country_subnational_division s ON s.country_subnational_division_id = m.subnational_division_id
+               JOIN
+            user u on u.user_id=m.participant_user_id
+               LEFT JOIN
+            default_mailing_address da ON m.address_id=da.address_id
+               LEFT JOIN
+            mailing_address_validation_status vs on m.validation_status_id=vs.mailing_address_validation_status_id
+			   JOIN
+            activity_instance_substitution ais on ais.variable_name = :addressGuidSubstitutionName and ais.value = m.address_guid
```
yufengwng: I'm curious about why this query need to look at activity substitutions? I'm sure we can write a more general query here to look for all addresses that are not marked as default?

pepper-apis/src/main/java/org/broadinstitute/ddp/export/ParticipantsResultSetUtil.java
```diff
@@ -0,0 +1,128 @@
...
+import org.broadinstitute.ddp.model.address.MailAddress;
+import org.broadinstitute.ddp.model.study.Participant;
+import org.jdbi.v3.core.Handle;
+
+/**
+ * Methods used for extracting {@link Participant}-s data from DB and setting it to List of
+ * {@link Participant}-s.
+ * This data used for exporting (in classes {@link DataExporter}, {@link CustomExporter}.
+ */
+public class ParticipantsResultSetUtil {
```
yufengwng: Does anyone else feel like we have way too many "util" classes laying around? I'm curious why we decided to move these out of `ExportUtil`?

pepper-apis/src/main/java/org/broadinstitute/ddp/export/ParticipantsResultSetUtil.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    /**
+     * Fetch non-default (snapshotted) {@link MailAddress}'es for all given {@link Participant}'s
+     * and then reduce the fetched data to stream of participants with Map(Long,MailAddress) where key = instanceId.
+     * Save maps with MailAddresses to a corresponding Participants.
+     */
+    private static void extractParticipantsNonDefaultAddresses(Handle handle, Map<String, Participant> participants) {
+        Set<Long> participantIds = participants.values().stream().map(p -> p.getUser().getId()).collect(Collectors.toSet());
```
yufengwng: For more efficiency, I recommend building this `Set` while iterating through the participants in `extractParticipantsFromResultSet()`. You can probably just add another `.peek()` to the stream pipeline. Even better, you can change `participants` to `Map<Long, Participant>` (id -> participant), so we can just do `participants.keySet()` and get all the participant ids for free.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMailAddress.java
```diff
@@ -94,4 +111,60 @@ default int updateAddress(String guid, MailAddress addressWithUpdatedFields,
...
+                        view.getColumn("zip", String.class),
+                        view.getColumn("phone", String.class),
+                        view.getColumn("pluscode", String.class),
+                        view.getColumn("description", String.class),
+                        DsmAddressValidationStatus.getByCode(view.getColumn("validationStatus", Integer.class)),
+                        false
+                );
+                container.computeIfAbsent(participantGuid, NonDefaultMailAddressesWrapper::new).unwrap().put(instanceId, mailAddress);
+            } catch (Exception e) {
+                throw new DDPException("Error during parsing a DB result with non-default MailAddresses ", e);
```
yufengwng: I recommend throwing `DaoException` when we're in DAO code.

pepper-apis/src/main/java/org/broadinstitute/ddp/customexport/export/CustomExporter.java
evgeniipr: Removed

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
evgeniipr: Removed

study-builder/studies/snippets/activity-general-form.conf
evgeniipr: Fixed

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiMailAddress.sql.stg
evgeniipr: But how to get instanceIds? We need to group it by instances and get proper substitution for the currently exported instance.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/ParticipantsResultSetUtil.java
evgeniipr: Moved to ExportUtil. 
My idea was that it is more convenient to have logically connected group of methods in a separate class.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMailAddress.java
evgeniipr: Fixed

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiMailAddress.sql.stg
yufengwng: We have `Map<Long, Map<String, String>> activityInstanceSubstitutions` in `Participant`.

pepper-apis/src/main/java/org/broadinstitute/ddp/export/ParticipantsResultSetUtil.java
evgeniipr: Fixed

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiMailAddress.sql.stg
evgeniipr: Reimplemented according to your suggestion

yufengwng (dismissed):
I think this looks about right!

yufengwng (dismissed):

yufengwng (approved):

yufengwng:
Merging this now so we can try it out on `dev`.

PR #901, opened 2021-07-22
pancan prequal refactoring to handle self/child prompts

ssettipalli:
refactored prequal cancer list collection question as two separate questions:
Some background:
Initially question_primary_cancer_list collects all diagnosed cancer list of self and child.
Question prompt was designed as content block so that diff text/prompt was displayed for self and child.
This caused issues for mobile while it looked fine in angular.

Refactor/Redesign
Separated  question_primary_cancer_list as question_primary_cancer_list_self and question_primary_cancer_list_child
Also had to separate question_primary_cancer which has ALL cancer list/options) into two questions: question_primary_cancer_self and question_primary_cancer_child
What this means:
Chosen cancer list options / answers are in two different questions/answers.
Other conf entries like study-events, pexx, workflow or anything else which refer to cancer list need to look at two question/answers.
DataExport (elastic and CSV) will contain two questions and answers with cancer list.
DSM will show as two questions/answers. Study staff user experience has impact with this design.

Some alternate thoughts:
Can we just show generic message which applies to both self/child/both and avoid designing as two questions ?

PR #905, opened 2021-07-23
Pancan Prequal cancer groups

ssettipalli:
Refactor pancan prequal cancer list to be picklist groups
This PR also includes changes from PR: https://github.com/broadinstitute/ddp-study-server/pull/901
which takes care of prequal refactoring cancer list to two separate questions for self and child. 
@evgeniipr : Grouped list is not entirely complete though.

**What's pending:**
Need to revisit and update PEX once we have full list of  cancer list/options (Once https://broadinstitute.atlassian.net/browse/DDP-6634 is completed)
Need to test once front-endPR to handle grouped PL options with render mode: AUTOCOMPLETE is ready.
Will do another PR once full cancer list groups are ready 

**Notes from other PR for reference**
refactored prequal cancer list collection question as two separate questions:
Some background:
Initially question_primary_cancer_list collects all diagnosed cancer list of self and child.
Question prompt was designed as content block so that diff text/prompt was displayed for self and child.
This caused issues for mobile while it looked fine in angular.

Refactor/Redesign
Separated question_primary_cancer_list as question_primary_cancer_list_self and question_primary_cancer_list_child
Also had to separate question_primary_cancer which has ALL cancer list/options) into two questions: question_primary_cancer_self and question_primary_cancer_child
What this means:
Chosen cancer list options / answers are in two different questions/answers.
Other conf entries like study-events, pexx, workflow or anything else which refer to cancer list need to look at two question/answers.
DataExport (elastic and CSV) will contain two questions and answers with cancer list.
DSM will show as two questions/answers. Study staff user experience has impact with this design.

Some alternate thoughts:
Can we just show generic message which applies to both self/child/both and avoid designing as two questions ?


study-builder/studies/pancan/study-events.conf
```diff
@@ -669,15 +669,37 @@
...
+        "type": "ACTIVITY_INSTANCE_CREATION",
+        "activityCode": ${id.act.about_cancer},
+        "createFromAnswer": true,
+        "sourceQuestionStableId": ${id.q.primary_cancer_list_child},
+        "targetQuestionStableId": ${id.q.diagnosis_type}
+      },
+      # If there's already about-cancer instances, then don't create any more.
+      "preconditionExpr": """!user.studies["cmi-pancan"].forms["ABOUT_CANCER"].hasInstance()
+      && (!user.studies["cmi-pancan"].forms["PREQUAL"].questions["DESCRIBE"].answers.hasOption("DIAGNOSED")
+            && user.studies["cmi-pancan"].forms["PREQUAL"].questions["DESCRIBE"].answers.hasOption("CHILD_DIAGNOSED"))""",
```
yufengwng: I think we can reuse our `${_pex...}` strings here.

yufengwng (dismissed):
Looks about right. I am a bit worried about how some of the cancer option labels/names are super long... these names/labels will be substituted into the About Cancer survey.

study-builder/studies/pancan/study-events.conf
evgeniipr: > I am a bit worried about how some of the cancer option labels/names are super long..

@yufengwng it is generated automatically from the cancer titles. I can suggest that it can be shorten somehow. 
I can suggest 2 options:
- manually assign names for each cancer (it can take some time);
- during generation of the names limit with some max length (for example 25 or 30 chars). But here needs to think about the logic a bit: to avoid duplicates etc.

yufengwng: Limiting by max length might not give the correct result. What if it cut off some of the abbreviations / aliases? Or if a single cancer name is longer than length and then it's split in half?

study-builder/studies/pancan/snippets/question-hidden-diagnosis-type.conf
```diff
@@ -1,6 +1,6 @@
 {
   include required("../../snippets/picklist-question-single-list.conf"),
-  include required("common/cancer-picklist-options.conf"),
+  include required("common/cancer-picklist-groups.conf"),
```
yufengwng, line 3: Nice!

yufengwng (approved):

study-builder/studies/pancan/study-events.conf
```diff
@@ -669,15 +669,36 @@
...
+      "action": {
+        "type": "ACTIVITY_INSTANCE_CREATION",
+        "activityCode": ${id.act.about_cancer},
+        "createFromAnswer": true,
+        "sourceQuestionStableId": ${id.q.primary_cancer_list_child},
+        "targetQuestionStableId": ${id.q.diagnosis_type}
+      },
+      # If there's already about-cancer instances, then don't create any more.
+      "preconditionExpr": """!user.studies["cmi-pancan"].forms["ABOUT_CANCER"].hasInstance()
+      && """ ${_pex.is_child_only},
```
yufengwng: Wait a minute... why do we need another one of this and have it look at child? If they have a `release` activity it means they are doing the self flow (only self or self+child). And if they have a `release_minor` activity then they're going through the child flow and they have answered the child cancer list question.

I don't think we need another one of these?

yufengwng (changes requested):

yufengwng (approved):

PR #908, opened 2021-07-26
DDP-6369 pancan: survey email reminder

yufengwng:
## Context

See DDP-6369. One caveat here is that we don't have pex expression to look at all instances, so right now the email event is not triggering exactly as we want, but it should be a good start.

## Checklist

## FUD Score

Redeploy pancan.

## Testing

### DIFF
--- a/study-builder/studies/pancan/study-events.conf
+++ b/study-builder/studies/pancan/study-events.conf
@@ -221,68 +221,6 @@
       "dispatchToHousekeeping": true,
       "order": 2
     },
-    ## Queue up 3 week reminders
-    {
-      "trigger": {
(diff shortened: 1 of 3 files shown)


study-builder/studies/pancan/study-events.conf
```diff
@@ -221,68 +221,6 @@
       "dispatchToHousekeeping": true,
       "order": 2
     },
-    ## Queue up 3 week reminders
```
yufengwng, line 224: According to the study staff's diagram, we're only doing 1 & 2 week reminders.

ssettipalli (approved):

PR #872, opened 2021-07-12
Pancan PREQUAL and new workflow STUDY_REDIRECT support  

ssettipalli:
PR covers below tickets:
https://broadinstitute.atlassian.net/browse/DDP-6324
https://broadinstitute.atlassian.net/browse/DDP-6313
https://broadinstitute.atlassian.net/browse/DDP-6326
https://broadinstitute.atlassian.net/browse/DDP-6328

For 6313 completion, we need couple of PEX related tickets (DDP-6322 and DDP-6323) to be implemented.
Will be revisiting and completing 6313 in a different PR
Couple of minor issues related to PEX and API Doc changes are still pending..

PR #915, opened 2021-07-27
DDP-6329 add full project names for CMI studies

yufengwng:
## Context

This PR adds in the full names for the existing CMI studies. Also fixes a bug. These name changes have already been applied to `dev` env so you can check it out there (once bug is fixed).

## Checklist

## FUD Score

## Testing

## Release

For deployed envs test/staging/prod, we'll need to run study-builder
```
$ ./render.sh v1 {env} {study}
$ java -Dconfig.file=... -jar target/StudyBuilder.jar ... {study} --run-task UpdateStudyDetails
```


pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
```diff
@@ -172,7 +172,7 @@ private String getStudyName(Handle handle, String userGuid, String studyGuid) {
         if (preferredLanguageCode != null) {
             List<StudyI18nDto> studyTranslations = translationDao.findTranslationsByStudyId(study.getId());
             for (StudyI18nDto dto : studyTranslations) {
-                if (dto.getLanguageCode() != null && dto.getLanguageCode() == preferredLanguageCode) {
+                if (dto.getLanguageCode() != null && dto.getLanguageCode().equals(preferredLanguageCode)) {
```
yufengwng, line 175: Found a bug: we should compare strings with `equals()`.

study-builder/studies/angio/study.conf
```diff
@@ -9,7 +9,7 @@
         "guid": "cmi"
     },
     "study": {
-        "name": "Angiosarcoma Project",
+        "name": "ascproject",
```
yufengwng, line 12: Just restoring the names in the config files. Note that we DID NOT push this `name` change to production when the file was changed, so we're just revert the change here.

ssettipalli (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/service/WorkflowService.java
bskinner: Nice catch! For anyone wondering why- when dealing with objects, `==` is reference equality (do both sides point to the same location in memory), whereas `equals(...)` is object equality (do the objects have the same content).

study-builder/studies/brain/study.conf
```diff
@@ -35,7 +35,13 @@
...
     },
 
-    "studyDetails": [],
+    "studyDetails": [
+        {
+            "language": "en",
+            "name": "Brain Tumor Project",
+            "summary": null
+        }
+    ],
```
bskinner, line 44: This is a handy section to have as a reference.

study-builder/studies/mbc/study.conf
```diff
@@ -37,7 +37,18 @@
...
+    {
+      "language": "en",
+      "name": "Metastatic Breast Cancer Project",
+      "summary": null
+    },
+    {
+      "language": "es",
+      "name": "Metastatic Breast Cancer Project",
+      "summary": null
+    },
```
bskinner, line 50: Do we need the `es` translation if they're both identical?

bskinner (approved):
Looks good! Just had a question about the `escproject` changes.

study-builder/studies/mbc/study.conf
yufengwng: I think so, because the code looks it up by language and if it doesn't find it, I don't think it will automatically assume to grab the English one.

bskinner: Do you think this should be tracked in Jira? From my understanding, this differs from our usual translation behavior of falling back to the default language.

yufengwng: Yeah, I agree! I think if we didn't find the "preferred language" one, we should fallback to the study's "default language" and try to look for "study details" for that language. And if that fails also, then we fallback to "study dto name". I made a comment about this in @ssettipalli's PR: https://github.com/broadinstitute/ddp-study-server/pull/872#discussion_r669087343. Sampath, did we create a ticket to track this?

yufengwng: In any case, I think there shouldn't be any issues with the configurations we have in this PR. For Spanish, I recall that we also want to show the English study name, so we should be good-to-go.

PR #935, opened 2021-07-30
Rarex july crs

YuriyPavlovBI:
Main files here are **RarexVersion2.java** and **patch_0721.conf**. Study configuration changed in accordance to the scripts for consistence of future updates.


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/RarexVersion2.java
```diff
@@ -0,0 +1,723 @@
...
+    private FormActivityDef findActivityDef(String activityCode, String versionTag) {
+        ActivityDto activityDto = jdbiActivity
+                .findActivityByStudyIdAndCode(studyId, activityCode).get();
+        ActivityVersionDto versionDto = jdbiVersion
+                .findByActivityCodeAndVersionTag(studyId, activityCode, versionTag).get();
+        return (FormActivityDef) activityDao
+                .findDefByDtoAndVersion(activityDto, versionDto);
+    }
+
+    private interface SqlHelper extends SqlObject {
```
zyme, line 500: There's some really helpful stuff here--it would be great to see this in its own class somewhere in our dao package.  Will be generally useful for incremental study changes!

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/RarexVersion2.java
```diff
@@ -0,0 +1,723 @@
...
+        detachQuestionFromBothSectionAndBlock(handle, "HD_LABOR_AND_DELIVERY_ISSUES");
+    }
+
+    private void updateExpressionTextByQuestionStableCode(long activityId, String questionStableCode, String expr) {
+        List<Long> ids = helper.findBlockExpressionIdsByQuestionStableCode(activityId, questionStableCode);
+        helper.updateExpressionText(expr, ids);
+    }
+
+    private void updateExpressionText(long activityId, String before, String after) {
+        List<Long> ids = helper.findBlockExpressionIdsByExpressionText(activityId, before);
```
zyme, line 360: Can you do checks on the row counts coming back from all these queries to ensure that the right number of things are being updated?  It's a good sanity check that can catch query errors that might update unrelated studies/forms.

YuriyPavlovBI: That's why it's narrowed down by activityId and also logged. But still it's a good idea, I'll do this.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/RarexVersion2.java
YuriyPavlovBI: Yes, some of the methods could be reused. Let's do that outside of this PR

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/RarexVersion2.java
```diff
@@ -0,0 +1,723 @@
...
+                "QOL_PATIENT_FRIENDS_HELP", "QOL_PATIENT_WANT_TO_BE_FRIENDS", "QOL_PATIENT_TROUBLE_SLEEPING_PAIN",
+                "QOL_PATIENT_PAY_ATTENTION_PAIN", "QOL_PATIENT_HARD_TO_RUN_PAIN", "QOL_PATIENT_HARD_TO_WALK_ONE_BLOCK_PAIN"
+        ));
+        addPicklistOption(handle, studyId, stableIds, option, 60);
+        stableIds.addAll(Set.of("QOL_SELF_PAIN_INTENSITY", "QOL_CHILD_PAIN_INTENSITY", "QOL_PATIENT_PAIN_INTENSITY"));
+        addValidation(handle, studyId, stableIds, ruleConfig);
+        addAsteriskToQuestionContentBlocks(qolSelfId, "QUALITY_OF_LIFE");
+        addAsteriskToQuestionContentBlocks(qolChildId, "CHILD_QUALITY_OF_LIFE");
+        addAsteriskToQuestionContentBlocks(qolPatientId, "PATIENT_QUALITY_OF_LIFE");
+        addBlockToActivity(handle, "QUALITY_OF_LIFE", qolSelfId, 7,
```
yufengwng, line 191: I don't see any new version tags or any new revision ids being used? Shouldn't we be creating new revisions of the activity definitions, or is RareX asking to get these changes in without revisioning?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/RarexVersion2.java
```diff
@@ -0,0 +1,723 @@
...
+                        + ".forms[\"HEALTH_AND_DEVELOPMENT\"].questions[\"HD_GENETIC_TEST\"].answers.hasAnyOption(\"YES\",\"UNSURE\")",
+                "user.studies[\"rarex\"].forms[\"ADD_PARTICIPANT\"].questions[\"PARTICIPANT_RELATIONSHIP\"].answers."
+                        + "hasAnyOption(\"PARENT\", \"GUARDIAN\")\n            && user.studies[\"rarex\"]"
+                        + ".forms[\"HEALTH_AND_DEVELOPMENT\"].questions[\"HD_GENETIC_TEST\"].answers.hasOption(\"YES\")");
+        updateExpressionText(healthAndDevelopmentId,
+                "user.studies[\"rarex\"].forms[\"HEALTH_AND_DEVELOPMENT\"].questions[\"HD_GENETIC_TEST\"].answers."
+                        + "hasAnyOption(\"YES\",\"UNSURE\")",
+                "user.studies[\"rarex\"].forms[\"HEALTH_AND_DEVELOPMENT\"].questions[\"HD_GENETIC_TEST\"].answers."
+                        + "hasOption(\"YES\")");
+
```
yufengwng, line 256: I recommend working through revisioning each activity one at a time. For each activity, create new version tag/revision ids, making the changes as part of that new revision, and move on to next activity.

study-builder/studies/rarex/patches/patch_0721.conf
```diff
@@ -0,0 +1,483 @@
...
+        "variables": []
+      },
+      "picklistLabelTemplate": {
+        "templateType": "TEXT",
+        "templateText": "",
+        "variables": []
+      },
+      "picklistOptions": [
+        {
+          "stableId": "PREFER_NO_TO_ANSWER",
```
yufengwng: Did we mean `PREFER_NOT_TO_ANSWER`?

study-builder/studies/rarex/quality-of-life-child.conf
```diff
@@ -970,12 +970,75 @@
...
+                    }
+                  ]
+                }
               }
             ]
           },
           "blockType": "QUESTION",
           "blockGuid": null,
-          "shownExpr": "true"
+          "shownExpr": """!user.studies["rarex"].forms["QUALITY_OF_LIFE"].questions["PAIN_INTENSITY_CHILD_PREFERENCE"].answers.hasOption("PREFER_NO_TO_ANSWER")"""
```
yufengwng: This is the child variant of this survey right? Shouldn't this be `CHILD_QUALITY_OF_LIFE`?

study-builder/studies/rarex/quality-of-life-patient.conf
```diff
@@ -961,12 +961,75 @@
...
+                    }
+                  ]
+                }
               }
             ]
           },
           "blockType": "QUESTION",
           "blockGuid": null,
-          "shownExpr": "true"
+          "shownExpr": """!user.studies["rarex"].forms["QUALITY_OF_LIFE"].questions["PAIN_INTENSITY_PATIENT_PREFERENCE"].answers.hasOption("PREFER_NO_TO_ANSWER")"""
```
yufengwng: Likewise, should probably be `PATIENT_QUALITY_OF_LIFE`.

study-builder/studies/rarex/add-participant.conf
```diff
@@ -320,6 +320,22 @@
               "templateCode": null,
               "templateText": "",
               "variables": []
+            },
+            "picklistLabelTemplate": {
```
yufengwng, line 324: I think adding the changes directly to the config files is nice, but one thing you lose out on is the ability to test out the different versions of the activities locally, since the config files now reflect the latest thing and there's no config files for the first version. Once we refactor the `RarexVersion2` task to do revisioning properly, I recommend not changing the config files.

Since the `RarexVersion2` should be doing the revisioning correctly, what we can do is then add `RarexVersion2` to the study's `patch-log.conf`. So then, in local development, we can get the first version of things, and then run `--patch` to apply `RareXVersion2` and get version 2 of things. For deployed environments we can `--run-task RarexVersion2` manually to get deployed study up-to-date.

ssettipalli:
@YuriyPavlovBI :Looks like good chunk of config changes without revisioning. 
Existing rareX users who already signed up / submitted forms will see new changes without revisioning.
If thats ok and approved by study staff, can you add some notes on how you tested these changes.
You need to make sure stuff looks ok for existing users/data in addition to new user sign ups. 
Basically you need to initialize old (existing prod version) of study, sign up users . 
Apply new config changes on the study, login as old users and make sure everything looks good.

study-builder/studies/rarex/add-participant.conf
YuriyPavlovBI: Yes, you're right, however my thoughts were: we can checkout any version of configuration for deployment if needed. But once all the environments have updated with patch we don't actually need the "old" version. Also, making next patch will be easier looking at configuration files rather than having updated environment only.
What do you think?

yufengwng: That sounds good. I think it will work better if studies are in their own repos so it's much easier to go off of git commits/tags to checkout an older version of things for local development. But right now studies are all in the same repo. Maybe create a new git tag like `rarex-v1` or something so we actually know which commit to checkout?

YuriyPavlovBI: I think that would work

ssettipalli (commented):
Other than question of revision or not overall config looks ok to me.

study-builder/studies/rarex/patches/patch_0721.conf
ssettipalli: @YuriyPavlovBI :
Are you going to take care of this change to change the stableID to PREFER_NOT_TO_ANSWER ?

study-builder/studies/rarex/quality-of-life-child.conf
ssettipalli: @YuriyPavlovBI :
Are you going to take care of this change to change the stableID to CHILD_QUALITY_OF_LIFE ?

YuriyPavlovBI: Sure

study-builder/studies/rarex/patches/patch_0721.conf
YuriyPavlovBI: Sure

ssettipalli (approved):

PR #945, opened 2021-08-05
DDP-6674 - provide synchronous execution of default address creation and the address snapshotting

evgeniipr:
## Context

Provided synchronous execution of default address creation and the address snapshotting (use reentrant locking mechanism for this).
There are 2 places in the code:
- `CreateMailAddressRoute#handleInputRequest()` - default address created (the process can be slower because it's called `easyPost` service);
- `AddressService#snapshotAddress()` (called from `PutFormAnswersRoute`) - created a copy of a default address (which should exist in DB at this moment).

Until the fix the `snapshotAddress()` was started before the default address creation completed. And this caused fatal error.
The fix is: it is used ReentrantLock mechanism (from the concurrency library) to lock bot places of code and ensure that  `snapshotAddress()` is started only after the default address creation completed.
The approach takes into account that it should be synchronized address creation/snapshotting for the same user. Therefore it is created a separate lock for each user (and this lock is deleted when it is not needed).

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
@@ -422,7 +422,7 @@ public class DataDonationPlatform {
 
         // User mailing address routes
         AddressService addressService = new AddressService(cfg.getString(ConfigFile.EASY_POST_API_KEY),
-                cfg.getString(ConfigFile.GEOCODING_API_KEY));
+                cfg.getString(ConfigFile.GEOCODING_API_KEY), true);
 
         post(API.PARTICIPANT_ADDRESS, new CreateMailAddressRoute(addressService), responseSerializer);
         get(API.PARTICIPANT_ADDRESS, new GetParticipantMailAddressRoute(addressService), responseSerializer);
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
@@ -1,7 +1,7 @@
 package org.broadinstitute.ddp.export;
 
-import static org.broadinstitute.ddp.export.ExportUtil.getSnapshottedMailAddress;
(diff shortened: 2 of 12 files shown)


MocanaAtBroad:
I don't think this solution will solve the problem: using Java lock assumes that address creates request and the PUT that COMPLETEs the activity are in the same VM. At least in production we will have multiple VMs handling requests, so you cannot count on it.

The other question is : is this really necessary anyways? 

It looks like your diagnosis of the problem is that the address might not have been saved or exist when the PUT for the activity is received and the snapshot copy is initiated.

I see in code that Yufeng had commented out the code to check the address. Maybe that just need to be restored?

I think it is OK for PUT to just send back an error code if address is supposed to be there but it isn't.

evgeniipr:
> I don't think this solution will solve the problem: using Java lock assumes that address creates request and the PUT that COMPLETEs the activity are in the same VM. At least in production we will have multiple VMs handling requests, so you cannot count on it.

I agree that it not solves the problem 100%. As a next request can be processed with another instance..
Initially I thought that within same HTTP session requests should be handled with the same instance but then I read documentation and it doesn't say that it can be true. But anyway this fix shouldn't do the things worse or add any side effects.
Btw, previously I thought that only 1 instance configured in Pepper: there are quite many thread synchronization keywords in the backend code, and other synchronization mechanisms like AtomicBoolean, AtomicInteger. In case of multiple instances the usage of thread synchronizations (not within one request) is not correct also.

> The other question is : is this really necessary anyways?

This is strange that 2nd request started before the previous request (address creation) is completed. It means that the frontend sends the requests asynchronously (it not waits until 1st request get a response and send next request?). Initially I asked Alexey but he said that the requests sent synchronously.. After that I decided to try to do this fix.
I think that more correct way to do the fix on front-end side (if possible).

> I think it is OK for PUT to just send back an error code if address is supposed to be there but it isn't.

Ok, then I'll reject this PR and create a new one where in case if default address not found then return an error. ok?

evgeniipr:
This fix will not be applied.
The following is implemented istead: https://github.com/broadinstitute/ddp-study-server/pull/951

PR #964, opened 2021-08-12
Pancan redirect workflow update of existing cmi studies depending on language

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-6782
redirect workflow update of existing cmi studies depending on language.


study-builder/studies/pancan/study-workflows.conf
```diff
@@ -40,31 +40,31 @@
...
-          "expression": ${_pex.is_self}"&&"${_pex.has_prostate}"&&"${_pex.is_advanced_prostate},
+          "expression": ${_pex.is_self}"&&"${_pex.has_prostate}"&&"${_pex.is_advanced_prostate}"&&"${_pex.is_lang_en},
         },
         {
           "type": "STUDY_REDIRECT",
           "studyGuid": "ANGIO",
           "redirectUrl": ${angioRedirectUrl},
-          "expression": ${_pex.is_self}"&&"${_pex.has_angio},
+          "expression": ${_pex.is_self}"&&"${_pex.has_angio}"&&"${_pex.is_lang_en},
         },
```
MocanaAtBroad, line 50: Can count on is_lang_en doing right thing? (I mean, do we have to worry that lang is not set at all, and then pex.is_lang_en returns "false" when, because en is the default language, it should return true?

study-builder/studies/pancan/subs.conf
```diff
@@ -167,19 +167,26 @@
...
+            && !user.studies["cmi-pancan"].forms["PREQUAL"].questions["DESCRIBE"].answers.hasOption("CHILD_DIAGNOSED")
+            && user.studies["cmi-pancan"].forms["PREQUAL"].questions["ADVANCED_BREAST"].answers.hasOption("YES")
+        )
+     || ( user.studies["cmi-pancan"].forms["PREQUAL"].questions["PRIMARY_CANCER_LIST_SELF"].numChildAnswers("PRIMARY_CANCER_SELF") == 1
+            && user.studies["cmi-pancan"].forms["PREQUAL"].questions["PRIMARY_CANCER_LIST_SELF"].children["PRIMARY_CANCER_SELF"].answers.hasOptionStartsWith("C_GENITOURINARY_S_PROSTATE")
             && !user.studies["cmi-pancan"].forms["PREQUAL"].questions["DESCRIBE"].answers.hasOption("CHILD_DIAGNOSED")
-            && (user.studies["cmi-pancan"].forms["PREQUAL"].questions["ADVANCED_BREAST"].answers.hasOption("YES")
-                || user.studies["cmi-pancan"].forms["PREQUAL"].questions["ADVANCED_PROSTATE"].answers.hasOption("YES"))
+            && user.studies["cmi-pancan"].forms["PREQUAL"].questions["ADVANCED_PROSTATE"].answers.hasOption("YES")
+            && user.profile.language() == "en"
```
MocanaAtBroad, line 189: Same question here regarding initializations of language. I HOPE that we can count on it being set to a default, but I don't know. Do you?

study-builder/studies/pancan/study-workflows.conf
ssettipalli: These users are temp users and while creating temp users, we fall back to english

study-builder/studies/pancan/subs.conf
ssettipalli: These users are temp users and while creating temp users, we fall back to english. Verified for users created in DB locally too.
  From CreateTemporaryUserRoute
  
        String languageCode = payload.getLanguageCode();
        if (languageCode == null) {
            //fall back to default language
            languageCode = LanguageStore.DEFAULT_LANG_CODE;
        }

MocanaAtBroad (approved):

PR #969, opened 2021-08-15
Ddp 6766 therapy list question

MocanaAtBroad:
Implemented therapy list as an text question with INCLUDED suggestions autocomplete.
Tested it locally and works.
List was generated by processing list from CMI. 
https://docs.google.com/spreadsheets/d/1HlvSyhEv6c3S9_42Ph9uIpvm3K7Zd-nZ2Q0wW28IQ2o/edit#gid=0

Approach taken to build list:
Create entry for every display name.
If there were abbreviations, add them to the INCLUDED option name enclosing in parentheses.
For alternate names, separate entries were created (independent of the "corresponding" display name).
Had to clean out duplicates (some names were duplicated in either "display name" column or in the "alternate names" colum).
Removed duplicates using case insensitive matching. Database enforces uniqueness in case insenstive manner, so if I tried to include 2 different suggestions for same question that look like:
FooBar
FOOBAR
foobar

We would only allowed to have one these options.
(This is OK for this particular list).


ssettipalli (approved):
looks good to me

PR #982, opened 2021-08-20
Add retries to retrieval of default address from snapshot functionality

MocanaAtBroad:
## Context
This is one part of trying to work around problem with snapshotting of addresses when processing activity PUT COMPLETE.

Examining carefully the GCP logs did show that, for failed PUT requests that complained about mailing address not being available, there was an overlap between the end of processing the mailing address POST and the activity PUT.
In other words, the PUT was being processed before the mailing address POST had completed.

The code in this PR simply retries 5 times with an interval of 2 secs if we fail to find the mailing address during the PUT processing if it is not found. 

We are also modifying the Angular client to introduce a pause between the time it says it thinks it is done processing the address save and the time it initiates the PUT. Existing code looked OK to me, but obviously the data points to a problem.

So hopefully, the retries will not be necessary. BUT don't want to mess around!

## Checklist

- [ X] I have assessed potential risks and labeled using the `R-*` labels.
- [X ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

- [X ] They are user-visible in dev as a regular user journey and require no additional instructions.

## Testing

- [ X] I have written zero automated tests but have poked around locally to verify proper functionality

## Release
Will need to be cherry picked into release branch!


pepper-apis/src/main/java/org/broadinstitute/ddp/service/AddressService.java
```diff
@@ -281,6 +282,23 @@ public boolean deleteAddress(Handle handle, String addressGuid) {
...
     }
 
+    public Optional<MailAddress> findDefaultAddressForParticipantWithRetries(Handle handle, String participantGuid, int retries) {
+        do {
+            Optional<MailAddress> optionalAddress = findDefaultAddressForParticipant(handle, participantGuid);
+            if (optionalAddress.isPresent()) {
+                return optionalAddress;
+            } else {
+                try {
+                    LOG.warn("Could not find default address for participant with guid {}. Waiting to retry", participantGuid);
```
MocanaAtBroad, line 292: Will be interesting to scan the logs to see how often retries were actually necessary.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/AddressService.java
```diff
@@ -281,6 +282,23 @@ public boolean deleteAddress(Handle handle, String addressGuid) {
...
+                    TimeUnit.SECONDS.sleep(2);
+                } catch (InterruptedException e) {
+                    LOG.error("Sleep while waiting to retry findDefaultAddressForParticipant was interrupted, e");
```
Rinchan20: logic lgtm, that amount of retries looks like it'll do the trick - one comment - I think the log.error() in the catch might have the ending quotation marks in the wrong place since it's at the end encasing the error message.

MocanaAtBroad: Thnx! Fixed here cf5a4444bc6d7cc8a7edda78fec3f8e326951d1e
I need have you review more of my PRs, @Rinchan20  !

Rinchan20 (approved):

ssettipalli (approved):

pepper-apis/src/main/java/org/broadinstitute/ddp/service/AddressService.java
evgeniipr: This approach works with transaction isolation level READ COMMITTED or lower.
I checked that in dev it is READ COMMITTED. Hopefully on production is the same level.

evgeniipr (approved):

evgeniipr:
@MocanaAtBroad I tested this locally - it works ok.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/AddressService.java
MocanaAtBroad: That is a good point. We set the isolation level in the JDBC URL and when setting connection pool.
```
jdbc:mysql:///pepperapisprod?cloudSqlInstance=XXXXXX&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=XXXXXX&password=XXXXXX&characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC&useSSL=false&sessionVariables=innodb_strict_mode=on,tx_isolation='READ-COMMITTED',sql_mode='TRADITIONAL'
```
and from our Java code:

https://github.com/broadinstitute/ddp-study-server/blob/fde8856a299b8f4525e883cc73225d16ada4af5c/pepper-apis/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java#L381

Presumably the JDBC will set the default for everything and the Java code sets is up for the specific connection. So I think we have it cover.

PR #997, opened 2021-08-30
DDP-6865,DDP-6885 - implement patches for adding rules LENGTH, UNIQUE

evgeniipr:
## Context

1) Implemented a patch which adds a validation rule of type `LENGTH` and `maxLength=500` for a question `OTHER_COMMENTS` in `about_you` activity.
2) Implemented a patch which adds a validation rule of type `UNIQUE` for a composite question with cancers list.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
+++ b/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
@@ -6,6 +6,7 @@ import java.nio.file.Path;
 
 import com.google.gson.Gson;
 import com.typesafe.config.Config;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.ddp.db.dao.JdbiQuestion;
 import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
 import org.broadinstitute.ddp.db.dao.ValidationDao;
@@ -41,10 +42,12 @@ public class PanCanUniqueCancerValidation implements CustomTask {
     @Override
     public void run(Handle handle) {
         String studyGuid = dataCfg.getString("studyGuid");
-        String questionStableId = dataCfg.getString("questionStableId");
+        String[] questionStableIds = StringUtils.split(dataCfg.getString("questionStableIds"), ',');
         Config ruleConfig = dataCfg.getConfig("uniqueRule");
         StudyDto studyDto = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
-        addValidation(handle, studyDto.getId(), questionStableId, ruleConfig);
+        for (String questionStableId :questionStableIds) {
+            addValidation(handle, studyDto.getId(), questionStableId, ruleConfig);
+        }
     }
 
     public static void addValidation(Handle handle, long studyId, String stableId, Config ruleConfig) {
--- a/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
+++ b/study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
@@ -1,0 +1,59 @@
+package org.broadinstitute.ddp.studybuilder.task;
+
+import static org.broadinstitute.ddp.studybuilder.task.util.TaskUtil.readConfigFromFile;
+
+import java.nio.file.Path;
+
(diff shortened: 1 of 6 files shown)


study-builder/studies/pancan/patches/unique-cancer-validation.conf
```diff
@@ -0,0 +1,22 @@
+{
+  "studyGuid": "cmi-pancan",
+  "questionStableId": "PRIMARY_CANCER_SELF",
+  "uniqueRule":
+    {
+      "ruleType": "UNIQUE",
+      "allowSave": false,
```
ssettipalli: allowSave is a non-nullable attribute in table: validation with a default value: false
Want to clarify that if the value is NOT set for any validation, its defaulting to "false" .
If you're expecting front-end to check this value and behave differently I don't think the approach will work.
If front-end behavior is same as every other validation which has default value of "false", we can remove explicitly setting it to "false" here.

ssettipalli (commented):
We could have combined both patches into one patch and added both validations from one patch.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
```diff
@@ -0,0 +1,59 @@
+package org.broadinstitute.ddp.studybuilder.task;
```
ssettipalli, line 1: We could have combined both patches into one patch and added both validations from one patch.

study-builder/studies/pancan/patches/unique-cancer-validation.conf
evgeniipr: Set allowSave=true (as for all other rules).

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
evgeniipr: Is it suitable to just combine it in `patch-log.conf`? 
```
{
  "patches": [
    "PanCanOtherCommentsValidation",
    "PanCanUniqueCancerValidation"
  ]
}
```
In such case it will be executed all (when option `--patch` is specified).

study-builder/studies/pancan/patches/unique-cancer-validation.conf
```diff
@@ -0,0 +1,22 @@
+{
+  "studyGuid": "cmi-pancan",
+  "questionStableIds": "PRIMARY_CANCER_LIST_SELF,PRIMARY_CANCER_LIST_ADD_CHILD",
```
evgeniipr: Specified a list of composite questions where to add UNIQUE rule:
`"PRIMARY_CANCER_LIST_SELF,PRIMARY_CANCER_LIST_ADD_CHILD"`

ssettipalli: How about "PRIMARY_CANCER_LIST_CHILD" . I believe we need validation for that list too.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
ssettipalli: I meant one java class to handle both but two/separate is ok. 
yes --patch will run all patches in the patch-log.conf file

study-builder/studies/pancan/patches/unique-cancer-validation.conf
evgeniipr: Fixed

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/PanCanUniqueCancerValidation.java
evgeniipr: I think about idea to defile a custom task to add any validation rules (to questions). Then it needs just to add parameters to them (name of a file with configuration).
One of options: to create a small Java class where just to invoke all these validations:
```
void run() {
  addValidator(config_file_1);
  addValidator(config_file_2);
}
```
But may be better to have special config file (placed in /patch subfolder) where to define such tasks for a patch:
```
{
 "tasks": [
   { "type" : "add_validator", "configFile" : "config_file_1" },
   { "type" : "add_validator", "configFile" : "config_file_2" }
 ]
}
```

ssettipalli (approved):

PR #1002, opened 2021-09-01
Add pubsub task to update user doNotContact

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-6896

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/event/HousekeepingTaskReceiver.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/event/HousekeepingTaskReceiver.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.ddp.event;
 
 import java.io.IOException;
+import java.util.Set;
 
 import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.MessageReceiver;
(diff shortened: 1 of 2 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/event/HousekeepingTaskReceiver.java
```diff
@@ -160,6 +166,34 @@ private void handleElasticExport(PubsubMessage message, AckReplyConsumer reply)
         }
     }
 
+    private void handleUserUpdateDoNotContact(PubsubMessage message, AckReplyConsumer reply) {
```
ssettipalli, line 169: @MocanaAtBroad :
This works for Housekeeping pub/sub 
Does this approach work for DSM ?

ssettipalli:
Sample message for this task:

gcloud --project={broad-ddp-dev pubsub} topics publish housekeeping-tasks --attribute 'type=UPDATE_USER_DNC' --message '{"hruids" : "ABCDE, XYZ" , "doNotContact" : true} '

doNotContact defaulted to true

ssettipalli:
Actually I created a new Housekeeping pub/sub task rather than extending UserProfile task .
Will close this one and create a new one.

ssettipalli:
Actually I created a new Housekeeping pub/sub task rather than extending UserProfile task .
Will close this one and create a new one.

PR #1005, opened 2021-09-02
DDP-6904 - fix Stackdriver monitoring of API routes

evgeniipr:
## Context

Fix `Stackdriver` monitoring of API routes. 

The cause of the problem described below:
In class `DataDonationPlatform` in each of Spark Routes registration calls the following method is called: `setupMDC(path, route)`;
This method looks like:
```
private static void setupMDC(String path, Route route) { 
     pathToClass.put(path, route.getClass().getSimpleName());     
     before(path, (request, response) -> MDC.put(MDC_ROUTE_CLASS, pathToClass.get(path))); 
}
```

So, it saves to map `pathToClass` (the info about routes where key=URL, value=route class).
`PutFormAnswerRoute` and `PatchFormAnswerRoute` have same URL: `"/pepper/v1/user/:userGuid/studies/:studyGuid/activities/:instanceGuid/answers"`
And also the method `before()` takes as a 1st parameter an URL which is the same for `PutFormAnswerRoute` and `PatchFormAnswerRoute`. 

In the DataDonationPlatform it is registered in the following order:
```
patch(API.USER_ACTIVITY_ANSWERS)
put(API.USER_ACTIVITY_ANSWERS)
```
Because of both have the same URL the last element saved into the `pathToClass` is `PutFormAnswerRoute`.
And when `StackdriverMetricsTracker` is created it always gets a value stored for `PutFormAnswerRoute`:
```
afterAfter((Request request, Response response) -> {
    String endpoint = MDC.get(MDC_ROUTE_CLASS);
    
    new StackdriverMetricsTracker(StackdriverCustomMetric.API_ACTIVITY,
            study, endpoint, httpMethod, client, participant, response.status(),
            PointsReducerFactory.buildSumReducer()).addPoint(1, Instant.now().toEpochMilli());
});
```

**The fix is:** it is added a prefix (HTTP method) before `path` in the `pathToClass` map. And during the url `before`-filtering an HTTP method is detected from a request and also added before `path` - in order to fetch proper element from `pathToClass`.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -629,43 +636,47 @@ private static void setupApiActivityFilter() {
         });
     }
 
-    private static void setupMDC(String path, Route route) {
-        pathToClass.put(path, route.getClass().getSimpleName());
-        before(path, (request, response) -> MDC.put(MDC_ROUTE_CLASS, pathToClass.get(path)));
+    private static void setupMDC(String path, Route route, String operation) {
```
MocanaAtBroad: Nitpick: can we reorder args so path and operation (Let's call `operation` **`httpMethod`** and `path` **`urlPath`**

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -629,43 +636,47 @@ private static void setupApiActivityFilter() {
...
+        pathToClass.put(operationAndPath(operation, path), route.getClass().getSimpleName());
+        before(path, (request, response) -> MDC.put(MDC_ROUTE_CLASS, pathToClass.get(operationAndPath(request.requestMethod(), path))));
+    }
+
+    private static String operationAndPath(String operation, String path) {
+        return operation + '_' + path;
```
MocanaAtBroad: IMO this operation just a few characters and not used anywhere else, so don't see the need need of a method for this.

MocanaAtBroad (commented):
This makes sense that stackdriver would report incorrectly the Route class, but not so sure that it would make difference in reporting the http method used (PUT vs PATCH). At any rate, could you try out your changes and see if the console on dev reports what we would expect? Not sure best way to do this, but I think it would be OK (if needed) to deploy your branch. If you can validate running locally, that is cool too.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
evgeniipr: Fixed. But I better leave name `path` because it is used in all other methods - just to easier understand that it is about same `path`.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
evgeniipr: Removed

evgeniipr:
@MocanaAtBroad 
> This makes sense that stackdriver would report incorrectly the Route class, but not so sure that it would make difference in reporting the http method used (PUT vs PATCH). At any rate, could you try out your changes and see if the console on dev reports what we would expect? Not sure best way to do this, but I think it would be OK (if needed) to deploy your branch. If you can validate running locally, that is cool too.

After the fix I did some experiment. In my local code added the following code to the `StackdriverMetricsTracker` (to the method `addPoint()`):
```
    static int putCount = 0;
    static int patchCount = 0;

    /**
     * Adds a point to this series, setting stackdriver start/end time
     * to epochTimeMillis
     */
    public void addPoint(long value, long epochTimeMillis) {
        boolean put = false;
        boolean patch = false;
        if (metricsSeries.getSeriesLabels().get("endpoint").contains("PutForm")) {
            put = true;
            ++putCount;
        }
        if (metricsSeries.getSeriesLabels().get("endpoint").contains("PatchForm")) {
            patch = true;
            ++patchCount;
        }
        if (put || patch) {
            LOG.info("Check_Endpoint: Put={}, Patch={}", putCount, patchCount);
        }
        if (DO_STACKDRIVER) {
            metricSeriesPoints.queuePoint(metricsSeries, value, epochTimeMillis);
        }
    }
```
Then I run the backend and frontend (PanCan) locally and entered some forms.
I got the following ratio between PUT and PATCH: `Check_Endpoint: Put=4, Patch=33`.
So it looks more better now.

pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
MocanaAtBroad: Did you reorder the method arguments to path, operation, route?

evgeniipr: Reordered

MocanaAtBroad (approved):

PR #1008, opened 2021-09-03
DDP-6896 . Add doNotContact to UpdateProfile pub/sub task

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-6896

refactored UpdateFirstLastNameHandler.java to UpdateProfileDataHandler so that it can handle all Profile attributes
Will update confluence once I merge the PR.


pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/updateprofile/UpdateProfileProcessor.java
```diff
@@ -52,6 +52,6 @@ protected void updateData(PubSubTask pubSubTask) {
 
         new UpdateEmailHandler().updateEmail(participantGuid, properties);
```
ssettipalli, line 53: May be we can merge EmailHandler into UpdateProfileDataHandler !
It does some additional work like updating Auth0 though. Should be ok to leave it for now

evgeniipr (approved):
Looks ok

MocanaAtBroad (approved):

MocanaAtBroad:
@evgeniipr  @ssettipalli Is this the only or best place we have today to document pub sub API?
https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/2159345665/PubSub+Tasks+API

PR #1000, opened 2021-08-31
Ddp 6920 fix recent db deadlocks

MocanaAtBroad:
## Context

Since Pancan launched we have a seen some occurrences of database deadlocks reported to pepper-prod-alerts.
This PR tries to address each of the deadlock situations we have observed.

I should be clear that I AM NOT SURE THAT THESE CHANGES FIX ALL THE DEADLOCKS but I am confident they do reduce the possibility of deadlocks occurring and that they are safe changes to apply.
Fixes here are mostly based on looking at stack traces and logs to try and figure out what might have gone wrong in moments of high activity.

I should also note that I believe the final piece to really put the deadlock problem to bed is to add a transaction retry on deadlocks being encountered. I have decided to try to do that in a separate PR as the fix will be a of a completely different nature than the work here.

But hope here is not to have any deadlock retries if we can help it!

The theory behind all these fixes is that deadlocks are caused by 2 different transactions locking the same records for UPDATE or INSERT but in different order so: 
transaction A has record X
transaction B has record Y
now transaction A is trying to get record Y (but B has it)
and transaction B is trying to get record X (but A has it).
So neither can proceed so DEADLOCK.

UPDATE FROM MARCO ON INTERESTING CASE:
Looking at logs more closely, it appears that some of these deadlock errors described below might be related. You can expand a series of log entries from GCP prod log:
<details>
<summary>Errors and HTTP calls for an A-T user that had a lot of problems</summary>

</details>
<details>
<summary>Query in broad-ddp-logs that generates log entries above</summary>
<code>
resource.type="gae_app"
resource.labels.module_id="pepper-backend"

resource.type="gae_app" AND
resource.labels.module_id="pepper-backend" AND
(
(textPayload=\~"processAllActionsForEventSignal" OR textPayload=\~"Deadlock" OR textPayload=\~"Lock wait timeout")
OR
(protoPayload.status=500 OR
httpRequest.status=500))
OR
protoPayload.resource=\~"70BL9Q4AC70Y0ZIBD64D"
OR
protoPayload.resource=\~"BTMQJZLT8Y"
OR severity=ERROR
OR
protoPayload.latency>"10000ms"
(timestamp>="2021-08-29T14:03:00.079902Z" AND timestamp<="2021-08-29T16:03:00.079902Z")
</code>
</details>

[...]

PR #1021, opened 2021-09-20
Ddp 6983 prevent activity instance creation deadlock

MocanaAtBroad:
## Context
This is sort of part 2 of this other PR:

https://github.com/broadinstitute/ddp-study-server/pull/1000

We encountered in production another similar deadlock caused by concurrent POST requests to create activity instances.
It is a bug on the front-end; it is incorrect to create multiple activity instances of the the same activity at the same time. Application should just permit one, wait for the response and then open the newly create activity instance.
Bugs have been filed for A-T client:
https://broadinstitute.atlassian.net/browse/DDP-6968
https://broadinstitute.atlassian.net/browse/DDP-6938
But server was still ending up in deadlocks and returning 500.

This PR attempts to solve problem by ensuring that the creation of activities of a given type for a user are serialized.

Proposed solution is to use MySQL database locks to enforce serialization.
1. A new table is created. 
https://github.com/broadinstitute/ddp-study-server/pull/1021/files#diff-1a16e9c6c02d856e04aebf7811b48e89c712fdb7b250f3bac95897c8ab701e89R1
2. When activity instance creation requested is created we will try to insert or update a row in this "mutex" table

https://github.com/broadinstitute/ddp-study-server/pull/1021/files#diff-27995f0a3b61bbea1dfcf65386ea7b1ec72a9de2be4fd25fe8e808467c1f0b0eR69

https://github.com/broadinstitute/ddp-study-server/pull/1021/files#diff-fd745b05134b872e7e8ae69cb59240299105dbebda378fbdfc61d5f67da0f8f1R22

Note that the unique key is a composite for participant id and study activity id; blocking should only occur on requests that are for same user and same activity code.

If it is an mutex table INSERT (when previously no activity instance of this type of activity has been created for this participant), the INSERT will create a gap lock that will prevent other concurrent requests for inserting a similar record.
So other requests will be blocked until the INSERTING transaction is completed.
In the case that the mutex record already exists, an UPDATE will create an exclusive lock and block other  transactions from proceeding.

3. This test seems to confirm the fix:
https://github.com/broadinstitute/ddp-study-server/pull/1021/files#diff-35539bfa0ac76f4f9f9feab2d7ad3d7e56731fdf55e7b75705029232314bfe93R206

Should note that the test above also discovered another bug (aside from the deadlocks): this query includes a count of existing instances of the activity of interest 
https://github.com/broadinstitute/ddp-study-server/pull/1021/files#diff-27995f0a3b61bbea1dfcf65386ea7b1ec72a9de2be4fd25fe8e808467c1f0b0eR70
 and used to decide whether the instance creation should be allowed (eg., max instance count of 1) However, because we use READ COMMITTED transaction isolation, it was possible that  2 concurrent creation requests would be pending at the same time and all the concurrent requests would see a count of 0 after which we would end up with  2 activity instances even though we are supposed to allow only 1.

This PR also fixes this problem: only one transaction is allowed to add an activity instance at a time, so the query is always get a correct count.

## Checklist

- [ X] I have considered security and privacy, and added `I-*` labels as needed
- [X ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.

## FUD Score

- [ X] :relaxed: All good, business as usual!

Can be seen in action by running test. Can also be tested using A-T client and double-clicking buttons to create activity instances (until the fix the problem!)

## Testing

- [X ] I have written automated positive tests
- [X ] I have written automated negative tests

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/ActivityInstanceDao.java
```diff
@@ -63,7 +63,7 @@
     Optional<ActivityInstanceCreationValidation> checkSuitabilityForActivityInstanceCreation(
             @Bind("studyId") long studyId,
             @Bind("activityCode") String activityCode,
-            @Bind("userGuid") String userGuid);
+            @Bind("participantUserId") long participantUserId);
```
MocanaAtBroad, line 66: Just refactoring. We have the user id, so might as well use it instead of the guid.

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
```diff
@@ -60,8 +60,15 @@ public Object handle(Request request, Response response, ActivityInstanceCreatio
             long participantId = found.getUser().getId();
 
             var activityInstanceDao = handle.attach(ActivityInstanceDao.class);
+
+            // Insert or update row for which this transaction will have to hold a lock before proceeding.
+            // This enforces that for a given activity and participant only one transaction is allowed to execute
+            // this code at a time.
+            // Lessens chance of deadlocks and ensures the counts of activity instances for a given activity are
+            // accurate so we can enforce max number of activity instances accurately
+            activityInstanceDao.upsertActivityInstanceCreationMutex(participantId, studyId, activityCode);
```
MocanaAtBroad: This where concurrent requests will be blocked. Only one will be allowed to proceed at at a time.
Note that the concurrency checks are by participant and study activity. 
Also, under normal correct operations, there should rarely (never?) be blocking at all, so this is really to work around client bugs.

pepper-apis/src/main/resources/org/broadinstitute/ddp/db/dao/ActivityInstanceDao.sql.stg
```diff
@@ -9,17 +9,43 @@ select act.study_activity_id,
...
+    (SELECT study_activity_id
+    FROM study_activity
+    WHERE
+        study_activity_code=:activityCode
+        AND study_id=:studyId),
+    :updateTime)
+ON DUPLICATE KEY UPDATE
+    update_time = :updateTime
+>>
+
```
MocanaAtBroad, line 36: insert if does not exist, update the time if it does exist. Both actions result in a database lock held through the completion of transaction.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateActivityInstanceRouteTest.java
```diff
@@ -39,7 +49,7 @@
 
 public class CreateActivityInstanceRouteTest extends IntegrationTestSuite.TestCase {
 
-    private static final Collection<String> activityInstancesToDelete = new HashSet<>();
+    private static final Collection<String> activityInstancesToDelete = ConcurrentHashMap.newKeySet();
```
MocanaAtBroad, line 52: had to make the map concurrent as we use parallel requests for our new test.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateActivityInstanceRouteTest.java
```diff
@@ -161,23 +171,24 @@ public void testCreateActivityShouldFailBecauseTooManyInstances() {
...
+            assertTrue(StringUtils.isNoneBlank(firstInstanceGuid));
+            activityInstancesToDelete.add(firstInstanceGuid);
 
-        String secondInstanceGuid = makeCreationRequest(parentActivity.getActivityCode())
-                .statusCode(200).and().extract().path("instanceGuid");
+            String secondInstanceGuid = makeCreationRequest(parentActivity.getActivityCode())
+                    .statusCode(200).and().extract().path("instanceGuid");
+
+            assertTrue(StringUtils.isNoneBlank(secondInstanceGuid));
+            activityInstancesToDelete.add(secondInstanceGuid);
```
MocanaAtBroad, line 189: Refactored this test to make sure things are cleaned up properly (not relevant to our change).

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateActivityInstanceRouteTest.java
```diff
@@ -191,6 +202,45 @@ public void testCreateActivityShouldHideExistingInstances() {
...
+    @Test
+    public void runCreateInstanceConcurrently() {
+        try {
+            // set constraint that only 1 activity instance of test activity is allowed
+            TransactionWrapper.useTxn(handle -> assertEquals(1, handle
+                    .attach(JdbiActivity.class)
+                    .updateMaxInstancesPerUserById(parentActivity.getActivityId(), 1)));
+            Gson gson = GsonUtil.standardGson();
+            // parallelize requests to test concurrent handling of this situation
+            List<Integer> returnCodes = IntStream.range(0, 4).parallel().mapToObj(i -> {
```
MocanaAtBroad, line 214: using parallel stream to have the concurrent requests in flight.

pepper-apis/src/test/java/org/broadinstitute/ddp/route/CreateActivityInstanceRouteTest.java
```diff
@@ -191,6 +202,45 @@ public void testCreateActivityShouldHideExistingInstances() {
...
+                    }
+                    return response.getStatusLine().getStatusCode();
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }).collect(toList());
+            // we attempted 3 creations, but only one should have succeeded
+            assertEquals(4, returnCodes.size());
+            assertEquals(1, returnCodes.stream().filter(each -> each == 200).count());
+            assertEquals(3, returnCodes.stream().filter(each -> each == 422).count());
```
MocanaAtBroad, line 232: What we are going for. And no database errors either.

evgeniipr:
@MocanaAtBroad 

Other ideas about deadlocks prevention:

What if to try to avoid deadlock by setting more strict transaction isolation level?
At least in `CreateActivityInstanceRoute`.
Currently in pepper set READ COMMITTED. What if to set to REPEATABLE READ or even SERIALIZABLE (from MySQL doc: " _SERIALIZABLE enforces even stricter rules than REPEATABLE READ, and is used mainly in specialized situations, such as with XA transactions and for troubleshooting issues with concurrency and **deadlocks**_".?
```
            handle.setTransactionIsolation(TransactionIsolationLevel.SERIALIZABLE);
```

Another possible approach: to use distributed cache (Redis). Before a transaction to store in Redis map an item with key = praticipant+study. After transaction (in final block) to delete this item. And before transaction check if item exist and if yes - return from method..

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
evgeniipr: Just a suggestion: to extract this code 
`activityInstanceDao.upsertActivityInstanceCreationMutex(participantId, studyId, activityCode);`
to method `preventMultipleCallsDeadlock()` (or invent a better name).
And put the comments as the method comments.
It could prevent method `handle()` to grow too big.

evgeniipr (dismissed):
The solution looks correct.

MocanaAtBroad:
> @MocanaAtBroad
> 
> Other ideas about deadlocks prevention:
> 
> What if to try to avoid deadlock by setting more strict transaction isolation level?
> At least in `CreateActivityInstanceRoute`.
> Currently in pepper set READ COMMITTED. What if to set to REPEATABLE READ or even SERIALIZABLE (from MySQL doc: " _SERIALIZABLE enforces even stricter rules than REPEATABLE READ, and is used mainly in specialized situations, such as with XA transactions and for troubleshooting issues with concurrency and **deadlocks**_".?
> 
> ```
>             handle.setTransactionIsolation(TransactionIsolationLevel.SERIALIZABLE);
> ```
> 

Tried this. Believe it or not, created more deadlocks!

> Another possible approach: to use distributed cache (Redis). Before a transaction to store in Redis map an item with key = praticipant+study. After transaction (in final block) to delete this item. And before transaction check if item exist and if yes - return from method..

Moving transactions to Redis could definitely be an option that could potentially work. Though at the end, seems to me simpler to depend on MySQL to do the concurrency. 
I figure we chose to use relational database with transactions for persistence. Might as well take advantage of them!

pepper-apis/src/main/java/org/broadinstitute/ddp/route/CreateActivityInstanceRoute.java
MocanaAtBroad: Good idea!
b075079b4

pepper-apis/src/main/resources/db-changes/schema/DDP-6983-activity-instance-creation-mutex-table.xml
```diff
@@ -0,0 +1,32 @@
...
+            <column name="activity_instance_creation_mutex_id" type="bigint" autoIncrement="true" startWith="1">
+                <constraints primaryKey="true" primaryKeyName="activity_instance_creation_mutex_pk"/>
+            </column>
+            <column name="participant_id" type="bigint">
+                <constraints nullable="false"
+                             references="user(user_id)"
+                             foreignKeyName="act_instance_mutex_participant_fk"/>
+            </column>
+            <column name="study_activity_id" type="bigint">
+                <constraints nullable="true"
```
ssettipalli, line 18: Why does study_activity_id need to be nullable ?
Every activity instance points to some study_activity_id , right ?  Probably should be a NOT NULL column.

MocanaAtBroad: Good question! I started with a table design with a composite primary key of study_activity_id and participant_id. So study_activity_id was NOT nullable.
But that version of my table did not pass the unit tests.
The test POSTs a request to create a activity instance for a study activity guid that does not exist.
So code had to handle bogus study activities.
So had two choices:
1) Add a query for the study activity to see if it exists and either return an error if bogus so as to avoid inserting a NULL in the mutex table.
2) Allow the null to be stored in the mutex table.
I did not want to add new queries (we do too many queries already!), so made the column nullable. The next query that checks the validation will produce the error.
And everything works fine/there is no harm with nullable column.
So that is the story.

evgeniipr (approved):

PR #1023, opened 2021-09-21
DDP-6950 Sort elastic activity instances to load latest instance first

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-6950

DSM seem to be looking for activities from elastic data and incases where any activity has multiple instances, its getting whatever first instance is in elastic.
DSS DataExport is loading by default order (instances created) , DSM is finding oldest activity instance.

Made change to export "latest" activity instance first and that fixed DDP-6950 issue


evgeniipr:
Does it make sense to do a fix in DSM also? The old exported activities still be taken oldest first?

evgeniipr (approved):

ssettipalli:
> Does it make sense to do a fix in DSM also? The old exported activities still be taken oldest first?

Once nightly uploads happen, new changes should reflect in DSM even for existing participants.

PR #1025, opened 2021-09-24
logging update

ssettipalli:
fix logging message string

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
@@ -261,9 +261,9 @@ public class UserRegistrationRoute extends ValidatedJsonInputRoute<UserRegistrat
                                        UserRegistrationPayload payload, StudyClientConfiguration clientConfig,
                                        Auth0ManagementClient mgmtClient) {
         String studyGuid = study.getGuid();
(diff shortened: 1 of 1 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -261,7 +261,7 @@ private UserPair signUpNewOperator(Response response, Handle handle, StudyDto st
                                        UserRegistrationPayload payload, StudyClientConfiguration clientConfig,
                                        Auth0ManagementClient mgmtClient) {
         String studyGuid = study.getGuid();
-        LOG.info(String.format("Attempting to register new user {}, with client {}  study {}  {}",
+        LOG.info(String.format("Attempting to register new user %s, with client %s  study %s  %s",
                 auth0UserId, clientConfig.getAuth0ClientId(), studyGuid,
                 payload.getTempUserGuid() != null ? " (temp user " + payload.getTempUserGuid() + ")" : " NO-Temp-User"));
```
MocanaAtBroad, line 267: I'll approve cuz this will work, but just FYI: you can do something like this with slf4j:
```java
LOG.info("[export] found {} activities for study {}", activities.size(), studyGuid);

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #1032, opened 2021-09-29
DDP-6997 Circadia PILOT: Changes for DLMO Instructions

YuriyPavlovBI:
As per DDP-6997 we need to change the trigger event for DLMO Instructions creation
CircadiaUpdateDLMOInstructionsTrigger needs to be run to update the trigger


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/CircadiaUpdateDLMOInstructionsTrigger.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.File;
+import java.nio.file.Path;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * One-off task to update configurations and events to support kit uploads and adhoc survey in deployed environments.
+ */
+public class CircadiaUpdateDLMOInstructionsTrigger implements CustomTask {
```
ssettipalli, line 46: Can we create a patch-log.conf and update it with patches we add.

ssettipalli (approved):

PR #1034, opened 2021-09-29
DDP-6941 - add translations processor to Study Builder: add translations for languages defined in a study

evgeniipr:
### What this PR is about?

This PR contains implementation of a `StudyBuilder` _new feature_: **translations' references automatic generation**. It is suggested to discuss this implementation: is it suitable 
_NOTE:_ **_this feature will work for new studies only!_** 
**For existing studies it will be created another functionality allowing to add translations for new languages directly to DB.**

#### How translations are defined now

In the `StudyBuilder` we can create a new `study` (set of config files). And define different translations: for activity titles, section titles, question prompts, etc. So, it needs to define `translation references` in a `study` `conf-files`.  
Like this:
```
      "promptTemplate": {
              "templateType": "HTML",
              "templateText": "$hint",
              "variables" : [
                  {
                    "name": "hint",
                       "translations": [
                           { "language": "en", "text": ${i18n.en.consent.readonly_hint} },
                           { "language": "es", "text": ${i18n.es.consent.readonly_hint} },
                       ]
                    }]
                  },
```
Where `${i18n.en.consent.readonly_hint}` a `translation reference`  (`name=value` pair) defined in a special `translation file` (usually named like `xx.conf` where xx - language code, example: `en.conf`). 
So it is required to define in every template translation references for each of known languages. And some templates contain more than one variable - in such case the translations definition becomes yet more complicated: needs to define a list of `translation references` for each variable.

#### How translations will be defined with the new approach

The new feature **translations' references automatic generation** simplifies this process: it no need to define `variables` with translations (but `variable names` should be equal to `translations references`. And the variables elements and it's `translation references` will be added automatically to templates during study objects generation (before writing it to DB).
In order to **translations' references automatic generation** it needs to define a `template` like this:
```
            "promptTemplate": {
              "templateType": "HTML",
              "templateText": """$prequal.primary_cancer_prompt_child <br/> <small><em>$prequal.primary_cancer_note_child</em></small>"""
            },
```
**So, it needs just to define an empty element `"variables"`.**

All `Velocity variables` are defined in `templateText`. 
And it should follow the rule: **`Velocity variables` which are defined in a `templateText` should coincide with `translations keys`.**
During `study` DB data generation for each of a `Velocity variable` an element `"variable"` (in `"variables"`) is generated, and for each `variable` generated a list of `translations` - for each of known `languages`.
And this generation based on the fact that `Velocity variables` defined in `templateText` equal to`translation references` (without prefix `i18n.xx` where `xx` - langCde).
It is expected that for each language defined a translation for variables specified in `templateText`.
So, if in a study defined languages `en`, `es`, then it is expected that in files `en.conf` and `es.conf`
specified properties:
`prequal.primary_cancer_prompt_child`,
`prequal.primary_cancer_note_child`.

There is a special case where `translations` are defined not inside a `Template` but as an _array of translations_. Such translations are defined for some of activity properties: title, subtitle, etc...
In order to define a `translation key` for such translations it needs to define it in an attribute `text` of a default language.
For example:
```
  "translatedNames": [
    { "language": "en", "text": "$prequal.name" }
  ],
```
In above example text contains `$prequal.name` and letter `$` means that this is not a real text but a reference to a `translation key`.
For such case will be generated translation references for each of known languages (and a default language translation text will be replaced with real translation value (for this language)).
NOTE: it is planned to get rid of such non-standard translations and wrap it into `Template` object (but it needs to preserve somehow a backward compatibility with old studies).

#### On the StudyBuilder side:
- class `ActivityDefTranslationsProcessor` which called after an object `ActivityDef` is built from a `conf-definition`: this class processes each translation and generates `translation references` for each of variables defined in `templateText`;
- 
- command-line option `process-translations` which enables the process of `translation references generation` (by default it is `false` because we do not want to execute it for old `studies`);

- `automatic detection` of `new languages` in a `study` (in `StudyBuilder`): if command-line option `i18n-path` is specified and contains valid folder with 18n-files then `StudyBuilder` detects i18n-translations not from  `subs.conf` section but read all `translation files` from a specified `folder`. If no any i18n-files are found in the specified folder then read it from `subs.conf`;

- unit-test `ActivityDefTranslationsProcessorTest` which builds object `ActivityDef` from a test `conf-file` and runs `ActivityDefTranslationsProcessor` and then after `translation references generation` it is rendered some of `templates` (defined in a test `conf-file`) to verify that rendered text is correct.

  
#### On DSS (Pepper backend) side:
- provided a possibility to correctly handle Velocity variables with dots in it's names;

### How to use the feature in Study Builder

There are 2 command-line arguments are introduced:
[1]
**process-translations** - if it is specified then translations' references will be generated (by default it is not generated - Study Builder works as before).
Example:
```
java -Dconfig.file=./output-config/application.conf -jar ./target/StudyBuilder.jar basil --vars ./output-config/basil-vars.conf ./studies/basil/study.conf --substitutions ./studies/basil/subs.conf --process-translations=PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS
```
(all possible values: PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS, PROCESS_ALL_TEMPLATES, NOT_PROCESS).
Note: PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS - can be used with existing and new studies because templates containing child variables[] are not processed (i.e. it's processed as before).

[2]
**i18n-path** - this option allows to specify a folder with i18n files: if such option specified and a folder contains i18n-files then study languages are detected automatically and it is no need to define i18n-files in `subs.conf`.
Example:
```
java -Dconfig.file=./output-config/application.conf -jar ./target/StudyBuilder.jar basil --vars ./output-config/basil-vars.conf ./studies/basil/study.conf --substitutions ./studies/basil/subs.conf --i18n-path ./studies/basil/i18n
```

### What else needs to be implemented

1) Implement a `translations references automatic generation test mode`: in such mode the generated translations and resolved templates could be written to an output file instead of direct saving to DB. It would provide early checking of translations correctness before saving it to DB.

2) Translations auto-generation for pdf files.

3) Translations auto-generation for emails.

## Checklist

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -38,7 +42,7 @@
 
     private VelocityEngine engine;
 
-    private static Long getDefaultLanguageId(Handle handle) {
+    private static Long getDefaultLanguageId() {
         if (defaultLangId != null) {
```
evgeniipr, line 46: Just removed unused parameter

pepper-apis/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -302,12 +306,27 @@ private String render(Handle handle, Long contentTemplateId, Long languageCodeId
...
+     *
+     * @param template  template text (stored in {@link Template#getTemplateText()}
+     * @param context   map with Velocity variables to be loaded to Velocity context
+     * @return a string with rendered template text
+     */
     public String renderToString(String template, Map<String, Object> context) {
-        VelocityContext ctx = new VelocityContext(context);
+
+        Map<String, Object> convertedContext = VelocityUtil.convertVariablesToValid(context);
+        String convertedTemplate = VelocityUtil.convertTemplateVariablesToValid(template, context.keySet());
```
evgeniipr: Before rendering a Velocity template convert in all variables '.' to '-'.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiActivityVersion.java
```diff
@@ -24,7 +24,7 @@
             + " where av.study_activity_id = :activityId"
             + " order by rev.start_date asc")
     @RegisterRowMapper(ActivityVersionDto.ActivityVersionDtoMapper.class)
-    List<ActivityVersionDto> findAllVersionsInAscendingOrder(long activityId);
+    List<ActivityVersionDto> findAllVersionsInAscendingOrder(@Bind("activityId") long activityId);
```
evgeniipr, line 28: Fixed unit test

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/ActivityBuilder.java
```diff
@@ -333,15 +331,25 @@ public Config readDefinitionConfig(String filepath, boolean allowUnresolved) {
...
+     *   <li>build an instance of {@link ActivityDef} from {@link Config} definition;</li>
+     *   <li>if command-line option `process-translations` is specified then run the process of
+     *       translations' references automatic generation;</li>
+     *   <li>validate the {@link ActivityDef}.</li>
+     * </ul>
+     */
+    private ActivityDef buildActivityDefFromConfig(Config definition) {
+        ActivityDef activityDef = gson.fromJson(ConfigUtil.toJson(definition), ActivityDef.class);
+        activityDefTranslationsProcessor.run(activityDef);
+        validateDefinition(activityDef);
```
evgeniipr, line 350: After the ActivityDef is built the translations processing is called (and it will be really called only if command-line option `process-translations` is specified. In unit tests this command-line flag not affects: the boolean flag set directly in the code.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
```diff
@@ -73,6 +76,7 @@ private void run(String[] args) throws Exception {
         options.addOption(null, OPT_EMAIL_KEYS, true, "comma-separated email keys, only create/update emails with these keys");
         options.addOption(null, OPT_RUN_TASK, true, "run a custom task");
         options.addOption(null, OPT_PATCH, false, "run patches for a study");
+        options.addOption(null, OPT_PROCESS_TRANSLATIONS, false, "run the process of translations assignment for all study languages");
```
evgeniipr: When this option is specified the translations references will be resolved automatically:
`StudyBuilder pancan --vars ../../configs/vars_pancan.conf -substitutions studies/pancan/subs.conf -process-translations`

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Map;
+import java.util.Properties;
+
+import com.typesafe.config.Config;
+import org.broadinstitute.ddp.studybuilder.translation.TranslationsUtil;
+
+
+public class StudyBuilderContext {
+
```
evgeniipr, line 19: This class holds the translations fetched from translation files (it is required during translations references auto-resolving).

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/ActivityDefTranslationsProcessor.java
```diff
@@ -0,0 +1,188 @@
...
+ *                 detect from such element a name of a translation and add translations for the rest of languages
+ *                 (which are still not defined);
+ *               - if variables array is empty then detect names of variables from a Template text and build
+ *                 translations for each of detected variables (for each of active languages).
+ *           </li>
+ *       </ol>
+ *     </li>
+ * </ul>
+ */
+public class ActivityDefTranslationsProcessor {
```
evgeniipr, line 59: This is a main class running the translations processing.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/ActivityDefTranslationsProcessor.java
```diff
@@ -0,0 +1,188 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ActivityDefTranslationsProcessor.class);
+
+    private final Map<String, Properties> allTranslations;
+
+    public ActivityDefTranslationsProcessor(Map<String, Properties> allTranslations) {
+        this.allTranslations = allTranslations;
+    }
+
+    public void run(ActivityDef activityDef) {
+        if (StudyBuilderContext.CONTEXT.isProcessTranslations()) {
```
evgeniipr: Translations processing will start only if context.processTranslations==true (which is assigned by cmd option `--process-translations`).

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/template/Template.java
```diff
@@ -31,7 +31,7 @@
 
     @NotNull
     @SerializedName("variables")
-    private final Collection<@Valid @NotNull TemplateVariable> variables = new ArrayList<>();
+    private Collection<@Valid @NotNull TemplateVariable> variables = new ArrayList<>();
```
evgeniipr, line 34: If in StudyBuilder a conf file serialized without an empty element `"variables":[]"` then this array assigned to null and eventually we get NPE. But we want to avoid an empty `"variables":[]"` declaration in a conf (we want to do a 
 template declaration as compact as possible) therefore it needs to check if this array is null and re-initiate it.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/template/Template.java
```diff
@@ -88,12 +88,19 @@ public String getTemplateText() {
 
     public void addVariable(TemplateVariable variable) {
         if (variable != null) {
+            if (variables == null) {
+                variables = new ArrayList<>();
+            }
```
evgeniipr, line 119: Add this check in order to avoid NPE (in case when an empty "variables":[]" is not declared in conf).

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/template/Template.java
```diff
@@ -88,12 +88,19 @@ public String getTemplateText() {
...
             variables.add(variable);
         }
     }
 
     public Optional<TemplateVariable> getVariable(String name) {
-        return variables.stream().filter(var -> var.getName().equals(name)).findFirst();
+        if (variables == null) {
+            return null;
+        } else {
```
evgeniipr, line 130: Add this check in order to avoid NPE (in case when an empty "variables":[]" is not declared in conf).

pepper-apis/src/main/java/org/broadinstitute/ddp/content/VelocityUtil.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     *   From this template text will be extracted variables:
+     *      'prequal.intro_thank_you',
+     *      'prequal.intro_auto_save'.
+     * </pre>
+     *
+     * @param templateText template text where to find Velocity variables
+     * @return list of String - list of variables
+     */
+    public static List<String> detectVelocityVariablesFromTemplate(String templateText) {
```
MocanaAtBroad:
```suggestion
    public static List<String> extractVelocityVariableExpressions(String templateText) {
```

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderContext.java
```diff
@@ -0,0 +1,38 @@
...
+ * <pre>
+ * - translations for all languages (stored in translation files);
+ * - boolean flag `processTranslations` indicating if translations references auto-resolving should happen or not.
+ * </pre>
+ */
+public class StudyBuilderContext {
+
+    public static final StudyBuilderContext CONTEXT = new StudyBuilderContext();
+
+    private boolean processTranslations = false;
```
MocanaAtBroad: I don't know what this means.

MocanaAtBroad (commented):
My impression is that code is still overly complicated, particularly because it tries to parse the expression itself (extract variable names), build a map of variable expressions (e.g., "prequal.title" -> "The title", etc)

I don't think any of this is necessary. We have standardized on Velocity templates which have variable expressions. And the HOCON configuration data is in a structure that can be easily converted to data structure that can be used from a Velocity context.
Let's delete all the parsing code and see what we can do with just Velocity templates and HOCON Config objects or JSON object or Maps of Maps of Objects to hold the translation content.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/ActivityDefTranslationsProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+            questionDef.getValidations().forEach(r -> enrichRuleWithTranslations(r));
+        }
+        switch (questionDef.getQuestionType()) {
+            case DATE:
+                if (((DateQuestionDef) questionDef).getRenderMode() != PICKLIST) {
+                    addTemplateTranslations(((DateQuestionDef) questionDef).getPlaceholderTemplate(), allTranslations);
+                }
+                break;
+            case BOOLEAN:
+                addTemplateTranslations(((BoolQuestionDef) questionDef).getTrueTemplate(), allTranslations);
```
MocanaAtBroad, line 184: I know we have this kind of code everywhere in the pepper backend: some piece of functionality that has a case statement that does something slightly different depending on the type of question or block or whatever.
This type of code precedes your arrival to the group, so your code is consistent with this pattern, but I think it is one of the things that makes pepper code harder to modify than it should be: if you try to introduce, say a new type of question, you have to touch code in a bunch of different places.

After we have figured out how to process the templates (without doing our own parsing), it might make sense to look and see if it would be possible to get rid of some the case statements and replace this reflective processing, perhaps (most likely) using annotations.
You see in this chunk of code, depending on which type of question we go object property by object property doing the same thing.
Another way to approach this is to have code that you give it the question object as an argument, and identifies itself what are the object properties that need to be processed.
The properties can be identified by either their type (e.g., the java object property is a Template) or an instance variable or method is has, say, "Template" annotation.

study-builder/src/test/resources/teststudy/activity.conf
```diff
@@ -0,0 +1,90 @@
...
+  # This activity should not be shown in dashboard -- only used for kicking off Add Child flow.
+  "excludeFromDisplay": true,
+  # Used by logged-in proxy to add child, so not going to allow unauthenticated.
+  "allowUnauthenticated": false,
+  "translatedNames": [
+    { "language": "en", "text": "$prequal.name" },
+    { "language": "fi", "text": "Finnish text" }
+  ]
+  "translatedTitles": [
+    { "language": "en", "text": "$prequal.title" }
```
MocanaAtBroad: Don't think it makes any sense to include Velocity expressions into translatedNames and translatedTitles given that these identify the language already.
(It is unfortunate that original design had two different ways to handle translations. I think this is caused by too many chefs making the soup).
If we use the Velocity expressions/templates in the configuration, then we don't need to also include the language.
We should keep translatedNames in the configuration for backwards compatibility, but would it make sense to just add to configuration?
"name": ${prequal.name}
"title": ${prequal.title}

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsUtil.java
```diff
@@ -0,0 +1,211 @@
...
+    public static Map<String, Properties> getTranslations(Config conf, String translationsPath) {
+        return groupTranslationsByLangs(toProperties(conf.getConfig(translationsPath).resolve()));
+    }
+
+    /**
+     * Convert Config to Properties
+     */
+    public static Properties toProperties(Config config) {
+        Properties properties = new Properties();
+        config.entrySet().forEach(e -> properties.setProperty(e.getKey(), config.getString(e.getKey())));
```
MocanaAtBroad, line 50: As I indicated, in main comment for PR, I would like to get rid of the conversion to `Properties` and just treat the translation structure as either JSON or map of maps.
However, this line of code makes me think that this MIGHT be a good strategy for storing the translations into the database IF using JSON is to troublesome.
Essentially we could "flatten" out the the translations config or JSON object into a one level Map with key and value and store it that way in the database.
Then when reading back, contruct the JSON or Map of Maps based on the key value (e.g., if the key value is prequal.title, then we create a key in top level map called prequal that points to a map.
And the second level map has a key called title that points to the actual string value.
This MIGHT be easier to implement using MySQL and JDBC/JDBI than to introduce JSON-specific database operations.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/ActivityDefTranslationsProcessor.java
evgeniipr: I fully agree that needs to do such refactoring.

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsUtil.java
evgeniipr: Yes, it definitely easier.

study-builder/src/test/resources/teststudy/activity.conf
evgeniipr: I suggest to introduce Template properties to Activity level (for title, subtitle) - nullable, and at the same use the old properties. And for new functionality use new properties.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/VelocityUtil.java
evgeniipr: Fixed

evgeniipr: @MocanaAtBroad I reimplemented parsing code: reused code from Velocity and also found some piece of code in the Internet (for variables to convert to map of maps of maps..).

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderContext.java
evgeniipr: This means that by default translations generation is turned off (it needs to specify a special command-line option to turn it on).

study-builder/src/test/resources/teststudy/activity.conf
evgeniipr: > would it make sense to just add to configuration?
> "name": ${prequal.name}
> "title": ${prequal.title}

Still not understand how to use it: when HOCON files parsed it will try to resolve these substitutions: and it throws errors as not finds ${prequal.name}...
The only possible solution: somehow to parse HOCON without substitutions resolving.. Is it possible?

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsProcessingType.java
```diff
@@ -0,0 +1,33 @@
...
+ * Defines different types of translations processing (translation references generation).
+ */
+public enum TranslationsProcessingType {
+
+    /**
+     * Not to run the translations processing.
+     * This is how the conf files processed an old way: translations
+     * should be defined inside variables in templates definitions.
+     */
+    NOT_PROCESS,
```
ssettipalli: Do we need NOT_PROCESS
Looks like default is to not process unless PROCESS_ALL_TEMPLATES or PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS is set with option --process-templates
OR if --process-templates is not passed, right ?

pepper-apis/src/main/java/org/broadinstitute/ddp/content/VelocityUtil.java
```diff
@@ -0,0 +1,154 @@
...
+import org.broadinstitute.ddp.model.activity.definition.template.Template;
+import org.broadinstitute.ddp.model.activity.definition.template.TemplateVariable;
+
+/**
+ * Utility methods used during Velocity {@link Template} rendering.
+ * It helps to render the templates where used Velocity variables containing '.'-symbol (this symbol
+ * is used in order to support translations references automatic resolving: as soon as this feature
+ * enabled in the study conf-files (in StudyBuilder) it needs to define Velocity variables with same names
+ * as translations references (without prefix `i18n.xx' where xx - langCde).
+ * For example, in StudyBuilder conf file thr translations defined in files en.conf, es.conf.. like:
```
ssettipalli: typo thr for the

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/template/Template.java
ssettipalli: Looks like we are initializing to empty ArrayList in the declaration:  variables = new ArrayList<>();

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/template/Template.java
ssettipalli: Looks like we are initializing to empty ArrayList in the declaration: variables = new ArrayList<>() and might not run into NPE

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PropertiesToMapTreeUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.apache.commons.lang.StringEscapeUtils;
+
+/**
+ * Converts properties to hierarchy of maps: properties names are splitted into parts which separated by '.'
```
ssettipalli: typo split

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PropertiesToMapTreeUtil.java
```diff
@@ -0,0 +1,55 @@
...
+ * </pre>
+ * and properties:
+ * <pre>
+ * Properties p = new Properties();
+ * p.put("prequal.sect1.name", "Sect_One");
+ * p.put("prequal.sect2.name", "Sect_Two");
+ * </pre>
+ * And the properties can be converted to a Map hierarchy and loaded to a Velocity context to evaluate a template.
+ */
+public class PropertiesToMapTreeUtil {
```
ssettipalli, line 27: Do we need a separate class or can this go into VelocityUtil.java !

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsEnricher.java
```diff
@@ -0,0 +1,143 @@
...
+     * The result translations list are returned.
+     *
+     * @param translations     existing array of translations (it can be empty or null)
+     * @param templateVariable template variable which equal to a translation key to be detected
+     * @param allTranslations  all translations for all languages of a study
+     * @return list of Translations built for all languages of a study.
+     */
+    public static List<Translation> addTranslations(
+            List<Translation> translations, String templateVariable, Map<String, Properties> allTranslations) {
+        String translationKey = templateVariable;
```
ssettipalli, line 118: wonder why this additional assignment to new variable !

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsProcessingType.java
ssettipalli: Looks like we don't need to pass this argument NOT_PROCESS to the StudyBuilderCli. Good.

pepper-apis/src/main/java/org/broadinstitute/ddp/content/VelocityUtil.java
evgeniipr: Fixed

evgeniipr:
> Looks like we are initializing to empty ArrayList in the declaration: variables = new ArrayList<>();

This needs to handle the following problem:
If we not specify an empty `variables[]` in conf-file (under a template definition) then during JSON de-serialization it causes that member variable `Template#variables` is set to null. But we don't want to define an empty `variables[]` - because the aim is to do a template definition as less verbose as possible. Therefore it needs to do this workaround: create new ArrayList() if it is null (otherwise we get NPE). 
I'll add this explanation in the method comments otherwise the code looks confusing..

evgeniipr:
> Looks like we are initializing to empty ArrayList in the declaration: variables = new ArrayList<>() and might not run into NPE
@ssettipalli the same problem is fixes there as described in this comment: https://github.com/broadinstitute/ddp-study-server/pull/1034/files#issuecomment-963167744

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PropertiesToMapTreeUtil.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/util/PropertiesToMapTreeUtil.java
evgeniipr: The idea was to define a util class which could be reused in other classes (not only for Velocity engine).
Now it is used in VelocityUtil only, but in future it could be used in other code: may be to convert properties to maps could be helpful somewhere else..

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsEnricher.java
evgeniipr: This was an attempt to emphasize the fact that parameter `templateVariable` comes from a templateText (to be defined like "$prequal.name") but we want to use it as a translation key (like "prequal.name").
This method (`addTranslations()`) is used to generate a list of template's child variables with translations inside them..
I'll try to describe the logic in more detail in the method comments..

study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/translation/TranslationsProcessingType.java
evgeniipr: I removed the NOT_PROCESS. It is really not needed.
So only 2 possible values: PROCESS_ALL_TEMPLATES, PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS.
But it looks like that it is enough only one: PROCESS_IGNORE_TEMPLATES_WITH_TRANSLATIONS...
But may be it could be possible to use PROCESS_ALL_TEMPLATES in some future when revisioning will be supported and updates in conf files could be applied to a DB..

ssettipalli (approved):

ssettipalli (dismissed):

ssettipalli (approved):

PR #1040, opened 2021-10-08
Fix error "Using existing user to upgrade..." in registration rule

aleksey-kuznetsov:
My previous update to this rule introduced a bug that wouldn't allow users to log in after first attempt.
I've added a couple of checks to append `temp_user_guid` from `user_metadata` only when no value was found in request AND user wasn't registered previously. This way subsequent log in attempts will no longer send `temp_user_guid` alongside normal request payload.

### DIFF
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -90,11 +90,17 @@ function (user, context, callback) {
         } else if (context.request.body.temp_user_guid) {
             pepper_params.tempUserGuid = context.request.body.temp_user_guid;
             console.log('Temp user guid passed in (via body) = ' + pepper_params.tempUserGuid);
-        } else if (user.user_metadata.temp_user_guid) {
-            pepper_params.tempUserGuid = user.user_metadata.temp_user_guid;
-            console.log('No temp user guid found in request, taking one from user metadata');
         } else {
-            console.log('No temp user guid found in request nor user metadata');
+            console.log('No temp user guid passed in request');
+        }
+
+        /**
+         * If `tempUserGuid` was not set with value from request
+         * AND user is not yet registered in pepper (`user.app_metadata.user_guid` is empty)
+         * take `tempUserGuid` from `user_metadata`
+         */
+        if (!pepper_params.tempUserGuid && !user.app_metadata.user_guid) {
+            pepper_params.tempUserGuid = user.user_metadata.temp_user_guid;
         }
 
         if (context.request.query.mode) {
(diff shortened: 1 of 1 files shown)


study-builder/tenants/register-user-in-pepper.js
```diff
@@ -90,11 +90,17 @@ function (user, context, callback) {
...
-            console.log('No temp user guid found in request nor user metadata');
+            console.log('No temp user guid passed in request');
+        }
+
+        /**
+         * If `tempUserGuid` was not set with value from request
+         * AND user is not yet registered in pepper (`user.app_metadata.user_guid` is empty)
+         * take `tempUserGuid` from `user_metadata`
+         */
+        if (!pepper_params.tempUserGuid && !user.app_metadata.user_guid) {
```
ssettipalli: You might also need to check if metadata has  temp_user_guid) before trying to use it , right   ? 
Something like adding another if condition:
if (!pepper_params.tempUserGuid && !user.app_metadata.user_guid && user.user_metadata.temp_user_guid) {

}

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -90,11 +90,21 @@ function (user, context, callback) {
...
+
+        /**
+         * If `tempUserGuid` was not set with value from request
+         * AND user is not yet registered in pepper (`user.app_metadata.user_guid` is empty)
+         * take `tempUserGuid` from `user_metadata` (if one exists)
+         */
+        if (
+            !pepper_params.tempUserGuid &&
+            !user.app_metadata.user_guid &&
+            !!user.user_metadata.temp_user_guid
```
ssettipalli, line 105: need to delete !! in !!user.user_metadata.temp_user_guid

bskinner: I don't think so- the `!!` is coercing the non-boolean `user.user_metadata.temp_user_guid` into a boolean type.

ssettipalli: Just thought it would be clean without it , functionally works.

ssettipalli (approved):

study-builder/tenants/register-user-in-pepper.js
bskinner: It may look cleaner, but removing the `!!` also makes the statement less safe.

bskinner: Would that decrease the safety of the statement? My understanding of the `!!` is it forces the value to an explicit boolean type, and prevents some type safety issues from cropping up.

Edit: Nope- `if` already coerces things to their truthy value, so it's redundant in this particular situation. TIL!

PR #1045, opened 2021-10-20
DDP-7039 - implement user full delete service

evgeniipr:
## Context

Implement a **user full delete** service; implement a new PubSub task of type USER_DELETE  (in order to be possible to delete user through PubSub).

**Basic changes:**
-  implemented service `UserFullDeleteService` providing user's all data deletion (it utilizes a service `UserDeleteService` which used to delete a user through a REST service;
- added a new PubSubTask type: `USER_DELETE` and implemented `UserDeleteProcessor` for an incoming PubSub task processing (user deletion).

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/KitTypeDao.java
```diff
@@ -27,7 +27,7 @@
 
     @SqlQuery("SELECT kit_type_id, name FROM kit_type WHERE name = :kitTypeName")
     @RegisterRowMapper(KitTypeRowMapper.class)
-    Optional<KitType> getKitTypeByName(@Bind String kitTypeName);
+    Optional<KitType> getKitTypeByName(@Bind("kitTypeName") String kitTypeName);
```
evgeniipr, line 30: Fixed text

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dto/EnrollmentStatusDto.java
```diff
@@ -12,8 +13,15 @@
...
-                               EnrollmentStatusType enrollmentStatus, long validFromMillis, Long validToMillis) {
+    public EnrollmentStatusDto(
+            @ColumnName("user_study_enrollment_id") long userStudyEnrollmentId,
+            @ColumnName("user_id") long userId,
+            @ColumnName("user_guid") String userGuid,
+            @ColumnName("study_id") long studyId,
+            @ColumnName("study_guid") String studyGuid,
+            @ColumnName("enrollment_status") EnrollmentStatusType enrollmentStatus,
+            @ColumnName("valid_from_millis") long validFromMillis,
+            @ColumnName("valid_to_millis") Long validToMillis) {
```
evgeniipr, line 24: Fixed test

pepper-apis/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/TaskPubSubPublisher.java
```diff
@@ -45,9 +45,9 @@ private PubsubMessage createPubSubMessage(String taskType, String eventPayload,
         var messageBuilder = PubsubMessage.newBuilder();
         messageBuilder.setData(ByteString.copyFromUtf8(eventPayload));
         messageBuilder.putAllAttributes(Map.of(
-                ATTR_TASK_TYPE, taskType,
-                ATTR_STUDY_GUID, studyGuid,
-                ATTR_PARTICIPANT_GUID, participantGuid
+                ATTR_NAME__TASK_TYPE, taskType,
+                ATTR_NAME__STUDY_GUID, studyGuid,
+                ATTR_NAME__PARTICIPANT_GUID, participantGuid
```
evgeniipr, line 50: Constants' names refactoring.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTask.java
```diff
@@ -11,7 +11,13 @@
...
 public class PubSubTask {
 
-    public static final String ATTR_TASK_TYPE = "taskType";
+    /** Name of a PubSub message attribute holding a PubSub task type (for example 'UPDATE_PROFILE', 'USER_DELETE') */
+    public static final String ATTR_NAME__TASK_TYPE = "taskType";
+
+    /** PubSubTask standard attribute: holds a study GUID */
+    public static final String ATTR_NAME__STUDY_GUID = "studyGuid";
+    /** PubSubTask standard attribute: holds a participant (user) GUID */
+    public static final String ATTR_NAME__PARTICIPANT_GUID = "participantGuid";
```
evgeniipr, line 21: Put all common (used by all tasks) attributes to `PubSubTask`.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/api/PubSubTaskResult.java
```diff
@@ -15,10 +15,10 @@
      * Name of an attribute to be added to result message.
      * It contains ID of a PubSubTask-message (incoming message)
      */
-    public static final String ATTR_TASK_MESSAGE_ID = "taskMessageId";
+    public static final String ATTR_TASK__MESSAGE_ID = "taskMessageId";
 
-    public static final String FIELD_RESULT_TYPE = "resultType";
-    public static final String FIELD_ERROR_MESSAGE = "errorMessage";
+    public static final String PAYLOAD_FIELD__RESULT_TYPE = "resultType";
+    public static final String PAYLOAD_FIELD__ERROR_MESSAGE = "errorMessage";
```
evgeniipr, line 21: Renamed constants: make it more readable.

pepper-apis/src/main/java/org/broadinstitute/ddp/event/pubsubtask/impl/PubSubTaskProcessorFactoryImpl.java
```diff
@@ -18,5 +19,9 @@ protected void registerPubSubTaskProcessors() {
                 TASK_TYPE__UPDATE_PROFILE,
                 new UpdateProfileProcessor()
         );
+        registerPubSubTaskProcessor(
+                TASK_TYPE__USER_DELETE,
+                new UserDeleteProcessor()
+        );
```
evgeniipr, line 25: Register new PubSubTask type

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserService.java
```diff
@@ -1,4 +1,4 @@
-package org.broadinstitute.ddp.service;
+package org.broadinstitute.ddp.service.userdelete;
```
MocanaAtBroad: Don't think this is a good name for a package. I suggest we stick with original location.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // check if user refers to revisions
+        long[] revisionIds = handle.attach(JdbiRevision.class).findByUserId(user.getId());
+        if (revisionIds.length > 0) {
+            throw new DDPException(format(ERROR_PREFIX_USER_DELETE + "the user has references to a revision history", user.getGuid()));
+        }
+    }
+
+    private void deleteUserSteps(Handle handle, User user) throws IOException {
+        deleteAuth0User(handle, user);
```
MocanaAtBroad: Since this in Auth0, I would prefer that this be the last thing we remove: if the full database transaction removing the user succeeds then we can proceed to remove from Auth0

MocanaAtBroad:
I am confused about having a UserFullDeleteService and a UserService.deleteUser method. Why can't we just have one method?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
```diff
@@ -0,0 +1,147 @@
...
+        deleteKitRequests(handle, user);
+        deleteUserStudyLegacyData(handle, user);
+        deleteUserAnnouncement(handle, user);
+        deleteActivityInstanceStatus(handle, user);
+        deleteAnswersByOperator(handle, user);
```
MocanaAtBroad: Are we sure that we want to delete answers only by operator id and not  participant id?
I am thinking of case whether a parent signs up child and we need to delete the childs data.
Then earlier check of not having any governed users will pass (the child does not govern any users).
But the deltreAnswersByOperator is not going to find the child's data because child is not operator.
Am I missing something? does userService.deleteUser delete the child's data?

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
```diff
@@ -0,0 +1,147 @@
...
+
+        userService.deleteUser(handle, user);
```
ssettipalli: We need a step to delete the user from elastic indices. user, participant and participant_structured indices.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
ssettipalli: Data is associated with child. 
I don't think we can delete a parent/proxy user without deleting ALL dependents/governed-users.
We should check if current user has any governed users first and exit with response that user has governed users and cannot be deleted until ALL governed users are deleted first which is what I believe we are doing here.

May be we should go with the assumption that ALL governed users will be deleted one at a time before attempting to delete the proxy/parent/guardian.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserService.java
evgeniipr: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
evgeniipr: Put the auth0 deletion at the end.

evgeniipr:
> I am confused about having a UserFullDeleteService and a UserService.deleteUser method. Why can't we just have one method?

Initially I didn't wanted to touch the existing deletion method in UserService (which used by Rest endpoint "delete user"). 
Or you suggest to use only one service and add some parameter (in constructor) which defines a level of deletion (for example delete/not delete auth0)...? Seems this is good approach. I like it.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
evgeniipr: These ES deletion steps are implemented in UserDelete service (was implemented earlier by Yury Pavlov) and which UserFullDeleteService uses.. But I agree with @MocanaAtBroad that better to define only one service for both kind of deletions.. I need to reimplement this service.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
evgeniipr: > Are we sure that we want to delete answers only by operator id and not participant id?

@MocanaAtBroad this is according to one of deletion statements from a confluence page you referred in the ticket:
`delete from answer where operator_user_id in (select user_id from delete_users);`

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,367 @@
...
+ *  <li>delete queued_event;</li>
+ *  <li>delete user_governance;</li>
+ *  <li>delete age_up_candidate;</li>
+ *  <li>delete user;</li>
+ *  <li>delete ES data;</li>
+ *  <li>delete auth0;</li>
+ *  <li>send data-sync request.</li>
+ * </ul>
+ */
+public class UserDeleteService {
```
evgeniipr, line 68: Instead of 2 services (UserFullDeleteService and UserService) only one service for user deletion is created: UserDeleteService.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
evgeniipr: Re-implemented services: created only one service - UserDeleteService.

evgeniipr:
@MocanaAtBroad I re-implemented services: created only one service - UserDeleteService.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,367 @@
...
+
+    private LogInfo deleteUserSteps(Handle handle, User user, boolean fullDelete) throws IOException {
+        LogInfo logInfo = new LogInfo();
+        UserCollectedData userCollectedData = new UserCollectedData();
+
+        if (fullDelete) {
+            deleteKitRequests(handle, user, logInfo);
+            deleteUserStudyLegacyData(handle, user, logInfo);
+            deleteUserAnnouncement(handle, user, logInfo);
+            deleteActivityInstanceStatus(handle, user, logInfo);
```
ssettipalli: SimpleDelete doesn't have activityInstanceStatus ? 
User with Activity Instance will have activity Instance status, right !!

evgeniipr: activity_instance_status is deleted during simpleDelete too. 
This ` deleteActivityInstanceStatus(handle, user, logInfo);` is additional deletion by operator_id.
I renamed it to `deleteActivityInstanceStatusByOperator`.
In the manual delete instruction (https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/856260664/User+Delete) there are 2 rows of status deletion:
```
delete from activity_instance_status where activity_instance_id in (select activity_instance_id from activity_instance where participant_id in (select user_id from delete_users));delete from activity_instance_substitution where activity_instance_id in (select activity_instance_id from activity_instance where participant_id in (select user_id from delete_users));

delete from activity_instance_status where operator_id in (select user_id from delete_users);
```
The 1st one called during both simpleDelete and fullDelete.
The 2nd one - only during fullDelete (at least in the initial logic of simple delete this statement was not executed..)

Btw, seems I need to re-create the PR: I created this new class UserDeleteService not quite correctly: added it as a new class and deleted old class `UserService`. Better to rename `UserService` to `UserDeleteService` in order to preserve the history of changes...

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+ *  <li>delete queued_notification_template_substitution;</li>
+ *  <li>delete queued_notification;</li>
+ *  <li>delete queued_event;</li>
+ *  <li>delete user_governance;</li>
+ *  <li>delete age_up_candidate;</li>
+ *  <li>delete user;</li>
+ *  <li>delete ES data;</li>
+ *  <li>delete auth0;</li>
+ *  <li>send data-sync request.</li>
+ * </ul>
```
MocanaAtBroad: Let's drop the steps from comment. Hopefully the code is clear enough to illustrate what the steps are. And this method will probably have to be updated as schema evolves.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+
+    public UserDeleteService(RestHighLevelClient esClient) {
+        this.esClient = esClient;
+    }
+
+    /**
+     * Used to delete a user which still not registered in auth0-system, not governed by others than a user
+     * who deletes this user, not having status ENROLLED.
+     */
+    public void simpleDelete(Handle handle, User user, String whoDeleted, String comment) throws IOException {
```
MocanaAtBroad, line 97: A full javadoc comment would be useful here, particularly explaining what this arguments mean and do. e.g. whoDeleted: is this just a comment string or an identifier like a guid or an email? I can't tell.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        // check if user refers to revisions
+        long[] revisionIds = handle.attach(JdbiRevision.class).findByUserId(user.getId());
+        if (revisionIds.length > 0) {
+            throw new DDPException(format(ERROR_PREFIX_USER_DELETE + "the user has references to a revision history", user.getGuid()));
+        }
+    }
+
+    private LogInfo deleteUserSteps(Handle handle, User user, boolean fullDelete) throws IOException {
```
MocanaAtBroad: Would rename to simply "deleteUser". The word "Steps" is misleading: I originally thought we were deletinguser  steps from the database.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/userdelete/UserFullDeleteService.java
MocanaAtBroad: OK. Please verify this, but what I am seeing is that this is not necessary. We block deleting users that have governed users already. And, I believe, the only case where the participant id is going to be different from the operator id is case when THERE ARE governed users. You delete the participants data a little later.
I don't think this line does any harm, but makes me nervous having it there. If my analysis is correct (please check my logic) then let's drop the deletion by operator id.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+        LogInfo logInfo = new LogInfo();
+        UserCollectedData userCollectedData = new UserCollectedData();
+
+        if (fullDelete) {
+            deleteKitRequests(handle, user, logInfo);
+            deleteUserStudyLegacyData(handle, user, logInfo);
+            deleteUserAnnouncement(handle, user, logInfo);
+            deleteActivityInstanceStatusByOperator(handle, user, logInfo);
+            deleteAnswersByOperator(handle, user, logInfo);
```
MocanaAtBroad: Hmm. Made comment about not needing this deleteAnswersByOperator elsewhere. github can be confusing....

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        deleteUserProfile(handle, user, logInfo);
+        deleteMedicalProvider(handle, user, logInfo);
+        deleteEnrollmentStatuses(handle, user, logInfo);
+        deleteUserAddresses(handle, user, logInfo);
+        deleteAnswersAndActivityInstances(handle, user, logInfo);
```
MocanaAtBroad: Might as well rename to deletePartcipantAnswersAndActivityInstances

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,381 @@
...
+            deleteAuth0User(handle, user, logInfo);
+        }
+
+        dataSyncRequest(handle, userCollectedData);
+
+        return logInfo;
+    }
+
+    private void deleteKitRequests(Handle handle, User user, LogInfo logInfo) {
+        logInfo.add("kit_request");
```
MocanaAtBroad: Collecting the tables touched by deletion into a LogInfo object is an interesting idea, but I don't think it is particularly useful to the caller of the method. There is nothing that the caller can do with the information. I would rather write to the real log the operations as we do them and see how many rows of each kind have been removed. And if something goes wrong with a step of the deletion, then a developer with access to logs can see how far it got from the log. The LogInfo is not even going to be returned in case of failure.

MocanaAtBroad:
Could you double check that we are indeed getting everything? This is the list of columns that have a foreign key reference to the `user` table. At bottom of this comment is the query I used to generate this.
I am not so sure about table `file_upload` being properly deleted.


|TABLE_NAME                            |COLUMN_NAME          |
|--------------------------------------|---------------------|
|activity_instance                     |participant_id       |
|activity_instance_creation_mutex      |participant_id       |
|activity_instance_status              |operator_id          |
|age_up_candidate                      |participant_user_id  |
|answer                                |operator_user_id     |
|data_sync_request                     |user_id              |
|default_mailing_address               |participant_user_id  |
|event_configuration_occurrence_counter|participant_user_id  |
|file_upload                           |operator_user_id     |
|file_upload                           |participant_user_id  |
|invitation                            |user_id              |
|kit_request                           |participant_user_id  |
|kit_schedule_record                   |participant_user_id  |
|mailing_address                       |operator_user_id     |
|mailing_address                       |participant_user_id  |
|queued_event                          |operator_user_id     |
|queued_event                          |participant_user_id  |
|revision                              |changed_by_user_id   |
|revision                              |terminated_by_user_id|
|study_admin                           |user_id              |
|study_exit_request                    |user_id              |
|temp_mailing_address                  |operator_user_id     |
|temp_mailing_address                  |participant_user_id  |
|user_announcement                     |user_id              |
|user_governance                       |operator_user_id     |
|user_governance                       |participant_user_id  |
|user_legacy_info                      |participant_user_id  |
|user_medical_provider                 |user_id              |
|user_profile                          |user_id              |
|user_study_enrollment                 |user_id              |
|user_study_legacy_data                |user_id              |

```sql
SELECT DISTINCT
    TABLE_NAME, COLUMN_NAME
FROM
    INFORMATION_SCHEMA.KEY_COLUMN_USAGE
WHERE
    REFERENCED_TABLE_NAME = 'user'
ORDER BY TABLE_NAME , COLUMN_NAME;
```

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Removed comments.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Added comments for parameters.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Renamed

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Renamed

evgeniipr:
> Could you double check that we are indeed getting everything?

Actually not everything: I checked that still needs to delete from: activity_instance_creation_mutex, file_upload, invitation, kit_schedule_record, study_admin, study_exit_request, user_legacy_info. 
Added deletion.
I didn't added deletion of `revision` because seems we decided that if user has reference to revisions then we cannot delete such user.

evgeniipr:
> OK. Please verify this, but what I am seeing is that this is not necessary. We block deleting users that have governed users already. And, I believe, the only case where the participant id is going to be different from the operator id is case when THERE ARE governed users. You delete the participants data a little later.
I don't think this line does any harm, but makes me nervous having it there. If my analysis is correct (please check my logic) then let's drop the deletion by operator id.

I agree that this delete by operator is not needed: anyway all answers are deleted by activity_instance_id's. 
So, I removed this delete.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -0,0 +1,416 @@
...
+
+    private void deleteUserData(Handle handle, User user, boolean fullDelete) throws IOException {
+        UserCollectedData userCollectedData = new UserCollectedData();
+
+        if (fullDelete) {
+            deleteKitRequests(handle, user);
+            deleteUserStudyLegacyData(handle, user);
+            deleteUserAnnouncement(handle, user);
+            deleteActivityInstanceStatusByOperator(handle, user);
+            deleteFileUpload(handle, user);
```
MocanaAtBroad: I would classify fileupload deletion as part of removing activity data and therefore make it part of "normal" deletion.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Replaced this logging approach to separate log calls before each deletion step.

pepper-apis/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
evgeniipr: Moved to "normal" deletion.

evgeniipr:
@MocanaAtBroad , @ssettipalli I updated PubSub task "USER_DELETE" format: removed from payload a property "whoDeleted" and added to attributes an attribute "operatorGuid" (optional).

ssettipalli (approved):

PR #1046, opened 2021-10-20
Enable Google Cloud Logging for Auth0 registration rule

bskinner:
## Context

This PR enables the [@google-cloud/logging](https://www.npmjs.com/package/@google-cloud/logging) module in the Auth0 `Register User in Pepper` rule to try and cover a hole in our logging coverage. Entries logged to GCP will include a label named `source` with the value `auth0`, and will include the full context from the execution environment as the `jsonPayload`.

In order to enable this feature, the tenant must have a key named `googleApplicationCredentials` present in the global `configuration` object for the rules (this is managed from Auth0's `Rules` page). The value of the `googleApplicationCredentials` key should be a set of keys for a Google Service Account with the `Logs Writer` role.

The changes are currently visible when using the `ddp-dev`. The log statements can be viewed using the [Auth0 Registration Rule](https://cloudlogging.app.goo.gl/r6KtjU1QAHXF19vM9) query.

## Testing

## Release

The script update can be deployed using the normal deployment process with [auth0-deploy-cli](https://auth0.com/docs/deploy/deploy-cli-tool), but the addition configuration value must be manually added at this time.

### DIFF
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -196,7 +196,7 @@ function (user, context, callback) {
             if (cloudLoggingEnabled) {
                 var severity = "INFO";
 
-                if ( !!pepper_params.mode ) {
+                if (pepper_params.mode) {
                     if ( (pepper_params.mode === 'signup') && (!pepper_params.tempUserGuid) ) {
                         severity = "ERROR";
                     }
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -194,7 +194,7 @@ function (user, context, callback) {
             }
 
             if (cloudLoggingEnabled) {
-                var severity = "DEBUG";
+                var severity = "INFO";
 
                 if ( !!pepper_params.mode ) {
                     if ( (pepper_params.mode === 'signup') && (!pepper_params.tempUserGuid) ) {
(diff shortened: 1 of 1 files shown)


aleksey-kuznetsov:
Do we have a page describing how to setup Google Cloud Logging on per study basis? Or maybe we can utilise same project for logging in various studies?

Otherwise looks good! Let's see what other think.

alextanlan (dismissed):

aglushko-broad (dismissed):

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -176,6 +193,28 @@ function (user, context, callback) {
                 console.log('User metadata has last name = ' + pepper_params.lastName);
             }
 
+            if (cloudLoggingEnabled) {
+                var severity = "DEBUG";
```
MocanaAtBroad: I think INFO level would be OK. It would be the only thing we capture in GCP logs from Auth0.

bskinner:
In this case, logging is done based on which project the backend is running underneath. Auth0 doesn't have a concept of studies, just tenants and clients, so I think setting this as a per-study flag would be an impedance mismatch with Auth0's architecture. Adding a per-client configuration is very doable in the future, though.

> Do we have a page describing how to setup Google Cloud Logging on per study basis? Or maybe we can utilise same project for logging in various studies?
> 
> Otherwise looks good! Let's see what other think.

ssettipalli (approved):
looks good to me

PR #1049, opened 2021-11-01
Pancan new activities

YuriyPavlovBI:
There are 3 new activities added to the cancan study and some platform-level functionality needed for these activities to work:
https://broadinstitute.atlassian.net/browse/DDP-6957
https://broadinstitute.atlassian.net/browse/DDP-6958
https://broadinstitute.atlassian.net/browse/DDP-6959

- [x] https://github.com/broadinstitute/ddp-study-server/pull/1050 Added reevaluation of the block visibilities for the parent activity when child activity is being added/removed. (In this case, we need to hide "I don't have children" checkbox if at least one child is added) 

- [x] https://github.com/broadinstitute/ddp-study-server/pull/1051 New question type added: Activity Instance Select - allows user to select an instance of some configured list of activities for that particular question. In this case it is used to select which parent side some relative is on. Stored value is the activity instance guid, shown value is a rendered activity instance name. Options can be obtained using `/user/{user}/studies/{study}/activity-instance-select/{stableId}` endpoint.

- [x] https://github.com/broadinstitute/ddp-study-server/pull/1052 Support for pre-selected answers. New attribute is introduced to the pick list options: **_default_**. If question has at least one default option, the default answer will be automatically written right after activity instance creation. 

- [x] https://github.com/broadinstitute/ddp-study-server/pull/1059 Implement new matrix question type

- [x] https://github.com/broadinstitute/ddp-study-server/pull/1064 DDP-7085 Extend API to support disabling activity block


ssettipalli:
@ YuriyPavlovBI
I see that PR already has 141 files . I see that more changes will be included in this PR. Please try to create separate PR's for different tickets. That way we can keep the PR's to reasonable and reviewable size.

I see that you added a new Question Type : ActivityInstanceSelect 
We usually have a separate ticket and PR for core platform changes like new QuestionType and study related changes.
cc @MocanaAtBroad

MocanaAtBroad:
Hi Yuriy!
Looks like you have been extremely busy! There is a ton of stuff here and it is going to be very difficult to review it all.

I am going to request that you break this P.R. in chunks, starting with just the proposed API changes, including explanations on why the API changes are needed/how they address new requirements/etc.

After we have processed that P.R. we can take a look at the back-end and front-end code changes needed to implement the API changes. If you are adding a new question type it would be very useful to create an activity in basil study that shows the new question type in action. Doesn't have to be anything elaborate: just an activity that has one or more variations of the question type in action.

Finally, in separate PRs, let's consider the study configuration changes. If you could do separate PRs for each survey, that would be best as it limits the size of the chunk that needs to be inspected.

Reach out via Slack if you have questions. Thanks!

YuriyPavlovBI:
Sounds reasonable! I've created 2 PRs, one per feature. Let's keep it PR as a "parent" PR. Once "child" PRs worked out and merged this one will contain just a configuration changes and we will split them up by activities if needed.

study-builder/studies/pancan/family-history/family-history-top-level.conf
```diff
@@ -0,0 +1,484 @@
...
+            "isRestricted": false,
+            "isDeprecated": false,
+            "hideNumber": true,
+            "groups": [],
+            "validations": [],
+            "promptTemplate": {
+              "templateType": "HTML",
+              "templateText": "",
+              "variables": []
+            },
```
bskinner, line 162: What's going on here? It looks like this question is setup to have an empty prompt.

study-builder/studies/pancan/family-history/family-history-additional-details.conf
```diff
@@ -0,0 +1,84 @@
...
+          "question": {
+            include required("../../snippets/text-question.conf"),
+            "stableId": "OTHER_FACTORS_CANCER_RISK",
+            "inputType": "ESSAY",
+            "hideNumber": true,
+            "promptTemplate": {
+              "templateType": "TEXT",
+              "templateText": "",
+              "variables": []
+            },
```
bskinner: This seems like a similar situation to my [other comment](https://github.com/broadinstitute/ddp-study-server/pull/1049#discussion_r751290694). There is a question, which is not part of a `COMPOSITE`, which lacks a prompt.

study-builder/studies/pancan/family-history/family-history-top-level.conf
YuriyPavlovBI: That's correct! This block is just a single checkbox. The "I do not have any information..." checkbox here:
image

study-builder/studies/pancan/family-history/family-history-additional-details.conf
YuriyPavlovBI: This is also made on purpose. Previous blocks contains it all.
This actually is the following section:
image

study-builder/studies/pancan/family-history/family-history-top-level.conf
bskinner: In that case, the structure needs to be reworked. If the "I do not have any information..." text is intended to be presented with the checkbox, it should be included as part of the question's data. At the moment, there is no way to know these two elements should be presented together, and only happens to work because of the way the Angular client renders things.

study-builder/studies/pancan/family-history/family-history-additional-details.conf
bskinner: Same thing as my other comment, this design needs to be reworked. A `CONTENT` block may not be used as a replacement for presenting a question's prompt.

YuriyPavlovBI: Moved text to the prompt

aleksey-kuznetsov (dismissed):
Let's go!

aleksey-kuznetsov (approved):

PR #1050, opened 2021-11-02
Add blockVisibility for parentActivity blocks during creation and del…

YuriyPavlovBI:
Added reevaluation of the block visibilities for the parent activity when child activity is being added/removed. (use case: we need to hide "I don't have children" checkbox if at least one child is added)

### DIFF
--- a/pepper-apis/docs/specification/src/components/responses.yml
+++ b/pepper-apis/docs/specification/src/components/responses.yml
@@ -35,7 +35,8 @@ ActivityAnswersPatchResponse:
           blockVisibility:
             type: array
             description: list of updated block visibilities
-            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
+            items:
+              $ref: '../pepper.yml#/components/schemas/Block.Visibility'
           validationFailures:
             type: array
             items:
@@ -112,7 +113,8 @@ ActivityInstanceDeletionResponse:
           blockVisibility:
             type: array
             description: list of updated parent instance block visibilities
-            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
+            items:
+              $ref: '../pepper.yml#/components/schemas/Block.Visibility'
 ActivityInstanceNotFoundResponse:
   description: the requested activity instance for user and study does not exist
   content:
--- a/pepper-apis/docs/specification/src/components/responses.yml
+++ b/pepper-apis/docs/specification/src/components/responses.yml
@@ -35,18 +35,7 @@ ActivityAnswersPatchResponse:
           blockVisibility:
             type: array
             description: list of updated block visibilities
-            items:
-              type: object
-              required:
-                - blockGuid
-                - shown
-              properties:
-                blockGuid:
-                  type: string
-                  description: identifier for the block
-                shown:
-                  type: boolean
-                  description: whether block is shown or not
+            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
           validationFailures:
             type: array
             items:
@@ -111,6 +100,19 @@ ActivitySummaryResponse:
     application/json:
       schema:
         $ref: '../pepper.yml#/components/schemas/Activity.Summary'
+ActivityInstanceDeletionResponse:
+  description: The activity was successfully deleted and return list of updated parent instance block visibilities
+  content:
+    application/json:
+      schema:
+        type: object
+        required:
+          - blockVisibility
+        properties:
+          blockVisibility:
+            type: array
+            description: list of updated parent instance block visibilities
+            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
 ActivityInstanceNotFoundResponse:
   description: the requested activity instance for user and study does not exist
   content:
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -756,18 +756,17 @@ Block.Content:
         body:
           type: string
 Block.Visibility:
-  items:
-    type: object
-    required:
-      - blockGuid
-      - shown
-    properties:
-      blockGuid:
-        type: string
-        description: identifier for the block
-      shown:
-        type: boolean
-        description: whether block is shown or not
+  type: object
+  required:
+    - blockGuid
+    - shown
+  properties:
+    blockGuid:
+      type: string
+      description: identifier for the block
+    shown:
+      type: boolean
+      description: whether block is shown or not
 #####
 #
 # Components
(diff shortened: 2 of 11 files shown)


pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -35,18 +35,7 @@ ActivityAnswersPatchResponse:
...
-                - blockGuid
-                - shown
-              properties:
-                blockGuid:
-                  type: string
-                  description: identifier for the block
-                shown:
-                  type: boolean
-                  description: whether block is shown or not
+            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
```
bskinner: This is an improperly defined array type. An array's items may take a `$ref`, but it should be in an `items` map.

Either review the [array](https://swagger.io/docs/specification/data-models/data-types/#array) data type, or look at one of the other examples of an array definition in this specification.

pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -111,6 +100,19 @@ ActivitySummaryResponse:
...
+    application/json:
+      schema:
+        type: object
+        required:
+          - blockVisibility
+        properties:
+          blockVisibility:
+            type: array
+            description: list of updated parent instance block visibilities
+            $ref: '../pepper.yml#/components/schemas/Block.Visibility'
```
bskinner: This is an improperly defined array.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -755,6 +755,19 @@ Block.Content:
           type: string
         body:
           type: string
+Block.Visibility:
+  items:
```
bskinner: This is an improperly defined [Schema](https://swagger.io/specification/#schema-object) object. The `items` key belongs in the array definition, not here.

pepper-apis/docs/specification/src/endpoints/user.studies.activities.yml
```diff
@@ -61,17 +61,22 @@ post:
...
                 - instanceGuid
+                - blockVisibility
               properties:
                 instanceGuid:
                   type: string
                   example: ABCDEFGH12
+                blockVisibility:
+                  type: array
+                  description: list of updated parent instance block visibilities
+                  $ref: '../pepper.yml#/components/schemas/Block.Visibility'
```
bskinner: This is an improperly defined array.

bskinner (dismissed):
Still doing a deeper review of the code, but I identified some issues in the OpenAPI spec changes.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
```diff
@@ -92,4 +105,36 @@ public static Answer getAnswer(Handle handle, long instanceId, String questionSt
                 .orElseThrow(() -> new DaoException(format(
                         "Error to detect answer: question stableId=%s, instanceId=%d", questionStableId, instanceId)));
     }
+
+    public static Optional<List<BlockVisibility>> getBlockVisibility(Handle handle, Response response,
```
MocanaAtBroad: For what I am seeing, in all the API changes the array BlockVisibilities is always required/not nullable property. Returning optional seems like it complicates the caller codes with conditionals. Might as well simply return an list every time, even if it is just empty.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
```diff
@@ -92,4 +105,36 @@ public static Answer getAnswer(Handle handle, long instanceId, String questionSt
...
+                                                                     String parentInstanceGuid, User participantUser, StudyDto studyDto,
```
MocanaAtBroad: Would it make sense to rename this argument to activityInstanceGuid? Just that, in the context of family history we do have activity instances that contain activity instances, so the `parent` prefix makes me think about that and don't think that is the intent at all.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
```diff
@@ -92,4 +105,36 @@ public static Answer getAnswer(Handle handle, long instanceId, String questionSt
...
+        if (parentInstanceDto != null) {
+
+            ActivityDefStore activityStore = ActivityDefStore.getInstance();
+            FormActivityDef parentActivity = ActivityInstanceUtil
+                    .getActivityDef(handle, activityStore, parentInstanceDto, studyDto.getGuid());
+
+            UserActivityInstanceSummary instanceSummary = RouteUtil.findUserActivityInstanceSummaryOrHalt(
+                    response, handle, participantUser.getGuid(), studyDto.getGuid(), parentInstanceGuid, isStudyAdmin);
+            result = formService.getBlockVisibilities(handle, instanceSummary, parentActivity,
+                    participantUser.getGuid(), operatorGuid, parentInstanceGuid);
```
MocanaAtBroad: Please verify, but it looks like the 3rd argument to `getBlockVisibilities` method ultimately is not used for anything in FormActivityService. If this is true, please refactor the argument out of the affected methods.

YuriyPavlovBI: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
YuriyPavlovBI: Fixed

pepper-apis/src/main/java/org/broadinstitute/ddp/util/QuestionUtil.java
YuriyPavlovBI: Fixed

pepper-apis/docs/specification/src/endpoints/user.studies.activities.yml
YuriyPavlovBI: Fixed

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Fixed

pepper-apis/docs/specification/src/components/responses.yml
YuriyPavlovBI: Fixed

pepper-apis/docs/specification/src/components/responses.yml
YuriyPavlovBI: Fixed

MocanaAtBroad (approved):

PR #1051, opened 2021-11-02
Activity instance select question type

YuriyPavlovBI:
New question type added: Activity Instance Select - allows user to select an instance of some configured list of activities for that particular question. In this case it is used to select which parent side some relative is on. Stored value is the activity instance guid, shown value is a rendered activity instance name. Options can be obtained using /user/{user}/studies/{study}/activity-instance-select/{stableId} endpoint.
Getting the options via endpoint is introduced to handle dynamic nature of what we select. For regular picklists we define the options once in the beginning. However, activity instance list is dynamic.  

https://user-images.githubusercontent.com/71830706/139815718-9add1f61-677d-4585-82e8-4b02d6659a6d.mov

Configuration: 
image


pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
```diff
@@ -222,6 +225,9 @@ private void collectQuestionMappings(Map<String, Object> props, Map<String, Obje
             case TEXT:
                 currProps.putAll(textFmt.mappings((TextQuestionDef) questionDef));
                 break;
+            case ACTIVITY_INSTANCE_SELECT:
```
bskinner, line 230: I don't see an OpenAPI spec update for this question type.

bskinner (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityResponseCollector.java
YuriyPavlovBI: Updated
image

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -311,6 +311,7 @@ Answer:
   discriminator:
     propertyName: type
     mapping:
+      ACTIVITY_INSTANCE_SELECT: '#/Answer.String'
```
bskinner: What is the actual data type here supposed to be? While the raw type is a string, there's no note of what the value is expected to represent.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1429,6 +1432,28 @@ Question:
...
+          type: string
+          default: ACTIVITY_INSTANCE_SELECT
+          enum:
+            - ACTIVITY_INSTANCE_SELECT
+        activityCodes:
+          description: 'List of activity codes for instance lookup'
+          type: array
+          items:
+            type: string
+        answers:
```
bskinner, line 1482: This needs a `description` field detailing what a client should be sending back as the `string`-valued answer.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1429,6 +1432,28 @@ Question:
...
+        questionType:
+          type: string
+          default: ACTIVITY_INSTANCE_SELECT
+          enum:
+            - ACTIVITY_INSTANCE_SELECT
+        activityCodes:
+          description: 'List of activity codes for instance lookup'
+          type: array
+          items:
+            type: string
```
bskinner, line 1480:
```suggestion
            type: string
            format: guid
```

bskinner (dismissed):
The code looks good so far, I'd still like to see @MocanaAtBroad's opinion on this feature. One thing I noticed is that there doesn't seem to be any checks preventing clients from creating an `ACTIVITY_INSTANCE_SELECT` answer with an arbitrary string- I didn't see any checks to make sure that the answer being saved corresponds to a participant's existing & accessible Activity Instance. Is this understanding accurate?

ssettipalli (commented):
@YuriyPavlovBI :
Since the answer saved for this question is activityInstanceGUID, want to check the behavior of this new question type:
1. CSV export: During the CSV export, if we populate the answer from DB (guid) it does not make much sense to the user/study staff. May be rendered activity instance name would make things clear
2. Elastic export. Depending on what DSM prefers or intends to do we might need to export guid or rendered instance name for the users. You might want to discuss this with Simone ahead of time

YuriyPavlovBI:
I have added the logic to exporter

YuriyPavlovBI:
> @YuriyPavlovBI : Since the answer saved for this question is activityInstanceGUID, want to check the behavior of this new question type:
> 
> 1. CSV export: During the CSV export, if we populate the answer from DB (guid) it does not make much sense to the user/study staff. May be rendered activity instance name would make things clear
> 2. Elastic export. Depending on what DSM prefers or intends to do we might need to export guid or rendered instance name for the users. You might want to discuss this with Simone ahead of time

I have added the logic to exporter

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Added

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/ActivityInstanceSelectQuestionFormatStrategy.java
```diff
@@ -38,13 +63,36 @@
...
+                        .filter(i -> i.getActivityInstanceGuid().equals(answer.getValue())).collect(Collectors.toList());
+                if (CollectionUtils.isNotEmpty(summaries)) {
+                    Map<String, FormResponse> responses = activityInstanceService.countQuestionsAndAnswers(
+                            handle, user.getGuid(), user.getGuid(), studyGuid, summaries);
+                    activityInstanceService.renderInstanceSummaries(handle, user.getId(), user.getGuid(), studyGuid, summaries, responses);
+                    return summaries.get(0).getActivityName();
+                }
+                return answer.getValue();
+            });
+        } else {
```
ssettipalli, line 92: @YuriyPavlovBI 
studyGuid will not be null. Trying to understand if userId will ever be null and If it does , what are the scenarios where userId can be null ? 
In these cases (else), are we using the answer guid?

YuriyPavlovBI: We use an answer value in these cases (in other words, selected instance GUID). The reason why I made these checks is that  the nulls still could be passed at some point of time even though we don't expect that. So my intention was to still maintain a minimal functionality instead of failing the whole export.

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Docs have been updated
image
image

YuriyPavlovBI:
I have created a simple usage example in basil as @MocanaAtBroad suggested

bskinner:
Any updates on this question, @YuriyPavlovBI? If the answer is not a valid GUID, the activity instance does not exist, or the user does have the privileges to access it, let's prevent an answer from being created for it.

>  I noticed is that there doesn't seem to be any checks preventing clients from creating an `ACTIVITY_INSTANCE_SELECT` answer with an arbitrary string- I didn't see any checks to make sure that the answer being saved corresponds to a participant's existing & accessible Activity Instance. Is this understanding accurate?

YuriyPavlovBI:
> Any updates on this question, @YuriyPavlovBI? If the answer is not a valid GUID, the activity instance does not exist, or the user does have the privileges to access it, let's prevent an answer from being created for it.
> 
> > I noticed is that there doesn't seem to be any checks preventing clients from creating an `ACTIVITY_INSTANCE_SELECT` answer with an arbitrary string- I didn't see any checks to make sure that the answer being saved corresponds to a participant's existing & accessible Activity Instance. Is this understanding accurate?

Oh, I thought I notified about it - it has been taken care of here https://github.com/broadinstitute/ddp-study-server/pull/1051/commits/6c53de8ffb98dc3d0b0cffc16a43d68b0c08eefc

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -500,16 +501,27 @@ private TextAnswer convertTextAnswer(String stableId, String guid, String actIns
...
         if (value != null && value.isJsonPrimitive() && value.getAsJsonPrimitive().isString()) {
             String textValue = value.getAsJsonPrimitive().getAsString();
-            return new ActivityInstanceSelectAnswer(null, stableId, guid, textValue, actInstanceGuid);
+            var jdbiQuestion = new JdbiQuestionCached(handle);
+            Set<String> activityCodes = new HashSet<>(jdbiQuestion
+                    .getActivityCodesByActivityInstanceSelectQuestionId(questionDto.getId()));
+            var instanceSummaries = handle.attach(org.broadinstitute.ddp.db.dao.ActivityInstanceDao.class)
+                    .findSortedInstanceSummaries(userGuid, studyGuid, activityCodes);
+            if (instanceSummaries.stream().noneMatch(summary -> summary.getGuid().equals(textValue))) {
+                throw ResponseUtil.haltError(response, 400, new ApiError(ErrorCodes.NOT_FOUND,
```
YuriyPavlovBI: @bskinner

bskinner: This should return a `404`. A `400` HTTP status code is `Bad Request`- there's nothing wrong with this request.

YuriyPavlovBI: @bskinner Changed to 404

ssettipalli:
Overall changes look ok to me :+1:

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1010,6 +1070,35 @@ default void insertQuestion(long activityId, TextQuestionDef textQuestion, long
...
+     * @param revisionId      the revision to use, will be shared by all created data
+     */
+    default void insertQuestion(long activityId, ActivityInstanceSelectQuestionDef activityInstanceSelectQuestion,
+                                long revisionId) {
+        insertBaseQuestion(activityId, activityInstanceSelectQuestion, revisionId);
+
+        int numInserted = getJdbiActivityInstanceSelectQuestion().insert(activityInstanceSelectQuestion.getQuestionId());
+
+        if (numInserted != 1) {
+            throw new DaoException("Inserted " + numInserted + " for activity instance select question "
```
bskinner: Let's expand on the error message here. The current message doesn't give a clear indication of what when wrong, just that _n_ questions were inserted and it was abnormal somehow.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1010,6 +1070,35 @@ default void insertQuestion(long activityId, TextQuestionDef textQuestion, long
...
+                    + activityInstanceSelectQuestion.getStableId());
+        }
+
+        if (CollectionUtils.isNotEmpty(activityInstanceSelectQuestion.getActivityCodes())) {
+            int[] ids = getJdbiActivityInstanceSelectActivityCodes().insert(activityInstanceSelectQuestion.getQuestionId(),
+                    activityInstanceSelectQuestion.getActivityCodes(),
+                    Stream.iterate(0, i -> i + DISPLAY_ORDER_GAP).iterator());
+            if (ids.length != activityInstanceSelectQuestion.getActivityCodes().size()) {
+                throw new DaoException("Inserted " + numInserted + " questions for "
```
bskinner: Same thing here as above- the exception should explicitly state the reason why it was thrown.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2578,3 +2632,17 @@ ParticipantsLookupProxyItem:
...
       format: email
+Suggestion.ActivityInstance:
+  type: object
+  required:
+    - guid
+    - name
+  properties:
+    guid:
+      description: Activity Instance GUID
+      type: string
```
bskinner, line 2643:
```suggestion
      type: string
      format: guid
```

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/AuthPathRegexUtil.java
```diff
@@ -43,6 +43,9 @@
     private static final Pattern CANCER_SUGGESTION_ROUTE_REGEX = Pattern.compile(
             BASE_REGEX + "\\/studies\\/" + GUID_PATTERN + "/suggestions/cancers" + "\\/?");
 
+    private static final Pattern DYNAMIC_ANSWER_BASED_ON_QUESTION_ROUTE_REGEX = Pattern.compile(
```
bskinner: This is the only place I'm seeing the path `/user/{user}/studies/{study}/dynamic-question/{stableId}`. Is this the correct path? In the docs above I see `/user/{user}/studies/{study}/activity-instance-select/{stableId}`

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/ActivityInstanceSelectAnswer.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return value == null || value.isEmpty();
```
bskinner: `value` is marked with `@NotNull` when it's declared.

pepper-apis/src/main/java/org/broadinstitute/ddp/pex/TreeWalkInterpreter.java
```diff
@@ -729,6 +735,36 @@ private Object applyChildTextAnswerPredicate(InterpreterContext ictx, PredicateC
...
+    private Object applyActivityInstanceSelectAnswerPredicate(InterpreterContext ictx, PredicateContext predicateCtx,
+                                                              String userGuid, long studyId, String activityCode,
+                                                              String instanceGuid, String stableId) {
+        if (predicateCtx instanceof PexParser.HasTextPredicateContext) {
+            String value = StringUtils.isBlank(instanceGuid)
+                    ? fetcher.findLatestActivityInstanceAnswer(ictx, userGuid, activityCode, stableId, studyId)
+                    : fetcher.findSpecificActivityInstanceSelectAnswer(ictx, activityCode, instanceGuid, stableId);
+            return StringUtils.isNotBlank(value);
+        } else if (predicateCtx instanceof PexParser.ValueQueryContext) {
+            throw new PexUnsupportedException("Getting Activity Instance Select answer value is not supported");
```
bskinner: Are the details about the `study`, `activityCode` and `stableId` easily identified in the logs when this error occurs? If not, let's extend the error message to include them.

pepper-apis/src/main/java/org/broadinstitute/ddp/pex/TreeWalkInterpreter.java
```diff
@@ -729,6 +735,36 @@ private Object applyChildTextAnswerPredicate(InterpreterContext ictx, PredicateC
...
+    }
+
+    private Object applyChildActivityInstanceSelectAnswerPredicate(InterpreterContext ictx, PredicateContext predicateCtx,
+                                                                   String userGuid, List<Answer> childAnswers) {
+        if (predicateCtx instanceof PexParser.HasTextPredicateContext) {
+            return childAnswers.stream()
+                    .map(child -> ((ActivityInstanceSelectAnswer) child).getValue())
+                    .anyMatch(StringUtils::isNotBlank);
+        } else if (predicateCtx instanceof PexParser.ValueQueryContext) {
+            throw new PexUnsupportedException("Getting Activity Instance Select answer value is not supported");
```
bskinner: Are the details about the `study`, `activityCode` and `stableId` easily identified in the logs when this error occurs? If not, let's extend the error message to include them.

bskinner (commented):
A couple of comments to address, but looks good otherwise.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/ActivityInstanceSelectAnswer.java
YuriyPavlovBI: Yep, but we have to implement this method. See other classes implementing `Answerable` interface

pepper-apis/src/main/java/org/broadinstitute/ddp/pex/TreeWalkInterpreter.java
YuriyPavlovBI: Extended

pepper-apis/src/main/java/org/broadinstitute/ddp/pex/TreeWalkInterpreter.java
YuriyPavlovBI: Extended

pepper-apis/src/main/java/org/broadinstitute/ddp/filter/AuthPathRegexUtil.java
YuriyPavlovBI: This is a leftover that was removed in another branch. Removed here as well

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
YuriyPavlovBI: Changed the wording

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
YuriyPavlovBI: Changed the wording

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/ActivityInstanceSelectAnswer.java
bskinner: I'm not certain we're on the same page. I understand the `boolean isEmpty()` method needs to be implemented in order to conform to the `Answerable` interface. My concern here is we're doing a `null` check on **value** but, when it's declared, it's annotated with an explicit `@NotNull` contract. If we're performing `null` checks, we're potentially masking bugs which would arise when that contract is being broken, or we're misleadingly stating the property will not contain a `null` value when in reality, it may.

I'm neutral on whether the field should have a `NotNull` contract or not, but I feel we should be consistent one way or the other.

YuriyPavlovBI: Ok, right. I checked the logic that uses this method and set returned value to be always `false`.

YuriyPavlovBI:
@bskinner @MocanaAtBroad Any concerns?

MocanaAtBroad:
I think it looks good! Thanks for the basil examples and the video! Helps understand what is going on.

MocanaAtBroad (approved):

ssettipalli (approved):

PR #1052, opened 2021-11-02
DDP-7074 Support for pre-selected answers

YuriyPavlovBI:
https://broadinstitute.atlassian.net/browse/DDP-7074 
Support for pre-selected answers. New attribute is introduced to the pick list option: **_default_**. If question has at least one default option, the default answer will be automatically written right after activity instance creation.


pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/PicklistOption.java
```diff
@@ -37,6 +37,9 @@
     @SerializedName("exclusive")
     private boolean isExclusive;
 
+    @SerializedName("isDefault")
```
bskinner: The API documentation needs to be added for this change if this is a new property being exposed to clients.

bskinner (changes requested):

bskinner:
In the description for the PR, you mention:

> If question has at least one default option, the default answer will be automatically written right after activity instance creation.

Who is doing the creation here? My understanding after looking over the code is that the backend is going to be taking care of creating the default set of answers. If that is the case, why does `isDefault` need to be exposed to clients? If this is not the case, what is the expected flow after a client receives the Activity response?

YuriyPavlovBI:
> In the description for the PR, you mention:
> 
> > If question has at least one default option, the default answer will be automatically written right after activity instance creation.
> 
> Who is doing the creation here? My understanding after looking over the code is that the backend is going to be taking care of creating the default set of answers. If that is the case, why does `isDefault` need to be exposed to clients? If this is not the case, what is the expected flow after a client receives the Activity response?

Yes, you're right, there is no need to expose `isDefault` since backend takes care of it. Removed

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/PicklistOption.java
YuriyPavlovBI: Removed

bskinner: Per [this comment](https://github.com/broadinstitute/ddp-study-server/pull/1052#issuecomment-960968833), `isDefault` is being removed from the client response so this conversation is moot.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ActivityInstanceUtil.java
```diff
@@ -147,4 +161,19 @@ public static boolean computeCanDelete(boolean canDeleteInstance, Boolean canDel
...
+                (questionDef instanceof PicklistQuestionDef) && !((PicklistQuestionDef) questionDef).getDefaultOptions().isEmpty());
+        for (QuestionDef question : questions) {
+            PicklistQuestionDef plQuestion = (PicklistQuestionDef) question;
+            var answer = new PicklistAnswer(null, plQuestion.getStableId(), null,
+                    plQuestion.getDefaultOptions().stream()
+                            .map(plOpt -> new SelectedPicklistOption(plOpt.getStableId()))
+                            .collect(Collectors.toList()));
+            answerDao.createAnswer(operatorId, instanceId, answer, plQuestion);
+        }
+    }
```
bskinner, line 178: Does this handle the case where an option has `isExclusive` set, or when there's a mismatch between the number of default options and any exclusivity or validation requirements?

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
```diff
@@ -499,23 +499,21 @@ default GroupAndOptionDtos findOrderedGroupAndOptionDtos(long questionId, long t
...
-                if (nestedOptMap.containsKey(parentDto.getId())) {
-                    parentDto.getNestedOptions().addAll(nestedOptMap.get(parentDto.getId()));
-                }
-            });
-        });
+        dtosMap.values().forEach(dto -> dto.getUngroupedOptions().forEach(parentDto -> {
+            if (nestedOptMap.containsKey(parentDto.getId())) {
+                parentDto.getNestedOptions().addAll(nestedOptMap.get(parentDto.getId()));
+            }
+        }));
```
bskinner: This looks like an unnecessary change. Unless there is an issue such as a linter failing, let's revert this as the changes make it more difficult to read.

pepper-apis/src/main/java/org/broadinstitute/ddp/util/ActivityInstanceUtil.java
YuriyPavlovBI: No, there is no any checks given the fact that it is a study design time feature in comparison to the runtime stuff like `isExclusive` and validations

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
YuriyPavlovBI: This change was just for clearing up all the warnings. I changed the shape back to make it looks like before

YuriyPavlovBI: This change was just for clearing up all the warnings. I changed the shape back to make it looks like before

bskinner (approved):
👍

PR #1055, opened 2021-11-03
DDP-7112 user registration/login rule fix

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7112

rule is used in ALL Auth0 tenants and hence need to test registration/login for ALL studies

### DIFF
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -257,7 +257,7 @@ function (user, context, callback) {
                     // all is well
                     var ddpUserGuid = body.ddpUserGuid;
                     user.app_metadata.user_guid = ddpUserGuid;
-
+                    //todo looks like updating on every re-login .. check and avoid the update when not needed
                     auth0.users.updateAppMetadata(user.user_id, user.app_metadata)
                         .then(function(){
                             context.idToken[pepperUserGuidClaim] = ddpUserGuid;
--- a/study-builder/tenants/register-user-in-pepper.js
+++ b/study-builder/tenants/register-user-in-pepper.js
@@ -121,7 +121,7 @@ function (user, context, callback) {
         if (
             !pepper_params.tempUserGuid &&
             !user.app_metadata.user_guid &&
-            !!user.user_metadata.temp_user_guid
+            user.user_metadata && !!user.user_metadata.temp_user_guid
         ) {
             pepper_params.tempUserGuid = user.user_metadata.temp_user_guid;
         }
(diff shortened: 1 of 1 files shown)


study-builder/tenants/register-user-in-pepper.js
```diff
@@ -121,7 +121,7 @@ function (user, context, callback) {
         if (
             !pepper_params.tempUserGuid &&
             !user.app_metadata.user_guid &&
-            !!user.user_metadata.temp_user_guid
+            user.user_metadata && !!user.user_metadata.temp_user_guid
```
ssettipalli, line 124: Bug fix to handle scenarios where has NO user_metada 
Added user.user_metadata &&  to explicitly check if user has user_metada to avoid: 
"TypeError: Cannot read property 'temp_user_guid' of undefined", 

Looks like !!user.user_metadata.temp_user_guid doesn't handle this scenario.

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -136,6 +136,20 @@ function (user, context, callback) {
             console.log('No Registration Mode passed in request');
         }
 
+        if (pepper_params.mode && pepper_params.mode === "login" &&
```
ssettipalli, line 139: During login check user_guid or temp_user_guid to check that user was registered before trying to login.
Block any further progress and return back if user was NOT registered

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -136,6 +136,20 @@ function (user, context, callback) {
...
+            !user.app_metadata.user_guid &&
+            !pepper_params.tempUserGuid
```
MocanaAtBroad, line 141: Could you double-check that we want && connecting everything here? Is it possible that we need is:
```javascript
if (pepper_params.mode && pepper_params.mode === "login" && (!user.app_metadata.user_guid ||
            !pepper_params.tempUserGuid)
```

ssettipalli: During login if it doesn't contain both
user.app_metadata.user_guid or pepper_params.tempUserGuid, then we consider it as NOT registered. 

Your expressions will be true even if one of them doesn't exist !

MocanaAtBroad (approved):

study-builder/tenants/register-user-in-pepper.js
bskinner: That's expected if `user_metadata` is not defined. The `!!` can only coerce an `undefined` or `null` value, and  `temp_user_guid` is neither if `user_metadata` is not defined.

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -136,6 +136,20 @@ function (user, context, callback) {
...
+        ) {
+            console.log('User not registered');
+            const loginErrPayload = {
+                code: 'user_not_registered',
+                message: 'User need to register first in order to login',
+                statusCode: 422,
```
bskinner, line 147: A `403 Forbidden` would fit better here than a `422 Unprocessable Entity`. There's nothing wrong with the request, the user is just not permitted to access the resource- regardless of their authentication state.

study-builder/tenants/register-user-in-pepper.js
```diff
@@ -243,7 +257,7 @@ function (user, context, callback) {
                     // all is well
                     var ddpUserGuid = body.ddpUserGuid;
                     user.app_metadata.user_guid = ddpUserGuid;
-
+                    //todo looks like updating on every re-login .. check and avoid the update when not needed
```
bskinner, line 260: The update itself is not required, but injecting the guid into the token claims is required for every run of this rule (see line 263 below). I don't feel adding more logic here would buy us anything.

PR #1058, opened 2021-11-08
DDP-5545 Brugada: Switch on Study Manager and Enable Kit Shipment

YuriyPavlovBI:
DDP-5545 Brugada: Switch on Study Manager and Enable Kit Shipment
@ssettipalli it's a patch for prod

### DIFF
--- a/study-builder/studies/brugada/patches/addKitConfigurationAndUpdateEnrolledEvent.conf
+++ b/study-builder/studies/brugada/patches/addKitConfigurationAndUpdateEnrolledEvent.conf
@@ -1,0 +1,30 @@
+"events": [
+   {
+      "trigger": {
+        "type": "ACTIVITY_STATUS",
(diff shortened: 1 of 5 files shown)


study-builder/studies/brugada/patches/addKitConfigurationAndUpdateEnrolledEvent.conf
```diff
@@ -0,0 +1,30 @@
+"events": [
```
ssettipalli, line 1: we might want to add these events to study-events.conf since we are not adding the patch to patch.log

YuriyPavlovBI: It's already set up and ready to work the same way either from scratch or after patch application.

ssettipalli (approved):

PR #1059, opened 2021-11-10
Matrix question type

YuriyPavlovBI:
Here's the new question type: MATRIX (DDP-7069)
It allows us to create grid-like questions:
image

Supports:
- multiple and single choices for row
- column grouping
- exclusive options (inside any group)
- usage in PEX using _.children[]_ notation
- REQUIRED validation

Tests and Doc update are on the way


bskinner:
Where is the OpenAPI proposal for this change?

YuriyPavlovBI:
> Where is the OpenAPI proposal for this change?

Working on this

YuriyPavlovBI:
> Where is the OpenAPI proposal for this change?

Doc update and tests are in place now!

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
```diff
@@ -0,0 +1,179 @@
...
+        </createTable>
+    </changeSet>
+
+    <changeSet author="ypavlov" id="matrix-question-table">
+        <createTable tableName="matrix_question">
+            <column name="question_id" type="bigint">
+                <constraints primaryKey="true" primaryKeyName="matrix_question_pk"
+                             references="question(question_id)" foreignKeyName="matrix_question_question_fk"/>
+            </column>
+            <column name="matrix_select_mode_id" type="bigint">
```
ssettipalli, line 24: Wondering if select mode : SINGLE is exactly same as option: IS_EXCLUSIVE if the select mode scope is entire question and NOT option in a ROW !!

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
```diff
@@ -0,0 +1,179 @@
...
+                <constraints nullable="false" references="matrix_select_mode(matrix_select_mode_id)"
+                             foreignKeyName="matrix_question_matrix_select_mode_fk"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="ypavlov" id="matrix-group-table">
+        <createTable tableName="matrix_group">
```
ssettipalli, line 32: @YuriyPavlovBI :
In the context of attached picture: Is HOW_OFTEN one matrix question and AMOUNT another matrix question and both grouped as MATRIX_GROUP 
OR
both HOW_OFTEN and AMOUNT is oneMATRIX question.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMatrixQuestion.java
```diff
@@ -0,0 +1,21 @@
...
+import org.broadinstitute.ddp.model.activity.types.MatrixSelectMode;
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.stringtemplate4.UseStringTemplateSqlLocator;
+
+public interface JdbiMatrixQuestion extends SqlObject {
+
+    @UseStringTemplateSqlLocator
+    @SqlUpdate("insertMatrixByModeIds")
```
ssettipalli: Name seem to be little misleading. Single insert with single ModeId. 
Its a simple insert, may be add the insert sql directly here rather than UseStringTemplateSqlLocator and adding insert sql in diff file

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMatrixQuestion.java
```diff
@@ -0,0 +1,21 @@
...
+    int insert(@Bind("questionId") long questionId,
+               @Bind("selectModeId") long selectModeId);
+
+    @UseStringTemplateSqlLocator
+    @SqlUpdate("insertMatrixByModeCodes")
```
ssettipalli: May be add the insert sql directly here rather than UseStringTemplateSqlLocator and adding insert sql in diff file

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
```diff
@@ -0,0 +1,180 @@
...
+            </column>
+            <column name="revision_id" type="bigint">
+                <constraints nullable="false" references="revision(revision_id)" foreignKeyName="matrix_option_revision_fk"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="ypavlov" id="matrix-option-uk">
+        <addUniqueConstraint tableName="matrix_option"
+                             columnNames="matrix_question_id, matrix_option_stable_id, revision_id"
```
MocanaAtBroad, line 100: If we are going to make the the option stable id unique for the whole question (nothing wrong with that!) then do we really need to have to include the optionStableId AND the groupStableId in the answer being PATCHed back by client? i.e., wouldn't be sufficient for a matrix answers to just have rowStableId and optionStableId (and drop groupStableId)?

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
```diff
@@ -0,0 +1,180 @@
...
+                <constraints nullable="false" references="matrix_row(matrix_row_id)"
+                             foreignKeyName="matrix_answer_row_fk"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="ypavlov" id="matrix-option-selection-uk">
+        <addUniqueConstraint tableName="matrix_answer" columnNames="answer_id, matrix_option_id, matrix_row_id"
+                             constraintName="matrix_answer_uk"/>
+    </changeSet>
```
MocanaAtBroad, line 163: OK. The answer table makes sense. I have been looking at the OpenAPI definition....
Screen Shot 2021-11-29 at 1 40 57 PM
:

bskinner:
The example given isn't a distinct question, it's asking 3 questions per row (Have you eaten this?, How much?, and How often?), and it looks like it's grouped into two categories ( the `first` and `second` groups). This is a confusing arrangement, and something that the show/hide block logic would be far more applicable to. Is the example question you provided representative of the types of questions being implemented in the survey?

> Here's the new question type: MATRIX ([DDP-7069](https://broadinstitute.atlassian.net/browse/DDP-7069)) It allows us to create grid-like questions: image
> 
> Supports:
> 
>     * multiple and single choices for row
> 
>     * column grouping
> 
>     * exclusive options (inside any group)
> 
>     * usage in PEX using _.children[]_ notation
> 
>     * REQUIRED validation
> 
> 
> Tests and Doc update are on the way

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1302,6 +1343,39 @@ PicklistOption:
...
+    groupId:
+      type: string
+      nullable: true
+    optionLabel:
+      type: string
+    tooltip:
+      type: string
+      description: tooltip for the option, in plain text
+    exclusive:
+      type: boolean
```
bskinner, line 1396: Let's add a description of what sort of behavior can be expected when the `exclusive` flag is `true`

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/MatrixQuestionFormatStrategy.java
```diff
@@ -0,0 +1,104 @@
...
+
+            groups.add(groupDef);
+        }
+        props.put("groups", groups);
+
+        List<Object> options = new ArrayList<>();
+        for (MatrixOptionDef optionDef : definition.getOptions()) {
+            Map<String, Object> stableIdTxt = new HashMap<>();
+            stableIdTxt.put("optionStableId", optionDef.getStableId());
+            stableIdTxt.put("optionText", HtmlConverter.getPlainText(optionDef.getOptionLabelTemplate().renderWithDefaultValues("en")));
```
bskinner, line 67: @ssettipalli What is the expected behavior when exporting fields which may contain HTML fragments? I'm not sure if the call to `HtmlConverter.getPlainText(String)` here is in-line with the other export methods.

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
YuriyPavlovBI: Just like in PICKLIST question, we can use SINGLE (radio buttons) and MULTIPLE (checkboxes) with IS_EXCLUSIVE. However, there is a difference other than look & feel: radio button cannot be cleared.

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
YuriyPavlovBI: It is the same matrix question with some groups. You can check the configuration of what on the picture in basil.

YuriyPavlovBI:
> The example given isn't a distinct question, it's asking 3 questions per row (Have you eaten this?, How much?, and How often?), and it looks like it's grouped into two categories ( the `first` and `second` groups). This is a confusing arrangement, and something that the show/hide block logic would be far more applicable to. Is the example question you provided representative of the types of questions being implemented in the survey?

Here's the real question needed for CMI:
image

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/MatrixQuestionDef.java
```diff
@@ -0,0 +1,163 @@
...
+
+public final class MatrixQuestionDef extends QuestionDef {
+
+    @NotNull
+    @SerializedName("selectMode")
+    private MatrixSelectMode selectMode;
+
+    @NotNull
+    @SerializedName("groups")
+    private List<@Valid @NotNull MatrixGroupDef> groups = new ArrayList<>();
```
bskinner: I've seen the generic type parameter annotated with `@NotNull` a couple times- what effect does that have?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1713,6 +1788,53 @@ Question.Picklist:
...
+        questionType:
+          type: string
+          default: MATRIX
+          enum:
+            - MATRIX
+        selectMode:
+          type: string
+          enum:
+            - MULTIPLE
+            - SINGLE
```
bskinner: What is the difference in expected behavior between the `MULTIPLE` and `SINGLE` select modes?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1713,6 +1788,53 @@ Question.Picklist:
...
+          items:
+            $ref: '#/MatrixOption'
+        matrixQuestions:
+          type: array
+          items:
+            $ref: '#/MatrixRowQuestion'
+        answers:
+          type: array
+          items:
+            $ref: '#/Answer.Picklist'
```
bskinner: Should this be referencing `#/Answer.Matrix`?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1713,6 +1788,53 @@ Question.Picklist:
...
+            properties:
+              identifier:
+                type: string
+              name:
+                type: string
+        matrixOptions:
+          type: array
+          items:
+            $ref: '#/MatrixOption'
+        matrixQuestions:
```
bskinner: Why does `matrixOptions` and `matrixQuestions` have a `matrix` prefix, while `groups` does not?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1713,6 +1788,53 @@ Question.Picklist:
...
+          type: array
+          items:
+            $ref: '#/MatrixRowQuestion'
```
bskinner: What is the `matrixQuestions` property and `MatrixRowQuestion` object used for? I don't see any references to a question identifier in other parts of the `Question.Matrix` object, and the `MatrixRowQuestion` doesn't refer to any `groups` or `matrixOptions`

bskinner (changes requested):

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: MULTIPLE allows to select as many columns in one row as you want (checkbox), SINGLE is just radio buttons.

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: We kept it as it is for all the question types, they all have `answers`

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: In Matrix question, `matrixQuestions` basically are rows, when `matrixOptions` are columns.

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Right! Fixed

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Let's document that please.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/MatrixQuestionDef.java
YuriyPavlovBI: It means that the list could not contain null items

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Sure! Done

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/MatrixQuestionDef.java
bskinner: Thank you!

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMatrixQuestion.java
YuriyPavlovBI: Inlined both queries

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMatrixQuestion.java
YuriyPavlovBI: Inlined both queries

study-builder/studies/basil/i18n/en.conf
```diff
@@ -11,6 +11,151 @@
...
+        "fruit": {
+          "prompt": "How often do you eat fruits?"
+        }
+      }
+    },
+    "matrix_multi_mode_w_groups": {
+      "prompt": "Matrix Question: Multiple Mode w/ Groups"
+      "groups": {
+        "first": {
+          "prompt": "HOW OFTEN"
```
bskinner, line 116: Is using `prompt` accurate here? I don't see any references to `prompt` when working with the `MatrixGroupDef`, `JdbiMatrixGroup`, or `MatrixGroup` types.

YuriyPavlovBI: It's just a variable name. The actual config is here:
image

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Oh, sorry, I mixed `groups` and `answers`
Added prefix

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1720,6 +1795,56 @@ Question.Picklist:
...
+          description: A list of groups used in the `matrixOptions` values. If there are no options which explicity request grouping, this will be an empty array.
+          minItems: 0
+          items:
+            type: object
+            required:
+              - identifier
+              - name
+            properties:
+              identifier:
+                type: string
```
bskinner, line 1833: If the API allows ungrouped options (options with a `groupId` set to `null`) to be mixed with grouped options, let's either allow for a `null` identifier in the `MatrixGroup` type, or require a non-`null` `groupId` in the `MatrixOption` type. The current contract allows for a set of ungrouped options which are unable to have a prompt set for them.

bskinner, line 1827: Is your intent here to prevent ungrouped (`MatrixOption` objects with a null `groupId`) from having a prompt?

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: I'd prefer going the other way and removing the prefix. The `matrix` prefix feels unnecessary as the properties are part of the `Matrix` question type- it's repetitive.

bskinner:
Based on the question you've given, and the API design you've proposed, is the below an accurate interpretation of how the example question would be presented via the API?

question-breakdown


> > The example given isn't a distinct question, it's asking 3 questions per row (Have you eaten this?, How much?, and How often?), and it looks like it's grouped into two categories ( the `first` and `second` groups). This is a confusing arrangement, and something that the show/hide block logic would be far more applicable to. Is the example question you provided representative of the types of questions being implemented in the survey?
> 
> Here's the real question needed for CMI: image

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: OK!

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Kind of, if option has no group, there will be no additional header over it

YuriyPavlovBI:
That's correct

> Based on the question you've given, and the API design you've proposed, is the below an accurate interpretation of how the example question would be presented via the API?
> 
> question-breakdown
>

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
YuriyPavlovBI: You're right, it seems redundant so I removed it

pepper-apis/src/main/resources/db-changes/schema/DDP-7069-matrix-question-type.xml
YuriyPavlovBI: You're right, it seems redundant so I removed it

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Added

pepper-apis/src/main/java/org/broadinstitute/ddp/export/collectors/MatrixQuestionFormatStrategy.java
YuriyPavlovBI: As I can see, it's a common practice. You can find similar code in strategies for other question types

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1339,6 +1375,40 @@ PicklistOption:
...
+      type: string
+      nullable: true
+    optionLabel:
+      type: string
+    tooltip:
+      type: string
+      description: tooltip for the option, in plain text
+    exclusive:
+      type: boolean
+      description: if `exclusive` option selected, other options will be cleared (applicable for MULTIPLE select mode)
```
bskinner: Does this de-selection cross group boundaries, or does this only impact selections within a single group?

YuriyPavlovBI: Within a group, of course. Updated description

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: Is there a reason we're special casing the ungrouped options like that? If not, I'd prefer that we treat everything the same- how difficult a change would it be to require that every option belong to a group _or_ allow a header to be declared for the `null` group?

YuriyPavlovBI: Well, most of the times in CMI there's no group usage at all (51 vs 1). It's just a concept that allow us to separate some columns and it does the thing. It's a whole lot of options to choose: without groups (null group only), null group + some groups, groups only. I believe it's quite flexible.

YuriyPavlovBI: Also, in the given example, there's no need for additional header over the `null` group. If we were in need of additional header, I'd just declare a group for it and that's it.
image

bskinner: > It's a whole lot of options to choose

Yes, that is my point. It's flexibility at the cost of client complexity, and doesn't seem be flexibility that is needed at this point. If an implementer wants to use groups, or needs the features provided by them, everything should be in a group. Otherwise, they can leave the `groupId` set to `null` and not opt-in to the extra behaviors _or_ everything is grouped from the start, and the client doesn't have to worry about it.

YuriyPavlovBI: I'm not sure I get what the client complexity is. `null` group is simply treated as a group without any additional header. @aleksey-kuznetsov you have written some code for that on the SDK side. What are your thoughts on the above?

aleksey-kuznetsov: From the UI perspective every possible case is covered and I don't see any complexities since ungrouped options (whose `groupId` is set to `null`) are actually a part of `null` group. So technically each and every option is a part of a group and there is no case like `ungrouped options` + `grouped options`.

aleksey-kuznetsov: And correct, for `null` group there is no header text and it cannot be configured. If one is needed -- just group these options in a named group and set a header for it.

bskinner: Using a `null` group id as short-hand for "This group doesn't have a header" is not a logical grouping of options- it is a UI treatment leaking into the API structure.

For me to feel comfortable approving this PR, I would like the following changes:

* `MatrixGroup.name` should be `nullable`
* If any groups are used, _all_ options should belong to a declared group
* If any options have a `null` `groupId`, all options should have a `null` group id.

@MocanaAtBroad or @ssettipalli , if you feel these changes are unnecessary, please dismiss my review and approve this PR at your discretion.

YuriyPavlovBI: I see it from different angle. As I said, most of the times there shouldn't be any groups at all (51 question VS 1 question with groups for CMI). However, sometimes we need additional headers AND separate selection within some set of options and this is achieved by these groups. Most of the times there shouldn't be any groups at all.
> Using a null group id as short-hand for "This group doesn't have a header"

YuriyPavlovBI: In other words, using group is short-hand for "This group should have a separate selection even if select mode is SINGLE, and it should have additional header so that user see that this set of options is separate from the rest"

aleksey-kuznetsov: - `MatrixGroup.name` should be `nullable`
With this change there will be only one usage of a group -- if we need an option to be `exclusive` among some specific options and not all of them. Since groups are not highlighted in any way (unless you style it on per study basis) it can be confusing without a name of a group to identify what options belong where. + what Yuriy mentioned when `selectMode` is `SINGLE`.

As for the points 2 and 3 a `groupId` of `null` is a shorthand for "This option is in unnamed group" so it not only doesn't have a header but all other specifics like `exclusive` flag will work within this unnamed group.

bskinner: >With this change there will be only one usage of a group -- if we need an option to be exclusive among some specific options and not all of them. Since groups are not highlighted in any way (unless you style it on per study basis) it can be confusing without a name of a group to identify what options belong where.

That's fine. If the UI is confusing, that is an issue with the UI, and not something the API should be dealing with. This is fine-tuning for one specific implementation. A group should be a logical collection of options with a set of defined behaviors for the options within that group.

If an implementer chooses to add 20 groups, all without headers, that's up to the implementer and it is not the API's place to enforce limitations on it. The header is a decoration element, not a logical one.

If there are issues visually presenting or distinguishing between groups, that is in the purview of the client implementation, and the API is not the place to address this. From my understanding, we're doing a great deal of per-study styling, and study-specific appearance tweaks wouldn't be out of the ordinary.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1339,6 +1375,40 @@ PicklistOption:
...
+    - stableId
+    - optionLabel
+    - exclusive
+    - groupId
+  properties:
+    stableId:
+      type: string
+    groupId:
+      type: string
+      nullable: false
```
bskinner:
```suggestion
```

`nullable` is `false` by default so we can remove this line.

bskinner (dismissed):
Minor tweak, but looks good otherwise. Thank you for your work in getting the changes into the PR! Let's get this in once the tests pass.

YuriyPavlovBI:
@bskinner All set up!

bskinner (approved):
Looks like this has drifted out of sync with `develop`- I'll keep an eye out and ✅ once everything's back up.

PR #1064, opened 2021-11-15
Ddp 7085 disabling activity block

YuriyPavlovBI:
DDP-7085 Extend API to support disabling activity block

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -587,6 +587,7 @@ Block:
     - blockType
     - blockGuid
     - shown
+    - enabled
   properties:
     blockType:
       type: string
@@ -594,6 +595,8 @@ Block:
       type: string
       format: guid
     shown:
+      type: boolean
+    enabled:
       type: boolean
 Block.Component:
   allOf:
@@ -760,6 +763,7 @@ Block.Visibility:
   required:
     - blockGuid
     - shown
+    - enabled
   properties:
     blockGuid:
       type: string
@@ -767,6 +771,9 @@ Block.Visibility:
     shown:
       type: boolean
       description: whether block is shown or not
+    enabled:
+      type: boolean
+      description: whether block is enabled or not
 #####
 #
 # Components
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -773,7 +773,7 @@ Block.Visibility:
       description: whether block is shown or not
     enabled:
       type: boolean
-      description: whether block is enabled or not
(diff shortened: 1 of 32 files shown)


ssettipalli:
wondering if this is something that can be handled from UI as one of cases. If checkbox checked , disable the edit button and avoid the API changes
@MocanaAtBroad

bskinner:
> wondering if this is something that can be handled from UI as one of cases. If checkbox checked , disable the edit button and avoid the API changes @MocanaAtBroad

How would a front-end client know _that specific checkbox_ is special?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -760,13 +763,17 @@ Block.Visibility:
...
   properties:
     blockGuid:
       type: string
       description: identifier for the block
     shown:
       type: boolean
       description: whether block is shown or not
+    enabled:
+      type: boolean
+      description: whether block is enabled or not
```
bskinner: What does `enabled` mean for a block? Adding a description of what changes with regards to the blocks behavior would be helpful.

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiBlockEnabledExpression.java
```diff
@@ -0,0 +1,34 @@
...
+    long insert(long blockId, long expressionId, long revisionId);
+
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryEnabledExprTextByBlockIdAndRevision")
+    Optional<String> getExpressionText(@Bind("blockId") long blockId,
+                                       @Bind("instanceGuid") String instanceGuid);
+
+    @SqlQuery("select * from block_enabled_expression as be"
+            + " join revision as rev on rev.revision_id = be.revision_id"
+            + " where be.block_id = :blockId and rev.end_date is null")
```
bskinner, line 28: Will there be cases where `end_date` is set to a date in the future, but not `null`?

ssettipalli (dismissed):
looks ok to me

bskinner (dismissed):

pepper-apis/src/main/java/org/broadinstitute/ddp/db/dao/JdbiBlockEnabledExpression.java
YuriyPavlovBI: I can imagine the situation you mentioned but for some reason it is made the same way for enabled expressions (see JdbiBlockExpression) so we did the same here. Revisions switch at the time we update the study. So, it seems ok for expressions.

pepper-apis/docs/specification/src/components/schemas.yml
YuriyPavlovBI: Added short description

ssettipalli (approved):

PR #1068, opened 2021-11-17
DDP-5634 Brugada: Configure New Medical Release Instance

YuriyPavlovBI:
DDP-5634 Brugada: Configure New Medical Release Instance


study-builder/src/main/java/org/broadinstitute/ddp/studybuilder/task/BrugadaUpdateActivityBaseSettings.java
```diff
@@ -0,0 +1,120 @@
...
+import org.broadinstitute.ddp.studybuilder.ActivityBuilder;
+import org.broadinstitute.ddp.util.ConfigUtil;
+import org.jdbi.v3.core.Handle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.nio.file.Path;
+
+public class BrugadaUpdateActivityBaseSettings implements CustomTask {
```
ssettipalli: Is the patch just to update 
"allowOndemandTrigger": true, in activity: MEDICAL_RECORD_RELEASE

We can just run the generic patch: UpdateActivityBaseSettings and that should update this allowOndemandTrigger value !

YuriyPavlovBI: You're right. Removed redundant files

ssettipalli (approved):

PR #1069, opened 2021-11-22
Readme file updated according to the last project prerequisites setup

Trimax:


pepper-apis/README.md
```diff
@@ -6,14 +6,18 @@ You'll need a few things to get up and running for local development:
 
 * Java 11 - see the [docs here](docs/java-11.md)
 * Maven - for building the project
-* Python - for running various scripts if needed
+* Python - for running various scripts if needed (latest 2.x.x version)
 * Ruby - for rendering configuration files from templates
```
MocanaAtBroad, line 10: Can you double check this against CloudSDK installation docs? Python for many years to get people to stop using 2.x and to use 3.x (there are code incompatibilities).

pepper-apis/README.md
```diff
@@ -6,14 +6,18 @@ You'll need a few things to get up and running for local development:
...
 * Google Cloud SDK - for running pubsub locally
 * Docker - for building images and optionally for running the test suite
 * Vault - for secrets management
 
 ### Setting up Google Cloud SDK
 
+Before installation of google components please make sure you have Python 2.x.x
+installed and environment variable CLOUDSDK_PYTHON is pointing to the python.exe
+(for example CLOUDSDK_PYTHON=c:\python\python.exe)
```
MocanaAtBroad, line 19: This looks very Windows-specific. Again please double check cloud sdk docs and if setting up this env variable is needed, do explain that it is for Windows (I don't have it in my MacOS machine but gcloud runs fine for me).

bskinner: The Cloud SDK [docs](https://cloud.google.com/sdk/docs/install) explicitly state a version of Python 3 between 3.5 and 3.8 is recommended.

pepper-apis/README.md
bskinner: The Cloud SDK [docs](https://cloud.google.com/sdk/docs/install) explicitly state a version of Python 3 between 3.5 and 3.8 is recommended.

Trimax: That's correct. But it didn't work with Python 3 on Windows

pepper-apis/README.md
Trimax: Will recheck this on Mac. If it works, then we can decline this PR

pepper-apis/README.md
Trimax: I believe the reason behind that is that MacOS has Python 2.x installed by default and Python 3.x is being installed during gcloud tools setup. So there are two python's: 2.x & 3.x. It could be checked by running python -version & python3 -version

Windows, on the other hand, has no Python distribution installed by default at all. That's why I had to install 2.x

PR #1085, opened 2021-12-10
DDP-7343-update-ddp-automation-log4j-version

Rinchan20:
security update just in case, I don't explicitly use log4j but it seems a dependency does, posting result of dependency tree:

```
mvn dependency:tree | grep log4j
[INFO] |  |  \- org.slf4j:slf4j-log4j12:jar:2.0.0-alpha5:compile
[INFO] |  +- log4j:log4j:jar:1.2.16:compile
[INFO]    |  |  +- org.apache.logging.log4j:log4j-api:jar:2.15.0:compile
```

### DIFF
--- a/ddp-automation/pom.xml
+++ b/ddp-automation/pom.xml
@@ -93,7 +93,7 @@
             <dependency>
                 <groupId>org.apache.logging.log4j</groupId>
                 <artifactId>log4j-api</artifactId>
-                <version>2.15.0</version>
+                <version>2.16.0</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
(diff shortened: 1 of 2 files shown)


Rinchan20:
tests are a bit old (but on way to being updated in a different branch)...import line taken out since class was changed since then (I think changed from JdbiAnswer to JdbiCompositeAnswer)

bskinner (dismissed):
Looks good to me

MocanaAtBroad (dismissed):

Rinchan20:
```
mvn dependency:tree | grep log4j
[INFO] |  |  \- org.slf4j:slf4j-log4j12:jar:2.0.0-alpha5:compile
[INFO] |  +- log4j:log4j:jar:1.2.16:compile
[INFO]    |  |  +- org.apache.logging.log4j:log4j-api:jar:2.16.0:compile
```

bskinner (approved):
👍

PR #1095, opened 2022-01-03
SendGrid dependencies update

ssettipalli:
Upgrade SendGrid dependencies to latest jars

### DIFF
--- a/pepper-apis/parent-pom.xml
+++ b/pepper-apis/parent-pom.xml
@@ -498,6 +498,7 @@
         <dependencies>
             <!-- jackson libraries below meant to upgrade  transitive dependencies    -->
             <!-- from sendgrid.com::sendgrid-java for infosec                         -->
+            <!-- shooting for jackson-databind version higher than 2.9.9.2            -->
             <!-- core, databind and annotations are maintained and compiled together, -->
             <!-- so updating all 3.                                                   -->
             <dependency>
--- a/pepper-apis/parent-pom.xml
+++ b/pepper-apis/parent-pom.xml
@@ -496,26 +496,6 @@
 
     <dependencyManagement>
         <dependencies>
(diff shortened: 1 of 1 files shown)


pepper-apis/parent-pom.xml
```diff
@@ -504,17 +504,17 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
-                <version>2.9.10.4</version>
```
ssettipalli, line 507: I believe these below 3 jackson transient dependencies can be removed since SendGrid jar 4.8.1 uses 2.12.1 !!
Doesn't hurt to leave them there though
@MocanaAtBroad

Rinchan20 (dismissed):

pepper-apis/parent-pom.xml
MocanaAtBroad: IMO If we are not using Jackson in our code directly, let's just remove them. If transient, then let sendgrid pull in what it needs.

ssettipalli: Looks like there were added for AppSec reasons. With updated jar, AppSec  check looks fine. 
removed the dependencies.

MocanaAtBroad: OK. In the diff I did not notice the comment explaining the reason we had the jackson libraries in the pom.
I think it would be safer to specify the minimum version that we want to be using. This would be to defend ourselves of the possibility of adding a new library that that pulls in a version of jackson that has the vulnerability we are worried about.
The proper way to do this in maven is to use the `dependencyManagement` section in Maven (note what we did to handle the log4j vulnerability fixes).
You can add the specific versions in dependencyManagement and (maybe? I am not 100% sure but look into it) you might be able to specify a minimum version number, which is really what we want to do for Jackson.

Rinchan20 (approved):

PR #1098, opened 2022-01-05
SendGrid event logging

ssettipalli:
Additional logging for SendGrid webhook event issue investigation


pepper-apis/src/main/java/org/broadinstitute/ddp/service/SendGridEventService.java
```diff
@@ -42,9 +42,8 @@ public void logSendGridEvents(SendGridEvent[] sendGridEvents) {
                         event.getAttempt(),
                         event.getSmtpId());
             } else {
-                LOG.info("SendGrid-EVENT[{}]: email={}, timestamp={}, sg_event_id={}, reason={}, status={}",
```
ssettipalli, line 45: email is sensitive info and we are displaying in logs hence removed it.

MocanaAtBroad (approved):

PR #1105, opened 2022-01-11
Pancan new activities bugfixes

YuriyPavlovBI:


study-builder/studies/pancan/study-events.conf
```diff
@@ -995,7 +995,7 @@
         "type": "ACTIVITY_INSTANCE_CREATION",
         "activityCode": ${id.act.stool_kit}
       },
-      "preconditionExpr": """!user.studies["cmi-pancan"].forms["STOOL_KIT"].hasInstance()""",
+      "preconditionExpr": ${_pex.has_colorectal}"""&& !user.studies["cmi-pancan"].forms["STOOL_KIT"].hasInstance()""",
```
grigol-brd, line 998: I understand how PEX expression and HOCON work separately, but this combined one looks like something new.

Why isn't `${_pex.has_colorectal}` inside string? Like this
`"""${_pex.has_colorectal} && !user.studies["cmi-pancan"].forms["STOOL_KIT"].hasInstance()"""`

Can you tell me what's the difference?
How is this combined expression evaluated? PEX is evaluated first and then HOCON or vice versa?

grigol-brd (approved):

study-builder/studies/pancan/study-events.conf
YuriyPavlovBI: That's how it works.
HOCON is the way to describe the config in the filesystem. HOCON substitutions work at config load time. PEX is being evaluated in runtime, when needed

grigol-brd: So even though it looks like the first part of the logical expression is outside of the string, the whole final value would get merged into a single logical expression and not separate strings?

PR #1110, opened 2022-01-14
DDP-7292 Add showActivityStatus to activity definition

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7292
New column will be used by DSM

Once we get confirmation if we want to set the new flag for existing studies, we might need to back fill/update the new column as needed.
Might need to tweak little bit later once we configure FON depending on DSM need

### DIFF
--- a/pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ActivityDef.java
+++ b/pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ActivityDef.java
@@ -360,7 +360,7 @@ public abstract class ActivityDef {
         return canDeleteFirstInstance;
     }
 
-    public boolean isShowActivityStatus() {
+    public boolean showActivityStatus() {
         return showActivityStatus;
     }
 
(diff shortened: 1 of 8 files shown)


pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ActivityDef.java
```diff
@@ -357,6 +360,10 @@ public Boolean getCanDeleteFirstInstance() {
         return canDeleteFirstInstance;
     }
 
+    public boolean isShowActivityStatus() {
```
bskinner: This reads strangely- `showActivityStatus` is already a true/false statement, prefixing the name with `is` isn't necessary

bskinner (dismissed):
Overall, this looks good! I did leave a comment about the naming of the getter for `showActivityStatus`.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/ActivityDef.java
ssettipalli: renamed

bskinner (approved):
👍

PR #1117, opened 2022-01-20
Update DSS API docs with Decimal types [DDP-7460]

bskinner:
## Context

This PR updates the API specification to include API definitions for the upcoming Decimal question and answer types. A definition for the Integer question type was also included as it was missing from the spec.

* Check out the associated branch
* `cd` into `pepper-apis/docs/specification`
* Run `npm run start` and follow the instructions to view the generated documentation locally

## Testing

## Release


josemanuel167 (dismissed):

ssettipalli (dismissed):

alextanlan:
LGTM

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -6,7 +6,7 @@ servers:
     description: Production server (uses real data)
 info:
   description: 'Pepper API specification'
-  version: '1.16.1-863fa496'
+  version: 'v1.25-46-gcd81179ef'
```
josemanuel167, line 9: is the version a random string?

bskinner: Nope. There isn't an overarching versioning scheme for the API, so this is the output from a git describe call prior to the commit. It's not precise, but close enough for our needs.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -345,17 +327,25 @@ Answer.Date:
     - $ref: '#/Answer'
     - type: object
       required:
+        - type
         - value
       properties:
+        type:
+          type: string
+          default: DATE
```
josemanuel167, line 335: does it make sense to have a type and a default when there is only a possible option? Could this be written something like this  `{...properties: { type : DATE}, {...properties: { type : COMPOSITE}`?

bskinner: Yes and no, respectively. These particular values are for display purposes in ReDoc, and to explicitly state what the value is for each object- the type isn't DATE, the type is `string` with a default value of the string `"DATE"`. The type field only allows JSON types (so complex types `array` and `object`, and primitive types `integer`, `boolean`, `number`, and `string`). Second, these values are used by the client to determine the type of the object being returned. It's a form of polymorphism which works in tandem with the discriminator properties you'll also see in the spec.

josemanuel167: I see it looks good on the docs
image
That double type on code got my attention

alextanlan (approved):

PR #1129, opened 2022-02-01
DDP-7306. Decimal question type

Trimax:
This PR contains the code that introduces a new type of question - the DECIMAL question. All (I assume) necessary classes were added to support the new type.

This is the second of three PRs to be raised.
The last one will contain the changes to question table with adding formatting stuff: how many digits after comma should be visible / requested


pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -631,6 +636,15 @@ private NumericAnswer convertNumericAnswer(Handle handle, NumericQuestionDto num
         }
     }
 
+    private DecimalAnswer convertDecimalAnswer(Handle handle, DecimalQuestionDto numericDto, String guid, String actInstanceGuid,
+                                               JsonElement value) {
+        BigDecimal decimalValue = null;
+        if (value != null && !value.isJsonNull()) {
+            decimalValue = value.getAsBigDecimal();
+        }
```
bskinner: BigDecimal is not a primitive type- this is taking advantage of a Java-specific API. What is the expected primitive type for the JsonElement here?

pepper-apis/src/test/java/org/broadinstitute/ddp/util/GsonBigDecimalDeserializationTest.java
```diff
@@ -0,0 +1,36 @@
...
+        final String s = "{\"stableId\":\"test\",\"answerGuid\":\"test\",\"value\":{\"unscaledValue\":10,\"scale\":2}}";
+
+        final AnswerSubmission answerSubmission = new Gson().fromJson(s, AnswerSubmission.class);
+        final JsonObject value = answerSubmission.getValue().getAsJsonObject();
+        final BigDecimal decimalValue = new BigDecimal(value.get("unscaledValue").getAsBigInteger(),
+                value.get("scale").getAsInt());
+
+        assertEquals("test", answerSubmission.getAnswerGuid());
+        assertEquals(0, BigDecimal.valueOf(0.1).compareTo(decimalValue));
+    }
```
bskinner, line 24: What is the purpose of this test? Nowhere else in this PR is the property name "unscaledValue" referenced.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/DecimalQuestion.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.function.Consumer;
+
+public final class DecimalQuestion extends Question<DecimalAnswer> {
+    @SerializedName("placeholderText")
+    private String placeholderText;
+
+    private transient Long placeholderTemplateId;
```
bskinner, line 18: I don't see any properties for defining the precision or scale of the expected answer value in the class definition. Is this what you were referring to in your description with "...and formatting thing"?

bskinner (changes requested):

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
Trimax: I know. Going to change the parsing to accept
`{"value":XXX,"scale":YYY}`
in following PR.

pepper-apis/src/test/java/org/broadinstitute/ddp/util/GsonBigDecimalDeserializationTest.java
Trimax: This is how it will work in the following PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/DecimalQuestion.java
Trimax: Yes, I wasn't clear in my PR description.
So, I was going to put in 3rd PR the following things:
1. Extend the question table by adding formatting stuff: how many digits after comma should be visible / requested
2. Deserialization BigDecimal using `{"value":XXX,"scale":YYY}` JSON structure
3. Extending Basil study by adding Decimal question/answer question section there

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
bskinner: Is this going to include updates to the serialization behavior, as well?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
```diff
@@ -0,0 +1,36 @@
+package org.broadinstitute.ddp.model.activity.instance.answer;
+
+import com.google.gson.annotations.SerializedName;
+import org.broadinstitute.ddp.model.activity.types.QuestionType;
+
+import java.math.BigDecimal;
+
+public final class DecimalAnswer extends Answer<BigDecimal> {
+    @SerializedName("value")
+    private BigDecimal value;
```
bskinner: What type is this going to be serialized to, or is this going to be updated in your next PR with the deserialization updates?

pepper-apis/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
Trimax: Yes

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
Trimax: Will be replaced with DecimalDef

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
```diff
@@ -1865,7 +1865,7 @@ public void testPatch_fileAnswer_createAndUpdate() {
 
         assertEquals(guid, nextGuid);
         answer = (FileAnswer) TransactionWrapper.withTxn(handle ->
-                new AnswerCachedDao(handle).findAnswerByGuid(guid).get());
+                new AnswerCachedDao(handle).findAnswerByGuid(guid).orElse(null));
```
bskinner, line 1867: There's a number of changes of this type which are unrelated to the current PR. What is the problem you're trying to work around here? This doesn't fix or guard around the NPE, it just moves it a line down. Given the context of the test, this _should_ crash if `.get()` returns a `null`. If you would like to make this safer, I would recommend doing something similar to the below:
```java
var answer = (Optional<FileAnswer>) TransactionWrapper.withTxn(handle -> new AnswerCachedDao(handle).findAnswerByGuid(guid));

assertTrue(answer.isPresent());
assertEquals(upload2.getFileName(), answer.get().getValue().getFilename());
```

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
```diff
@@ -2167,14 +2167,6 @@ private ActivityInstanceStatusDto queryStatus(String instanceGuid) {
         });
     }
 
-    /**
```
bskinner, line 2022: Has this code been causing problems? If not, this code is unrelated to the current PR- what is the purpose of the reorganization?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/validation/DecimalRangeRuleDef.java
```diff
@@ -0,0 +1,44 @@
...
+    private DecimalDef max;
+
+    public DecimalRangeRuleDef(Template hintTemplate, DecimalDef min, DecimalDef max) {
+        super(RuleType.DECIMAL_RANGE, hintTemplate);
+        if (min != null && max != null && min.compareTo(max) > 0) {
+            throw new IllegalArgumentException("Must provide a valid range");
+        }
+        this.min = min;
+        this.max = max;
+    }
```
MocanaAtBroad, line 26: Not seeing a DECIMAL_RANGE in the current ReDoc docs. @bskinner

MocanaAtBroad: And this is a question: FOR THE PURPOSES OF THE API: Does it make sense to have a a separate range rule type or decimals and integers? i.e., isn't the client simply going to compare whatever answer was entered with the number included in the rule definition?
More specifically: would it make sense to just re-implement the existing rule range so it can deal with any numeric type?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public void setValue(DecimalDef value) {
+        this.value = value;
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return value == null;
+    }
```
MocanaAtBroad, line 37: The answer definition differs from what we have in ReDoc:
Screen Shot 2022-02-07 at 12 24 57 PM
@bskinner @alextanlan  Does it make sense to have those additional attributes (precision and scale) in the answer? (i.e., wouldn't the client just send back whatever the user entered and just echo back the precision and scale values to make the API happy).
The specific question: can we drop precision and scale from the answer attributes in the API?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/DecimalRangeRule.java
```diff
@@ -0,0 +1,59 @@
...
+import org.broadinstitute.ddp.model.activity.instance.answer.DecimalAnswer;
+import org.broadinstitute.ddp.model.activity.instance.question.Question;
+import org.broadinstitute.ddp.model.activity.types.RuleType;
+
+import java.math.BigDecimal;
+
+/**
+ * A validation rule that checks numeric integer value is within an optional min/max range, inclusive.
+ */
+public class DecimalRangeRule extends Rule<DecimalAnswer> {
```
MocanaAtBroad, line 13: Same question as above about having separate range rules. I can understand why we might have separate Java implementations (I think) but not sure it buys us anything in the API.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
bskinner: Precision is purely for display & validation purposes and only needs to be present in the question  definition- it's the server's responsibility to validate that and, as such, isn't necessary in the answer (and isn't present in the current docs & implementation). As for the scale- that's an intrinsic part of the BigDecimal type, and the number can not be represented without it. We could send the representation as a string, but that assumes whatever client libraries we are interacting with will be able to encode and decode a string representation of a BigDecimal, and that's not a given.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/DecimalRangeRule.java
bskinner: The Integer range validation has integer ranges- that's about the only limitation I can see. If we're ok with that we can I don't see any reason we can't reuse the validation (at least, API-wise)

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
MocanaAtBroad: Oh wait: the type for the answer is int64? Is just using the "number" type a problem?

bskinner: Yes, it is. RFC-8259 does not define a representation of the `number` type but, in practice, most systems use the interpretation of a `number` type as either an integer, or a IEEE 754 double-precision 64-bit floating point value (including the ECMAScript standard)

GSON is a bit of an outlier here in that it provides an API for directly converting a JSON Number to a BigDecimal value, which most other libraries do not provide.

bskinner: Oh, huh- I just took another look at your screenshot. The `precision` property there _is_ an error, and I'll get that fixed up ASAP.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/validation/DecimalRangeRuleDef.java
bskinner: That's a known issue- @Trimax and I noticed it was missing when we were going over this PR.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
MocanaAtBroad: OK. So thinking how this will work in Angular:
User enters a number (e.g, "4.75") in the input field.
Client would then: count the digit characters after the . (in this case, 2).
Transform to a string without the . character => "475"
send answer: 
```json
{answerGuid: "blah", value: 475, scale: 2}
```
Am I getting this right?

bskinner: Yup- that's the general idea.

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/definition/validation/DecimalRangeRuleDef.java
Trimax: I'm not sure I got what you mean. Are you offering to remove the NumericRange and use DecimalRange everywhere since it's more general?

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/DecimalAnswer.java
Trimax: Yep. The example is in Basil study in this PR...

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
Trimax: Yes. My idea was not to avoid NPE but to handle it in the first assert since it's a more correct way IMHO. Your approach is similar to mine, but if you think its better, I'll convert it to Optional

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
Trimax: The method was unused, so I decided to remove it

pepper-apis/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/DecimalRangeRule.java
Trimax: If we're going to reuse the validation, why we're not going to just replace the Numeric type from Integer to Decimal? This also will work. I don't like half/solutions. It will look less clear from my POV. If we do what you offer, we will have two hierarchies of the classes: for Numeric and for Decimal. But Decimal hierarchy will partially reuse Numeric's one.
I guess there are just two options to keep the code readable:
1. Completely replace the type used in Numeric (Integer -> Decimal) and have only one question type - Numeric based on BigDecimal type
2. Keep two independent hierarchies of the classes (like in this PR).

I would say that both solutions are fine. Should we discuss this?

pepper-apis/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
bskinner: My mistake! This looks fine- thank you for the cleanup!

PR #1133, opened 2022-02-02
Feature/new endpoint definition for server options

josemanuel167:
Add new endpoint definition and url prototype for remote options to questions


pepper-apis/docs/specification/src/endpoints/user.studies.activity.instance.question-options.yml
```diff
@@ -0,0 +1,36 @@
...
+        type: string
+    - in: path
+      name: stableId
+      required: true
+      description: the activity instance select question's unique identifier
+      schema:
+        type: string
+  responses:
+    200:
+      $ref: '../pepper.yml#/components/responses/ActivityInstanceSelectAnswersResponse'
```
bskinner: Is this the correct response type for the 200 code?

bskinner:
This looks like a good start, but the response schema for a 200 status code does not seem appropriate for this endpoint. That will need to be fleshed out before approval.

bskinner (changes requested):

pepper-apis/docs/specification/src/endpoints/user.studies.activity.instance.question-options.yml
josemanuel167: Not sure. I think we could re use this class `ActivityInstanceSelectAnswerSubmission` and just change the file's name for a more appropiate one:

image

What do you think?

pepper-apis/docs/specification/src/endpoints/user.studies.activity.instance.question-options.yml
```diff
@@ -0,0 +1,36 @@
...
+      required: true
+      description: the unique identifier for the instance
+      schema:
+        type: string
+    - in: path
+      name: stableId
+      required: true
+      description: the activity instance select question's unique identifier
+      schema:
+        type: string
```
bskinner: There should also be a `q` query string parameter for filtering down the results- @MocanaAtBroad mentioned it in the most recent conversation between the three of us.

pepper-apis/docs/specification/src/endpoints/user.studies.activity.instance.question-options.yml
bskinner: `ActivityInstanceSelectAnswerSubmission` is not an appropriate template for this situation. First off, the `ActivityInstanceSelectAnswerSubmission` response is used when responding to the requests PATCHed to the `/answers` endpoint- we aren't doing that here. Further, at a minimum, the response from the `/options` endpoint should be returning a list of `PicklistOption` objects which match the request's `q` query parameter. Given that the `PicklistOption`s may also be part of a group, those groups should be included in the response as well.

bskinner: I'd suggest checking out the `getDrugSuggestions` and `getCancerSuggestions` endpoints for the general pattern we've used in the past.

josemanuel167: @bskinner Is it possible to do this :  `List<PicklistOption, PicklistGroup> results` in java?

image

bskinner: No- the [java.util.List](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html) class only supports a single type as a generic parameter. Structures like tuples are not supported in Java.

bskinner (dismissed):

MocanaAtBroad (dismissed):
Looks good!

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #1139, opened 2022-02-04
DDP7182 unique value validation

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7182
UNIQUE_VALUE validation rule checks for same value within All participants same questions answers of the study
Scope of the uniqueness is within the entire study same question answers

There is a angular ticket that's pending to handle this new validation type: https://broadinstitute.atlassian.net/browse/DDP-7525


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -99,6 +99,7 @@ ValidationType:
     - REQUIRED
     - UNIQUE
     - YEAR_REQUIRED
+    - UNIQUE_VALUE
```
ssettipalli, line 102: Considered validation rule name STUDY_UNIQUE_VALUE so that scope of the uniqueness will be evident in the name but felt like adding notes in the code and API Doc might be good enough.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2205,6 +2207,17 @@ Validation.Unique:
...
+  allOf:
+    - $ref: '#/Validation'
+    - type: object
+      properties:
+        rule:
+          type: string
+          default: UNIQUE_VALUE
+          enum:
+            - UNIQUE_VALUE
+          description: check for uniqueness in same question answers among all participants of the study
```
bskinner: Let move this key up a few levels so it's at the same depth as the `properties` key above (the description will be associated with the object schema, and not the `rule` property)

pepper-apis/src/main/resources/db-changes/seed/DDP-7182-add-unique-value-validation-type.xml
```diff
@@ -0,0 +1,24 @@
...
+            <column name="translation_text" value="Value should be unique with the study question answers."/>
+        </insert>
+        <insert tableName="i18n_validation_msg_trans">
+            <column name="validation_type_id" valueComputed="(select validation_type_id from validation_type where validation_type_code = 'UNIQUE_VALUE')"/>
+            <column name="language_code_id" valueComputed="(select language_code_id from language_code where iso_language_code = 'es')"/>
+            <column name="translation_text" value="El valor debe ser único con las respuestas a las preguntas del estudio."/>
+        </insert>
+    </changeSet>
+
+</databaseChangeLog>
```
bskinner: Is this file necessary? It looks like it's in the hierarchy prior to the Maven module refactor.

bskinner (commented):
Look good, but I had a couple of questions.

pepper-apis/src/main/resources/db-changes/seed/DDP-7182-add-unique-value-validation-type.xml
ssettipalli: oops. Thank you.
I deleted it and looks like I added it back.

bskinner (approved):
👍

PR #1143, opened 2022-02-07
`[DDP-7379] added family history survey for osteo study

grigol-brd:
https://broadinstitute.atlassian.net/browse/DDP-7379


nsakande (approved):
LGTM

PR #1154, opened 2022-02-08
study-builder build plugins added, test resources added

archilsharashenidzebroad:
After refactor studybuilder-cli was failing to build. It was missing build plugins in pom.xml and some extra resources in tests.


ssettipalli (approved):
looks good to me

PR #1159, opened 2022-02-09
DDP-7306. Decimal support (rebased)

Trimax:
Same changes, but rebased


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/DecimalRangeRule.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * A validation rule that checks numeric integer value is within an optional min/max range, inclusive.
+ */
+public class DecimalRangeRule extends Rule<DecimalAnswer> {
+
+    @SerializedName("min")
+    private BigDecimal min;
+
+    @SerializedName("max")
+    private BigDecimal max;
```
bskinner: What does the JSON serialization of an object of this class look like?

bskinner (commented):
Looks good- just a question about the JSON serialization of the `DecimalRangeRule`.

ssettipalli (dismissed):
looks good to me

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/DecimalRangeRule.java
Trimax: Good point. This should be changed to the DecimalDef as well. Like for Answer

ssettipalli (approved):

PR #1168, opened 2022-02-12
Fix GitHub status check

MocanaAtBroad:
## Context
This is my attempt at fixing GitHub status checks to merge PR branches into develop.
Strategy is to have only one CircleCi workflow for all the tests that are run in a PR.
All the individual test jobs for DSM, DSS and API docs will run every time in this new workflow, but we will try to do as little work as possible.
Please see comments in PR for further explanation.


.circleci/config.yml
```diff
@@ -28,8 +28,9 @@ workflows:
           base-revision: develop
           config-path: .circleci/main-config.yml
           mapping: |
-            pepper-apis/.* action "build-test"
+            .* action "build-test"
```
MocanaAtBroad, line 31: To get a successful status check a workflow needs to run. This change ensures that we will trigger build-test workflow if ANY change in the repo has occurred.
So a change in a, e.g., study configuration file will trigger build-test even though we really don't have tests for these kind of files.

.circleci/config.yml
```diff
@@ -28,8 +28,9 @@ workflows:
...
             pepper-apis/pom.xml dss true
+            pepper-apis/((dss-.*)|(ddp-.*)|(pex-antlr)|(housekeeping)|(studybuilder-cli))/.* test_parallelism 15
```
MocanaAtBroad, line 33: This is a bit of an optimization.
More explanation below.

.circleci/main-config.yml
```diff
@@ -773,12 +770,16 @@ jobs:
...
+      - run: echo "Doing build-api-docs-job << pipeline.parameters.api-docs >>"
+      - when:
+          condition: << pipeline.parameters.api-docs >>
+          steps:
+            - checkout
+            - run:
+                name: Build docs
+                command: |
+                  cd << parameters.spec_dir >>
+                  ./build.sh documentation
```
MocanaAtBroad, line 782: Our original configuration was completely NOT running this job if there were no changes in api-docs.
This PR is about combining all the jobs into a single workflow that can be status checked.
So this job HAS to run.
But if there are no api-docs changes the job can skip all the real steps.
CircleCI forces us to at least execute one step, so that is why we have the run: echo

.circleci/main-config.yml
```diff
@@ -789,38 +790,42 @@ jobs:
...
-          save_m2_cache: true    #REMOVE WHEN DONE
-      - save-ddp-cache
-      - persist_to_workspace:
-          root: *root_path
-          paths:
-            - repo
+      - run: echo 'Doing dsm-compile-and-run-tests-job << pipeline.parameters.dsm >>'
+      - when:
+          condition: << pipeline.parameters.dsm >>
+          steps:
```
MocanaAtBroad, line 796: The ONLY thing that changed in the added conditional statement. Bogus "run echo" step and the conditional.

.circleci/main-config.yml
```diff
@@ -835,26 +840,30 @@ jobs:
...
-          phase: test
-          skip_tests: false
-          skip_checkstyle: true
-          additional_options: '-Dtest=ServerInSameProcessIntegrationTestSuite'
-          restore_m2_cache: false
-      - save-ddp-cache
+      - run: echo "Doing dss-compile-and-run-test-suite-job << pipeline.parameters.dss >>"
+      - when:
+          condition: << pipeline.parameters.dss >>
+          steps:
```
MocanaAtBroad, line 846: Same thing as the other jobs above. Only thing new is the conditional check.

.circleci/main-config.yml
```diff
@@ -1067,6 +1079,10 @@ parameters:
     type: boolean
     default: false
 
+  test_parallelism:
+    type: integer
+    default: 1
+
```
MocanaAtBroad, line 1085: This is an optimization. Explained further below.

.circleci/main-config.yml
```diff
@@ -1079,35 +1095,15 @@ parameters:
 workflows:
   version: 2
 
-  # Run all tests on DSS project
-  dss-build-test-workflow:
+  # Triggered from a local script to test PR changes
```
MocanaAtBroad, line 1098: Idea of this whole PR is that the github status check can be setup so that a single workflow passes.
So the separate workflows we had to run tests for DSS, DSM and API docs have been combined into this one workflow.
It is unfortunate that  CircleCi does not allow conditional logic to run jobs that are part of a workflow conditionally, so we will run all the jobs, but, as you saw in previous changes, each job will do as little as possible.

.circleci/main-config.yml
```diff
@@ -1079,35 +1095,15 @@ parameters:
...
-          parallelism: *test_parallelism
-
-  # Run tests on DSM
-  dsm-build-test-workflow:
-    when:
-      and:
-        - equal: ["build-test", << pipeline.parameters.action >>]
-        - << pipeline.parameters.dsm >>
-    jobs:
+          parallelism: << pipeline.parameters.test_parallelism >> # the caller of this workflow will tell us the parallelism
```
MocanaAtBroad, line 1105: This is an optimization `test_parallelism` pipeline parameter is by default 1.
Expectation is that if setup workflow (in config.yml file) does indeed want to run dss tests, then it will also set the parallelism level to something high (15 in our case).
But if there is no need to run dss tests, then we start the test job on the 1 instance and skip all the steps.

ssettipalli (approved):

bskinner:
Looking this over now, but it seems like `combined-build-test-workflow` is the new job name we should be using for the Github verification?

MocanaAtBroad:
> Looking this over now, but it seems like `combined-build-test-workflow` is the new job name we should be using for the Github verification?

Exactly! Essentially will configure the Github status check to look for successful execution of `combined-build-test-workflow`.

bskinner (approved):
This looks good- let's get it in!

PR #1169, opened 2022-02-14
DDP-7569. Lombok introduction

Trimax:
This PR doesn't contain any business logic changes (please verify that).
This is the first PR from the series (I hope) that will introduce wider usage of Lombok.
This PR contains changes to DTO classes. Instead of using vanilla java's getters, setters and constructors we use Lombok annotations that will generate all the stuff in preprocessing stage.

If you're using IntelliJ IDEA there is a Lombok plugin available. Please make sure you installed it.
Снимок экрана 2022-02-14 в 11 25 31


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/dsm/DsmKitRequest.java
```diff
@@ -1,79 +1,17 @@
...
-
-    /**
-     * The identifier for the kit request used by DSM as a handle.
-     * Check elsewhere for what we set this to.
-     *
-     * @return the kit request id
-     */
-    public String getKitRequestId() {
-        return kitRequestId;
-    }
```
MocanaAtBroad, line 47: Some of these javadoc on the getters look useful and important.  Can we move the jdoc comments to the instance variable definitions?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/dsm/DsmStudyParticipant.java
```diff
@@ -20,141 +22,6 @@
...
-    public void setLegacyShortId(String legacyShortId) {
-        this.legacyShortId = legacyShortId;
-    }
-
-    /**
-     * Report the validation status of the mailing address.
-     * “0” for an address that EasyPost says is invalid.
-     * “1” for an address that EasyPost says is valid and contains suggested EasyPost corrections.
-     * “2” for an address that EasyPost says is valid but does not contain EasyPost suggested corrections.
-     */
```
MocanaAtBroad, line 112: This looks like use javadoc  stuff too.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/housekeeping/BackupJobDto.java
```diff
@@ -1,47 +1,25 @@
 package org.broadinstitute.ddp.db.dto.housekeeping;
 
+import lombok.AllArgsConstructor;
+import lombok.Data;
 import org.jdbi.v3.core.mapper.reflect.ColumnName;
 import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
 
+@Data
+@AllArgsConstructor(onConstructor = @__(@JdbiConstructor))
```
MocanaAtBroad, line 9: Am I right that the generated constructor would have arguments in the same order as the instance variables are listed in the class definition (i.e., (String runName, long startTime, etcetc)?

MocanaAtBroad (commented):
We using the "right" version of Lombok already?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/dsm/DsmKitRequest.java
Trimax: Yep, I'll move the docs. But I don't think that javadoc will work properly if you would like to generate it.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/housekeeping/BackupJobDto.java
Trimax: Yes, you're right.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/dsm/DsmKitRequest.java
MocanaAtBroad: We have never generated javadocs. More interested that comments are preserved when browsing the code.

MocanaAtBroad (approved):

PR #1173, opened 2022-02-14
DDP-7306. Decimal question extended with scale

Trimax:
I hope this is the last PR from the series of DECIMAL PRs.

This PR contains the change to the decimal question table and corresponding classes. The field scale is now in use. It could be set on the definition of the study, imported to the database, and provided to the front-end side to render and restrict the numbers properly.


bskinner (approved):
Looks good! 👍

PR #1174, opened 2022-02-15
moving LiquibaseUtil to ddp-common and all related classes

SimoneMaiwald:
## Context

Moving LiquibaseUtil and related classes into ddp-common


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
```diff
@@ -1,6 +1,6 @@
 package org.broadinstitute.ddp.constants;
 
-import static org.broadinstitute.ddp.util.MiscUtil.fmt;
+import org.broadinstitute.ddp.util.MiscUtil;
```
MocanaAtBroad: You don't like static imports?

SimoneMaiwald: StupidJ didn't like it… I don't have any strong feelings about it

MocanaAtBroad (approved):
Looks good! 👍

PR #1175, opened 2022-02-15
DDP-7173 DSM auth0 support

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7173

DSS changes to let through DSM requests with DSM Tokens from DSM Auth0 Tenant
Some StudyBuilder support to assign clients with NON-Study default Tenant


study-builder/studies/study-example.conf
```diff
@@ -76,7 +76,8 @@
       # Current restrictions:
       # - Only one client can have a redirect URL set.
       # - This client with the redirect URl must have a database connection.
-      "passwordRedirectUrl": "string"   # Optional
+      "passwordRedirectUrl": "string"   # Optional,
+      "clientDomain": "string"   # Optional and Tenant should exist in DB
```
bskinner: Should this be `domain`? I don't see a `clientDomain` key used elsewhere in this PR.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilder.java
```diff
@@ -337,14 +337,28 @@ private StudyDto getStudyOrInsert(Handle handle, long tenantId, long umbrellaId)
...
+                    throw new DDPException("Client Domain :" + clientDomain + " not found ");
+                }
+                clientTenantId = dto.getId();
+                LOG.info("Using client domain: {} ", clientDomain);
+            }
+
+            long finalTenantId = (clientTenantId != null) ? clientTenantId : tenantId;
             ClientDto clientDto = jdbiClient
-                    .findByAuth0ClientIdAndAuth0TenantId(clientId, tenantId)
+                    .findByAuth0ClientIdAndAuth0TenantId(clientId, finalTenantId)
```
bskinner, line 355: Is there a case where this could end up using the same `clientId` for two different tenants?

ssettipalli: I don't think so unless configured incorrectly

MocanaAtBroad:
I see that you are allowing having multiple tenants per study. Presumably this is to add dsm as a second tenant to any study?
I think that is fine. But my question is: 
Will we be treating dsm tenants/tokens created by dsm differently in the backend code? Or if we have 2 tenants associated with one study,  our backend code will treat both tokens identically/does not care which tenant it belongs to?

I ask because then the question is whether we need to somehow flag the dsm tenant in the configuration as "special" in some way so backend knows to treat tokens that use this tenant differently.

ssettipalli:
> I see that you are allowing having multiple tenants per study. Presumably this is to add dsm as a second tenant to any study? I think that is fine. But my question is: Will we be treating dsm tenants/tokens created by dsm differently in the backend code? Or if we have 2 tenants associated with one study, our backend code will treat both tokens identically/does not care which tenant it belongs to?
> 
> I ask because then the question is whether we need to somehow flag the dsm tenant in the configuration as "special" in some way so backend knows to treat tokens that use this tenant differently.

@MocanaAtBroad 
Backend will  verify if Token is valid and findPermittedStudyGuidsByAuth0ClientIdAndAuth0Domain will check if this tenant (DSM or any other client/Tenant) is authorized to this study and will let it through. 
NO special flagging of DSM client/tenant needed.
Study configuration will insert additional clients (DSM) so that DSS backend code during authorization (findPermittedStudyGuidsByAuth0ClientIdAndAuth0Domain) will identify ALL clients(Tenants) permitted to the study.

MocanaAtBroad (approved):

bskinner (approved):
👍

PR #1186, opened 2022-02-18
Lombok: DTOs migration

Trimax:
This PR contains refactoring for the rest of the DTO classes. All DTO model classes are now defined using Lombok annotations. A lot of boilerplate code was removed. Unused methods, constants, constructors were removed.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
```diff
@@ -13,332 +16,106 @@
...
 import org.broadinstitute.ddp.model.event.PdfAttachment;
 import org.broadinstitute.ddp.model.user.EnrollmentStatusType;
+import org.jdbi.v3.core.mapper.reflect.ColumnName;
+import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
 
 /**
  * A DTO representing the event configuration, left joined with possible trigger and action configuration info
  */
+@Value
+@AllArgsConstructor(onConstructor = @__(@JdbiConstructor))
```
bskinner, line 26: This is listed as an experimental feature in Lombok's [docs](https://projectlombok.org/features/experimental/onX). I'm hesitant about relying on this feature. What are your thoughts, @MocanaAtBroad

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
```diff
@@ -13,332 +16,106 @@
...
-    private int executionOrder;
-    private String label;
-    private String gcpTopic; // FIXME should not be a base field of EventAction
+    long eventConfigurationId;
+    String label;
+    EventTriggerType eventTriggerType;
+    EventActionType eventActionType;
+    Integer postDelaySeconds;
+
+    @Accessors(fluent = true)
```
bskinner, line 34: [@Accessors](https://projectlombok.org/features/experimental/Accessors) is considered an experimental feature. It looks like the feature is on-track to being moved to `stable` and I think we're clear with the potential behavioral changes, but there are potentially backwards incompatible changes [in-flight](https://github.com/projectlombok/lombok/issues/3105). Is `@Accessors` a requirement, or can similar effects be achieved using the current [Stable](https://projectlombok.org/features/all) features?

MocanaAtBroad: @bskinner I think that the alternative is to have the accessors for instance variables of type boolean to use the `is` prefix (the Java bean standard), in this case `isDispatchToHousekeeping()` which I believe you appropriately flagged as reading awkward.
Again, I am inclined to go forward with its usage as I think worst case we can globally/mechanically refactor code to be whatever the new thing is.

MocanaAtBroad (approved):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
bskinner: I'm good with that- My confusion here was misunderstanding the role of the `@Accessors` annotation. It's just controlling how the `@Getter` and `@Setter` annotations synthesize the methods.

My understanding is that the behavior preventing this from being declared a `Stable` feature is (mostly) a change in how the configurations handle cascading the default values and AFAIK we aren't using that. My question was whether `@Accessors` was the only option. It looks like a potential alternative was to set the global `lombok.accessors.fluent` configuration to `true`, but that disables the prefixing of setters with `set` and I think that alone rules it out for this case.

bskinner (approved):
👍

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
MocanaAtBroad: Thanks for flagging this, @bskinner . I am inclined to go ahead and use it anyways because it really is a very useful option.
Just thinking that one of 3 things could happen:
1) Capability remains as is (in this case we are golden)
2) Capability is modified changing the annotation name/syntax. (In this case we could do a global replace).
3) Capability is totally dropped from Lombok (In this case we would have to add back all the constructor stuff).

1 or 2 seem to A LOT more probable than 3 , so would be no op or minor refactor. And 3 would not be the end of the world.
I say let's go for it. I am just miffed that we are up to Java 9001 and we still need to write all this mindless boilerplate.

bskinner: > Thanks for flagging this, @bskinner . I am inclined to go ahead and use it anyways because it really is a very useful option.

Agreed with it being a useful feature- I'm all for reducing the amount of boilerplate we're writing.

>  I am just miffed that we are up to Java 9001 and we still need to write all this mindless boilerplate.

No kidding- It'd be lovely when we can finally drop this. When Objective-C introduced the `@synthesize` keyword and Automatic Property Synthesis, it was a game changer.

@Trimax :
Walking through the docs again, I ran across this blurb in the overview section of the [onX](https://projectlombok.org/features/experimental/onX) docs:
```
The syntax is a little strange and depends on the javac you are using.
On javac7, to use any of the 3 onX features, you must wrap the annotations to be applied to the constructor / method / parameter in @__(@AnnotationGoesHere). To apply multiple annotations, use @__({@Annotation1, @Annotation2}). The annotations can themselves obviously have parameters as well.
On javac8 and up, you add an underscore after onMethod, onParam, or onConstructor. 
```

It _seems_ like things are working, but do we need to change the annotation, or is this an outdated note? The use of `onConstructor` without an underscore seems to be counter to my understanding of the docs given we're on Java 11.

Trimax: I tested it even with Java 17 and it is working as expected. Besides, no one canceled backward compatibility :)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
Trimax: Definition of the global variable requires us to change the field names accordingly. Otherwise, getters and setters will be generated a bit differently. From my POV using @Accessors is the most elegant way to solve this. As for the alternatives, there are several:
1. Define methods on our own. 
Let say we have Boolean writeOnce. Lombok will generate getWriteOnce & setWriteOnce(). In addition to that we could manually define shouldWriteOnce() method.
2. Rename all the fields and methods. In this case we will not have shouldXXX canXXX we have to give the readable name starting with is

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/EventConfigurationDto.java
bskinner: I'm not sure where backward compatibility comes into this?

Edit: Did some more digging- My impression from the Lombok history is that the bug in Java 8 that required the trailing underscore workaround was fixed in Java 11, so we're good here!

PR #1192, opened 2022-02-22
DDP-7180. Cross-form validation

Trimax:
This PR contains the implementation of the cross-form validation for the answers to the comparable-type questions.


bskinner (changes requested):
Looks good, and the DB modifications apply cleanly! This PR does need tests added for the new rule type, though. The tests we have for the other rule types are located in the `org.broadinstitute.ddp.model.activity.instance.validation` package, if examples would help.

Trimax:
> Looks good, and the DB modifications apply cleanly! This PR does need tests added for the new rule type, though. The tests we have for the other rule types are located in the `org.broadinstitute.ddp.model.activity.instance.validation` package, if examples would help.

Thanks for the suggestion. Will add the test

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonDateRuleTest.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.List;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@SuppressWarnings({"unchecked", "rawtypes"})
+public class ComparisonDateRuleTest extends TxnAwareBaseTest {
+    private static TestDataSetupUtil.GeneratedTestData testData;
+    private static Pair<Long, String> questionIdInstanceGuidPair;
+    private static Question unused;
```
bskinner: Let's rename this- `unused` is misleading.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRule.java
```diff
@@ -0,0 +1,99 @@
...
+
+            return validate(question.getQuestionType(), answer, referencedAnswer.get());
+        });
+    }
+
+    private boolean validate(final QuestionType type, final Answer actualAnswer, final Answer referencedAnswer) {
+        if (actualAnswer.getQuestionType() != referencedAnswer.getQuestionType()) {
+            log.debug("The answers are incompatible. Actual: {}. Referenced: {}",
+                    actualAnswer.getQuestionType(), referencedAnswer.getQuestionType());
+            return false;
```
bskinner: Let's make a lot more noise here.

To my current understanding, entering this method with 2 distinct types of answers is indicative of a larger problem such as an incorrect study design or an issue in the database, and we shouldn't try to limp on. Let's throw and wash our hands of this.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRule.java
```diff
@@ -0,0 +1,99 @@
...
+                return validate(DecimalDef::compareTo, (DecimalDef) referencedAnswer.getValue(), (DecimalDef) actualAnswer.getValue());
+            case DATE:
+                final Optional<LocalDate> referencedDate = ((DateValue) referencedAnswer.getValue()).asLocalDate();
+                final Optional<LocalDate> originalDate = ((DateValue) actualAnswer.getValue()).asLocalDate();
+
+                return referencedDate.isEmpty() || originalDate.isEmpty()
+                        || validate(LocalDate::compareTo, referencedDate.get(), originalDate.get());
+            default:
+                log.debug("The question type is not comparable. This is impossible situation. Nothing to validate Actual: {}", type);
+                return false;
```
bskinner: Same thing as above. If the situation is impossible, but we're returning a `false` value, we're returning a value the caller can't trust. Let's throw here as well.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
```diff
@@ -0,0 +1,113 @@
...
+        testData = TransactionWrapper.withTxn(TestDataSetupUtil::generateBasicUserTestData);
+        unused = new NumericQuestion("sid", 1L, 2L, false, false, false, null, null, null, List.of(), List.of());
+        questionIdInstanceGuidPair = TransactionWrapper.withTxn(ComparisonRuleNegativeTest::prepareTestData);
+    }
+
+    @Test
+    public void testValidate_noValue() {
+        ComparisonRule rule = ComparisonRule.builder().build();
+        assertFalse(rule.validate(null, null));
+        assertTrue(rule.validate(null, new NumericAnswer(1L, "q", "a", null)));
```
bskinner: Let's split this out into two separate tests as the two cases are different types of "no value".

In the former case (`rule.validate(null, null)`), the question doesn't have any Answers (signaling that maybe the user hasn't even seen the question yet), and latter case a client responded (a `PATCH` was sent to create the Answer), but the value of the answer is `null`.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test(expected = RuntimeException.class)
+    public void testValidate_incorrectComparisonType() {
+        var comparisonRule = ComparisonRule.builder()
+                .type(RuleType.COMPARISON)
+                .referenceQuestionId(questionIdInstanceGuidPair.getLeft())
+                .build();
+
+        assertTrue(comparisonRule.validate(unused, new NumericAnswer(null, testData.getStudyGuid(),
+                "abc", 0L, questionIdInstanceGuidPair.getRight())));
```
bskinner: The behavior being tested here is unclear- Is it expected to throw a `RuntimeException`, or is the `validate(...)` call expected to return `true`?

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
```diff
@@ -0,0 +1,113 @@
...
+                .build();
+
+        assertFalse(comparisonRule.validate((Question) new TextQuestion("sid", 1L,
+                2L,  List.of(), List.of(), TextInputType.TEXT),
+                new NumericAnswer(null, testData.getStudyGuid(), "abc", 0L,
+                        questionIdInstanceGuidPair.getRight())));
+    }
+
+    @Test
+    public void testValidate_incompatibleQuestions() {
```
bskinner: Let's add some short javadoc explaining what these two tests are looking for. It's not clear to me what the difference is between an incompatible and an incorrect question type.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.math.BigInteger;
+import java.util.List;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@SuppressWarnings({"unchecked", "rawtypes"})
```
bskinner: Let's fix the warnings instead of masking them.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
```diff
@@ -0,0 +1,141 @@
...
+        TestFormActivity act = TestFormActivity.builder()
+                .withNumericIntQuestion(true)
+                .build(handle, testData.getUserId(), testData.getStudyGuid());
+
+        var activityInstance = createInstance(handle, act.getDef().getActivityId());
+
+        var created = handle.attach(AnswerDao.class)
+                .createAnswer(testData.getUserId(), activityInstance.getId(),
+                        new NumericAnswer(null, act.getNumericIntQuestion().getStableId(), null,
+                                10L, activityInstance.getGuid()));
```
bskinner, line 127: What are your thoughts on moving the expected value of the answer for this question to a top-level ivar or constant in the class? The expected value of the question is nested pretty deeply in the setup callstack, and is easy to miss in the noise. I'm also open to other suggestions you may have here.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @Test
+    public void testValidateEqual() {
+        var comparisonRule = ComparisonRule.builder()
+                .type(RuleType.COMPARISON)
+                .referenceQuestionId(questionIdInstanceGuidPair.getLeft())
+                .comparisonType(ComparisonType.EQUAL)
+                .build();
+
+        assertTrue(comparisonRule.validate(unused, createAnswer(10L, questionIdInstanceGuidPair.getRight())));
+        assertFalse(comparisonRule.validate(unused, createAnswer(20L, questionIdInstanceGuidPair.getRight())));
```
bskinner: The `assertFalse` isn't needed here- `testValidateNotEqual` has it covered.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @Test
+    public void testValidateNotEqual() {
+        var comparisonRule = ComparisonRule.builder()
+                .type(RuleType.COMPARISON)
+                .referenceQuestionId(questionIdInstanceGuidPair.getLeft())
+                .comparisonType(ComparisonType.NOT_EQUAL)
+                .build();
+
+        assertTrue(comparisonRule.validate(unused, createAnswer(20L, questionIdInstanceGuidPair.getRight())));
+        assertFalse(comparisonRule.validate(unused, createAnswer(10L, questionIdInstanceGuidPair.getRight())));
```
bskinner: Same as the `testValidateEqual` case- the other test has the `assertFalse` case here already covered so we can skip it.

bskinner (changes requested):
Glad to see the tests in, and the benefits of the Lombok integration! Left a few comments to address.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRule.java
Trimax: Fair enough. Agree.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
Trimax: assertTrue is redundant. Will remove it

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
Trimax: Ok. Incompatible: when you're trying to compare decimal and date answers for instance. Incorrect type when you try to compare, for instance, text answers

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
Trimax: All other tests just leave them as is. There's no strict type casting.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
Trimax: Shouldn't we test each case separately? If we have an "equals" operator, we expect it to return true for equal objects and false for not equal. This will be full coverage. Otherwise, we do two independent tests with only a positive scenario check.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonDateRuleTest.java
Trimax: It was a copy-paste from another test. Anyway, done.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
Trimax: Simply removed the annotation

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
Trimax: Why? First of all, when we create NumericAnswer we reuse the identifiers that were generated in runtime a few lines up. Second, we don't need the answer itself directly, we need just the produced identifiers. And identifiers are stored on a top-level static fields.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
bskinner: You are correct- we do need to check both cases here. Disregard my above comment and let's leave this as-is!

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
bskinner: Disregard this comment.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonNumericRuleTest.java
bskinner: In this case I was specifically referring to the `10L` value, not the Answer object itself. My intended goal was to make it easier for folks (especially those unfamiliar with the codebase) to verify what the "correct" value is without digging deep into the setup methods. We're using `10L` in a number of places, and it's easy to overlook the baseline truth given the surrounding context.

Let's keep this as-is for now, but if you have ideas about alternative ways to handle this situation let's have a chat!

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/model/activity/instance/validation/ComparisonRuleNegativeTest.java
bskinner: Gotcha- Thank you for the clarification!

bskinner (approved):

PR #1195, opened 2022-02-24
DDP-7547_DSM_addingClaims

pegahtah:
working with RSA token, calling the M2M application of auth0 for getting a new token with DSM claims and fixes


pegahtah:
I made the PR and then realized it was based on the wrong old DSM branch, made another one based off of develop and closed this one

PR #1196, opened 2022-02-24
DDP-7547 DSM with new auth0 tenant

pegahtah:
Changed the authentication method to be RSA, calling the M2M application in auth0 to get the new token with DSM claims, added the needed DSS claims, created test class to test the verification that dss doess on dsm token an to call a dss endpoint


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/JWTRouteFilterTest.java
```diff
@@ -216,4 +238,50 @@ private String createTokenWithRoles(String secret, long invalidAfter, String rol
...
+
+            jwkProvider = new JwkProviderBuilder(auth0Domain).cached(100, 3L, MINUTES).build();
+            jwkProviderMap.put(auth0ClientId, jwkProvider);
+        }
+
+        DecodedJWT validToken = verifyDDPToken(jwt, jwkProvider);
+        String ddpUserGuid = validToken.getClaim(userId).asString();
+        Assert.assertNotNull(ddpUserGuid);
+    }
+
```
pegahtah, line 288: I copied the dss part of token verification here and cleared the unnecessary parts (user info, etc) to test our token with that verification method

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -473,8 +427,9 @@ protected void setupDB(@NonNull Config config) {
         TransactionWrapper.init(new TransactionWrapper.DbConfiguration(TransactionWrapper.DB.DSM, maxConnections, dbUrl));
 
         logger.info("Running DB update...");
-        LiquibaseUtil.runLiquibase(dbUrl, TransactionWrapper.DB.DSM);
-        LiquibaseUtil.releaseResources();
+        // TODO Pegah fix and uncomment before merge (throws error so should eb commented now)
```
ssettipalli: need to uncomment liquibase like mentioned in TODO

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -565,9 +527,8 @@ protected void setupCustomRouting(@NonNull Config cfg) {
 
                     boolean isTokenValid = false;
                     if (StringUtils.isNotBlank(tokenFromHeader)) {
-                        isTokenValid =
-                                new CookieUtil().isCookieValid(req.cookie(cookieName), cookieSalt.getBytes(), tokenFromHeader, jwtSecret);
-                        isTokenValid = new JWTRouteFilter(jwtSecret, null).isAccessAllowed(req);
+                        isTokenValid = new CookieUtil().isCookieValid(req.cookie(cookieName), cookieSalt.getBytes(), tokenFromHeader, auth0Domain);
```
ssettipalli: isTokenValid is overwritten with value evaluated from next line (JWTRouteFilter)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/auth0/Auth0M2MRequest.java
```diff
@@ -0,0 +1,36 @@
...
+    private final String clientSecretKey = "client_secret";
+    private final String audienceKey = "audience";
+
+    public Auth0M2MRequest(@NonNull String clientId,@NonNull String grantType,@NonNull String clientSecret,@NonNull String audience, Map<String, String> claims, String audienceNameSpace){
+
+        //adding the request values with key name (audienceNameSpace + key) because auth0 requires claims to have a valid url as the name
+        this.params = new ArrayList<>();
+        for (String key: claims.keySet()){
+            String k = key;
+            if(key.indexOf(audienceNameSpace) == -1)
```
ssettipalli: Why are we looking up with audienceNameSpace inside the keySet loop iteration ? 
Do we meant to lookup audienceNameSpace + k ?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
```diff
@@ -19,6 +20,7 @@
 import lombok.NonNull;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.security.JWTConverter;
```
ssettipalli: wonder if we can use common JWTConverter for DSS & DSM !

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/JWTRouteFilter.java
```diff
@@ -43,11 +44,11 @@
      *                     must be present in the roles claim in the token.
      *                     If empty or null, the roles claim is not checked.
      */
-    public JWTRouteFilter(String jwtSecret, Collection<String> allowedRoles) {
+    public JWTRouteFilter(String jwtSecret, Collection<String> allowedRoles, String auth0Domain) {
```
MocanaAtBroad: Consider implementing the spark.Filter

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/SecurityUtil.java
```diff
@@ -25,12 +29,17 @@
...
+    private static String auth0Domain;
+    private static String auth0Namespace;
+    private static String auth0Signer;
 
-    public SecurityUtil(@NonNull String secret) {
-        this.secret = secret;
+    public SecurityUtil(@NonNull String auth0Domain, @NonNull String auth0Namespace, @NonNull String auth0Signer) {
+        SecurityUtil.auth0Domain = auth0Domain;
+        SecurityUtil.auth0Namespace = auth0Namespace;
+        SecurityUtil.auth0Signer = auth0Signer;
```
MocanaAtBroad: Consider getting rid of this constructor you are not using instances of this object anywhere as far as I can see.
Looks like intent of constructor is to set static variables. Might make more sense to do something like:
```java
public static init(@NonNull String auth0Domain, @NonNull String auth0Namespace, @NonNull String auth0Signer)

Comment (MocanaAtBroad) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/SecurityUtil.java]: And we have been using convention of capital letter names for static variables.  Appears this convention was already being used in class.

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java [hunk 8]
```diff
@@ -80,24 +97,61 @@ public Object handle(Request request, Response response) throws Exception {
...
-
-                DSMToken authResponse = new DSMToken(jwtToken);
-                return authResponse;
-            } else {
-                throw new RuntimeException("UserIdentity not found");
+                String dsmToken = getNewAuth0TokenWithCustomClaims(claims, clientSecret, auth0ClientId, auth0Domain, auth0MgmntAudience, audienceNameSpace);
+                if (dsmToken != null) {
+                    return new DSMToken(dsmToken);
+                } else{
+                    throw new RuntimeException("Not authorized user");
```
MocanaAtBroad: What is the response code and body content to client if you throw a RuntimeException? Is it a 500?
Consider returning meaningful http codes like 400 (Bad request) or 422 (Unprocessable Entity)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -80,24 +97,61 @@ public Object handle(Request request, Response response) throws Exception {
...
+                        () -> {
+                            throw new RuntimeException("Missing " + clientId + " in auth0 claims");
+                        });
+                auth0Util.getClaimValue(auth0Token, userId, auth0Domain).ifPresentOrElse(claim -> claims.put(userId, claim.asString()),
+                        () -> {
+                            throw new RuntimeException("Missing " + userId + " in auth0 claims");
+                        });
 
                 long auth0Expiration = auth0UserInfo.getTokenExpiration();
                 int cookieAgeInSeconds = new Long(auth0Expiration - new Double(System.currentTimeMillis() / 1000d).intValue()).intValue();
```
MocanaAtBroad: Looks like this variable is never used.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -59,7 +76,7 @@ public Object handle(Request request, Response response) throws Exception {
         String auth0Token = jsonObject.get(payloadToken).getAsString();
```
MocanaAtBroad: `new JSONParser().parse(` throws a `JSONSyntaxException` . We should `catch` it and return a 400 with some error message (don't know what DSM uses for serialization of error messages) so client knows that there was a problem with the JSON. Optionally log the error.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
```diff
@@ -121,13 +128,11 @@ private void verifyUserConnection(@NonNull String userId, @NonNull String email)
...
         Map<String, Claim> auth0Claims = new HashMap<>();
         try {
-            Algorithm algorithm = Algorithm.HMAC256(decodedSecret);
-            JWTVerifier verifier = JWT.require(algorithm).withIssuer(account).build(); //Reusable verifier instance
-            DecodedJWT jwt = verifier.verify(auth0Token);
-            auth0Claims = jwt.getClaims();
+            DecodedJWT validToken = JWTConverter.verifyDDPToken(auth0Token, auth0Domain);
+            auth0Claims = validToken.getClaims();
         } catch (Exception e) {
             throw new RuntimeException("Could not verify auth0 token.", e);
```
MocanaAtBroad: Consider returning a 400 error when this happens, so might have to catch it from caller in that case.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
```diff
@@ -121,13 +128,11 @@ private void verifyUserConnection(@NonNull String userId, @NonNull String email)
...
 
-    private Map<String, Claim> verifyAndParseAuth0TokenClaims(String auth0Token) {
+    private Map<String, Claim> verifyAndParseAuth0TokenClaims(String auth0Token, String auth0Domain) {
         Map<String, Claim> auth0Claims = new HashMap<>();
         try {
-            Algorithm algorithm = Algorithm.HMAC256(decodedSecret);
-            JWTVerifier verifier = JWT.require(algorithm).withIssuer(account).build(); //Reusable verifier instance
-            DecodedJWT jwt = verifier.verify(auth0Token);
-            auth0Claims = jwt.getClaims();
+            DecodedJWT validToken = JWTConverter.verifyDDPToken(auth0Token, auth0Domain);
```
MocanaAtBroad: JWTConverter.verifyDDPToken code appears could return a null. Should check for it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -80,24 +97,61 @@ public Object handle(Request request, Response response) throws Exception {
...
             throw new RuntimeException("There was no token in the payload");
         }
     }
 
+    private String getNewAuth0TokenWithCustomClaims(Map<String, String> claims, String clientSecret, String clientId, String auth0Domain, String auth0Audience, String audienceNameSpace) {
+        String api = "/oauth/token";
+        String contentType = "application/x-www-form-urlencoded";
+        String clientCredentials = "client_credentials";
+
+        String sendRequest = "https://" + auth0Domain + api;
```
MocanaAtBroad:
```suggestion
        String requestUrl = "https://" + auth0Domain + api;
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/auth0/Auth0M2MRequest.java
```diff
@@ -0,0 +1,36 @@
...
+    private final String grantTypeKey = "grant_type";
+    private final String clientSecretKey = "client_secret";
+    private final String audienceKey = "audience";
+
+    public Auth0M2MRequest(@NonNull String clientId,@NonNull String grantType,@NonNull String clientSecret,@NonNull String audience, Map<String, String> claims, String audienceNameSpace){
+
+        //adding the request values with key name (audienceNameSpace + key) because auth0 requires claims to have a valid url as the name
+        this.params = new ArrayList<>();
+        for (String key: claims.keySet()){
+            String k = key;
```
MocanaAtBroad: Unnecessary assignment. Just use `key` or rename iteration variable `k`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/auth0/Auth0M2MRequest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+@Data
+public class Auth0M2MRequest {
+    public List<NameValuePair> params;
+    private final String clientIdKey = "client_id";
+    private final String grantTypeKey = "grant_type";
+    private final String clientSecretKey = "client_secret";
+    private final String audienceKey = "audience";
+
+    public Auth0M2MRequest(@NonNull String clientId,@NonNull String grantType,@NonNull String clientSecret,@NonNull String audience, Map<String, String> claims, String audienceNameSpace){
```
MocanaAtBroad: For what I am seeing this object does not do anything useful except to create the params list.
Consider getting rid of thus class and refactor code somewhere else to a method called that looks like so:
```java
List<NameValuePair> buildTokenRequestBody((@NonNull String clientId,@NonNull String grantType,@NonNull String clientSecret,@NonNull String audience, Map<String, String> claims, String audienceNameSpace)

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/auth0/Auth0M2MRequest.java (hunk 4 above)
Comment (pegahtah) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/auth0/Auth0M2MRequest.java]: This is because DSS needed claims that are added per `RULES` to the first token ( token obtained while doing the Google login) already has the namespace added to them (they are set as `audienceNameSpace + \ + claimName` in auth0 rule ), so I check to make sure I don't add the URL name space to them twice

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java (hunk 3 above)
Comment (pegahtah) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java]: Great catch!

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java [hunk 16]
```diff
@@ -80,24 +97,61 @@ public Object handle(Request request, Response response) throws Exception {
                 claims.put(authUserId, String.valueOf(userDto.getId()));
                 claims.put(authUserName, userDto.getName().orElse(""));
                 claims.put(authUserEmail, email);
+                auth0Util.getClaimValue(auth0Token, tenantDomain, auth0Domain).ifPresentOrElse(claim -> claims.put(tenantDomain, claim.asString()),
```
MocanaAtBroad: Line 89 above probably just returns a 500 error if user is not found in database and client and we will never find out who that user was. We should return something like 422 or 400 and log the email that was not found as an error.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -30,6 +34,9 @@
     private final String authUserEmail = "USER_MAIL";
     private final String userAccessRoles = "USER_ACCESS_ROLE";
     private final String userSettings = "USER_SETTINGS";
+    private final String clientId = "https://datadonationplatform.org/cid";
+    private final String userId = "https://datadonationplatform.org/uid";
+    private final String tenantDomain = "https://datadonationplatform.org/t";
```
MocanaAtBroad, line 54: All these variables above look like they should be static rather than instance variables. Please take a look and decide.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -578,9 +539,15 @@ protected void setupCustomRouting(@NonNull Config cfg) {
         });
         setupDDPConfigurationLookup(cfg.getString(ApplicationConfigConstants.DDP));
 
-        AuthenticationRoute authenticationRoute =
-                new AuthenticationRoute(auth0Util, jwtSecret, cookieSalt, cookieName, userUtil, cfg.getString("portal.environment"));
-        post(UI_ROOT + RoutePath.AUTHENTICATION_REQUEST, authenticationRoute, new JsonTransformer());
+        AuthenticationRoute authenticationRoute = new AuthenticationRoute(auth0Util,
```
MocanaAtBroad, line 600: This object is not being used.

MocanaAtBroad: Actually, I am not seeing the AuthenticationRoute being used at all. Did we set a route path for it somewhere I am missing?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -38,9 +45,14 @@
     private final String cookieName;
     private final UserUtil userUtil;
     private final String environment;
```
MocanaAtBroad: environment, cookieSalt, cookieName, and jwtSecret are unused.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/DDPRequestUtil.java
```diff
@@ -309,4 +313,19 @@ public static Integer postRequest(String sendRequest, Object objectToPost, Strin
...
+        logger.info("Requesting data from " + name + " w/ " + sendRequest);
+        org.apache.http.client.fluent.Request request = SecurityUtil.createPostRequestWithHeaderNoToken(sendRequest, header, objectToPost);
+
+        T objects = null;
+        try {
+            Response response = request.execute();
+            objects = response.handleResponse(res -> getResponse(res, responseClass, sendRequest));
+        }
+        catch (IOException e) {
+            e.printStackTrace();
```
MocanaAtBroad: Let's log the error . I don't know what this will do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -80,24 +97,61 @@ public Object handle(Request request, Response response) throws Exception {
...
+
+        String sendRequest = "https://" + auth0Domain + api;
+        Map<String, String> headers = new HashMap<>();
+        headers.put("content-type", contentType);
+
+        Auth0M2MRequest requestModel = new Auth0M2MRequest(clientId, clientCredentials, clientSecret, auth0Audience, claims, audienceNameSpace);
+
+        Auth0M2MResponse response;
+        try {
+            response = DDPRequestUtil.postRequestWithResponse(Auth0M2MResponse.class, sendRequest, requestModel.getParams(), "auth0 M2M", headers);
```
MocanaAtBroad: postRequestWithResponse as written can return a null. Refactor by either throwing an exception or checking for a null.

MocanaAtBroad (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -534,10 +542,7 @@ protected void setupCustomRouting(@NonNull Config cfg) {
 
                     boolean isTokenValid = false;
                     if (StringUtils.isNotBlank(tokenFromHeader)) {
-                        isTokenValid =
-                                new CookieUtil().isCookieValid(req.cookie(cookieName), cookieSalt.getBytes(), tokenFromHeader, jwtSecret);
```
pegahtah, line 538: @SimoneMaiwald I removed this line because it wasn't doing anything., I am not sure how to use this with the current authentication method. The method `CookieUtil.createSecureCookieForToken` is only used in tests in develop so I think this is unnecessary .

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: Sorry, My bad! It is back on L563!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
pegahtah: They are just the keys for claim names. Do we need them really to be static?

MocanaAtBroad: Well, it is long standing standard practice in Java to assign "variables" that are constants to static variables.
Has to do with language constraint on only having with Class object in a given class loader, so you would not be duplicating the space.
More importantly it is at this point a common expectation in Java programming and not doing this way jumps out immediately as "something not right here".

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -30,72 +35,158 @@
...
+                        claims = getDSSClaimsFromOriginalToken(auth0Token, auth0Domain, claims);
+
+                        try {
+                            String dsmToken = getNewAuth0TokenWithCustomClaims(claims, clientSecret, auth0ClientId, auth0Domain, auth0MgmntAudience, audienceNameSpace);
+                            if (dsmToken != null) {
+                                return new DSMToken(dsmToken);
+                            }
+                            else {
+                                response.status(401);
+                                throw new RuntimeException("DSMToken was null! Not authorized user");
```
MocanaAtBroad: Unless Spark does something surprising, I don't expect that your status codes will be sent to the client.
Setting the code in response is always followed by RuntimeException, so doubt Spark will get a chance to send the response you intended. More likely, client will get a 500/System error.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DSSTestingRoute.java
```diff
@@ -0,0 +1,46 @@
...
+        HttpEntity entity = objects.getEntity();
+        String responseString = EntityUtils.toString(entity, "UTF-8");
+        System.out.println(responseString);
+        if (objects != null) {
+            logger.info("Got response back");
+        }
+        return responseString;
+    }
+}
+
```
MocanaAtBroad, line 46: Ah! This is marked for removal..

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/LoggingFilter.java
```diff
@@ -24,25 +25,30 @@
...
+                    Claim userEmailClaim = decodedJWT.getClaim(claimNameSpace+"USER_MAIL");
+
+                    if (userEmailClaim != null) {
+                        String userEmail = userEmailClaim.asString();
+                        if (StringUtils.isNotBlank(userEmail)) {
+                            MDC.put(USER_EMAIL, userEmail);
+                        }
                     }
-                }
+                }, () ->{throw  new RuntimeException("Unable to verifiy token");});
```
MocanaAtBroad: Is intent to block the request from proceeding to execute a route?
If so, proper way to do this can be found here:
https://github.com/broadinstitute/ddp-study-server/blob/1ef16842cc27f6d286c1ec8cfc7558b4b7452d75/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/StudyAdminAuthFilter.java#L29

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -0,0 +1,47 @@
...
+        //        claims.put("https://datadonationplatform.org/uid", "C2JYXJUUBC2KP9R10G1U");
+        //        String jwtToken = new SecurityHelper().createToken(auth0Token, cookieAgeInSeconds + (System.currentTimeMillis() / 1000) + (60 * 5),
+        //                claims);
+
+        Map<String, String> authHeaders = new HashMap<>();
+        authHeaders.put("Authorization", "Bearer " + auth0Token);
+        String reqAddress = "https://pepper-dev.datadonationplatform.org/pepper/v1/user/"+user_id+"/studies/cmi-pancan/activities";
+        HttpResponse response = TestUtil.performGet(reqAddress, "", authHeaders).returnResponse();
+        System.out.println(response.toString());
+    }
```
MocanaAtBroad, line 42: This test is intended to a be temporary/to try out stuff too, right?
I think you DO want a test for your authentication route; it's critical infrastructure!
(My advice because of the urgency in getting this functionality merged is to create a test in separate PR).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DSSTestingRoute.java
pegahtah: Oh my todo is written wrong, what I meant is we have to keep this for Alexei's test and then when dss and dsm is really connected through the angular dss, we remove this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/LoggingFilter.java
pegahtah: Thank you!! changes applied!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
pegahtah: done and fixed. Thanks for pointing it out

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
pegahtah: I am now using `halt` like how DSS does it, I think this should work now 
(Also taking  note to make sure to write tests for all of these cases)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DSSTestingRoute.java
pegahtah: I think I wrote the todo bad (fixed now) , what I meant is, let's keep this until the frontend code is complete, this may come handy for testing, and then we should remove

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
pegahtah: I will put that in my todo list. I think that's a good idea, will definitely look into it

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
pegahtah: yes! I plan to change the unit tests we have for the old token to be used for testing new Authentication. As you said, maybe a separate PR is better.

MocanaAtBroad (dismissed):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -453,17 +455,19 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
         }
 
         //  capture basic route info for logging
-        before("*", new LoggingFilter());
+        before("*", new LoggingFilter(auth0Domain, auth0claimNameSpace));
         afterAfter((req, res) -> MDC.clear());
 
         before(API_ROOT + "*", (req, res) -> {
-            if (!new JWTRouteFilter(bspSecret, null).isAccessAllowed(req)) {
+            if (!new JWTRouteFilter(null, auth0Domain).isAccessAllowed(req)) {
```
SimoneMaiwald: why are we changing the gp routes token check?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/DSMAuthenticationException.java
```diff
@@ -0,0 +1,11 @@
+package org.broadinstitute.dsm.exception;
+
+public class DSMAuthenticationException extends RuntimeException {
```
SimoneMaiwald: you could remove DSM from the class name

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -1,101 +1,190 @@
...
 
-                long auth0Expiration = auth0UserInfo.getTokenExpiration();
-                int cookieAgeInSeconds = new Long(auth0Expiration - new Double(System.currentTimeMillis() / 1000d).intValue()).intValue();
+        return claims;
+    }
+
+    private String getNewAuth0TokenWithCustomClaims(Map<String, String> claims, String clientSecret, String clientId, String auth0Domain, String auth0Audience, String audienceNameSpace) throws DSMAuthenticationException {
+        String api = "/oauth/token";
+        String contentType = "application/x-www-form-urlencoded";
+        String clientCredentials = "client_credentials";
```
SimoneMaiwald: can we move these strings into statics as well or is there a reason why we do them here?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DSSTestingRoute.java
SimoneMaiwald: Thank you for the todo! otherwise I would have commented on having another test route in dsm…

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -490,10 +494,14 @@ protected void setupCustomRouting(@NonNull Config cfg) {
             throw new RuntimeException("appRoute was not configured correctly.");
         }
 
-        String jwtSecret = cfg.getString(ApplicationConfigConstants.BROWSER_JWT_SECRET);
         String cookieSalt = cfg.getString(ApplicationConfigConstants.BROWSER_COOKIE_SALT);
         String cookieName = cfg.getString(ApplicationConfigConstants.BROWSER_COOKIE_NAME);
```
SimoneMaiwald: I think they can get removed now?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/CookieUtil.java
```diff
@@ -36,9 +36,9 @@ private String hashToken(String token, byte[] salt) throws NoSuchAlgorithmExcept
         return new String(Base64.getEncoder().encode(hash));
     }
 
-    public boolean isCookieValid(String cookieValue, byte[] cookieSalt, String jwtToken, String tokenSecret)
+    public boolean isCookieValid(String cookieValue, byte[] cookieSalt, String jwtToken, String auth0Domain)
```
SimoneMaiwald: are we still using the cookie method?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
```diff
@@ -60,19 +60,23 @@ public Auth0Util(@NonNull String account, @NonNull List<String> connections, boo
...
-            isEmailVerified = value == null ? false : value;
-        }
-        Auth0UserInfo userInfo = new Auth0UserInfo(auth0Claims.get("email").asString(), auth0Claims.get("exp").asInt(),
-                isEmailVerified);
-        verifyUserConnection(auth0Claims.get("sub").asString(), userInfo.getEmail());
+    public Auth0UserInfo getAuth0UserInfo(@NonNull String idToken, String auth0Domain) throws DSMAuthenticationException {
+        try {
+            Map<String, Claim> auth0Claims = verifyAndParseAuth0TokenClaims(idToken, auth0Domain);
+            boolean isEmailVerified = false;
+            Claim emailVerifiedClaim = auth0Claims.getOrDefault("email_verified", null);
```
SimoneMaiwald: do we care if email is verified?

pegahtah: not really, we just extract it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/CookieUtil.java
pegahtah: I will remove this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: I wasn't sure if we will need them back , I will remove

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
SimoneMaiwald: all dsm users are unverified, maybe we should change the auth0util to not care for that anymore

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/JWTRouteFilter.java
```diff
@@ -38,27 +41,23 @@
...
      * @param allowedRoles If given, at least one of the roles in the list
      *                     must be present in the roles claim in the token.
      *                     If empty or null, the roles claim is not checked.
      */
-    public JWTRouteFilter(String jwtSecret, Collection<String> allowedRoles) {
-        if (StringUtils.isBlank(jwtSecret)) {
-            throw new IllegalArgumentException("jwtSecret is required");
-        }
-        this.jwtSecret = jwtSecret;
+    public JWTRouteFilter(Collection<String> allowedRoles, String auth0Domain, String bspSecret) {
```
SimoneMaiwald: JWTRouteFilter is always called with allowedRoles = null, so you could just remove that from the class

pegahtah: all removed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/Auth0Util.java
pegahtah: I removed it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -24,78 +35,158 @@
...
+            params.add(new BasicNameValuePair(finalKey, claims.get(key)));
         }
+        params.add(new BasicNameValuePair(clientIdKey, clientId));
+        params.add(new BasicNameValuePair(grantTypeKey, grantType));
+        params.add(new BasicNameValuePair(clientSecretKey, clientSecret));
+        params.add(new BasicNameValuePair(audienceKey, audience));
+        return params;
     }
 
     private static class DSMToken {
```
SimoneMaiwald, line 192: to make it easier, do we still need the DSMToken class? could we just return the string from method `getNewAuth0TokenWithCustomClaims`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/ApplicationConfigConstants.java
```diff
@@ -35,10 +35,8 @@
 
     //security information
     public static final String BSP_SECRET = "bsp.secret";
-    public static final String BROWSER_JWT_SECRET = "browser_security.jwt_secret";
     public static final String BROWSER_COOKIE_SALT = "browser_security.cookie_salt";
     public static final String BROWSER_COOKIE_NAME = "browser_security.cookie_name";
```
SimoneMaiwald, line 39: I think they can be removed now as well

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -453,17 +502,19 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
         }
 
         //  capture basic route info for logging
-        before("*", new LoggingFilter());
+        before("*", new LoggingFilter(auth0Domain, auth0claimNameSpace));
         afterAfter((req, res) -> MDC.clear());
 
         before(API_ROOT + "*", (req, res) -> {
-            if (!new JWTRouteFilter(bspSecret, null).isAccessAllowed(req)) {
+            if (!new JWTRouteFilter(null, auth0Domain, bspSecret).isAccessAllowed(req, false)) {
```
SimoneMaiwald: JWTRouteFilter constructor only wants 2 values

SimoneMaiwald (dismissed):

SimoneMaiwald (approved):

PR #1200, opened 2022-02-24
DDP-7424: remove naming ability

SimoneMaiwald:
## Context

study staff can not set column name anymore. DSM will make that automatically in pascal snake case with a counter if it already exists

like test_field and test_field_1

changed FieldSettingsRoute into two separate classes


SimoneMaiwald:
column_name will be now
like TEST_FIELD and TEST_FIELD_1

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/FieldSettings.java
```diff
@@ -139,10 +143,12 @@ public static void saveFieldSettings(@NonNull String realm, @NonNull Map<String,
             for (FieldSettings fieldSetting : settingsOfType) {
                 String settingId = fieldSetting.getFieldSettingId();
                 try {
-                    if (StringUtils.isNotBlank(settingId)) {
-                        updateFieldSetting(settingId, fieldSetting, userId);
-                    } else {
-                        addFieldSetting(realm, fieldSetting, userId);
+                    if (StringUtils.isNotBlank(fieldSetting.columnDisplay)) {
+                        if (StringUtils.isNotBlank(settingId) && !fieldSetting.isAddedNew() && fieldSetting.isChanged()) {
```
gmakharat, line 147: extracting if condition into method with meaningful name will help for quick understanding in future.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/FieldSettings.java
```diff
@@ -259,4 +276,26 @@ private static void addFieldSetting(@NonNull String realm, @NonNull FieldSetting
         return values;
     }
 
+    private static String getUniqueFieldName(Collection<FieldSettings> settingsOfType, String columnDisplay) {
```
gmakharat, line 279: I think we need to restrict this from frontend as well, because there is a chance that someone may enter very different pattern and we don't/should't handle all of them, snakeCase and UNDER_scores are allowed, I think instead of transforming here we just need to add restriction on frontend and also backend side that if someone entered different type pattern we should response with error message that such pattern is not allowed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/settings/GetFieldSettingsRoute.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Object processRequest(Request request, Response response, String userId) throws Exception {
+        String realm = request.params(RequestParameter.REALM);
+        if (StringUtils.isBlank(realm)) {
+            throw new IllegalArgumentException("Realm cannot be empty");
+        }
+
+        String userIdRequest = UserUtil.getUserId(request);
+        if (RoutePath.RequestMethod.GET.toString().equals(request.requestMethod())) {
```
gmakharat, line 28: I am getting bad feeling about checking whether request method is `GET` that in future this class may get very ugly, that someone will follow this pattern and use this class for `POST` method and we will dive into ugly, stair like `if-else` statements.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/settings/GetFieldSettingsRoute.java
```diff
@@ -0,0 +1,41 @@
...
+            if (UserUtil.checkUserAccess(realm, userId, "mr_view", userIdRequest) || UserUtil.checkUserAccess(realm, userId, "pt_list_view",
+                    userIdRequest)) {
+                return FieldSettings.getFieldSettings(realm);
+            } else {
+                response.status(500);
+                return new Result(500, UserErrorMessages.NO_RIGHTS);
```
gmakharat, line 34: I remember that we agreed not to use `Result` anymore?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -761,9 +762,8 @@ private void setupMedicalRecordRoutes(@NonNull Config cfg, @NonNull Notification
 
         get(UI_ROOT + RoutePath.LOOKUP, new LookupRoute(), new JsonTransformer());
 
-        FieldSettingsRoute fieldSettingsRoute = new FieldSettingsRoute();
-        get(UI_ROOT + RoutePath.FIELD_SETTINGS_ROUTE, fieldSettingsRoute, new JsonTransformer());
-        patch(UI_ROOT + RoutePath.FIELD_SETTINGS_ROUTE, fieldSettingsRoute, new JsonTransformer());
+        get(UI_ROOT + RoutePath.FIELD_SETTINGS_ROUTE, new GetFieldSettingsRoute(), new JsonTransformer());
+        patch(UI_ROOT + RoutePath.FIELD_SETTINGS_ROUTE, new PatchFieldSettingsRoute(), new JsonTransformer());
```
gmakharat, line 766: just saw these, I believe we don't need anymore to check in classes whether request is `GET` or `PATCH`, it will be handled automatically by this.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/settings/PatchFieldSettingsRoute.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        String userIdRequest = UserUtil.getUserId(request);
+        if (RoutePath.RequestMethod.PATCH.toString().equals(request.requestMethod())) {
+            if (UserUtil.checkUserAccess(realm, userId, "field_settings", userIdRequest)) {
+                try {
+                    String requestBody = request.body();
+                    Type settingsType = new TypeToken<Map<String, Collection<FieldSettings>>>() {
+                    }.getType();
+                    Map<String, Collection<FieldSettings>> fieldSettingsLists = new Gson().fromJson(requestBody, settingsType);
```
gmakharat, line 41: `ObjectMapperSingleton` here as well instead of `Gson`.

PR #1204, opened 2022-02-28
DDP-7610

SimoneMaiwald:
## Context

getting mr tab to display

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
@@ -20,12 +20,18 @@ import com.google.gson.annotations.SerializedName;
 import lombok.Data;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.db.dao.ddp.instance.DDPInstanceDao;
+import org.broadinstitute.dsm.db.dto.ddp.instance.DDPInstanceDto;
+import org.broadinstitute.dsm.db.dto.ddp.tissue.ESTissueRecordsDto;
 import org.broadinstitute.dsm.db.structure.ColumnName;
 import org.broadinstitute.dsm.db.structure.DbDateConversion;
 import org.broadinstitute.dsm.db.structure.SqlDateConverter;
 import org.broadinstitute.dsm.db.structure.TableName;
+import org.broadinstitute.dsm.model.elastic.export.Exportable;
+import org.broadinstitute.dsm.model.elastic.export.painless.UpsertPainlessFacade;
 import org.broadinstitute.dsm.model.patch.Patch;
 import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.statics.ESObjectConstants;
 import org.broadinstitute.dsm.statics.QueryExtension;
 import org.broadinstitute.dsm.util.DBUtil;
 import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
@@ -68,8 +74,7 @@ public class OncHistoryDetail {
             + "LEFT JOIN ddp_participant as p on (p.participant_id = inst.participant_id) "
             + "LEFT JOIN ddp_instance as ddp on (ddp.ddp_instance_id = p.ddp_instance_id) "
             + "LEFT JOIN ddp_medical_record as m on (m.institution_id = inst.institution_id AND NOT m.deleted <=> 1) "
-            + "LEFT JOIN ddp_onc_history_detail as oD on (m.medical_record_id = oD.medical_record_id) "
-            + "WHERE p.participant_id = ?";
+            + "LEFT JOIN ddp_onc_history_detail as oD on (m.medical_record_id = oD.medical_record_id) " + "WHERE p.participant_id = ?";
     public static final String STATUS_REVIEW = "review";
     public static final String STATUS_SENT = "sent";
     public static final String STATUS_RECEIVED = "received";
@@ -87,8 +92,6 @@ public class OncHistoryDetail {
     public static final String PROBLEM_OTHER = "other";
     public static final String PROBLEM_OTHER_OLD = "Other";
     private static final Logger logger = LoggerFactory.getLogger(OncHistoryDetail.class);
-    private static final String SQL_CREATE_ONC_HISTORY =
-            "INSERT INTO ddp_onc_history_detail SET medical_record_id = ?, request = ?, last_changed = ?, changed_by = ?";
     private static final String SQL_SELECT_ONC_HISTORY =
             "SELECT onc_history_detail_id, medical_record_id, date_px, type_px, location_px, histology, accession_number, facility,"
                     + " phone, fax, notes, additional_values_json, request, fax_sent, fax_sent_by, fax_confirmed, fax_sent_2, "
@@ -99,6 +102,7 @@ public class OncHistoryDetail {
             "SELECT tissue_received FROM ddp_onc_history_detail WHERE onc_history_detail_id = ?";
     private static final String SQL_INSERT_ONC_HISTORY_DETAIL =
             "INSERT INTO ddp_onc_history_detail SET medical_record_id = ?, request = ?, last_changed = ?, changed_by = ?";
+
     @ColumnName(DBConstants.ONC_HISTORY_DETAIL_ID)
     private long oncHistoryDetailId;
 
@@ -396,7 +400,7 @@ public class OncHistoryDetail {
         return (OncHistoryDetail) results.resultValue;
     }
 
-    public static String createNewOncHistoryDetail(@NonNull String medicalRecordId, @NonNull String changedBy) {
+    public static String createNewOncHistoryDetail(@NonNull String medicalRecordId, @NonNull String changedBy, Patch patch) {
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement stmt = conn.prepareStatement(SQL_INSERT_ONC_HISTORY_DETAIL, Statement.RETURN_GENERATED_KEYS)) {
@@ -430,7 +434,18 @@ public class OncHistoryDetail {
             throw new RuntimeException("Error adding new oncHistoryDetail for medicalRecord w/ id " + medicalRecordId,
                     results.resultException);
         } else {
-            return (String) results.resultValue;
+            String oncHistoryDetailId = (String) results.resultValue;
+            if (StringUtils.isNotBlank(oncHistoryDetailId)) {
+                DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(patch.getRealm()).orElseThrow();
+                String participantGuid = Exportable.getParticipantGuid(patch.getDdpParticipantId(), ddpInstanceDto.getEsParticipantIndex());
+                ESTissueRecordsDto esTissueRecordsDto =
+                        new ESTissueRecordsDto(patch.getDdpParticipantId(), Integer.getInteger(oncHistoryDetailId), null, null, null, null,
+                                null, null, null, null);
+
+                UpsertPainlessFacade.of(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS, esTissueRecordsDto, ddpInstanceDto,
+                        ESObjectConstants.TISSUE_RECORDS_ID, ESObjectConstants.DOC_ID, participantGuid).export();
+            }
+            return oncHistoryDetailId;
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
@@ -22,7 +22,6 @@ import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.dao.ddp.instance.DDPInstanceDao;
 import org.broadinstitute.dsm.db.dto.ddp.instance.DDPInstanceDto;
-import org.broadinstitute.dsm.db.dto.ddp.tissue.ESTissueRecordsDto;
 import org.broadinstitute.dsm.db.structure.ColumnName;
 import org.broadinstitute.dsm.db.structure.DbDateConversion;
 import org.broadinstitute.dsm.db.structure.SqlDateConverter;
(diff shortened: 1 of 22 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -100,8 +97,13 @@ private void fetchAndPrepareDataByFilters(DDPInstance ddpInstance, Map<String, S
                 }
             }
         }
-        esData = elasticSearchable.getParticipantsByRangeAndFilter(ddpInstance.getParticipantIndexES(), participantWrapperPayload.getFrom(),
+        esData = elasticSearchable.getParticipantsByRangeAndFilter(participantIndexES, participantWrapperPayload.getFrom(),
                 participantWrapperPayload.getTo(), boolQueryBuilder);
+        if (StringUtils.isNotBlank(userIndexES)) {
```
SimoneMaiwald, line 102: adding getting proxy data here as well @gmakharat @Anzori-Ghurtchumelia

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -111,122 +113,68 @@ private boolean isUnderDsmKey(String source) {
...
                 participantWrapperDto.setAbstractionActivities(Collections.emptyList());
                 participantWrapperDto.setAbstractionSummary(Collections.emptyList());
 
-                result.add(participantWrapperDto);
+                List<KitRequestShipping> kitRequestShipping = esDsm.getKitRequestShipping();
+                participantWrapperDto.setKits(kitRequestShipping);
 
+                ESProfile esProfile = elasticSearchParticipantDto.getProfile().orElseThrow();
+                String ddpParticipantId = esProfile.getGuid();
+                if (proxiesByParticipantIds != null) {
```
SimoneMaiwald, line 171: using the map of proxies here from the fetch… methods to not ask ES for user info again @gmakharat @Anzori-Ghurtchumelia

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/MedicalRecordUtil.java
```diff
@@ -10,58 +10,81 @@
...
+                }
+                logger.info("Added new medical record for institution w/ id " + institutionId);
             } catch (SQLException e) {
                 throw new RuntimeException("Error inserting new medical record ", e);
             }
         } else {
             throw new RuntimeException("DB connection was null");
         }
+
+        if (mrId != null) {
```
SimoneMaiwald, line 77: adding upsert to ES here to add the mr object to ES even if it is empty except of the medicalRecordId

SimoneMaiwald: @gmakharat @Anzori-Ghurtchumelia

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -430,7 +433,18 @@ public static String createNewOncHistoryDetail(@NonNull String medicalRecordId,
             throw new RuntimeException("Error adding new oncHistoryDetail for medicalRecord w/ id " + medicalRecordId,
                     results.resultException);
         } else {
-            return (String) results.resultValue;
+            String oncHistoryDetailId = (String) results.resultValue;
```
SimoneMaiwald, line 436: adding upsert here to add new oncHistoryDetail object to ES even if it is empty and only has oncHistoryDetailId
@gmakharat @Anzori-Ghurtchumelia

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -430,7 +433,18 @@ public static String createNewOncHistoryDetail(@NonNull String medicalRecordId,
...
+            if (StringUtils.isNotBlank(oncHistoryDetailId)) {
+                DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(patch.getRealm()).orElseThrow();
+                String participantGuid = Exportable.getParticipantGuid(patch.getDdpParticipantId(), ddpInstanceDto.getEsParticipantIndex());
+                OncHistoryDetail oncHistoryDetail = new OncHistoryDetail();
+                oncHistoryDetail.setOncHistoryDetailId(Integer.getInteger(oncHistoryDetailId));
+                oncHistoryDetail.setDdpParticipantId(patch.getDdpParticipantId());
+
+                UpsertPainlessFacade.of(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS, oncHistoryDetail, ddpInstanceDto,
+                        ESObjectConstants.TISSUE_RECORDS_ID, ESObjectConstants.DOC_ID, participantGuid).export();
```
gmakharat, line 445: should it be `oncHistoryDetailId` instead of tissue?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -396,7 +399,7 @@ public static OncHistoryDetail getOncHistoryDetail(@NonNull String oncHistoryDet
         return (OncHistoryDetail) results.resultValue;
     }
 
-    public static String createNewOncHistoryDetail(@NonNull String medicalRecordId, @NonNull String changedBy) {
+    public static String createNewOncHistoryDetail(@NonNull String medicalRecordId, @NonNull String changedBy, Patch patch) {
```
gmakharat, line 402: this method is called only by `OncHistoryDetailPatch`, since patch already does upsert to ES, do we need to add upsert complexity to this method as well?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
gmakharat: @SimoneMaiwald I think it will be good idea if you pull sorting branch here, because I think proxies problem is solved already on that branch.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -111,122 +113,68 @@ private boolean isUnderDsmKey(String source) {
...
                 List<OncHistoryDetail> oncHistoryDetails = esDsm.getOncHistoryDetail();
                 List<Tissue> tissues = esDsm.getTissue();
                 mapTissueToProperOncHistoryDetail(oncHistoryDetails, tissues);
-
-                List<KitRequestShipping> kitRequestShipping = esDsm.getKitRequestShipping();
-
-                List<String> proxyGuids = elasticSearchParticipantDto.getProxies();
-                String usersIndexES = ddpInstance.getUsersIndexES();
-                ElasticSearch participantsByIds = elasticSearchable.getParticipantsByIds(usersIndexES, proxyGuids);
-                List<ElasticSearchParticipantDto> proxies = participantsByIds.getEsParticipants();
```
gmakharat, line 213: piece of code from L210 to L214 is responsible to get proxies data in both case whether participant list is requested with filter or without filter, I don't see the point of taking this out from here and putting separately.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -111,122 +113,68 @@ private boolean isUnderDsmKey(String source) {
...
     }
 
-    private void fetchAndPrepareData(DDPInstance ddpInstance) {
-        esData = elasticSearchable.getParticipantsWithinRange(ddpInstance.getParticipantIndexES(), participantWrapperPayload.getFrom(),
+    private void fetchAndPrepareData(String participantIndexES, String userIndexES) {
+        logger.info("Fetch participant data...");
+        esData = elasticSearchable.getParticipantsWithinRange(participantIndexES, participantWrapperPayload.getFrom(),
                 participantWrapperPayload.getTo());
-        participants = getParticipantsFromEsData();
-        if (ddpInstance.isHasRole()) {
```
gmakharat, line 118: don't we care anymore about roles?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
gmakharat: what's the point? `ParticipantWrapper` class fully follows object oriented idea, on every request new object of `ParticipantWrapper` is created, thus every field of it will be re-created.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -242,11 +190,11 @@ private void mapTissueToProperOncHistoryDetail(List<OncHistoryDetail> oncHistory
 
     void sortBySelfElseById(Collection<List<ParticipantData>> participantDatas) {
         participantDatas.forEach(pDataList -> pDataList.sort((o1, o2) -> {
-            Map<String, String> participantData = new Gson().fromJson(o1.getData().orElse(StringUtils.EMPTY),
-                    new TypeToken<Map<String, String>>() {
-                }.getType());
-            if (Objects.nonNull(participantData)
-                    && FamilyMemberConstants.MEMBER_TYPE_SELF.equals(participantData.get(FamilyMemberConstants.MEMBER_TYPE))) {
+            Map<String, String> participantData =
+                    new Gson().fromJson(o1.getData().orElse(StringUtils.EMPTY), new TypeToken<Map<String, String>>() {
```
gmakharat, line 194: we introduced `ObjectMapperSingleton` class to replace `Gson` it would be good idea to eliminate `Gson` from the project, because during this time of refactoring we understood that `Gson` has some problems, for example it gets every number as a double, does not matter number is integer, long, etc, thus it made problems to us and we decided not to use it anymore, `ObjectMapperSingleton` uses `Jackson` which is better alternative of `Gson`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -266,7 +214,7 @@ void sortBySelfElseById(Collection<List<ParticipantData>> participantDatas) {
     }
 
     Map<String, List<ElasticSearchParticipantDto>> getProxiesWithParticipantIdsByProxiesIds(String esUsersIndex,
-                                                      Map<String, List<String>> proxiesIdsByParticipantIds) {
+                                                                                            Map<String, List<String>> proxiesIdsByParticipantIds) {
```
gmakharat, line 217: I think it is better to discuss about proxies once you review PR of sorting.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateBSPDummyKitRoute.java
```diff
@@ -31,22 +32,24 @@ public Object handle(Request request, Response response) throws Exception {
         logger.info("Found kitlabel " + kitLabel + " in Mercury request");
         int ddpInstanceId = (int) DBUtil.getBookmark(DUMMY_REALM_NAME);
         logger.info("Found ddp instance id for mock test " + ddpInstanceId);
-        DDPInstance mockDdpInstance = DDPInstance.getDDPInstanceById(ddpInstanceId);
-        if (mockDdpInstance != null) {
-            logger.info("Found mockDdpInstance " + mockDdpInstance.getName());
+        DDPInstanceDto mockDdpInstanceDto =
```
gmakharat, line 35: if this is dummy route and will never be used, why should we add complexity of saving all this dummy instance names into database and fetching it and etc? we can just mock it here directly using Builder of `DDPInstanceDto`, or I did not get the idea of this route?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -344,14 +345,23 @@ public static ElasticSearchParticipantDto getElasticSearchForGivenMatch(String i
         return esData;
     }
 
-    public static Map<String, Map<String, Object>> getFilteredDDPParticipantsFromES(@NonNull DDPInstance instance, @NonNull String filter) {
-        String index = instance.getParticipantIndexES();
-        if (StringUtils.isNotBlank(index)) {
+    public static Map<String, Map<String, Object>> getFilteredDDPParticipantsFromES(@NonNull DDPInstanceDto ddpInstanceDto,
```
gmakharat, line 348: main idea of creating `ElasticSearchable` interface was to eliminate `ElasticSearchUtil` class, of course it can't be done immediately, but every time we need to add/edit method which will work with ES, it is better to use `ElasticSearchable` interface and its subclass `ElasticSearch` or create new subclass, depends what task requires.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/MedicalRecordUtil.java
```diff
@@ -173,19 +176,16 @@ public static void writeInstitutionIntoDb(@NonNull Connection conn, @NonNull Str
         }
     }
 
-    private static void insertInstitution(@NonNull Connection conn, @NonNull PreparedStatement insertInstitution, @NonNull String id,
-                                          boolean setDuplicateFlag) {
+    private static void insertInstitution(@NonNull Connection conn, @NonNull PreparedStatement insertInstitution, String participantId,
+                                          String ddpParticipantId, String instanceName) {
```
gmakharat, line 180: what is the difference between `participantId` and `ddpParticipantId`? does `participantId` come from some internal table which keeps data of participants?

pepper-apis/dsm-core/src/main/resources/application.conf
```diff
@@ -210,7 +215,92 @@ portal {
...
+            and inRol.instance_role_id = role.instance_role_id
+            and role.name = ?
+            and realm.ddp_instance_id = inst.ddp_instance_id) as 'has_role'
+        from
+        ddp_instance inst,
+        ddp_kit_request_settings kSetting,
+        kit_type kType
+        where inst.ddp_instance_id = kSetting.ddp_instance_id
+        and kType.kit_type_id = kSetting.kit_type_id
+        and inst.is_active = 1
```
gmakharat, line 303: this sql looks very very dangerous :sweat_smile:

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
SimoneMaiwald: here it is getting the proxy data for each participant seperately, which could mean worst case: for a pt list to show 50 pt -> 50 calls to the user index to get proxy data. if we use the map we already have from L125 we need only 1 call and then get the proxy data from the map

SimoneMaiwald:
will rebase to sorting branch

PR #1207, opened 2022-03-01
changing service

SimoneMaiwald:
## Context

DSM from the "old" repo can be stopped and DSM from the "new" repo should get that service name now


MocanaAtBroad (approved):
I think that really is it!

PR #1210, opened 2022-03-02
DDP-7620_StackDriverAndMissingCode

pegahtah:
Added the missing code that was throwing error

### DIFF
--- a/pepper-apis/dsm-core/pom.xml
+++ b/pepper-apis/dsm-core/pom.xml
@@ -251,10 +251,6 @@
             <groupId>com.brsanthu</groupId>
             <artifactId>google-analytics-java</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.brsanthu</groupId>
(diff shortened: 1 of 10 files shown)


pepper-apis/dsm-core/pom.xml
```diff
@@ -247,6 +247,14 @@
...
             <!-- <version>2.8.5</version> -->
         </dependency>
+        <dependency>
+            <groupId>com.brsanthu</groupId>
+            <artifactId>google-analytics-java</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.brsanthu</groupId>
+            <artifactId>google-analytics-java</artifactId>
+        </dependency>
```
SimoneMaiwald: double or am I missing something?

pegahtah: oh yes, I think intellij added it while importing too, will fix now

SimoneMaiwald (approved):

PR #1217, opened 2022-03-04
Ddp 7534 file answer one to many

GiCharkviani:


MocanaAtBroad (changes requested):
Please add description of what this PR is supposed to be using and and comments to allow us to understand how you are trying to solve the problem.

LKapanad:
This PR adds support for storing multiple files per answer.
There was a one-to-one relationship between answer and file_upload in file_answer table. 
DDP-7534-file-answer-one-to-many.xml liquibase script contains the following modification:
Create temporary file_answer_temp table with the updated schema
Copy file_answer data -> file_answer_temp
drop file_answer table
create file_answer with the updated schema
copy file_answer_temp data -> file_answer
delete file_answer_temp table.
Java code is updated according to the database change, instead of storing the reference to single file in answer object, we store the List of files.

LKapanad:
> Please add description of what this PR is supposed to be using and and comments to allow us to understand how you are trying to solve the problem.



> 



> Please add description of what this PR is supposed to be using and and comments to allow us to understand how you are trying to solve the problem.

This PR adds support for storing multiple files per answer.
There was a one-to-one relationship between answer and file_upload in file_answer table.
[DDP-7534](https://broadinstitute.atlassian.net/browse/DDP-7534)-file-answer-one-to-many.xml liquibase script contains the following modification:
Create temporary file_answer_temp table with the updated schema
Copy file_answer data -> file_answer_temp
drop file_answer table
create file_answer with the updated schema
copy file_answer_temp data -> file_answer
delete file_answer_temp table.
Java code is updated according to the database change, instead of storing the reference to single file in answer object, we store the List of files.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -149,9 +149,11 @@ private void createAnswerValue(long operatorId, long instanceId, long answerId,
     }
 
     private void createAnswerFileValue(long answerId, FileAnswer answer) {
-        FileInfo info = answer.getValue();
-        Long uploadId = info == null ? null : info.getUploadId();
-        DBUtils.checkInsert(1, getAnswerSql().insertFileValue(answerId, uploadId));
+        List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
```
Trimax, line 155: According to line 152, it is possible that the answer can have null as the value. That means that it's better to check it here as well since answer.getValue().stream() might produce NPE.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -1036,9 +1036,9 @@ private QuestionRecord createRecordForQuestion(QuestionDef question, FormRespons
             DateValue value = (DateValue) answer.getValue();
             return new DateQuestionRecord(question.getStableId(), value);
         } else if (answer.getQuestionType() == QuestionType.FILE) {
-            FileInfo info = ((FileAnswer) answer).getValue();
-            String uploadGuid = info != null ? info.getUploadGuid() : null;
-            return new SimpleQuestionRecord(type, question.getStableId(), uploadGuid);
+            List<FileInfo> fileInfos = ((FileAnswer) answer).getValue();
+            List<Long> uploadIds = fileInfos.stream().map(FileInfo::getUploadId).collect(Collectors.toList());
```
Trimax: Please add not null check here

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -608,17 +608,21 @@ private DateAnswer convertDateAnswer(String stableId, String guid, String actIns
...
+        if (isNull || value.isJsonArray()) {
+            List<FileInfo> fileInfos = new ArrayList<>();
             if (!isNull) {
-                String uploadGuid = value.getAsString();
-                info = handle.attach(FileUploadDao.class).findFileInfoByGuid(uploadGuid).orElse(null);
-                if (info == null) {
-                    throw ResponseUtil.haltError(response, 400, new ApiError(ErrorCodes.FILE_ERROR,
-                            "Could not find file upload with guid " + uploadGuid));
+                for (JsonElement element : value.getAsJsonArray()) {
+                    FileInfo info = handle.attach(FileUploadDao.class)
```
Trimax: Let's move handle.attach our of the cycle.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PatchFormAnswersRoute.java
```diff
@@ -762,29 +766,31 @@ private void checkAnswerConstraints(Response response, Answer answer) {
...
-                        "File is infected and cannot be associated with answer"));
-            case SIZE_MISMATCH:
-                throw ResponseUtil.haltError(response, 400, new ApiError(ErrorCodes.FILE_ERROR,
-                        "File uploaded size does not match expected size"));
-            case OK:
-                LOG.info("File upload with id {} is uploaded and can be associated with participant's answer", uploadId);
-                break;
-            default:
-                throw new DDPException("Unhandled file check result: " + verifyResult);
+        List<FileInfo> fileInfos = answer.getValue();
```
Trimax, line 769: Let's break down this method into two separate. Since you're just adding a cycle, you are able to reuse the previously defined method.
For instance:
`verifyFileUpload() {
  for (FileInfo info : fileInfos)
    verifyFileInfo(info);
}`

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java
```diff
@@ -1852,10 +1852,10 @@ public void testPatch_fileAnswer_createAndUpdate() {
         assertEquals(guid, answer.get().getAnswerGuid());
         assertEquals(stableId, answer.get().getQuestionStableId());
         assertEquals(QuestionType.FILE, answer.get().getQuestionType());
-        assertEquals(upload1.getFileName(), ((FileAnswer)answer.get()).getValue().getFileName());
-        assertEquals(upload1.getFileSize(), ((FileAnswer)answer.get()).getValue().getFileSize());
+        assertEquals(upload1.getFileName(), ((FileAnswer)answer.get()).getValue().get(0).getFileName());
```
Trimax: We can extend FileAnswer with getInfo(int index) method. This will simplify the usage. Instead of .getValue().get(i) we will use .getInfo(i). And inside this method we can easily introduce not null check

study-builder/studies/singular/activities/medical-record-file-upload.conf
```diff
@@ -68,6 +68,18 @@
                 },
               ]
             },
+            "tooltipTemplate": {
+               "templateType": "TEXT",
+               "templateText": "$add_participant_incapacitated_tooltip",
+               "variables": [
+                  {
+                     "name": "add_participant_incapacitated_tooltip",
+                     "translations":[
```
Trimax, line 77: Shouldn't we use new translations approach? @MocanaAtBroad, what do you think? Or at some point we will have kind of refactoring task for that?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7534-file-answer-one-to-many.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ssettipalli, line 1: All this to add a  new auto generated primarykey/ID and get rid of answer_id as PK. 

Can we evaluate option:
Delete existing FK, then  existing PK
and make combination PK (answer_id + file_upload_id) , assuming this combo is UNIQUE.

Other thing to check is:
May be deleting existing FK, then PK, adding new auto generated PK, filling it and later making it not null and PK can make not needing a temp table, deleting existing table , recreating table and updating from temp table.
This could avoid deleting prod table with prod data and recreating

ssettipalli:
Looks good to me

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -149,9 +149,12 @@ private void createAnswerValue(long operatorId, long instanceId, long answerId,
     }
 
     private void createAnswerFileValue(long answerId, FileAnswer answer) {
-        FileInfo info = answer.getValue();
-        Long uploadId = info == null ? null : info.getUploadId();
-        DBUtils.checkInsert(1, getAnswerSql().insertFileValue(answerId, uploadId));
+        if (answer.getValue() == null) return;
```
bskinner: This is a style violation- `if` constructs must use braces.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/FileAnswer.java
```diff
@@ -3,33 +3,40 @@
...
         this.value = value;
     }
 
     @Override
     public boolean isEmpty() {
         return value == null;
     }
+
+    public FileInfo getInfo(int index) {
+        if (value != null && value.size() > index) return value.get(index);
```
bskinner: Checkstyle violation here- `return` must be on a new line, and brace-less `if` constructs are not permitted.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/FileAnswer.java
```diff
@@ -3,33 +3,40 @@
...
+        throw new RuntimeException("File info with index " + index +" doesn't exist");
```
bskinner: Another checkstyle violation- the second `+` should have whitespace on either side

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -1036,9 +1037,11 @@ private QuestionRecord createRecordForQuestion(QuestionDef question, FormRespons
             DateValue value = (DateValue) answer.getValue();
             return new DateQuestionRecord(question.getStableId(), value);
         } else if (answer.getQuestionType() == QuestionType.FILE) {
-            FileInfo info = ((FileAnswer) answer).getValue();
-            String uploadGuid = info != null ? info.getUploadGuid() : null;
-            return new SimpleQuestionRecord(type, question.getStableId(), uploadGuid);
+            List<FileInfo> fileInfos = ((FileAnswer) answer).getValue();
+            List<Long> uploadIds = fileInfos == null?
```
bskinner: Another style violation. When using the ternary operator, `?` must be surrounded by whitespace and, if the statement spans multiple lines, the `?` element may only appear at the beginning of a line.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/FileAnswer.java
```diff
@@ -3,33 +3,40 @@
...
 
     @Override
-    public void setValue(FileInfo value) {
+    public void setValue(List<FileInfo> value) {
         this.value = value;
     }
 
     @Override
     public boolean isEmpty() {
         return value == null;
```
bskinner: Is this still accurate given the changes? `isEmpty` will return `false` even if `value` is an empty list.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/db/dao/AnswerDaoTest.java
```diff
@@ -292,17 +294,17 @@ public void testCreateUpdateDelete_file() {
                     testData.getStudyId(), testData.getUserId(), testData.getUserId(),
                     "blob", "mime", "file", 123L);
             fileDao.markVerified(upload.getId());
-            var info = fileDao.findFileInfoByGuid(upload.getGuid()).get();
+            FileInfo info = fileDao.findFileInfoByGuid(upload.getGuid()).get();
```
bskinner: The only change on this line is from `var` to `FileInfo`. Was `var` causing problems here?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7534-file-answer-one-to-many.xml
LKapanad: It's a nicer solution, thanks

ssettipalli (dismissed):
looks good to me

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/answer/FileInfo.java
```diff
@@ -4,6 +4,7 @@
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.Positive;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
```
MocanaAtBroad: This is not used (we don't use Jackson)

MocanaAtBroad (dismissed):

ssettipalli (approved):

PR #1226, opened 2022-03-09
DDP-7632. Derived question (API + grammar)

Trimax:
This PR contains:
1. Changes to the API schema
2. Expression grammar definition


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1791,6 +1792,26 @@ Question.Decimal:
...
+Question.Equation:
+  allOf:
+    - $ref: '#/Question'
+    - type: object
+      required:
+        - answers
+      properties:
+        expression:
+          type: string
+          description: the expression that must be computed.
```
bskinner: Why does the client need knowledge of the equation? As far as I understand, there shouldn't be anything for a client to do aside from present the value computed by the server.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1791,6 +1792,26 @@ Question.Decimal:
...
+        scale:
+          type: integer
+          description: the number of digits allowed to the right of the decimal point.
+          minimum: 0
+          default: 0
+        readonly: true
```
bskinner: This is not valid OpenAPI- a [Schema][schema] object is required here.


[schema]: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schema-object

bskinner (dismissed):

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/equation/lang/EquationParserTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class EquationParserTest {
+    private static final String EXPRESSION = "3.14 + 2.71 * (x + y ^ 2)";
+    private static final String TREE = "(expression (expression (atom (number 3.14))) + " +
+            "(expression (expression (atom (number 2.71))) * (expression ( (expression (expression (atom (variable x))) + " +
+            "(expression (expression (atom (variable y))) ^ (expression (atom (number 2))))) ))))";
```
bskinner: This is a checkstyle violation- the `+` characters should be placed immediately after a newline, not before it.
```suggestion
    private static final String TREE = "(expression (expression (atom (number 3.14))) + "
            + "(expression (expression (atom (number 2.71))) * (expression ( (expression (expression (atom (variable x))) + "
            + "(expression (expression (atom (variable y))) ^ (expression (atom (number 2))))) ))))";
```

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: I supposed that we might want to display the formula to let the client know what's going on and how we get the number.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/equation/lang/EquationParserTest.java
Trimax: Ok, will fix. Thanks

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Hm... Will check

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: I like the idea of including the formula and can see it being useful at some point. But we don't need it at this time, so let's leave it out.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: What was the goal you were trying to reach here?

Trimax: Let the user know that the question doesn't require any answer, e.g. the question is read-only.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: The formula will be stored in the DB anyway, but ok, I'll remove it from the definition

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: OpenAPI doesn't support specifying fixed values in schemas- using the `default` property is the closest I've gotten so far. ~Making a note in the `readonly` property schema's `description` field should be good enough.~

On further thought, let's skip mentioning the fixed value altogether. If we get into a weird situation where the server returns `false`, we're already in an error state as sending a `PATCH` for an `EQUATION`'s answer is undefined. A client should be able to handle that situation gracefully.

MocanaAtBroad:
Did we settle on how the result of the equation will be returned to the client after an answer is PATCHed? I see schema change includes the answer that would be returned with GET activity instance. But we need to get the answer to client immediately after that is entered.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: So, as far as I understand, we don't need readonly at all here. So I removed it

Trimax:
> Did we settle on how the result of the equation will be returned to the client after an answer is PATCHed? I see schema change includes the answer that would be returned with GET activity instance. But we need to get the answer to client immediately after that is entered.



> 

I just extended the patch response. Please have a look.

pepper-apis/pex-antlr/src/main/antlr4/org/broadinstitute/ddp/pex/lang/Equation.g4
```diff
@@ -0,0 +1,88 @@
...
+   : number
+   | variable
+   ;
+
+number
+   : NUMBER
+   ;
+
+variable
+   : VARIABLE
```
MocanaAtBroad, line 23: OK. An important use case (maybe the most important one in FON) that I have emphasized enough is that the calculation involves two or more child questions within a composite, e.g. some pseudo-modeling below (the shape of the model is the point I am bringing up here/disregard my syntax):

```
parent question guid: dimensions
  question type: composite
  child questions:
     child question 1
          question type: decimal
           guid: length
     child question 2
           question type: decimal
           guid: width
    child question 3
           question type: equation
           equationExpression: width * height
           guid: area
 
```
So for the equation, we need to be able to refer to other children within a composite.

Do keep in mind that a case that our composite questions allow is to have multiple rows of child answers, so:

dimensions
row 1: 1.6   3.7    5.92
row 2: 2.8  5.0    14.0

So it is important that both in the expression language and the serialization of the results in the PATCH response can clearly refer to the composite question/answer.

Take a look at PEX to see how it happens there.

Trimax: I see that the PEX grammar allows us to use constructions like ABC[234] to refer to the specific row. But it's not our case as far as I understand. When you specify the expression, you just need to refer to the question by its stable id. In case of multiple answers/rows we will need to compute the equation for each row correspondingly.

So, in your example, grammar should be the same, but the return value will contain two answers for each child question: 2 lengths, 2 widths, 2 computed areas.

Am I missing anything?

pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -32,6 +33,23 @@ ActivityAnswersPatchResponse:
...
+                - values
+              properties:
+                stableId:
+                  type: string
+                  description: equation question identifier that the answer is for (in case of composite, this is a child question id)
+                values:
+                  type: array
+                  description: the real computed values (in case of composite can have multiple values)
+                  items:
+                    $ref: 'schemas.yml#/DecimalNumber'
```
bskinner, line 52: +1 to reusing the DecimalNumber schema! Thank for walking us through this morning.

Have you checked the flattened schema to make sure the `$ref` reference is resolved properly? In prior versions of the tools, the path needed to be relative to, and descended from, the root OpenAPI document in order for the schemas to be properly resolved in the flattened file.

If using a file-relative reference like this is working and ReDoc is able to present it properly, that is _awesome_, and there's going to be some refactoring coming up.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1791,6 +1792,22 @@ Question.Decimal:
...
           items:
             $ref: '#/Answer.Decimal'
+Question.Equation:
+  allOf:
+    - $ref: '#/Question'
+    - type: object
+      required:
+        - answers
+      properties:
+        scale:
```
bskinner: `scale` should be required.

pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -32,6 +33,23 @@ ActivityAnswersPatchResponse:
                 answerGuid:
                   type: string
                   description: identifier for the answer
+          equations:
+            type: array
+            description: list of computed equations
```
bskinner: Minor tweak to the wording with the intent to make it more explicit about what's being included here.

```suggestion
            description: list of computed equation values
```

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1791,6 +1792,22 @@ Question.Decimal:
...
+          type: integer
```
bskinner, line 1803: Needs a `format: int64` property to be in-line with the other `scale` properties.

bskinner (changes requested):
Beyond a couple of minor changes and a question, it looks good!

pepper-apis/docs/specification/src/components/responses.yml
Trimax: Could you please advise how to check the flattened schema? I just had a look at the references in similar files and did it in the same way.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: It is not the scale that is the exponent. It is the number of digits after the comma to be displayed. Several lines above (in Decimal Question) it is defined as integer as well

Trimax: Maybe we can think about a better name to avoid names ambiguation?

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: It is not required In the decimal question definition. This means We'll display all the digits we store in the database.

pepper-apis/docs/specification/src/components/responses.yml
bskinner: If you `cd` into `pepper-apis/spec/documentation` and run `build.sh documentation`, both the rendered documentation and the flattened OpenAPI schema will be output to the `build` folder as `pepper.html`, and the flattened schema as `pepper.yaml`. If this schema appears properly in both locations this should be good!

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: I'd be inclined to label that an error in the Decimal question spec- at the very least; the scale is intrinsic to the Decimal question type (even if it has a default value set). There shouldn't be any cases where we are sending a response from the server in which we do not have a value for this property in the response.

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: > It is not the scale that is the exponent.

The scale is the exponent? It's `unscaled * base ^ scale`, and we're fixed in base 10.

I agree that we need a new naming for this, though, as is a confusing mix. The usage here seems to be setting the maximum value for the scale/number of digits to the right of the point. What would you suggest?

pepper-apis/docs/specification/src/components/responses.yml
Trimax: Wow, super. Didn't know that. Let me check

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: I guess I wasn't clear enough in my explanation.
We have **value** and **scale** in the definition of DecimalNumber type. They have an int64 type and are defined in another file. And those value and scale represent exactly what you described: `value * base ^ scale`.

The scale defined here is required by the front-end to cut off redundant digits after the comma. For instance, in the database, we store 123456789 as the value and 5 as the scale. If the scale defined here is not set, the front-end will render the number as 1234.56789. But if the scale is set to 2, for example, then the front-end will render the number as 1234.57 (or 1234.56) depending on rounding mode.

P.S. As for the exponent I mentioned. I was sure that in the English language **b** in expressions like **c = a ^ b** are is called the exponent. Google translate's it in this way as well. Could you please advise what is the right term I should use?

Trimax: The scale defined here might be renamed to **decimalPlacesCount**. I believe you can suggest a better name since you know what sounds better in English.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Well, the client doesn't need to set this value. This is for formatting/rendering only. This is the read-only property for the front-end. Based on this property value the front-end will render the exact number of the digits after the comma, no matter how many of them do we have.

Trimax: Discussed personally

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Discussed personally.

bskinner (approved):
Looks good! 👍

PR #1241, opened 2022-03-11
DDP-7636: adding missing send grid

SimoneMaiwald:
changed log level back to info
adding some more checkstyle fixes
bringing back the missing SendGridClient class but reduced a lot to only what DSM needs and no LDDP needs


pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -0,0 +1,125 @@
...
+public class SendGridClient implements EmailClient {
+
+    private static final Logger logger = LoggerFactory.getLogger(SendGridClient.class);
+
+    private static final String LOG_PREFIX = "SendGrid";
+
+    private String sendGridKey;
+    private JsonObject settings = null;
+
+    public void configure(String sendGridKey, JsonObject settings, @NonNull String portalUrl, @NonNull String environment) {
```
gmakharat: are `portalUrl` and `environment` used at all?

SimoneMaiwald: Good catch I forgot to rip these ones out! 👍

gmakharat (approved):

PR #1242, opened 2022-03-11
add script project

josemanuel167:
Add nodeJs script to generate picklist options from RxNorm .txt


ssettipalli:
Looks like script generated output option stableIDs are numerics like below
{
      "stableId": 1000000,
      "optionLabelTemplate": {
        "templateType": "TEXT",
        "templateText": "TRIBENZOR Oral Pill"
      }
    },
    {
      "stableId": 1000001,
      "optionLabelTemplate": {
        "templateType": "TEXT",
        "templateText": "amLODIPineHydrochlorothiazideOlmesartan Oral Pill"
      }
    },
    
I believe we need better stableIDs that reflect option text.
May be user one word or combo or two words capitalized TRIBENZOR_ORAL OR TRIBENZOR_ORAL_PILL

bskinner:
The `dist` directory seems to contain compiled code- does this need to be committed to the repository?

ssettipalli:
Script should take input / output file names as parameters and remove pushing actual input and output data files to repo.

josemanuel167:
@ssettipalli I updated the script to take input/output and removed them from repo. For stable IDs I'm not sure if we should create our owns, adding @MocanaAtBroad for help with this.
Removed /dist from repo cc..@bskinner

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,41 @@
+import * as XLSX from "xlsx";
+import { RxNormDrugOption } from "./models";
+import { parseToHOCON, writeFile } from "./utils/common";
+
+const inputFileName = 'RxTerms202202.txt';
+const outputFileName = 'output.txt'
```
bskinner: Let's take the files from the command line instead of hard-coding the paths.

bskinner (commented):
Looks good- adding two items to address from my end:
* Let's add a short README describing, at a high level, what the script does and how to invoke it.
* The input and output filenames should not be hard coded into the script.

scripts/ddp-RxTerms-script/src/index.ts
jm167: Sounds good. Added

jm167: Readme added

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      name: "files",
+      message: "choose RxNorm .txt file"
+    },
+    {
+      type: "input",
+      name: "outputName",
+      message: "Enter output file name. File will be generated on TXT extension"
+    }
+  ])
+  .then((answers: InputOutput) => {
```
bskinner: This is going to be difficult to incorporate into any sort of non-interactive environment- We should not be prompting for input. If it helps, `npm run` allows for command line arguments to be passed using the `npm run <command> [<args>]` format, where `<args>` is one or more key-value pairs. Any kv pairs provided in `args` will be exposed inside `process.env` under a property named `npm_config_$arg`. For example, `--foo=bar` will be exposed via `process.env.npm_config_foo` and be set to the value `bar`.

If this doesn't work, positional arguments via `process.argv` would be acceptable as well (`npm run start -- <inputPath> <outputPath>`).

Let me know if you have any questions.

josemanuel167: Added arguments from console and defaults

bskinner:
Thank you for the changes- let's get this in 🙂

bskinner (dismissed):
Looks good- let's get this in 🙂

scripts/ddp-RxTerms-script/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+  }
+}
```
MocanaAtBroad: Had a difficult time with this configuration. Please consider replacing with this more concise one that worked for me:
```json
{
  "compilerOptions": {
    "lib": ["es2021"],
    "module": "commonjs",
    "target": "es2021",
    "types": [
      "node"
    ],
    "outDir": "dist",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  }
}

File: scripts/ddp-RxTerms-script/package.json [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "tsc && node dist/index.js"
+  },
+  "author": "broad-institute",
+  "license": "ISC",
+  "devDependencies": {
+    "@types/node": "^17.0.21"
+  },
+  "dependencies": {
+    "hocon-parser": "^1.0.1",
```
MocanaAtBroad: You don't need hocon-parser. Have some suggetions in code.

scripts/ddp-RxTerms-script/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "tsc && node dist/index.js"
+  },
+  "author": "broad-institute",
+  "license": "ISC",
+  "devDependencies": {
+    "@types/node": "^17.0.21"
+  },
```
MocanaAtBroad, line 15: Let's add Typescript

```suggestion
      "typescript": "^4.4.4"
  },
```

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      name: drug_name
+      translations: [
+         ${translations.map(lang => `{language:  ${lang}  text:  "${current.optionLabelTemplate.templateText.replace(/'/g, "")}"}` )}
+      ]
+    }
+  ]  
+}},`;
+  return (prev += stringBuilder);
+}, "");
+
```
MocanaAtBroad, line 94: All the stuff above I don't think makes sense to do.
You are creating here a gigantic HOCON string, but then, when you write it out, you convert to an object (using HOCON parser) and THEN to JSON string. 
So complicated AND we throw the work away.

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+}));
+
+const pickListOptions = filteredRxNormOptions.map((option) => ({
+  stableId: option.stableId,
+  optionLabelTemplate: {
+    templateType: "TEXT",
+    templateText: option.displayName
+  }
+}));
+
```
MocanaAtBroad, line 53: I strongly suggest go directly from your rxNormArray to the list of picklistOptions that has the exact shape needed in the configuration and that is what you use to generate the JSON to export, e.g.:
```typescript
const pickListOptions = rxNormArray.map((drug) => ({
    stableId: drug.RXCUI.toString(),    // <--- I noticed we were exporting numbers instead of strings!
    optionLabelTemplate: {
        templateType: 'TEXT',
        templateText: '$drug_name',
        variables: [{
            name: 'drug_name',
            translations: languageCodes.map(lang => ({ language: lang, text: drug.DISPLAY_NAME.replace(/'/g, '') }))
        }]
    }
}));

Comment (MocanaAtBroad) [reply scripts/ddp-RxTerms-script/src/index.ts:53]: You can then do something like:

JSON.stringify({pickListOptions: pickListOptions}, null, 2) 

and that is the contents of the file. One iteration over data, no HOCON generation or parseHOCON necessary.

File: scripts/ddp-RxTerms-script/src/index.ts [hunk 8]
```diff
@@ -0,0 +1,53 @@
+import * as XLSX from "xlsx";
+import { RxNormDrugOption } from "./models";
+import { parseToHOCON, writeFile } from "./utils/common";
+
+const inputFileName = process.argv[2] || "RxTerms202202.txt";
+const outputFileName = process.argv[3] || "output.txt";
+
+const translations: string[] =  process.argv.filter(arg => ['en', 'es', 'ru'].includes(arg));
```
MocanaAtBroad: I think we are unnecessarily limiting the the codes that are acceptable.
Just do something like this:
```typescript
const languageCodes: string[] | void = process.argv[4]?.split(",");

if (!languageCodes) {
    console.error("Comma separated language separators were not provided");
    process.exit(1)
}

Now users can say en,es,fn,etc and we will use it.

File: scripts/ddp-RxTerms-script/src/index.ts [hunk 9]
```diff
@@ -0,0 +1,53 @@
+import * as XLSX from "xlsx";
+import { RxNormDrugOption } from "./models";
+import { parseToHOCON, writeFile } from "./utils/common";
+
+const inputFileName = process.argv[2] || "RxTerms202202.txt";
```
MocanaAtBroad, line 5: I recommend checking that the data file actually exists. So something like this:
```typescript

import * as fs from 'fs';

if (!fs.existsSync(inputFileName)) {
    console.error(`Input file with name ${inputFileName} does not exist`);
    process.exit(1);
}

File: scripts/ddp-RxTerms-script/src/utils/common.ts [hunk 10]
```diff
@@ -0,0 +1,12 @@
+export const parseToHOCON = require("hocon-parser");
```
MocanaAtBroad: Not needed. You can delete.

scripts/ddp-RxTerms-script/src/utils/common.ts
```diff
@@ -0,0 +1,12 @@
...
+const fs = require("fs");
+
+export const writeFile = (fileName: string, value: string) => {
+  return fs.writeFile(fileName, value, (err: string) => {
+    if (err) {
+      return console.log(err);
+    }
+    console.log("The file was saved!");
+  });
+};
```
MocanaAtBroad: Consider just getting rid of this file and move write file to index.ts . We are just calling fs.writefile. Up to you.

scripts/ddp-RxTerms-script/src/utils/common.ts
```diff
@@ -0,0 +1,12 @@
+export const parseToHOCON = require("hocon-parser");
+
+const fs = require("fs");
```
MocanaAtBroad: You can do this TypeScript style:
```typescript
import * as fs from 'fs';

File: scripts/ddp-RxTerms-script/src/index.ts [hunk 13]
```diff
@@ -0,0 +1,53 @@
...
+const outputFileName = process.argv[3] || "output.txt";
+
+const translations: string[] =  process.argv.filter(arg => ['en', 'es', 'ru'].includes(arg));
+
+const workBook = XLSX.readFile(inputFileName);
+
+const workSheet = workBook.Sheets[workBook.SheetNames[0]];
+
+const rxNormArray = XLSX.utils.sheet_to_json(workSheet) as RxNormDrugOption[];
```
MocanaAtBroad, line 31: The parsing seems to work well! The one problem I am seeing is that it converts the drug ids to numbers because they all look like integers. But we need strings for stable ids. Suggested way to handle it below in the suggest code.

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  ]  
+}},`;
+  return (prev += stringBuilder);
+}, "");
+
+
+writeFile(
+  outputFileName,
+  JSON.stringify(parseToHOCON(`picklistOptions : [${picklistString}]`), null, 2)
```
MocanaAtBroad: I suggest above we just build the needed Javascript object and we stringify/ no HOCON.

MocanaAtBroad (changes requested):

scripts/ddp-RxTerms-script/tsconfig.json
josemanuel167: This was a fresh file autogenerated, it was nice to know all available options but sure I can replace it

scripts/ddp-RxTerms-script/src/index.ts
josemanuel167: Not really. Hocon parser library allows to provide strings without commas or quotes: 
image
I parse it to HOCON just to double check, and add all needed symbols.
Not creating a HOCON's string particularly just a string,

scripts/ddp-RxTerms-script/src/utils/common.ts
josemanuel167: Please read explanation above

scripts/ddp-RxTerms-script/src/index.ts
josemanuel167: Sounds good, thanks for that. I'm not familiar with ocon and wanted make sure it had the right format.

scripts/ddp-RxTerms-script/src/utils/common.ts
josemanuel167: Agree, removed

scripts/ddp-RxTerms-script/tsconfig.json
MocanaAtBroad: The options above I mostly got from the nodejs development team. Just types is added.

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const translations: string[] =  process.argv.filter(arg => ['en', 'es', 'ru'].includes(arg));
+
+const workBook = XLSX.readFile(inputFileName);
+
+const workSheet = workBook.Sheets[workBook.SheetNames[0]];
+
+const rxNormArray = XLSX.utils.sheet_to_json(workSheet) as RxNormDrugOption[];
+
+const pickListOptions = rxNormArray.map((drug) => ({
```
ssettipalli: In the generated output it should be picklistOptions (small `l`)

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      templateType: 'TEXT',
+      templateText: '$drug_name',
+      variables: [{
+          name: 'drug_name',
+          translations: translations.map(lang => ({ language: lang, text: drug.DISPLAY_NAME.replace(/'/g, '') }))
+      }]
+  }
+}));
+
+fs.writeFile(outputFileName, JSON.stringify({pickListOptions}, null, 2), err => {
```
ssettipalli: In the generated output it should be picklistOptions (small `l`)

MocanaAtBroad (dismissed):

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -1,33 +1,89 @@
...
+  return text.replace(/'/g, "");
+}
+
+function generateStableID(drugDisplayName: string): string {
+  drugDisplayName = drugDisplayName
+    .trimEnd()
+    .replace(/[^a-zA-Z0-9 ]/g, "")
+    .replaceAll(" ", "_")
+    .toUpperCase();
+
```
MocanaAtBroad, line 65: I see that a lot of entries use the "\", e.g.:
```
Capsaicin/Lidocaine/Menthol/Methyl salicylate
```
Current implementation just gets rid of the "/". I think would be easier to read if we replace the / for a _ in the final string.
So would like to see something like:
```
CAPSAICIN_LIDOCAINE_MENTHOL_METHYL_SALICYLATE
```

MocanaAtBroad (commented):
Looks good. I am requesting a small enhancement.

scripts/ddp-RxTerms-script/src/index.ts
```diff
@@ -1,33 +1,89 @@
...
 const rxNormArray = XLSX.utils.sheet_to_json(workSheet) as RxNormDrugOption[];
 
-const picklistOptions = rxNormArray.map((drug) => ({
-  stableId: drug.RXCUI.toString(),
+const filteredRxNormsArray = groupBy(rxNormArray, (x) => x.DISPLAY_NAME).map(
+  (group) => group[0]
+);
+
+const picklistOptions = filteredRxNormsArray.map((drug) => ({
+  stableId: generateStableID(drug.DISPLAY_NAME),
```
MocanaAtBroad, line 38: I think you are going to have to check that the generated stable id is really unique/cannot count on the current transformation giving you a unique ids.

ssettipalli (approved):
tested it & looks good to me.

PR #1246, opened 2022-03-14
DDP-7185 remote auto complete

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7185

Features of new Picklist RENDER_MODE: REMOTE_AUTOCOMPLETE 
PL Options are NOT included in the activity response to client/angular
Clients invoke a new Route (GetOptionsForActivityInstanceQuestionRoute) to get PL options/suggestions that match the string. 
REMOTE_AUTOCOMPLETE picklist templates are retrieved from cached ActivityDef:QuestionDef and NO database query for each call to new route (GetOptionsForActivityInstanceQuestionRoute)
Picklist label template rendering happens during run time depending on users current language. May be render in All supported languages and cache it (WiIl plan for it later)
Elastic activity_definition export should include options
Patch should be able to query REMOTE_AUTOCOMPLETE options to create / update answers


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Comparator;
+import java.util.regex.Pattern;
+
+/**
+ * Comparator that can be used to sort results for
+ * the picklist option typeahead.  The general approach is favoring leftwards
+ * occurrences, in particular word starts, allowing for a starting
+ * parentheses.
+ */
+public class StringSuggestionTypeaheadComparator implements Comparator<String> {
```
ssettipalli, line 12: Separated String comparison logic into separate class so that it can reused.
CancerTypeaheadComparator and DrugTypeaheadComparator can be refactored to use this one.
All of them use same logic

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ *  This route returns the list of suggestions for the supplied Picklist Option label
+ */
+public class GetOptionsForActivityInstanceQuestion implements Route {
```
MocanaAtBroad: I think we always use  the suffix "Route" for Route implementation classes. Our test suite looks for Routes using the suffix, so it is important to stick with this convention.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
```diff
@@ -101,6 +101,8 @@
                 PathParam.USER_GUID, PathParam.STUDY_GUID, PathParam.INSTANCE_GUID);
         public static final String USER_ACTIVITY_UPLOADS = fmt(BASE + "/user/%s/studies/%s/activities/%s/uploads",
                 PathParam.USER_GUID, PathParam.STUDY_GUID, PathParam.INSTANCE_GUID);
+        public static final String USER_QUESTION_OPTIONS = fmt(BASE + "/user/%s/studies/%s/activities/%s/questions/%s/options",
```
ssettipalli, line 104: new route to get REMOTE_AUTOCOMPLETE picklist options

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -505,6 +506,7 @@ private static void start() throws MalformedURLException {
                 responseSerializer
         );
         post(API.USER_ACTIVITY_UPLOADS, new CreateUserActivityUploadRoute(fileUploadService), responseSerializer);
+        get(API.USER_QUESTION_OPTIONS, new GetOptionsForActivityInstanceQuestion(i18nContentRenderer), responseSerializer);
```
ssettipalli: new route to get REMOTE_AUTOCOMPLETE picklist options. Option labels should be rendered

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/collectors/PicklistQuestionFormatStrategy.java
```diff
@@ -73,7 +73,7 @@
         props.put("groups", groups);
 
         List<Object> options = new ArrayList<>();
-        for (PicklistOptionDef optionDef : definition.getPicklistOptions()) {
+        for (PicklistOptionDef optionDef : definition.getPicklistOptionsIncludingRemoteAutoComplete()) {
```
ssettipalli: Include REMOTE_AUTOCOMPLETE picklist options during elastic export

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/RemoteAutoCompleteResponse.java
```diff
@@ -0,0 +1,27 @@
+package org.broadinstitute.ddp.json;
+
+import com.google.gson.annotations.SerializedName;
+import org.broadinstitute.ddp.model.activity.instance.question.PicklistOption;
+
+import java.util.List;
+
+public class RemoteAutoCompleteResponse {
```
ssettipalli, line 8: RemoteAutoCompleteResponse as per ApiSpec

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/PicklistQuestionDef.java
```diff
@@ -121,6 +122,14 @@ public Template getPicklistLabelTemplate() {
     }
 
     public List<PicklistOptionDef> getPicklistOptions() {
+        if (renderMode == PicklistRenderMode.REMOTE_AUTOCOMPLETE) {
```
ssettipalli, line 129: By default donot include PL options if render mode is REMOTE_AUTOCOMPLETE

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/PicklistQuestionDef.java
```diff
@@ -121,6 +122,14 @@ public Template getPicklistLabelTemplate() {
...
+            return Collections.emptyList();
+        } else {
+            return picklistOptions;
+        }
+    }
+
+    public List<PicklistOptionDef> getPicklistOptionsIncludingRemoteAutoComplete() {
```
ssettipalli, line 144: Scenarios like elastic activity definition  export  and PATCH answers need REMOTE_AUTOCOMPLETE PL options included.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+
+            //load All options first
+            List<PicklistOption> allOptions = new ArrayList<>();
+            List<PicklistOption> suggestions;
+            JdbiPicklistOption jdbiPicklistOption = handle.attach(JdbiPicklistOption.class);
+            JdbiQuestion jdbiQuestion = handle.attach(JdbiQuestion.class);
+            Optional<Long> questionId = jdbiQuestion.findIdByStableIdAndInstanceGuid(questionStableId, instanceGuid);
+            List<PicklistOptionDto> optionDtos = jdbiPicklistOption.findAllActiveOrderedOptionsByQuestionId(questionId.get());
+            for (PicklistOptionDto optionDto : optionDtos) {
+                PicklistOption option = new PicklistOption(optionDto.getStableId(), optionDto.getStableId(),
```
MocanaAtBroad: I think first argument is groupStableId, to the option stable id

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
+package org.broadinstitute.ddp.route;
```
ssettipalli: new route to query from DB REMOTE_AUTOCOMPLETE picklist options for give question and string

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+
+        RemoteAutoCompleteResponse result = TransactionWrapper.withTxn(handle -> {
+
+            //load All options first
+            List<PicklistOption> allOptions = new ArrayList<>();
+            List<PicklistOption> suggestions;
+            JdbiPicklistOption jdbiPicklistOption = handle.attach(JdbiPicklistOption.class);
+            JdbiQuestion jdbiQuestion = handle.attach(JdbiQuestion.class);
+            Optional<Long> questionId = jdbiQuestion.findIdByStableIdAndInstanceGuid(questionStableId, instanceGuid);
+            List<PicklistOptionDto> optionDtos = jdbiPicklistOption.findAllActiveOrderedOptionsByQuestionId(questionId.get());
```
ssettipalli: load all options, render the option labels to compare and return matches

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+            for (PicklistOptionDto optionDto : optionDtos) {
+                PicklistOption option = new PicklistOption(optionDto.getStableId(), optionDto.getStableId(),
+                        optionDto.getOptionLabelTemplateId(), optionDto.getTooltipTemplateId(), optionDto.getDetailLabelTemplateId(),
+                        optionDto.isAllowDetails(), optionDto.isExclusive(), optionDto.isDefault());
+                allOptions.add(option);
+            }
+            // Render labels using current time to get latest templates.
+            long timestamp = Instant.now().toEpochMilli();
+            long langCodeId = preferredUserLanguage.getId();
+            renderer.bulkRenderAndApply(handle, allOptions, style, langCodeId, timestamp);
```
ssettipalli: render option labels

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+
+            if (StringUtils.isBlank(autoCompleteQuery)) {
+                LOG.info("Option suggestion query is blank, returning all results");
+                suggestions = allOptions.stream().limit(limit).collect(Collectors.toList());
```
ssettipalli: returns limit (first 100 default)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/PicklistOptionTypeaheadComparator.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Comparator;
+
+/**
+ * Comparator that can be used to sort results for
+ * the picklist option typeahead.  The general approach is favoring leftwards
+ * occurrences, in particular word starts, allowing for a starting
+ * parentheses.
+ */
+public class PicklistOptionTypeaheadComparator implements Comparator<PicklistOption> {
```
ssettipalli: Comparator to match and filter picklist options by search term

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
```diff
@@ -0,0 +1,72 @@
...
+     * Create a new one, using the given search term
+     * @param searchTerm a simple search term, provided by the
+     *                   user.  Not a regex.
+     */
+    public StringSuggestionTypeaheadComparator(String searchTerm) {
+        this.searchTerm = searchTerm.toUpperCase();
+    }
+
+    @Override
+    public int compare(String string1, String string2) {
```
ssettipalli, line 27: search logic to place matches with start of work first
same logic used in CancerTypeaheadComparator and DrugTypeaheadComparator

pepper-apis/dss-database/src/main/resources/db-changes/seed/DDP-7185-add-picklist-render-mode-remote-autocomplete.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="sampath" id="20220302-add-picklist-render-mode-remote-autocomplete">
```
ssettipalli, line 7: Add new RENDER MODE

study-builder/studies/snippets/picklist-question-single-remote-autocomplete.conf
```diff
@@ -0,0 +1,14 @@
+{
+  "questionType": "PICKLIST",
+  "selectMode": "SINGLE",
+  "renderMode": "REMOTE_AUTOCOMPLETE",
```
ssettipalli, line 4: PL Question with new RENDER_MODE

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
MocanaAtBroad: We want to avoid reading all the options into memory as the list can be quite large.
We need a lookup implementation that can query the database for matches without having to read every single option into memory to find them.
Let me know if you would like to chat about some implementation alternatives.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+        String participantGuid = request.params(RouteConstants.PathParam.USER_GUID);
+        String instanceGuid = request.params(RouteConstants.PathParam.INSTANCE_GUID);
+        String questionStableId = request.params(RouteConstants.PathParam.QUESTION_STABLE_ID);
+
+        DDPAuth ddpAuth = RouteUtil.getDDPAuth(request);
+        String operatorGuid = StringUtils.defaultIfBlank(ddpAuth.getOperator(), participantGuid);
+        boolean isStudyAdmin = ddpAuth.hasAdminAccessToStudy(studyGuid);
+        String autoCompleteQuery = request.queryParams(RouteConstants.QueryParam.TYPEAHEAD_QUERY);
+        String queryLimit = request.queryParams(RouteConstants.QueryParam.TYPEAHEAD_QUERY_LIMIT);
+        int limit = StringUtils.isNotBlank(queryLimit) ? Integer.valueOf(queryLimit) : DEFAULT_LIMIT;
```
MocanaAtBroad: Let's set a max limit on options, i.e. let's say our limit is 500 and they ask for 10000, we return an error. Don't want to let a single client kill us!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        // now sort the matches in a way that puts left-most matches near the top, favoring word start matches
+        List<PicklistOption> sortedSuggestions = new ArrayList<>();
+        optionMatches.stream().sorted(new PicklistOptionTypeaheadComparator(autoCompleteQuery)).limit(limit).forEach(option -> {
+            sortedSuggestions.add(option);
+        });
+
+        return sortedSuggestions;
+    }
```
MocanaAtBroad: Probably this whole thing can be a single lambda expression. Give it a shot!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
```diff
@@ -0,0 +1,72 @@
...
+        // bias search results so that matches at the start of the name are favored
+        if (string1StartsWithSearchTerm && string2StartsWithSearchTearm) {
+            return string1Name.compareTo(string2Name);
+        } else if (string1StartsWithSearchTerm) {
+            return -1;
+        } else if (string2StartsWithSearchTearm) {
+            return 1;
+        } else {
+            // if the search is not at the start of the term, prefer hits that are at start of words
+            Pattern startOfWordMatch = Pattern.compile("\\b(\\))*" + Pattern.quote(searchTerm));
```
MocanaAtBroad: Had no idea you could use the + operator on Pattern! Had to try it out.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
```diff
@@ -0,0 +1,72 @@
...
+                        return -1;
+                    } else if (string2MatchPosition > -1) {
+                        return 1;
+                    } else {
+                        return 0;
+                    }
+                }
+            }
+        }
+    }
```
MocanaAtBroad, line 72: Lot's of sorting criteria! This is good. Do check the the Regex though.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
MocanaAtBroad: I don't know what exactly the + operator does BUT you can probably avoid compiling the \b each time by assigning Pattern.compile("\\b(\\) expression to a static variable.

MocanaAtBroad: I don't think this pattern is doing what the comment says it should be doing
Screen Shot 2022-03-14 at 5 59 47 PM
Note that it matches "hello" but not the last example that has the word hello.

MocanaAtBroad (commented):
Have a few comments. I think the big one is that we need to redo implementation so that we don't have to read all the options into memory to find matches.
I understand that this requirement makes the implementation harder. It might make sense to fix whatever needs to be fixed here, merge it so Jose can do some integration testing and come back and reimplement the lookup. Let me know what you think.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestion.java
ssettipalli: How about just sticking to default as MAX limit too if the ask is more than default ?
Trying to avoid erroring out and sticking to default silently. 
If needed we can update APIDocs that max results will be default value of 100

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/StringSuggestionTypeaheadComparator.java
ssettipalli: Its clone of existing in-use DrugTypeaheadComparator and CancerTypeaheadComparator. Planned to refactor those to use this Comparator and avoid code duplication.
Will check it out any way to see if its missing startOfWord matches

ssettipalli: > Had no idea you could use the + operator on Pattern! Had to try it out.

Two String's are combined and then converted into Pattern 
Pattern.compile(regexString + searchString);

ssettipalli: Code seem to work fine as intended.
DrugTypeaheadComparatorTest covers the scenarios that you mentioned Marco

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+            return new RemoteAutoCompleteResponse(autoCompleteQuery, suggestions);
+        });
+
+        return result;
+    }
+
+    private List<PicklistOption> getPicklistOptions(Handle handle, String studyGuid,
+                                                    String activityInstanceGuid, String questionStableId) {
+
+        ActivityInstanceDto instanceDto = handle.attach(JdbiActivityInstance.class)
```
ssettipalli: Get QuestionDef from cached ActivityDef and get the picklist options.
Options are retrieved from already cached ActivityDef and no database query each time.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/CancerTypeaheadComparator.java
```diff
@@ -14,6 +13,7 @@
 public class CancerTypeaheadComparator implements Comparator<Cancer> {
```
ssettipalli: Refactored to use StringTypeaheadComparator and avoid code duplication.
Cancer, Drug uses exact same logic to compare & sort the suggestions

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DrugTypeaheadComparator.java
```diff
@@ -22,58 +22,13 @@
      */
     public DrugTypeaheadComparator(String searchTerm) {
         this.searchTerm = searchTerm.toUpperCase();
+        this.stringComparator = new StringSuggestionTypeaheadComparator(searchTerm);
     }
 
     @Override
     public int compare(Drug drug1, Drug drug2) {
```
ssettipalli: Refactored to use StringTypeaheadComparator and avoid code duplication.
Cancer, Drug uses exact same logic to compare & sort the suggestions

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+            long timestamp = Instant.now().toEpochMilli();
+            long langCodeId = preferredUserLanguage.getId();
+            List<PicklistOption> suggestions = new ArrayList<>();
+            List<PicklistOption> allOptions = getPicklistOptions(handle, studyGuid, instanceGuid, questionStableId);
+            if (StringUtils.isBlank(autoCompleteQuery)) {
+                LOG.info("Option suggestion query is blank, returning all results size to default limit");
+                suggestions = allOptions.stream().limit(limit).collect(Collectors.toList());
+                renderer.bulkRenderAndApply(handle, allOptions, style, langCodeId, timestamp);
+            } else {
+                renderer.bulkRenderAndApply(handle, allOptions, style, langCodeId, timestamp);
```
ssettipalli, line 86: Render Picklist label templates depending on users current language.
May be caching rendered templates for all possible languages might be an improvement to pursue but will look into it later.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+
+            //Get all options first from ActivityDef (already in memory) then render and match
+            long timestamp = Instant.now().toEpochMilli();
+            long langCodeId = preferredUserLanguage.getId();
+            List<PicklistOption> suggestions = new ArrayList<>();
+            List<PicklistOption> allOptions = getPicklistOptions(handle, studyGuid, instanceGuid, questionStableId);
+            if (StringUtils.isBlank(autoCompleteQuery)) {
+                LOG.info("Option suggestion query is blank, returning all results size to default limit");
+                suggestions = allOptions.stream().limit(limit).collect(Collectors.toList());
+                renderer.bulkRenderAndApply(handle, allOptions, style, langCodeId, timestamp);
```
ssettipalli: render only needed options templates since no search needed

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/question/PicklistQuestionDef.java
MocanaAtBroad: I think it can be a little confusing to call getPicklistOptions and not get anything back.
Can I suggest we have a method called getLocalPicklistOptions() that returns only the "regular" options, another called getRemotePicklistOptions() that returns the remote ones and then just leave getPicklistOptions() to return whatever option we have? (local/remote).

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+        String instanceGuid = request.params(RouteConstants.PathParam.INSTANCE_GUID);
+        String questionStableId = request.params(RouteConstants.PathParam.QUESTION_STABLE_ID);
+
+        DDPAuth ddpAuth = RouteUtil.getDDPAuth(request);
+        String operatorGuid = StringUtils.defaultIfBlank(ddpAuth.getOperator(), participantGuid);
+        boolean isStudyAdmin = ddpAuth.hasAdminAccessToStudy(studyGuid);
+        String autoCompleteQuery = request.queryParams(RouteConstants.QueryParam.TYPEAHEAD_QUERY);
+        String queryLimit = request.queryParams(RouteConstants.QueryParam.TYPEAHEAD_QUERY_LIMIT);
+        int limit = (StringUtils.isNotBlank(queryLimit) && Integer.valueOf(queryLimit) <= DEFAULT_LIMIT)
+                ? Integer.valueOf(queryLimit) : DEFAULT_LIMIT;
```
MocanaAtBroad: Reason I was suggesting that requesting more rows than are supported should produce an error is that, say they ask for 1000 rows and we return 100 (because 100 is our limit), then it could be interpreted by client as meaning that 100 was all the available matches (when their could have been 200 or 300 matches).
It is likely that this sort of error would only be seen once by the client. It is likely that the limit would be set as a constant in the client too, so once the error is encountered they would adjust and set it to the "real" limit once.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
MocanaAtBroad: We have exact same line of code on both the `if` and `else`. Consider refactoring so we only have the line appear once.

--> updated first line to use correct list so that only needed option templates are rendered

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+                                                    String activityInstanceGuid, String questionStableId) {
+
+        ActivityInstanceDto instanceDto = handle.attach(JdbiActivityInstance.class)
+                .getByActivityInstanceGuid(activityInstanceGuid)
+                .orElseThrow(() -> new DaoException("Could not find activity instance using guid " + activityInstanceGuid
+                        + " while loading activity definition "));
+        FormActivityDef formActivityDef = ActivityDefStore.getInstance().findActivityDef(handle, studyGuid, instanceDto)
+                .orElseThrow(() -> new DaoException("Could not find activity def for studyGuid: " + studyGuid
+                        + " and instanceGuid: " + activityInstanceGuid));
+        PicklistQuestionDef questionDef = (PicklistQuestionDef) formActivityDef.getQuestionByStableId(questionStableId);
```
MocanaAtBroad: This is an unsafe type casting: imagine that someone calls the with the wrong stableId (e.g., one that corresponds to a string question). This line would cause a ClassCastException. Consider checking that it is indeed a PicklistQuestionDef and emit an apppropriate http error code otherwise.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -0,0 +1,121 @@
...
+
+        return result;
+    }
+
+    private List<PicklistOption> getPicklistOptions(Handle handle, String studyGuid,
+                                                    String activityInstanceGuid, String questionStableId) {
+
+        ActivityInstanceDto instanceDto = handle.attach(JdbiActivityInstance.class)
+                .getByActivityInstanceGuid(activityInstanceGuid)
+                .orElseThrow(() -> new DaoException("Could not find activity instance using guid " + activityInstanceGuid
```
MocanaAtBroad: Would this result in a 500 error code? let's save those for truly unexepected server errors. This is probably closer to a 422.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DrugTypeaheadComparator.java
MocanaAtBroad: I think this is a really good idea!
Let me suggest you go further with the refactor and delete DrugTypeAheadComparator and CancerEtcEtc and in their place just use a local lambda expression that implements comparator.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
ssettipalli: returning error now

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetDsmDrugSuggestionsRoute.java
```diff
@@ -65,11 +66,16 @@ private DrugSuggestionResponse getDrugSuggestions(String drugQuery, int limit) {
```
MocanaAtBroad, line 63: Not your code, but all the code above this line in this method can be reduced to a simple lambda expression. Consider doing the refactor.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetDsmDrugSuggestionsRoute.java
```diff
@@ -65,11 +66,16 @@ private DrugSuggestionResponse getDrugSuggestions(String drugQuery, int limit) {
...
+        var suggestionComparator = new StringSuggestionTypeaheadComparator(upperDrugQuery);
         List<DrugSuggestion> sortedSuggestions = new ArrayList<>();
-        drugMatches.stream().sorted(new DrugTypeaheadComparator(drugQuery)).limit(limit).forEach(drug -> {
-            int offset = drug.getName().toUpperCase().indexOf(upperDrugQuery);
-            sortedSuggestions.add(new DrugSuggestion(drug, Collections.singletonList(new PatternMatch(offset, drugQuery.length()))));
-        });
+        drugMatches.stream()
+                .sorted((lhs, rhs) -> suggestionComparator.compare(lhs.getName(), rhs.getName()))
+                .limit(limit)
+                .forEach(drug -> {
```
MocanaAtBroad, line 71: I believe that if you convert the forEach to `map` you won't need to do an `add` to a blank list. You can just follow the `map` in pipeline with a collect(toList()). More compact and no mutation of objects (in this case we are mutating the list).

MocanaAtBroad (dismissed):
Looks good to me. I added a couple of code suggestions, but they would not affect functionality. I am approving now. And if you want to make those changes, I'll approve as soon as you complete them again.

MocanaAtBroad (approved):

PR #1251, opened 2022-03-14
Possible NPE fix

Trimax:
This PR contains fix to NPE that happens during the processing of validation rules

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/Question.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/Question.java
@@ -225,21 +225,25 @@ public abstract class Question<T extends Answer> implements Renderable {
     }
 
     protected boolean requiredRulesPass(List<T> answerValues) {
-        Optional<Rule> res = validations.stream()
+        final Optional<Rule> res = validations.stream()
                 .filter(Objects::nonNull)
                 .filter(rule -> RuleType.REQUIRED.equals(rule.getRuleType()))
                 .findFirst();
-        if (res.isPresent()) {
-            RequiredRule<T> required = (RequiredRule<T>) res.get();
-            if (answerValues.isEmpty()) {
+        if (res.isEmpty()) {
+            return true;
+        }
+
+        final RequiredRule<T> required = (RequiredRule<T>) res.get();
+        if (answerValues.isEmpty()) {
+            return false;
+        }
+
+        for (final T answer : answerValues) {
+            if (!required.validate(this, answer)) {
                 return false;
             }
-            for (T answer : answerValues) {
-                if (!required.validate(this, answer)) {
-                    return false;
-                }
-            }
         }
+
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/question/Question.java
```diff
@@ -224,7 +224,11 @@ protected boolean allowSaveRulesPass(List<T> answerValues) {
     }
 
     protected boolean requiredRulesPass(List<T> answerValues) {
-        Optional<Rule> res = validations.stream().filter(rule -> rule.getRuleType().equals(RuleType.REQUIRED)).findFirst();
+        if (validations == null) {
```
MocanaAtBroad: Looks like `validations` instance variable is initialized to an empty array. And has a @NotNull annotation (which I think is used for deserialization, but still).
I would much rather ensure that wherever caller/constructor/whatever that sets validation to null is modified so that it does not get set to null.

Trimax: Well, it can't be the case. Validation list being initialized together with a definition. And the only one set method checks if the list is not null.

Trimax: It seems that either rule is null or rule type is null

Trimax: Btw, is it possible to upgrade Java to v17? 17th version shows exactly the cause for NPEs.

MocanaAtBroad: I am seeing that the validations instance variable is referenced directly in some subclasses. And these subclasses code count on `validations` being non-null.
Also getValidations() is referenced in a way that assumes it is not null.
I would encourage you ensure that the value of `validations` is always non-null. Will avoid problems in a lot of places including this very specific bug being fixed.

By the way: noticing in the setValidations method that there is a null check too. I do not like that null check there either.

Trimax: @MocanaAtBroad, not sure I completely understand your concern here. There are only two places where this field is assigned:
1. Inline initialization: validations = new ArrayList(). This guarantees it is always notnull
2. Setter that has the check and guarantees that if the variable will be overwritten, then the new value is not null.

Taking into consideration those two points, we can be sure that validations are always not null.

As for NPE, my assumption is that some specific validation is lacking in type. And we're doing something like null.equals(...). Simple changing of arguments will solve this.

MocanaAtBroad: So why is this line needed?
```java
if (validations == null) {

MocanaAtBroad: My point is that I don't want to check for nulls. I want to ensure that this instance variable is never null in the first place.

Trimax: Removed. This check is redundant

MocanaAtBroad (approved):

PR #1267, opened 2022-03-18
DDP-7183. Equation

Trimax:
This PR contains:
1. The definition of the equation question and its import functionality (with the expression's validation).
2. The evaluation functionality

### DIFF
--- a/pepper-apis/dss-core/pom.xml
+++ b/pepper-apis/dss-core/pom.xml
@@ -41,6 +41,11 @@
         </dependency>
 
         <dependency>
+            <groupId>ch.obermuhlner</groupId>
+            <artifactId>big-math</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>com.aventrix.jnanoid</groupId>
             <artifactId>jnanoid</artifactId>
         </dependency>
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
@@ -28,6 +28,7 @@ import org.broadinstitute.ddp.db.dto.FileQuestionDto;
 import org.broadinstitute.ddp.db.dto.FormBlockDto;
 import org.broadinstitute.ddp.db.dto.NumericQuestionDto;
 import org.broadinstitute.ddp.db.dto.DecimalQuestionDto;
+import org.broadinstitute.ddp.db.dto.EquationQuestionDto;
 import org.broadinstitute.ddp.db.dto.PicklistGroupDto;
 import org.broadinstitute.ddp.db.dto.PicklistOptionDto;
 import org.broadinstitute.ddp.db.dto.PicklistQuestionDto;
@@ -49,6 +50,7 @@ import org.broadinstitute.ddp.model.activity.definition.question.DateQuestionDef
 import org.broadinstitute.ddp.model.activity.definition.question.FileQuestionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.NumericQuestionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.DecimalQuestionDef;
+import org.broadinstitute.ddp.model.activity.definition.question.EquationQuestionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.PicklistGroupDef;
 import org.broadinstitute.ddp.model.activity.definition.question.PicklistOptionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.PicklistQuestionDef;
@@ -81,6 +83,7 @@ import org.broadinstitute.ddp.model.activity.instance.question.DateQuestion;
 import org.broadinstitute.ddp.model.activity.instance.question.FileQuestion;
 import org.broadinstitute.ddp.model.activity.instance.question.NumericQuestion;
 import org.broadinstitute.ddp.model.activity.instance.question.DecimalQuestion;
+import org.broadinstitute.ddp.model.activity.instance.question.EquationQuestion;
 import org.broadinstitute.ddp.model.activity.instance.question.PicklistGroup;
 import org.broadinstitute.ddp.model.activity.instance.question.PicklistOption;
 import org.broadinstitute.ddp.model.activity.instance.question.PicklistQuestion;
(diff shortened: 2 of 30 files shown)


pepper-apis/dss-core/pom.xml
```diff
@@ -40,6 +40,11 @@
             <artifactId>guava</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>ch.obermuhlner</groupId>
+            <artifactId>big-math</artifactId>
+        </dependency>
+
```
bskinner, line 47: This dependency has not had any releases since Nov 2019. This could be an issue going forward. What does this package provide that the JRE classes do not?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1646,6 +1718,25 @@ default void disableDecimalQuestion(long questionId, RevisionMetadata meta) {
...
 
+    /**
+     * End currently active numeric question by terminating common data and numeric specific data.
+     *
+     * @param questionId the question id
+     * @param meta       the revision metadata used for terminating data
+     */
+    default void disableEquationQuestion(long questionId, RevisionMetadata meta) {
+        QuestionDto dto = getJdbiQuestion().findQuestionDtoById(questionId).orElse(null);
+        EquationQuestionDto questionDto = dto == null ? null : (EquationQuestionDto) dto;
```
bskinner: Is the ternary operator required here, or can we cast the return type of `orElse` to `EquationQuestionDto` directly?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/EquationEvaluator.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public BigDecimal visitMultiplication(final EquationParser.MultiplicationContext ctx) {
+        return visit(ctx.expression(0)).multiply(visit(ctx.expression(1)), MathContext.UNLIMITED);
+    }
+
+    @Override
+    public BigDecimal visitDivision(final EquationParser.DivisionContext ctx) {
+        return visit(ctx.expression(0)).divide(visit(ctx.expression(1)), MathContext.DECIMAL64);
```
bskinner, line 33: What is the reason for mixing `UNLIMITED` and `DECIMAL64` precision types here?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7183-equation-question.xml
```diff
@@ -0,0 +1,34 @@
...
+
+    <changeSet author="Dmitrii Eliseev" id="DDP-7183-add-equation-question-table">
+        <createTable tableName="equation_question">
+            <column name="question_id" type="bigint">
+                <constraints primaryKey="true" primaryKeyName="equation_question_pk"/>
+            </column>
+            <column name="expression" type="text">
+                <constraints nullable="false" />
+            </column>
+            <column name="maximum_decimal_places" type="bigint" defaultValue="0" />
```
bskinner: After our [discussions](https://github.com/broadinstitute/ddp-study-server/pull/1226/files#r828977353) during the last PR, do you still feel that `0` is a valid default for this field?

My perspective here is defaulting to `0` would require a user to explicitly set the `maximumDecimalPlaces` to `null` in order to have the question present a DecimalNumber how it received it (unless the front-end is special-casing `0`), and that seems like unwieldy and unnatural behavior for the study builders to contend with.

pepper-apis/pex-antlr/src/main/antlr4/org/broadinstitute/ddp/pex/lang/Equation.g4
```diff
@@ -1,88 +1,20 @@
...
-LPAREN
-   : '('
+   :  expression op='^' expression # Power
+   |  expression op='*' expression # Multiplication
+   |  expression op='/' expression # Division
+   |  expression op='+' expression # Addition
+   |  expression op='-' expression # Subtraction
+   |  '(' expression ')'           # Parentheses
+   |  VARIABLE                     # Variable
+   |  NUMBER                       # Number
```
bskinner, line 11: _Nice_ compression ratio!

bskinner (commented):
Looks good- just a few questions!

pepper-apis/dss-core/pom.xml
Trimax: BigDecimal.pow method with BigDecimal argument.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
Trimax: I guess we can do it in another way. It's just a copy-paste from the similar block next to this one.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/EquationEvaluator.java
Trimax: In the case of addition, subtraction, multiplication you can't produce significantly different values (I mean with a lot of digits after comma). In case of division or power you can. So with the UNLIMITED context, we'll have an exception, so we need to specify the limit for the result.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7183-equation-question.xml
Trimax: I agree. Haven't noticed copy-paste issue. Will remove default value.

pepper-apis/pex-antlr/src/main/antlr4/org/broadinstitute/ddp/pex/lang/Equation.g4
Trimax: :)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/EquationEvaluator.java
Trimax: We can use DECIMAL64 everywhere. What do you think?

bskinner: Ahhh- I see. Things like `1/3` will cause `UNLIMITED` precision a great deal of unhappiness. Let's stick with `DECIMAL64` across the board for now

Trimax: Yeah, exactly. Ok, I will change it to DECIMAL64

bskinner (dismissed):
👍

alextanlan (approved):

PR #1273, opened 2022-03-21
Remote autocomplete update

ssettipalli:
For REMOTE_AUTOCOMPLETE backend doesn't send picklist option defs as part of activity / activity_instance requests.
Angular needs a way to get the selected PL option def's (option lables) to display the selected answer(s).

Updated backend to include PL options defs of selected options  (from answer) during activity instance loading.
Angular need to look at the option stableID, get the option label from activity Def and display the answer.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/QuestionCreatorHelper.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/QuestionCreatorHelper.java
@@ -19,6 +19,7 @@ import org.broadinstitute.ddp.model.activity.definition.question.PicklistQuestio
 import org.broadinstitute.ddp.model.activity.definition.question.MatrixQuestionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.TextQuestionDef;
 import org.broadinstitute.ddp.model.activity.definition.question.ActivityInstanceSelectQuestionDef;
+import org.broadinstitute.ddp.model.activity.instance.answer.Answer;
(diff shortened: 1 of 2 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/QuestionCreatorHelper.java
```diff
@@ -246,7 +251,22 @@ PicklistQuestion createPicklistQuestion(AIBuilderContext ctx, PicklistQuestionDe
...
-        List<PicklistOption> picklistOptions = CollectionMiscUtil.createListFromAnotherList(questionDef.getLocalPicklistOptions(),
+        //localPicklistOptions doesn't include RemotePicklistOptions
+        //include option defs of any already selected options (answers)
+        List<PicklistAnswer> answers = questionCreator.getAnswers(ctx, questionDef.getStableId());
+        List<PicklistOptionDef> options = questionDef.getLocalPicklistOptions();
+        if (questionDef.getRenderMode() == PicklistRenderMode.REMOTE_AUTOCOMPLETE && !answers.isEmpty()) {
+
+            List<String> selectedOptsStableIds = answers.stream()
+                    .flatMap(answer -> answer.getValue().stream().map(SelectedPicklistOption::getStableId))
+                    .collect(Collectors.toList());
```
Trimax: Isn't it safer to rewrite it this way:
`
List<String> selectedOptsStableIds = answers.stream()
                    .map(Answer::getValue)
                    .flatMap(ValueClass::steam)
                    .map(SelectedPicklistOption::getStableId))
`
In this case we could probably avoid NPE in case answer's value is null for instance.

Trimax (dismissed):

MocanaAtBroad (approved):

PR #1275, opened 2022-03-21
DDP-7382 create os somatic addendum adult pediatric

archilsharashenidzebroad:
https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/836?modal=detail&selectedIssue=DDP-7382


grigol-brd (dismissed):

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/OsteoSomaticConsentAddendum.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(OsteoSomaticConsentAddendum.class);
+    private static final String DATA_FILE = "patches/somatic-consent-addendum-val.conf";
+    private static final String CMI_OSTEO = "CMI-OSTEO";
+
+    private Path cfgPath;
+    private Config cfg;
+    private Config dataCfg;
+    private Config varsCfg;
+    private String versionTag;
```
nsakande, line 40: no need to have versionTag as global variable since it's only used in one method.
Make it local to that method.

nsakande:
@archilsharashenidzebroad please write here one more time the idea that we were talking about, on the morning's meeting.
Describing the revision mechanism you used here and whyu you think it might have problems.

archilsharashenidzebroad:
The main idea, was behind the process of versioning, in this particular case I made a version change in activity via method handle.attach(ActivityDao.class).changeVersion(activityId, versionTag, meta) there is possibility of version mismatch after this execution, because after this patch will be change the version to v2, it might become problematic for next patches that make search for activity based on previous version and make revision, because previous version of activity might not persist in storage anymore.

nsakande (approved):
LGTM

PR #1278, opened 2022-03-23
DDP-7185 . Always include options with allowDetails=true

ssettipalli:
As part of DDP-7185 REMOTE_AUTOCOMPLETE picklist options:
If user enters some label/text that doesn't match any PL option labels, if there is a Picklist option with allowDetails=true, client need to select that option and send the entered text/label in details.

For ex:
Picklist list options are apple and orange and
the user enters “banana”, client need to send back to the server the stableId an answer with the stableId of the option that has “allowDetails=true” and “banana” as details.

For client/angular to know if there is a picklist option with allowDetails=true, backend need to include allowDetails=true option(s) . Usually study PL question conf will have one. 
Either restrict to one during study init OR client can always pick first one if more than one exists.

This PR has change to include  picklist option with allowDetails=true while returning selected matches.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
@@ -131,6 +131,9 @@ public class GetOptionsForActivityInstanceQuestionRoute implements Route {
         List<PicklistOption> optionMatches = options.stream().filter(option -> option.getOptionLabel().toUpperCase()
                 .contains(autoCompleteQuery.toUpperCase())).collect(Collectors.toList());
 
+        //include options with allowDetails = true
+        optionMatches.addAll(options.stream().filter(option -> option.isDetailsAllowed()).collect(Collectors.toList()));
+
         // now sort the matches in a way that puts left-most matches near the top, favoring word start matches
         // apply limit and return results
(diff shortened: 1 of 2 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -131,6 +131,9 @@ public RemoteAutoCompleteResponse handle(Request request, Response response) {
         List<PicklistOption> optionMatches = options.stream().filter(option -> option.getOptionLabel().toUpperCase()
                 .contains(autoCompleteQuery.toUpperCase())).collect(Collectors.toList());
 
+        //include options with allowDetails = true
+        optionMatches.addAll(options.stream().filter(option -> option.isDetailsAllowed()).collect(Collectors.toList()));
```
ssettipalli, line 135: Debated to include allowDetails = true options only when NO matches OR always. These will be at the end of sorted list any way.  Change includes always

Trimax: Could you please replace lambda with a method reference? Takes less space and is more readable from my POV.
`
options.stream().filter(PicklistOption::isDetailsAllowed()).collect(Collectors.toList())
`

Trimax (approved):

PR #1285, opened 2022-03-24
Get remote picklist suggestions/options fix

ssettipalli:
Running into edge case where user enters text from Picklist options  with "allowDetails=true" , these "allowDetails=true" picklist options are duplicated in the response.

Include options with allowDetails=true only if option is NOT already included in the response options list


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetOptionsForActivityInstanceQuestionRoute.java
```diff
@@ -131,8 +131,10 @@ public RemoteAutoCompleteResponse handle(Request request, Response response) {
         List<PicklistOption> optionMatches = options.stream().filter(option -> option.getOptionLabel().toUpperCase()
                 .contains(autoCompleteQuery.toUpperCase())).collect(Collectors.toList());
 
-        //include options with allowDetails = true
-        optionMatches.addAll(options.stream().filter(option -> option.isDetailsAllowed()).collect(Collectors.toList()));
+        //include options with allowDetails=true if no matches
+        if (optionMatches.isEmpty()) {
```
ssettipalli, line 135: Include options with allowDetails=true if only NO matches/suggestions found for entered text.

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #1286, opened 2022-03-24
Data exporter extended by records for decimal & numeric questions

Trimax:


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -1027,14 +1027,20 @@ private QuestionRecord createRecordForQuestion(QuestionDef question, FormRespons
...
         if (answer == null) {
             return null;
         }
-        QuestionType type = answer.getQuestionType();
+
+        final QuestionType type = answer.getQuestionType();
         if (type == QuestionType.DATE) {
             DateValue value = (DateValue) answer.getValue();
             return new DateQuestionRecord(question.getStableId(), value);
+        } else if (answer.getQuestionType() == QuestionType.NUMERIC) {
```
ssettipalli, line 1040: else by default covers these Numeric & Decimal questions if no additional stuff needed and transforms to SimpleQuestionRecord.

PR #1291, opened 2022-03-25
Payload classes migration to Lombok

Trimax:
Payload classes were migrated to Lombok


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserActivity.java
```diff
@@ -2,41 +2,25 @@
...
-            String isoLanguageCode
-    ) {
-        this.guid = guid;
-        this.title = title;
-        this.subtitle = subtitle;
-        this.typeName = typeName;
-        this.status = status;
-        this.isoLanguageCode = isoLanguageCode;
-    }
+    String isoLanguageCode;
```
bskinner, line 23: Can we add a `@Getter(onMethod = @__(@Override))` to this field and remove the manual getter implementation?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/medicalprovider/PostPatchMedicalProviderRequestPayload.java
```diff
@@ -1,55 +1,32 @@
...
-    public String getState() {
-        return state;
-    }
-
     public static class Fields {
         public static final String INSTITUTION_NAME = "institutionName";
         public static final String PHYSICIAN_NAME = "physicianName";
         public static final String CITY = "city";
         public static final String STATE = "state";
     }
```
bskinner, line 31: Is this class referenced elsewhere? If not, let's just move the strings directly into the `SerializedName` annotations above.

bskinner (commented):
It's great to see all those manually implemented getters _gone_! Couple of minor changes, then I think we're good to merge!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/GovernedParticipant.java
```diff
@@ -1,32 +1,25 @@
...
-        return userGuid;
-    }
-
-    public String getAlias() {
-        return alias;
-    }
-
-    public Profile getUserProfile() {
-        return userProfile;
+        this.userProfile = Optional.ofNullable(userProfile).map(Profile::new).orElse(null);
```
MocanaAtBroad, line 23: You sure about the constructor in `map` doing the right thing with argument?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserActivity.java
Trimax: I'll just remove the getter itself. It's redundant. I don't think we need to specify the annotation explicitly. The override annotation has retention policy SOURCE, so it doesn't really matter.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/medicalprovider/PostPatchMedicalProviderRequestPayload.java
Trimax: Yep, it is used in the PatchMedicalProviderRoute

bskinner (approved):
👍

MocanaAtBroad (commented):
All looks good. Just saw one that looks suspicious.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/GovernedParticipant.java
Trimax: I'm sure. It's just a method reference to the class's constructor

Trimax: `.map(X::new)` is equivalent to `.map(x -> new X(x))`

MocanaAtBroad (approved):

PR #1298, opened 2022-03-28
Single answer support for the equations

Trimax:
This PR contains following changes:

1. Patch response now contains the equations with the computed values if possible
2. The equations that are refers to the questions with single answer only is now supported


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+    private final Map<String, EquationResponse> values = new HashMap<>();
+    private final Handle handle;
+    private final String instanceGuid;
+
+    public EquationResponse evaluate(final EquationQuestionDto equation) {
+        StreamEx.of(EquationVariablesCollector.builder().build().collect(equation.getExpression()))
+                .remove(values::containsKey)
+                .forEach(this::fetchVariableValue);
+
+        if (!StreamEx.of(EquationVariablesCollector.builder().build().collect(equation.getExpression()))
```
bskinner: Can we break this up? This `if` conditional is a mouthful to read.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+                .allMatch(values::containsKey)) {
+            log.info("The equation can't be evaluated. Not all variables were populated");
+            return null;
+        }
+
+        final var evaluatorBuilder = EquationEvaluator.builder();
+        StreamEx.of(EquationVariablesCollector.builder().build().collect(equation.getExpression()))
```
bskinner: There are multiple repetitions of `EquationVariablesCollector.builder().build().collect(equation.getExpression())` in this method. Do we need to rebuild this collection each time or can this be created locally once and reused or copied?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+
+        addValue(question.get().getType(), variable, answer.get().getValue());
+    }
+
+    private void addValue(final QuestionType type, final String variable, final Object value) {
+        switch (type) {
+            case NUMERIC:
+            case PICKLIST:
+                values.put(variable, new EquationResponse(variable,
+                        Collections.singletonList(new DecimalDef((Integer) value))));
```
bskinner: PICKLIST questions natively have a string type, not an integer one so this typecast will always fail.

bskinner (commented):
Couple minor comments, but looks good otherwise.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
```diff
@@ -0,0 +1,101 @@
...
+public class QuestionEvaluator {
+    private final Map<String, EquationResponse> values = new HashMap<>();
+    private final Handle handle;
+    private final String instanceGuid;
+
+    public EquationResponse evaluate(final EquationQuestionDto equation) {
+        final var variables = new EquationVariablesCollector(equation.getExpression()).collect();
+
+        StreamEx.of(variables).remove(values::containsKey).forEach(this::fetchVariableValue);
+        if (!StreamEx.of(variables).allMatch(values::containsKey)) {
```
bskinner, line 29: This is far easier to read!

bskinner (approved):
Let's merge it! 👍

PR #1305, opened 2022-03-28
DDP-7694 elastic search settings/mapping update

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7694

DSM will be exporting more DSM data into participants_structured index.
This seem to increase number of fields that will end up in the index.
We are running into issues during elastic export in Dev 
Updated the setting to increase number of field from the default: 1000


elasticsearch/templates/participants_structured.json
```diff
@@ -2,6 +2,11 @@
   "index_patterns": ["participants_structured.*"],
   "settings": {
     "index": {
+      "mapping": {
+        "total_fields": {
+          "limit": "2000"
```
ssettipalli, line 7: increased from default value: 1000

MocanaAtBroad (approved):

PR #1308, opened 2022-03-29
Missed equation switch statements

Trimax:


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -125,6 +125,8 @@ private void createAnswerValue(long operatorId, long instanceId, long answerId,
         } else if (type == QuestionType.DECIMAL) {
             DecimalAnswer ans = (DecimalAnswer) answer;
             DBUtils.checkInsert(1, answerSql.insertDecimalValue(answerId, ans.getValueAsBigDecimal()));
+        } else if (type == QuestionType.EQUATION) {
+            throw new DaoException("Equation question doesn't require any answer");
```
ssettipalli, line 129: I wonder if throwing exception works ! If we want to move on, may be check and ignore it !

Trimax: Isn't it an answer insertion to the database? We're not going to insert anything into the database, even if the user sends the data to us. Shouldn't we report an error like this?

bskinner: I agree with @Trimax on this- `EQUATION` questions do not support answers. If a dev or client attempts to create one, this should be a full stop, and we shouldn't try to recover from it- it's a client error.

ssettipalli (approved):

PR #1309, opened 2022-03-29
DDP-7653 initial participant list export

LKapanad:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -32,19 +34,21 @@
...
+            tableAlias = StringUtils.isNotBlank(sortBy.getOuterProperty()) ? sortBy.getOuterProperty() : sortBy.getInnerProperty();
+        } else {
+            tableAlias = sortBy.getTableAlias();
         }
-        return alias;
+        return of(tableAlias);
+    }
+
+    public static Alias of(String tableAlias) {
+        return Enums.getIfPresent(Alias.class, tableAlias.toUpperCase()).or(ACTIVITIES);
```
gmakharat: I don't see the point of adding some 3rd party library complexity here, I believe `try/catch` was doing its job properly and is easily understandable special-case pattern.

gmakharat: I don't see the point of adding some 3rd party library complexity here, I believe `try/catch` was doing its job properly and is easily understandable special-case pattern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -21,8 +23,8 @@
     EX(String.join(DBConstants.ALIAS_DELIMITER, ESObjectConstants.DSM, ESObjectConstants.PARTICIPANT), false),
     DSM(ElasticSearchUtil.DSM, false), STATUS(ElasticSearchUtil.STATUS, false), PROFILE(ElasticSearchUtil.PROFILE, false),
     ADDRESS(ElasticSearchUtil.ADDRESS, false), INVITATIONS(ElasticSearchUtil.INVITATIONS, false), PROXY(ElasticSearchUtil.PROFILE, false),
-    ACTIVITIES(ElasticSearchUtil.ACTIVITIES, true), REGISTRATION(ElasticSearchUtil.ACTIVITIES, true);
-
+    ACTIVITIES(ElasticSearchUtil.ACTIVITIES, true), REGISTRATION(ElasticSearchUtil.ACTIVITIES, true),
+    DATA("", false);
```
gmakharat: I am afraid that adding `DATA` will break something, since we had special check in `of` method for case of `data`, why have you had to add `DATA` as enum?

gmakharat: I am afraid that adding `DATA` will break something, since we had special check in `of` method for case of `data`, why have you had to add `DATA` as enum?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -32,19 +34,21 @@
...
-        return alias;
+        return of(tableAlias);
+    }
+
+    public static Alias of(String tableAlias) {
+        return Enums.getIfPresent(Alias.class, tableAlias.toUpperCase()).or(ACTIVITIES);
     }
 
+
+    public static Alias of(ParticipantColumn column) {
```
gmakharat: I think this method is not used at all.

gmakharat: I think this method is not used at all.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
```diff
@@ -68,8 +69,9 @@ protected void prepareNecessaryData(QueryParamsMap queryParamsMap) {
         } else if (savedFilters != null) {
             filters = savedFilters;
         }
-        this.from = Integer.parseInt(queryParamsMap.get(LIST_RANGE_FROM).value());
-        this.to = Integer.parseInt(queryParamsMap.get(LIST_RANGE_TO).value());
+        this.from = NumberUtils.toInt(queryParamsMap.get(LIST_RANGE_FROM).value(), 0);
+        this.to = NumberUtils.toInt(queryParamsMap.get(LIST_RANGE_TO).value(), 10000);
```
gmakharat: I think it would be better to extract `0` and `10000` as constants with meaningful names to avoid so called magic numbers.

gmakharat: I think it would be better to extract `0` and `10000` as constants with meaningful names to avoid so called magic numbers.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/DownloadParticipantListPayload.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.stream.Collectors;
+
+import com.google.gson.annotations.SerializedName;
+import lombok.Getter;
+import lombok.Setter;
+import org.broadinstitute.dsm.model.ParticipantColumn;
+
+@Getter
+@Setter
+public class DownloadParticipantListPayload {
```
gmakharat: I think this class should be in `org.broadinstitute.dsm.model.excel` and not `org.broadinstitute.dsm.model.participant`, since it belongs to the exporting excel feature.

gmakharat: I think this class should be in `org.broadinstitute.dsm.model.excel` and not `org.broadinstitute.dsm.model.participant`, since it belongs to the exporting excel feature.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+import org.broadinstitute.dsm.model.participant.DownloadParticipantListPayload;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperDto;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperResult;
+import org.broadinstitute.dsm.security.RequestHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.Request;
+import spark.Response;
+
+public class DownloadParticipantListRoute extends RequestHandler {
```
gmakharat, line 24: I think producing excel feature deserves to have its own class, rather than having everything directly in Route

gmakharat, line 24: I think producing excel feature deserves to have its own class, rather than having everything directly in Route

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(DownloadParticipantListRoute.class);
+    private static final String FILE_DATE_FORMAT = "ddMMyyyy";
+
+    public DownloadParticipantListRoute() {
+    }
+
+    @Override
+    public Object processRequest(Request request, Response response, String userId) throws Exception {
+        DownloadParticipantListPayload payload = new Gson().fromJson(request.body(), DownloadParticipantListPayload.class);
```
gmakharat: because of we faced strange behavior from `gson` when it converts all the numbers to `double` we decided to slowly eliminate it from the project and replace it with `jackson`, there is a special classes created for it called `ObjectMapperSingleton`, it is preferable to use it.

gmakharat: because of we faced strange behavior from `gson` when it converts all the numbers to `double` we decided to slowly eliminate it from the project and replace it with `jackson`, there is a special classes created for it called `ObjectMapperSingleton`, it is preferable to use it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+
+        List<ParticipantColumn> columnNames = payload.getColumnNames();
+        List<String> esAliases = columnNames.stream()
+                .map(column -> {
+                    Alias esAlias;
+                    if (column.getObject() != null) {
```
gmakharat: personally I prefer to use `Objects.nonNull()` method for such comparison in case of readability.

gmakharat: personally I prefer to use `Objects.nonNull()` method for such comparison in case of readability.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+                        esAlias = Alias.of(column.getObject());
+                    } else {
+                        esAlias = Alias.of(column.getTableAlias());
+                    }
+                    return esAlias.getValue().isEmpty()? column.getName() : esAlias.getValue() + "." + column.getName();
+                })
+                .collect(Collectors.toList());
```
gmakharat: extracting piece of code from L52 to L63 with meaningful name would be better.

gmakharat: extracting piece of code from L52 to L63 with meaningful name would be better.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+        List<ParticipantColumn> columnNames = payload.getColumnNames();
+        List<String> esAliases = columnNames.stream()
+                .map(column -> {
+                    Alias esAlias;
+                    if (column.getObject() != null) {
+                        esAlias = Alias.of(column.getObject());
+                    } else {
+                        esAlias = Alias.of(column.getTableAlias());
+                    }
+                    return esAlias.getValue().isEmpty()? column.getName() : esAlias.getValue() + "." + column.getName();
```
gmakharat: there is a constant for `.`, `DBConstants.ALIAS_DELIMITER`

gmakharat: there is a constant for `.`, `DBConstants.ALIAS_DELIMITER`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+                })
+                .collect(Collectors.toList());
+
+        Filterable filterable = FilterFactory.of(request);
+        ParticipantWrapperResult filteredList = (ParticipantWrapperResult) filterable.filter(request.queryMap());
```
gmakharat, line 45: what kind of payload we get here in `queryMap`? asking because `FilterFactory.of` is already pretty messy method and adding new case of `RoutePath.DOWNLOAD_PARTICIPANT_LIST_ROUTE` is kinda out of context for it, can't we directly use `ManualFilterParticipantList` class here? which would help us not to add new case statement in `FilterFactory.of`?

gmakharat, line 45: what kind of payload we get here in `queryMap`? asking because `FilterFactory.of` is already pretty messy method and adding new case of `RoutePath.DOWNLOAD_PARTICIPANT_LIST_ROUTE` is kinda out of context for it, can't we directly use `ManualFilterParticipantList` class here? which would help us not to add new case statement in `FilterFactory.of`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+        Workbook workbook = new XSSFWorkbook();
+        List<String> headerNames = payload.getHeaderNames();
+        Sheet sheet = workbook.createSheet("Participant List");
+        Row header = sheet.createRow(0);
+        IntStream.range(0, headerNames.size()).forEach(i -> {
+            Cell cell = header.createCell(i);
+            cell.setCellValue(headerNames.get(i));
+            sheet.autoSizeColumn(i);
+        });
```
gmakharat: extracting method from L70 to L75 with meaningful name like `buildExcelHeader` or something similar would be straightforward.

gmakharat: extracting method from L70 to L75 with meaningful name like `buildExcelHeader` or something similar would be straightforward.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+        int currentRow = 1;
+        for (ParticipantWrapperDto participant : filteredList.getParticipants()) {
+            List<Object> excelRow = new ArrayList<>();
```
gmakharat: is there any chance to introduce some abstraction instead of having `Object`? something like interface which would be implemented by concrete classes and would have one abstract method which would produce necessary data?

gmakharat: is there any chance to introduce some abstraction instead of having `Object`? something like interface which would be implemented by concrete classes and would have one abstract method which would produce necessary data?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+            Map<String, Object> esDataAsMap = participant.getEsDataAsMap();
+            List<String> row = new ArrayList<>();
```
gmakharat: `row` is never used I guess.

gmakharat: `row` is never used I guess.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+            }
+            currentRow = createRecord(sheet, currentRow, new ParticipantRecord(excelRow));
+        }
+        responseFile(response, workbook);
+        return response.raw();
+    }
+
+
+    private void responseFile(Response response, Workbook workbook) throws Exception {
+        File currDir = new File(".");
```
gmakharat: having here `try-with-resource` statement to handle `AutoCloseable` type objects (`Workbook`, `FileOutputStream`, `ServletOutputStream`) instead of closing them manually would be better.

gmakharat: having here `try-with-resource` statement to handle `AutoCloseable` type objects (`Workbook`, `FileOutputStream`, `ServletOutputStream`) instead of closing them manually would be better.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,181 @@
...
+        response.raw().setHeader("Content-Disposition", "attachment; filename=" + file.getName());
+        byte[] encoded = Files.readAllBytes(Paths.get(fileLocation));
+        ServletOutputStream os = response.raw().getOutputStream();
+        os.write(encoded);
+        os.close();
+        Files.deleteIfExists(file.toPath());
+    }
+
+    private String getFormattedDate() {
+        Date date = new Date();
```
gmakharat: `Date` is outdated I would say, please change it to `LocalDate` and its relevant formatter.

gmakharat: `Date` is outdated I would say, please change it to `LocalDate` and its relevant formatter.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
gmakharat: any idea about that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
gmakharat: still have the same question

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -0,0 +1,119 @@
...
+            Alias alias = Alias.of(column);
+            columnAliasEsPathMap.computeIfAbsent(alias, paths -> new ArrayList<>())
+                    .add(column);
+        });
+
+        Filterable filterable = FilterFactory.of(request);
+        ParticipantWrapperResult filteredList = (ParticipantWrapperResult) filterable.filter(request.queryMap());
+
+        ParticipantRecordData rowData = new ParticipantRecordData(columnAliasEsPathMap);
+        for (ParticipantWrapperDto participant : filteredList.getParticipants()) {
```
gmakharat: all these manipulation details would be perfectly encapsulated in its own class of generating excel

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -114,6 +114,9 @@
     public static final String PROPERTIES = "properties";
     public static final byte OUTER_FIELD_INDEX = 0;
     public static final byte INNER_FIELD_INDEX = 1;
+    public static final String STABLE_ID = "stableId";
```
gmakharat: I think `ESObjectConstants` is better place for this constant.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/export/ExcelUtils.java
```diff
@@ -0,0 +1,74 @@
...
+        String strDate = getFormattedDate();
+        String fileLocation = String.format("%sParticipant-%s.xlsx", path.substring(0, path.length() - 1), strDate);
+        try (FileOutputStream outputStream = new FileOutputStream(fileLocation)){
+            workbook.write(outputStream);
+        }
+        File file = new File(fileLocation);
+        response.raw().setContentType("application/octet-stream");
+        response.raw().setHeader("Content-Disposition", "attachment; filename=" + file.getName());
+        byte[] encoded = Files.readAllBytes(Paths.get(fileLocation));
+        try(ServletOutputStream os = response.raw().getOutputStream()) {
```
gmakharat: I think having two different try-with-resource statement is not necessary since try-with-resource supports having several AutoClosable type object into one statement separated by semicolon 
`try(FileOutputStream outputStream = new FileOutputStream(fileLocation); ServletOutputStream os = response.raw().getOutputStream(); workbook)`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -21,8 +25,8 @@
     EX(String.join(DBConstants.ALIAS_DELIMITER, ESObjectConstants.DSM, ESObjectConstants.PARTICIPANT), false),
     DSM(ElasticSearchUtil.DSM, false), STATUS(ElasticSearchUtil.STATUS, false), PROFILE(ElasticSearchUtil.PROFILE, false),
     ADDRESS(ElasticSearchUtil.ADDRESS, false), INVITATIONS(ElasticSearchUtil.INVITATIONS, false), PROXY(ElasticSearchUtil.PROFILE, false),
-    ACTIVITIES(ElasticSearchUtil.ACTIVITIES, true), REGISTRATION(ElasticSearchUtil.ACTIVITIES, true);
-
+    ACTIVITIES(ElasticSearchUtil.ACTIVITIES, true), REGISTRATION(ElasticSearchUtil.ACTIVITIES, true),
+    DATA("", false);
```
gmakharat: Let's use `StringUtils.EMPTY` constant instead of hard coded empty literal of string

gmakharat (approved):

pepper-apis/dsm-core/pom.xml
```diff
@@ -239,6 +239,16 @@
             <artifactId>gson</artifactId>
             <!-- <version>2.8.5</version> -->
         </dependency>
+        <dependency>
+            <groupId>org.apache.poi</groupId>
+            <artifactId>poi</artifactId>
+            <version>5.2.0</version>
```
SimoneMaiwald, line 245: versions should go into https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/pom.xml

PR #1330, opened 2022-03-31
updated pancan new activites precondition expressions for events

nsakande:
Updated preconditions to events introduced for new pancan activities to support child with CRC cancer from regular flow and.
Before this for child flow, new activites were appearing only for children added from the dashboard and not initial flow ( chosen children from prequal ).


archilsharashenidzebroad (approved):
LGTM

PR #1345, opened 2022-04-01
DDP-7707: Extend BOOLEAN question that can be rendered as a checkbox

alextanlan:
## Context

This PR updates the API specification to extend API definitions for the Boolean question. It can have 2 render modes now: **RADIO_BUTTONS** (current one) and **CHECKBOX** (new one).

## Release

### DIFF
--- a/pepper-apis/docs/specification/package.json
+++ b/pepper-apis/docs/specification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pepper-api-spec",
-  "version": "1.27.370",
+  "version": "1.27.0-370",
   "private": true,
   "description": "",
   "dependencies": {
--- a/pepper-apis/docs/specification/package.json
+++ b/pepper-apis/docs/specification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pepper-api-spec",
-  "version": "1.27-370-ge7f3453f7",
+  "version": "1.27.370",
   "private": true,
   "description": "",
   "dependencies": {
--- a/pepper-apis/docs/specification/package.json
+++ b/pepper-apis/docs/specification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pepper-api-spec",
-  "version": "1.25.0-71",
+  "version": "1.27-370-ge7f3453f7",
   "private": true,
   "description": "",
   "dependencies": {
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1642,10 +1642,10 @@ Question.Boolean:
           - CHECKBOX
           - RADIO_BUTTONS
       trueContent:
-        type: string | null (for CHECKBOX renderMode)
+        type: string
         example: 'Yes'
       falseContent:
-        type: string | null (for CHECKBOX renderMode)
+        type: string
         example: 'NO'
       answers:
         type: array
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1628,17 +1628,24 @@ Question.Boolean:
       - answers
       - trueContent
       - falseContent
+      - renderMode
     properties:
       questionType:
         type: string
         default: BOOLEAN
         enum:
           - BOOLEAN
-      trueContent:
+      renderMode:
         type: string
+        default: RADIO_BUTTONS
+        enum:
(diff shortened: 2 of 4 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1628,17 +1628,24 @@ Question.Boolean:
...
           - BOOLEAN
-      trueContent:
+      renderMode:
         type: string
+        default: RADIO_BUTTONS
+        enum:
+          - CHECKBOX
+          - RADIO_BUTTONS
+      trueContent:
+        type: string | null (for CHECKBOX renderMode)
```
bskinner: This is not a valid `type` value. The `type` property must be one of the supported [data types](https://spec.openapis.org/oas/v3.0.3.html#data-types). This field should not be nullable in any case as the `trueContent` and `falseContent` are required.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1628,17 +1628,24 @@ Question.Boolean:
...
         example: 'Yes'
       falseContent:
-        type: string
+        type: string | null (for CHECKBOX renderMode)
```
bskinner: Same as above for the `trueContent`

bskinner (changes requested):

pepper-apis/docs/specification/src/components/schemas.yml
alextanlan: Should it be an empty string in this case (for CHECKBOX mode) ?

alextanlan: Or should I use `nullable` like below ?
image

bskinner: Making those fields `nullable` is a breaking API change. How is the question supposed to appear? I was under the impression that this was supposed to mirror the `renderMode` of a `PICKLIST` question. Radio-like behavior using a checkbox appearance. This is sounding like the behavior is changing to be more in-line with the `AGREEMENT` question type- a single checkbox where `True` is checked and `False` is unchecked.

bskinner: Since DDP-7707 doesn't have any acceptance criteria I'm falling back to DDP-7706. My understanding of the behavior described in DDP-7706 does seems to agree with the latter situation. If the front-end is just going to ignore these values, I don't see any need to make the fields `nullable`. Let's just leave the definition as-is and add the `renderMode` enum.

alextanlan: > If the front-end is just going to ignore these values

Right, for **CHECKBOX** mode the **_trueContent_**/**_falseContent_** values will be ignored.

Fixed in ef3be1c55660bf96b18a8a623b4bb90bac131fb6

pepper-apis/docs/specification/src/components/schemas.yml
alextanlan: Fixed

pepper-apis/docs/specification/src/pepper.yml
```diff
@@ -6,7 +6,7 @@ servers:
     description: Production server (uses real data)
 info:
   description: 'Pepper API specification'
-  version: 'v1.25-46-gcd81179ef'
+  version: 'v1.27-370-ge7f3453f7'
```
alextanlan, line 9: The version value is the output from git describe, like:
```
cd dss-study-server
git pull --all --tags
git describe --tags
```

pepper-apis/docs/specification/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pepper-api-spec",
-  "version": "1.25.0-71",
+  "version": "1.27.370",
```
alextanlan: The version value is a value from https://github.com/broadinstitute/ddp-study-server/pull/1345/files#r841690210,
which is formatted in order to be a valid [SemVer](https://semver.org/).

bskinner: Small edit to get this a bit closer to the `dss-server` version (since we make use of the `PATCH` field)
```suggestion
  "version": "1.27.0-370",
```

alextanlan: Fixed in 4a4fecd9eedb9dc3b591a8a6645f444092190da9

bskinner (approved):
Looks good

PR #1364, opened 2022-04-04
updated osteo event expression conditions with correct users

nsakande:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-7743


LKapanad (approved):
LGTM

PR #1385, opened 2022-04-07
DDP-7735 hide family aliases in ES export

LKapanad:
https://broadinstitute.atlassian.net/browse/DDP-7735

Created a config file with question stableIds
In the ES export flow real values are override with default values


pepper-apis/dss-core/src/main/resources/ESprotectedColumns.conf
```diff
@@ -0,0 +1,9 @@
+"columnsWithDefaultValues" : {
```
ssettipalli: I believe what we need to skip loading into elastic are study specific and we don't need to run through these for All study exports. 
Can we capture these columns by study and export can check only for those studies which has values that need to be skipped. 
If it's hard to get that info in configuration, may be we can add a  database table to capture it.
study_id, alias, stable_id (or whatever is needed)

ssettipalli: If its in DB, later some point of time we can have study-builder config to populate this table during study configuration too. Also its not something that changes by env and hence can be in DB.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7735-family-alias.xml
```diff
@@ -0,0 +1,58 @@
...
+            <column name="alias" type="varchar(200)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="stable_id" type="varchar(45)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint columnNames="study_guid, stable_id" tableName="es_hidden_alias"/>
+
+        <insert tableName="es_hidden_alias">
```
ssettipalli: @LKapanad 
Looks good one minor renaming: Can we please rename table name to study_data_alias rather than es_hidden_alias. we don't need it be elastic specific. We are hiding data from CSV upload too and in future it might be something else.
Please rename the Dao as well to reflect table name.
May be I should have suggested name yesterday, sorry for additional refactor

ssettipalli (changes requested):

ssettipalli (approved):

PR #1389, opened 2022-04-07
Try to make Vault access more resilient

MocanaAtBroad:
A couple of changes to our consul template rendering trying to avoid vault errors.
Have tried multiple times running the parallel tests and have not encountered the vault error, so this might be a workaround.

### DIFF
--- a/pepper-apis/configure.rb
+++ b/pepper-apis/configure.rb
@@ -128,7 +128,7 @@ def read_secret_from_path(path, field = nil)
   end
 
   # Not sure why Vault requires the -1 flag, but it does.
-  curl_cmd = ["curl","-1", "-H", "X-Vault-Token: #{$vault_token}", "#{$vault_url_root}/#{path}"]
+  curl_cmd = ["curl","-1", "--retry", "-H", "X-Vault-Token: #{$vault_token}", "#{$vault_url_root}/#{path}"]
   Open3.popen3(*curl_cmd) { |stdin, stdout, stderr, wait_thread|
     if wait_thread.value.success?
       json = JSON.load(stdout)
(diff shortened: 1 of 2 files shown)


pepper-apis/configure.rb
```diff
@@ -128,7 +128,7 @@ def read_secret_from_path(path, field = nil)
   end
 
   # Not sure why Vault requires the -1 flag, but it does.
-  curl_cmd = ["curl","-1", "-H", "X-Vault-Token: #{$vault_token}", "#{$vault_url_root}/#{path}"]
+  curl_cmd = ["curl","-1", "--retry", "5", "-H", "X-Vault-Token: #{$vault_token}", "#{$vault_url_root}/#{path}"]
```
MocanaAtBroad, line 131: We were seeing some curl call fail. We are going to retry multiple times.

pepper-apis/configure.rb
```diff
@@ -200,6 +200,7 @@ def render_from_path(path, output_file_name = nil)
                   $dsde_toolbox_image_name,
                   "consul-template", "-config=/etc/consul-template/config/config.json",
                   "-template=#{file_name}:#{output_file_name}",
+                  "-retry=30",
```
MocanaAtBroad, line 203: For consul-template executable, best we can do is to extend the timeout time (the param is called retry, but it really is a timeout)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiValidation.java
```diff
@@ -10,6 +10,7 @@
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
+
```
MocanaAtBroad, line 13: Totally bogus change so there is a Java code change that will allow the execution of the tests.

ssettipalli (approved):

bskinner (approved):
👍

PR #1408, opened 2022-04-11
[DSS] DDP-7630: composite dates mapping fix

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7630
Mapping changes to handle edge cases where dates are captured as non Date fields in a composite questions.
Explicitly called out in mapping to handle as text rather than dates while exporting in new style for dsm.

Tried to go through old alerts and studies to identify these possible questions but if any missed , need to add them later .


ssettipalli:
changes merged from diff PR

PR #1428, opened 2022-04-12
Resolved DDP-7792.

leonidrakitin:
https://broadinstitute.atlassian.net/browse/DDP-7792


archilsharashenidzebroad:
The patch that updates these values is already created in patches/self-consent.conf. For instance you are updating value of this variable "v2_osteo_consent_s3_additional_agree_item5". It is already present in patches/self-consent.conf. No need for separate OsteoAdultConsentFixes class. Just update values in patches/self-consent.conf, as these patches are not yet applied to prod, we can update them directly. @leonidrakitin

archilsharashenidzebroad (approved):
LGTM

PR #1429, opened 2022-04-12
DDP-7824 . Mapping fixes to handle dates in composite questions

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-7824
Mapping changes to handle dates in composite questions that are not captured as dates while exporting in new style needed for DSM.
Explicitly call out these edge cases in mapping to export as text 
tested and looks good

### DIFF
--- a/elasticsearch/templates/participants_structured.json
+++ b/elasticsearch/templates/participants_structured.json
@@ -344,6 +344,30 @@
                     }
                   }
                 },
+                "LONGITUDINAL_VACCINE_RECEIVED_LIST": {
+                  "type": "text"
+                },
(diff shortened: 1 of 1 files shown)


elasticsearch/templates/participants_structured.json
```diff
@@ -344,6 +344,30 @@
                     }
                   }
                 },
+                "LONGITUDINAL_VACCINE_RECEIVED_LIST": {
```
ssettipalli, line 347: Explicitly call out these composite questions in mapping to export as text in new style

Rinchan20 (approved):

PR #1435, opened 2022-04-13
[DSS] Derived question: added support of multiple answers for children of a composite question

Trimax:
Well. This PR contains the code modification that allows us to support equations as children of a COMPOSITE question.
The thing is that the composite question might have several rows with the answers to children's questions.

For example, let's say that the COMPOSITE question is _RECTANGLE_. Such a question will have the children:
1. _RECTANGLE_WIDTH_ question
2. _RECTANGLE_HEIGHT_ question
3. _RECTANGLE_AREA_ equation (_RECTANGLE_WIDTH_ * _RECTANGLE_HEIGHT_)

The user might provide 3 rows of the answers. For instance like this:
[{width=2, height=3},{width=null, height=3},{width=2, height=null}]

And as a result, we should expect the _RECTANGLE_AREA_ equation as follows:
[6, null, null]

Described means that we should use the Lists of the values instead of scalar values when computing the equations.
Before this PR our equations were just the functions that map a scalar real value to a scalar real value (f: R->R).

After merging this PR, the equation will be a more general instrument. It will be the function that maps a vector value (f: Rˆn -> R^n).

**This code should be tested on the DEV server after merging, so please don't close the ticket once PR is merged**


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/SingleValueList.java
```diff
@@ -0,0 +1,23 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+
+@Value
+@AllArgsConstructor
+public class SingleValueList extends ArrayList<BigDecimal> {
```
bskinner: I have a few concerns with this class:
 * It's public
 * It descends from ArrayList, but bypasses most of its logic
 * and is being returned as a `List<>` in several methods, including public ones

Using one of `Collections::singletonList`, `List::of`, using an internal private data class to wrap the list (or item), or a similar method would be preferable.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1545,7 +1546,10 @@ default void insertQuestion(long activityId, PicklistQuestionDef picklist, long
 
     default void insertQuestion(long activityId, CompositeQuestionDef compositeQuestion, long revisionId) {
         if (!compositeQuestion.isAcceptable()) {
-            throw new DaoException("Composites only support DATE, PICKLIST, TEXT, NUMERIC and DECIMAL child questions");
+            throw new DaoException("Composite questions are only support following types for children: "
```
ssettipalli: Need to remove "are" from the sentence
Composite questions only support following types for children:

ssettipalli: Rest all looks good to me

ssettipalli (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/EquationEvaluator.java
```diff
@@ -79,4 +87,32 @@ private EquationParser buildParser(final String expression) {
...
+
+    private List<BigDecimal> apply(final List<BigDecimal> left, final List<BigDecimal> right,
+                                   final BiFunction<BigDecimal, BigDecimal, BigDecimal> componentFunction) {
+        if (left instanceof SingleValueList) {
+            return apply(IntStreamEx.range(0, right.size()).mapToObj(left::get).toList(), right, componentFunction);
+        }
+
+        if (right instanceof SingleValueList) {
+            return apply(left, IntStreamEx.range(0, left.size()).mapToObj(right::get).toList(), componentFunction);
+        }
```
bskinner: Let's see if we can find a way to do this without the `instanceof SingleValueList` checks.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/equation/compute/EquationEvaluationTest.java
```diff
@@ -64,36 +66,52 @@ public void testEvaluateDivisionByZero() {
...
         return EquationEvaluator.builder().build().evaluate(expression);
     }
+
+    private List<BigDecimal> expected(final String value) {
+        return expected(new BigDecimal(value));
+    }
+
+    private List<BigDecimal> expected(final BigDecimal value) {
+        return Collections.singletonList(value);
+    }
```
bskinner: That's a nice touch! It helps with the readability in the test cases.

bskinner (changes requested):
Looks good, but there are a couple comments to address.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
Trimax: Thank you :)
Removed.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/SingleValueList.java
Trimax: In this case, I will have to implement two internal classes:
1. Reimplementation of the list
2. Implementation of the "infinite single value list"

I agree that the class might be package-private.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/EquationEvaluator.java
Trimax: Well, reimplementation will work. It was my original idea to implement my own lists but later decided not to reinvent the wheel. But it seems you're right. The readability and safety of reimplemented classes will be higher. Will do.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/SingleValueList.java
Trimax: Done

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/containers/SingleValueVector.java
```diff
@@ -0,0 +1,30 @@
+package org.broadinstitute.ddp.equation.containers;
+
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+
+import java.math.BigDecimal;
+import java.util.Collections;
+import java.util.List;
```
bskinner, line 9: These imports should go above the lombok imports. While it's not an error, Checkstyle does expect the import ordering to be as detailed below (and does log warnings if it's not):
* static imports
* Java/javax
* All others

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/containers/SingleValueVector.java
```diff
@@ -0,0 +1,30 @@
...
+
+@EqualsAndHashCode
+@AllArgsConstructor
+public final class SingleValueVector implements AbstractVector {
+    private final BigDecimal value;
+
+    @Override
```
bskinner, line 16: My understanding is this is conforming to an interface, not overriding a super implementation. Is `@Override` required (this comment applies to `MultiValueVector` as well)?

bskinner (approved):
Other than a few style comments, this is good to go.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/containers/SingleValueVector.java
Trimax: You're right. The annotation is not required.

PR #1492, opened 2022-04-20
Ddp 7717 dss prion medical release info since mid 2021 seems borked

leonidrakitin:
Resolved DDP-7717


archilsharashenidzebroad (approved):
LGTM

PR #1504, opened 2022-04-21
[DSS] DDP-7616. API specification for the study audit endpoint

Trimax:
Example of the query to the endpoint:

POST /study/aaaa-bbbb-cccc-dddd/audit
```
{
  "dateRange":{"from":"2020-05-01T10:00:00Z"},
  "actionType":"CREATED",
  "limitation":{"from":3,"to":26},
  "sorting":{"column":"dateTime", "type":"ASCENDING"} 
}
```


pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,95 @@
+post:
```
ssettipalli, line 1: If we are querying data, it should be a "GET" and not "POST"

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,95 @@
...
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            actionType:
+              description: the action
+              type: string
```
ssettipalli: looks like requirement is to support multiple operations with a OR .
May be action can be array type where multiple actions can be specified. If null any action should be queried in implementation.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,95 @@
...
+              nullable: true
+              enum:
+                - CREATE
+                - READ
+                - UPDATE
+                - DELETE
+                - EXPORT
+            entityType:
+              description: the entity
+              type: string
```
ssettipalli: looks like requirement is to support multiple entities with a OR .
May be entities can be array type where multiple entities can be specified. If null all entities should be queried in implementation.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,95 @@
...
+              nullable: true
+            activityCode:
+              description: the activity code (must be specified in case entity type was set to Activity)
+              nullable: true
+              type: string
+            questionStableId:
+              description: the question stable id (must be specified in case entity type was set to Answer)
+              nullable: true
+              type: string
```
ssettipalli: Need to support multiple questionStableIds, need to be an array and must be optional.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,95 @@
...
+              nullable: true
+            userGuid:
+              description: the subject user's GUID (must be specified in case entity type was set to User)
+              type: string
+              nullable: true
+            activityCode:
+              description: the activity code (must be specified in case entity type was set to Activity)
+              nullable: true
+              type: string
```
ssettipalli: Need to support multiple activity codes, need to be an array and must be optional.

ssettipalli (changes requested):

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
Trimax: Then how it will be possible to send a complex query body? The GET method path length is limited.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
Trimax: I thought about that. If we're going to process them independently, then yes. I mean
`(x IN (1,2,3)) AND (y IN ('a','b'))`

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,96 @@
...
+              type: array
+              items:
+                type: string
+            questionStableId:
+              description: the question stable id (must be specified in case entity type was set to Answer)
+              nullable: true
+              type: array
+              items:
+                type: string
+            limitation:
```
alextanlan: This is for pagination purpose, right ?

Trimax: Right. I assume we would like to have something like infinite scroll (or pagination (from BE POV it's the same)), so I just propose to expose the limit constraint

ssettipalli (dismissed):

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,96 @@
...
+              description: the limitation constraint
```
bskinner: What exactly is this a limitation of, and how is this expected to function? It is not immediately obvious from the description.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1992,6 +1992,45 @@ Section:
...
+    - dateTime
+    - description
+  properties:
+    studyGuid:
+      type: string
+      format: guid
+    operatorGuid:
+      type: string
+      format: guid
+    operatorEMail:
```
bskinner: The casing of `EMail` is mixed- there are both `EMail` and `Email`. Let's switch all the cases to `Email` for consistency (so, `operatorEmail` and `userEmail`

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2894,3 +2933,49 @@ Suggestion.ActivityInstance:
...
+    - to
+  properties:
+    from:
+      type: string
+      description: the lower constraint of the range. This value is inclusive. The date format is a full, extended ISO-8601 Calendar date and time in the format `YYYY-MM-DDTHH:MM:SSZ` in the UTC time zone.
+      nullable: true
+      format: date-time
+    to:
+      type: string
+      description: the upper constraint of the range. This value is inclusive. The date format is a full, extended ISO-8601 Calendar date in the format `YYYY-MM-DD` in the UTC time zone.
```
bskinner: The date format described in this field does not match the one specified in the `from` property

bskinner (changes requested):

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
Trimax: This is a representation of the LIMIT clause from SQL. Literally, the client might specify the start row and the number of rows.

Trimax: Changed description accordingly.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Done

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Fixed. Good catch.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2894,3 +2933,49 @@ Suggestion.ActivityInstance:
...
+    from:
+      type: integer
+      description: the lower constraint of the range. This value is inclusive.
+      nullable: true
+      format: int64
+    to:
+      type: integer
+      description: the upper constraint of the range. This value is inclusive.
+      nullable: true
+      format: int64
```
MocanaAtBroad, line 3054: I don't think the from/to is enough to enable pagination.
Client is likely to have a limit of items that wants to display at a time, but does not know how many total rows there will be in a particular result set.
So I think that one capability we would want is for client to know on the first query what is the total number of results so it it know if it should allow user to request another page of results or not.
This could be achieved by including a "limit" in the query and returning a "total" in the results. 
A follow up query then could specify an offset/index to start reading the next set of values (in first query, the offset default value is "0").
(I am not saying that we should do exactly what I wrote, just suggesting that this is the sort of interaction that would be needed with client).
Another problem that would be SUPER NICE if we dealt with it in API and implementation is if we can keep the result set "locked" during pagination.
What I mean is that the audit trail data set will continually grow/always new rows are going to be added and can be a little confusing for users to do an initial query, find that there are 350 results, click for the next page and find out that there are 352 results now and that the second page includes some of the same results from the first page.
There are multiple strategies that can be used to keep this dataset locked while paging.
I recommend doing some Googling to see how other folks have resolved this REST API paging problem and see what we can learn from prior art.
A reminder: this audit table is going to become very quickly the largest in Pepper! So this paging piece is super-important. But no pressure!

Trimax: Well, a couple of comments from my side.
1. Regarding the pagination: I'm not quite sure I follow you entirely. When you implement infinite scroll, once the user reaches the end of the page FE sends another query with the shifted start index. And this happens until BE returns not empty result. To avoid delays in rendering, we can use "double buffering" technics and preload the next page in advance.
2. As for the result set, the simplest thing we can do here is to create temporary table or view and assign a hash to each view. Then, based on the hash we can fetch and narrow down the result set from the view. So, filter will be extended by result set hash.

Trimax: But do we really need this? I mean most likely user will request the logs in natural order (id ASC or date_created ASC). So the situation you described wouldn't be possible, right?

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,96 @@
...
+              description: the question stable id (must be specified in case entity type was set to Answer)
+              nullable: true
+              type: array
+              items:
+                type: string
+            limitation:
+              description: the limitation constraint (specifies the first and the last row of the range to be fetched)
+              nullable: true
+              type: object
+              $ref: '../pepper.yml#/components/schemas/IntegerRange'
```
bskinner: Assuming this structure does not change from the discussion with @MocanaAtBroad from above- Can we rename this to pagination, or something along those lines? `limitation` is extremely vague and an unusual use of the term given the context. It's also unclear what needs to be specified here- the description does specify "the first and last row", but doesn't specify what about the rows is being specified.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
```diff
@@ -0,0 +1,96 @@
...
+            sorting:
+              description: the sorting options
+              nullable: true
+              type: object
+              $ref: '../pepper.yml#/components/schemas/Sorting'
+  responses:
+    200:
+      $ref: '../pepper.yml#/components/responses/StudyAuditResponse'
+    400:
+      description: invitation not valid
```
bskinner: Looks like there's some copy & paste shrapnel here.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1992,6 +1992,45 @@ Section:
...
+  required:
+    - studyGuid
+    - operatorGuid
+    - operatorEmail
+    - userGuid
+    - userEmail
+    - entityType
+    - dateTime
+    - description
+  properties:
```
bskinner, line 2061: Can we add a unique identifier field here so the client can reliably identify each record?

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: OK. I don't think we will be doing infinite scrolling. We will want to do an initial query and the client will want to know if there are more results. And if there are more results how many there are.
Ideally, we client will be able to calculate how many pages of results there are to display (maybe we will have a Next and a Previous button or maybe a list of buttons with the page numbers).
One of the things that I am requesting we enable in the design of the API (and later, the implementation), is that, after the initial query, if the client goes to page 1 of results and the page 2 and 3, if they go back to page 2 they will see the exact same results they had seen before on page 2 UNLESS they do a brand new query.

MocanaAtBroad: The default order should actually be DESC by time stamp: the most recent events at the top. But even if sorting is not required (we need to check with David O on sorting), you can see that if you go back to the first page of results AND a new event has occurred since the query, there can be potentially new items that were not there before.

MocanaAtBroad: I think this dude had a good discussion of the issues involved in designing this sort of APIs.
https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/

Trimax: Agreed with DESC by default.
I went through the article you sent me and as far as I understood, what we need is [Keyset pagination](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#keyset-pagination). And it is totally possible to implement using current API schema by the combination of `limitation` statement and `dateRange` restriction.

pepper-apis/docs/specification/src/endpoints/studies.audit.yml
Trimax: Got it. Sure, I will rename it.

pepper-apis/docs/specification/src/components/schemas.yml
Trimax: Yep. Will expose PK

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2894,3 +2933,49 @@ Suggestion.ActivityInstance:
...
+      description: the upper constraint of the range. This value is inclusive.
+      nullable: true
+      format: int64
+Sorting:
+  type: object
+  properties:
+    column:
+      description: the column name
+      nullable: true
+      type: string
```
MocanaAtBroad, line 3061: This description might need to be more specific. the value of `column` has to be one of the attribute names returned with each audit trail record right? (e.g., studyGuid, operatorGuid, etc from your AuditTrailResponse definition).

Trimax: Right. I was thinking the we might define enum (mapping between column name and database column name) or just get the DB column name using reflection mechanism.

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: Is IntegerRange at line 2893 being used anymore?

MocanaAtBroad: Still not seeing how the client will know how many pages of results are available after initial query. Don't we need a count of the total number of results available? How will client know that there are more results?

Trimax: Probably we need to implement one more method with the same signature - **count**. Accepting the filter will return the only number - total results matching the criteria.

I was a bit confused by pagination since it is quite an outdated approach as far as I know.

Answering your question: how the client will know if there are more results:
Front-end performs two parallel requests:
1. Search(0; 10)
2. Search(10, 10)

Front-end renders only the first one. Once the page is scrolled to the bottom it executes the next search (20; 10) and renders a buffered one. And so on, until the number of results is less than the requested count. Once we received fewer results than queried that means it was the last batch.

It is some kind of double-buffering. So the data will be loaded smoothly.

Trimax: If we still want to support the pages, then my suggestion is to implement pair of the methods:
1. List<R> search(F filter) - does the real search
2. Integer count(F filter) - counts the number of results matching the filter (ignoring the Pagination property)

Having those, two UI approaches might be implemented: infinite scroll and pagination. Pagination might be reached by calling count just once and dividing the number of the results by page size.

ssettipalli (approved):

ssettipalli (approved):

ssettipalli (approved):

Trimax:
@bskinner, could you please re-approve the PR. Looks like you requested some changes and I believe everything was fixed some time ago.

PR #1505, opened 2022-04-21
[DSS] DDP-7711: Add definition of a tabular block

alextanlan:
## Context

This PR updates the API specification to add API definitions for a new TABULAR container block.

See related Angular PR [(PR-1263)](https://github.com/broadinstitute/ddp-angular/pull/1263) for displaying the tabular block.


MocanaAtBroad:
I apologize, but I have changed my mind about how we should go about designing and implementing the tabular container.
I think it might be easier to try to implement it first in Angular AND THEN design the API.
Approach that I am going to suggest is that we just mock some JSON data in Angular that looks like an activity with a tabular container and use that to try and generate a tabular display in Angular.
I think that might be easier than trying to figure out in the abstract what the server should send to the client and we can change things before we settle on a final API.
Don't think this would be the best approach in most cases, but this one feels different.
If you disagree with approach I am proposing here, please DO TELL ME!

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -861,6 +862,77 @@ Block.Content:
...
+    firstColumnBooleanQuestion:
+      allOf:
+        - $ref: '#/Question.Boolean'
+        - description: A boolean question that can be situated in first cell of a row as to enable/disable the whole row
+        - nullable: true
+    questionLabel:
+      type: string
+      description: An alternative label for a question row (instead of the question prompt)
+    mainQuestion:
+      $ref: '#/Question'
```
MocanaAtBroad: I am not seeing that we have to create a special case for the boolean checkbox. It's just another question that happens to be in same row. No? At least in the mockups I did looked like we could get away with not treating it special.

MocanaAtBroad (commented):
I also think it is OK (if you think it would simplify things) to extend our existing QUESTION definition to include additional properties, like numberOfColumns that would help with the layout. Might be simpler than keeping all the layout info in a separate JSON object.

alextanlan:
Ok, let's do it. 
Starting FE implementation then.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -861,6 +862,77 @@ Block.Content:
...
+          items:
+            $ref: '#/Block.Tabular.ContentItem'
+
+Block.Tabular.Header:
+  type: object
+  required:
+    - label
+    - startColumn
+    - endColumn
+  properties:
```
Trimax, line 897: Could you please give some details on that? What is startColumn & endColumn in terms of Header block?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1697,6 +1746,9 @@ Question.Composite:
         type: array
         items:
           $ref: '#/Answer.Composite'
+      tabularSeparator:
+        type: string
+        description: a symbol/word to display between composite children in a tabular block
```
alextanlan, line 1753: An example (`Or` word):

image

pepper-apis/docs/specification/src/components/schemas.yml
alextanlan: Sure. 
Those numbers are used for setting [grid-column css property](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column).
Generally it is a vertical track (invisible border between cells) in grid-layout. 

Here in API spec:
`StartColumn` - from where the header should start,
`EndColumn` - till what column it should be spread.

e.g.
Untitled3
Data model for the headers above: 
```
headers: [
  {
    label: 'Type',
    startColumn: 1, 
    endColumn: 2
  },
  {
    label: 'First onset or event (it can be very looooooooooooooong title)',
    startColumn: 3, 
    endColumn: 5
  },
]

```

Trimax (dismissed):

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -861,6 +862,54 @@ Block.Content:
...
+        columnsCount:
+          type: integer
+          minimum: 1
+          description: The total number of the tabular component columns.
+        headers:
+          type: array
+          items:
+            $ref: '#/Block.Tabular.Header'
+        content:
+          type: array
```
Trimax, line 888: Am I correct in assuming that each row (each ContentRow) must have exactly **columnsCount** objects?

Trimax (approved):

PR #1519, opened 2022-04-26
[SB] DDP-7384 Fix: Removed composite question (added component)

leonidrakitin:
DDP-7384 Fix: Removed composite question (returned Institution component)


archilsharashenidzebroad (approved):
LGTM

PR #1523, opened 2022-04-26
[DSM] DDP-7722 fix rgp export

LKapanad:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/excel/ParticipantRecordData.java
```diff
@@ -175,4 +203,13 @@ private Collection<?> getNestedValue(String fieldName, Map<String, Object> esDat
...
     }
+
+    private boolean isJsonValue(String jsonInString) {
+        try {
+            ObjectMapperSingleton.instance().readTree(jsonInString);
+            return true;
+        } catch (IOException e) {
+            return false;
+        }
+    }
```
SimoneMaiwald: I don't think this method is used anywhere

SimoneMaiwald (approved):

PR #1525, opened 2022-04-26
[DSM] DDP-7888: abstraction

SimoneMaiwald:
fixing abstraction by adding abstraction activity to abstraction wrapper


pegahtah (dismissed):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
```diff
@@ -45,7 +45,7 @@ public Object doPatch() {
 
     @Override
     public Object patchNameValuePairs() {
-        ESProfile profile =
+        profile =
```
SimoneMaiwald, line 48: @gmakharat @Anzori-Ghurtchumelia I think this will fix [DDP-7889](https://broadinstitute.atlassian.net/browse/DDP-7889)

SimoneMaiwald: please check

pegahtah (approved):

PR #1530, opened 2022-04-27
[DSS] DDP-7248: Audit Trail Service implementation

Trimax:
The audit trail service is implemented in terms of this PR. The service logs events happening with different objects in the database.

JDBI library version was updated in order to support optional nested beans.


pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
```diff
@@ -0,0 +1,67 @@
...
+    <changeSet author="Dmitrii Eliseev" id="20220427-audit-table">
+        <createTable tableName="audit_trail">
+            <column name="audit_trail_id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" primaryKeyName="audit_trail_pk"/>
+            </column>
+
+            <column name="study_id" type="bigint" />
+
+            <column name="operator_id" type="bigint">
+                <constraints nullable="true" />
```
ssettipalli: Isn't nullable=true by default any way ?
Same applies to all below columns too

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
```diff
@@ -0,0 +1,67 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+    <changeSet author="Dmitrii Eliseev" id="20220427-audit-table">
+        <createTable tableName="audit_trail">
+            <column name="audit_trail_id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" primaryKeyName="audit_trail_pk"/>
+            </column>
+
+            <column name="study_id" type="bigint" />
```
ssettipalli: Do you mean to make this Not null ? 
We need to set nullable="false" , default is true right ?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
Trimax: Changed

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
Trimax: What is your suggestion for the **onDelete** trigger of FK:
```
        <addForeignKeyConstraint baseTableName="audit_trail" baseColumnNames="study_id"
                                 referencedTableName="umbrella_study" referencedColumnNames="umbrella_study_id"
                                 constraintName="audit_trail_study_id_fk" onDelete="SET NULL"/>
```

Currently, study_id will be set to null. If we set the constraint to not null, then on delete of study we have to perform CASCADE DELETE of the audit log. But I'm not sure we want to delete the log.

ssettipalli: I don't believe we delete studies, we invalidate them using StudyBuilder --invalidate option

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
```diff
@@ -0,0 +1,49 @@
...
+            <column name="subject_user_guid" type="varchar(50)" />
+
+            <column name="description" type="text" />
+        </createTable>
+
+        <addForeignKeyConstraint baseTableName="audit_trail" baseColumnNames="study_id"
+                                 referencedTableName="umbrella_study" referencedColumnNames="umbrella_study_id"
+                                 constraintName="audit_trail_study_id_fk" onDelete="SET NULL"/>
+
+        <addForeignKeyConstraint baseTableName="audit_trail" baseColumnNames="operator_id"
```
ssettipalli, line 36: We have DeleteUser feature which deletes users and any dependent data.
These user FKs impact that functionality.
We need to either delete these audit logs or need to do something different if we ever want to delete users who have audit logs.
 @MocanaAtBroad 
 Do we want to delete audit logs as part fo User deletion !

Trimax: It has the trigger `ON DELETE SET NULL`, so it shouldn't be an issue. Am I missing anything?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
Trimax: So, then I will update this trigger to `ON DELETE RESTRICT` right? Is that what you suggest @ssettipalli ?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
ssettipalli: ON DELETE SET NULL is good

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
ssettipalli: I was thinking may be make the column study_id NOT NULL and remove onDelete="SET NULL" because we never delete complete study and any log entry is related to some study.
This is with assumption that we have log entries associated with some study for sure.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
Trimax: Ok, will update the PR in a couple of minutes.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7248.xml
Trimax: Ok, will update the PR in a couple of minutes.

ssettipalli (approved):

PR #1539, opened 2022-04-28
[DSS] Equation question inside composite (test added)

Trimax:
1. Manual test (incorrect result)
Снимок экрана 2022-04-28 в 15 52 57

2. SQL query returns the correct result (for WIDTH & HEIGHT questions)
Снимок экрана 2022-04-28 в 15 54 13

3. Implemented test (with the same input) returns the correct result

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
@@ -45,7 +45,6 @@ import org.broadinstitute.ddp.db.dto.ActivityInstanceSelectQuestionDto;
 import org.broadinstitute.ddp.db.dto.TypedQuestionId;
 import org.broadinstitute.ddp.db.dto.validation.RuleDto;
 import org.broadinstitute.ddp.equation.QuestionEvaluator;
-import org.broadinstitute.ddp.json.EquationResponse;
 import org.broadinstitute.ddp.model.activity.definition.QuestionBlockDef;
 import org.broadinstitute.ddp.model.activity.definition.question.AgreementQuestionDef;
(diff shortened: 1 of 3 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -918,9 +917,9 @@ default Question getEquationQuestion(EquationQuestionDto dto, String activityIns
                 dto.getAdditionalInfoHeaderTemplateId(),
                 dto.getAdditionalInfoFooterTemplateId(),
                 StreamEx.of(getJdbiEquationQuestion().findEquationsByActivityInstanceGuid(activityInstanceGuid))
+                        .filterBy(EquationQuestionDto::getStableId, dto.getStableId())
```
Trimax, line 920: Just an optimization. No need to compute not relevant equations

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/QuestionCreatorHelper.java
```diff
@@ -268,9 +268,9 @@ EquationQuestion createEquationQuestion(AIBuilderContext ctx, EquationQuestionDe
                         ctx, questionDef.getAdditionalInfoFooterTemplate()),
                 StreamEx.of(new QuestionCachedDao(ctx.getHandle()).getJdbiEquationQuestion()
                                 .findEquationsByActivityInstanceGuid(ctx.getInstanceGuid()))
+                        .filterBy(EquationQuestionDto::getStableId, questionDef.getStableId())
```
Trimax, line 271: Just an optimization. No need to compute not relevant equations

alextanlan (approved):

PR #1545, opened 2022-04-29
[DSM] DDP-7174: DSM_use_role_migration

pegahtah:
DDP-7174


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/DdpDBUtils.java
```diff
@@ -0,0 +1,58 @@
...
+     * Helper to check uniqueness of generated guids against database table.
+     */
+    private static String uniqueGuid(Supplier<String> generator, Handle handle, String table, String guidColumn) {
+        String query = String.format(IS_GUID_UNIQUE_QUERY_TEMPLATE, table, guidColumn);
+        String candidate = null;
+        boolean unique = false;
+
+        while (!unique) {
+            candidate = generator.get();
+            try (PreparedStatement stmt = handle.getConnection().prepareStatement(query)) {
```
gmakharat, line 31: Does not DSS already have class which is responsible to make queries to the database to check whether guid is unique or not?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/roles/UserRoleDao.java
```diff
@@ -0,0 +1,124 @@
...
+            return dbVals;
+        });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Error getting user by id ", results.resultException);
+        }
+        return (List<String>) results.resultValue;
+    }
+
+    public static HashMap<Long, AssigneeDto> getAssigneeMap(String realm) {
```
gmakharat: why do we have this method as `static`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/roles/UserRoleDao.java
```diff
@@ -0,0 +1,124 @@
...
+                        assigneeDto.getEmail().orElseThrow()));
+            }
+            return null;
+        });
+
+        logger.info("Found " + assignees.size() + " assignees ");
+        return assignees;
+    }
+
+    public static Collection<AssigneeDto> getAssignees(String realm) {
```
gmakharat: `static` here as well

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
```diff
@@ -65,4 +69,83 @@ public int delete(int id) {
         }
         return (int) results.resultValue;
     }
+
+    public Map<Integer, String> getAllUserMap() {
+        SimpleResult results = TransactionWrapper.withTxn(TransactionWrapper.DB.SHARED_DB, handle -> {
+            SimpleResult dbVals = new SimpleResult();
+            List<UserDto> res = handle.attach(JdbiUser.class).getUserMap();
+            Map<Long, String> map = new HashMap<>();
+            if (res != null) {
```
gmakharat, line 84: as far as I know if method of jdbi has `List` as return value and there are no results from `SQL` query it returns an empty list, maybe @Trimax can help us to clarify that, if so, we can ignore check on null.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
```diff
@@ -3,28 +3,37 @@
...
 
-@Setter
+@Data
 public class UserDto {
     @SerializedName ("user_id")
-    private int userId;
+    private long userId;
+    private String guid;
+    private String firstName;
+    private String lastName;
```
gmakharat: what's the point of having `name` if we have `firstName` and `lastName`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -54,24 +54,27 @@ public class AuthenticationRoute implements Route {
...
+                               UserSettingsDao userSettingsDao) {
 
         this.auth0Util = auth0Util;
-        this.userUtil = userUtil;
         this.auth0Domain = auth0Domain;
         this.clientSecret = clientSecret;
         this.auth0ClientId = auth0ClientId;
         this.auth0MgmntAudience = auth0MgmntAudience;
         this.audienceNameSpace = audienceNameSpace;
+        this.userDao = new UserDao();
```
gmakharat, line 77: if we pass `UserSettingsDao` through the constructor let's pass `UserDao` through the constructor to make this class fully dynamic.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -89,24 +92,25 @@ public Object handle(Request request, Response response) {
                         logger.info("User (" + email + ") was found ");
                         Gson gson = new Gson();
```
gmakharat, line 94: there is a singleton class called `ObjectMapperSingleton` which is wrapper of `jackson`, let's use it everywhere instead of `gson`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java
```diff
@@ -121,7 +125,7 @@ public Object handle(Request request, Response response) {
                             haltWithErrorMsg(401, response, "DSMToken was null! Not authorized user", e);
                         }
                     } else {
-                        haltWithErrorMsg(400, response, "user was null");
+                        haltWithErrorMsg(400, response, "user info in token was null");
```
gmakharat, line 129: this class will rot, we can survive if we change nested if to guard conditions like:
`if (StringUtils.isBlank(auth0Token)) { haltWithErrorMsg(400, response, "user info in token was null"); }`
such structure to avoid nested if statements

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/UserSettingRoute.java
```diff
@@ -27,8 +34,8 @@ public Object processRequest(Request request, Response response, String userId)
                     || UserUtil.checkUserAccess(null, userId, "pt_list_view", userIdRequest)) {
                 if (StringUtils.isNotBlank(userIdRequest)) {
                     String requestBody = request.body();
-                    UserSettings userSettings = new Gson().fromJson(requestBody, UserSettings.class);
-                    UserSettings.editUserSettings(userIdRequest, userSettings);
+                    UserSettingsDto userSettingsDto = new Gson().fromJson(requestBody, UserSettingsDto.class);
```
gmakharat, line 37: let's use `ObjectMapperSingleton` here instead of `gson`

PR #1550, opened 2022-04-29
[DSS] DDP-7921: Comparison rule API specification updated

Trimax:
1. Update API specification by adding COMPARISON_RULE type
2. The ComparisonRule model class updated by adding referenceQuestionStableId

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -2275,6 +2275,9 @@ Validation.Comparison:
   allOf:
     - $ref: '#/Validation'
     - type: object
+      required:
+        - reference_question_stable_id
+        - comparison_validation_type
       description: compares the value of the given answer with the answer to the referenced question
       properties:
         rule:
@@ -2284,6 +2287,7 @@ Validation.Comparison:
             - COMPARISON
         reference_question_stable_id:
           type: string
+          format: guid
         comparison_validation_type:
           type: string
           enum:
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -2041,6 +2041,7 @@ Validation:
     mapping:
       AGE_RANGE: '#/Validation.AgeRange'
       COMPLETE: '#/Validation.Complete'
+      COMPARISON: '#/Validation.Comparison'
       DATE_RANGE: '#/Validation.DateRange'
       DAY_REQUIRED: '#/Validation.DayRequired'
       INT_RANGE: '#/Validation.IntRange'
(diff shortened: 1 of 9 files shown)


ssettipalli (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/ValidationDao.java
```diff
@@ -253,6 +254,11 @@ private Rule getValidationRule(RuleDto dto, String message, String hint) {
                         .correctionHint(hint)
                         .allowSave(dto.isAllowSave())
                         .referenceQuestionId(comparisonRule.getReferenceQuestionId())
+                        .referenceQuestionStableId(TransactionWrapper.withTxn(handle ->
+                                handle.attach(QuestionDao.class).getJdbiQuestion()
+                                        .findQuestionDtoById(comparisonRule.getReferenceQuestionId())
+                                        .map(QuestionDto::getStableId)
+                                        .orElse(null)))
                         .comparisonType(comparisonRule.getType())
```
bskinner: What options do we have aside from triggering another database round-trip _just_ for the `stableId` here? My impression is all of the Rule DTOs are composed using the results from the `select_all_question_rule_dtos` statement. Can the `select_all_question_rule_dtos` query be extended to include the `stableId` in the initial round trip?

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2270,6 +2271,28 @@ Validation.UniqueValue:
...
+    - type: object
+      description: compares the value of the given answer with the answer to the referenced question
+      properties:
+        rule:
+          type: string
+          default: COMPARISON
+          enum:
+            - COMPARISON
+        reference_question_stable_id:
+          type: string
```
bskinner, line 2289: Let's add a `format: guid` property to this field

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -2270,6 +2271,28 @@ Validation.UniqueValue:
           default: UNIQUE_VALUE
           enum:
             - UNIQUE_VALUE
+Validation.Comparison:
+  allOf:
+    - $ref: '#/Validation'
+    - type: object
+      description: compares the value of the given answer with the answer to the referenced question
+      properties:
```
bskinner, line 2282: Which of the below properties are `required`?

bskinner (changes requested):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/ValidationDao.java
Trimax: Actually, this is a good idea. Let me update DTO & SQL

bskinner (approved):
Thanks for the changes! I'm good with merging this.

MocanaAtBroad (approved):

PR #1558, opened 2022-04-29
[DSM] DDP-7913: abstraction

SimoneMaiwald:
adding primaryKeyId to response map
changing participantDataMap forEach because of double
only trying to write workflow if nameValue was of d.data


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -13,15 +13,12 @@
 import org.broadinstitute.dsm.util.NotificationUtil;
 import org.broadinstitute.dsm.util.PatchUtil;
 import org.broadinstitute.dsm.util.UserUtil;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
 
 //Class needs to be refactored as soon as possible!!!
```
gmakharat, line 19: I guess we can remove this comment already :smile:

gmakharat (approved):

ScalaChamp (approved):

PR #1575, opened 2022-05-03
[DSM] Create separate value provider and renderer for all type of questions

LKapanad:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -67,17 +68,18 @@ public class Filter {
...
+    public static final String DATE = "DATE";
+    public static final String DATE_SHORT = "DATE_SHORT";
+    public static final String ADDITIONAL_VALUES = "ADDITIONALVALUE";
+    public static final String NUMBER = "NUMBER";
+    public static final String BOOLEAN = "BOOLEAN";
+    public static final String CHECKBOX = "CHECKBOX";
+    public static final String COMPOSITE = "COMPOSITE"; //ES type
+    public static final String JSON_ARRAY = "JSONARRAY"; //Sample result
+    public static final String AGREEMENT = "AGREEMENT";
+    public static final String MATRIX = "MATRIX";
```
gmakharat, line 82: I think `enum` is the best case to use for those constants, there is also enum class of `DSS` `QuestionType` you may find it useful as well.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/excel/renderer/ValueProvider.java
```diff
@@ -0,0 +1,145 @@
...
+            if (((Collection<?>) o).isEmpty()) {
+                return Collections.singletonList(StringUtils.EMPTY);
+            }
+            return (Collection<?>) o;
+        } else {
+            return Collections.singletonList(o);
+        }
+    }
+
+}
```
gmakharat, line 155: I think having all this calculation in interface does not look really clean.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/excel/renderer/ValueProvider.java
```diff
@@ -0,0 +1,145 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.model.Filter;
+import org.broadinstitute.dsm.model.ParticipantColumn;
+import org.broadinstitute.dsm.model.elastic.sort.Alias;
+import org.broadinstitute.dsm.statics.ESObjectConstants;
+import org.broadinstitute.dsm.util.ElasticSearchUtil;
+import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
+
+public interface ValueProvider {
```
gmakharat, line 22: Since refactoring is started for better design, I would love to see unit tests as well.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Filter.java
LKapanad: @gmakharat I wanted to use QuestionType class but it's in dss module and currently there's no dependency in dsm

gmakharat: let's create ours then

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -35,11 +37,12 @@ public DownloadParticipantListRoute() {
     public Object processRequest(Request request, Response response, String userId) throws Exception {
```
SimoneMaiwald, line 37: looking at the processRequest: Can we please try to come up with a way to do counting and processing in one? In a study like TestBoston with 10.000 patients there would need to be 20 calls to ES for counting and another 20 calls for processing.  that seems excessive and it would be a big hit to performance.
Also even if we end up doing it that way, can we please move the whole block into a separate method and not just copy paste the code?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Filter.java
Trimax: The QuestionType might be moved to dsm-common (or core module) that is common for us.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -91,6 +93,7 @@ public class Filter {
     private String[] selectedOptions;
     private ParticipantColumn participantColumn;
     private String additionalType;
+    private List<NameValue> options;
 
     public Filter() {
```
Trimax, line 98: NoArgsContructor

Trimax: Looks like this class might be a bit optimized using Lombok's Data annotation

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Filter.java
SimoneMaiwald: if there is a class in dss-core that dsm needs as well, we should move it to ddp-common!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/QuestionType.java
```diff
@@ -0,0 +1,29 @@
+package org.broadinstitute.dsm.model;
```
SimoneMaiwald, line 1: try to reuse the dss QuestionType and move it to ddp-common

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
gmakharat: Maybe we can introduce Java's parallelism here or Producer/Consumer technique to speed up the process and avoid processing huge amount of data by only one thread

devonbush:
closing as superseded by #1884

PR #1594, opened 2022-05-06
comparison validation fix

ssettipalli:
fix for https://broadinstitute.slack.com/archives/GC80PJLTZ/p1651797690687739
Basil bug with comparison rule/validation

1. SQL fix 
2. As per Basil configuration:
Question in LOVED_ONE activity is referring to (comparing to) question in another activity NUMERIC_ACTIVITY
Code in ValidationRuleCreator.createComparisonRule is looking for question with referring stableID in same activity instance (LOVED_ONE) and hence not able to find the question.
ValidationRuleCreator.createComparisonRule should be updated to lookup referred question  probably by stableId, user and study and NOT activityInstance

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/ValidationRuleCreator.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/ValidationRuleCreator.java
@@ -9,6 +9,7 @@ import org.broadinstitute.ddp.db.ActivityDefStore;
 import org.broadinstitute.ddp.db.TransactionWrapper;
 import org.broadinstitute.ddp.db.dao.QuestionDao;
 import org.broadinstitute.ddp.db.dao.ValidationDao;
+import org.broadinstitute.ddp.db.dto.QuestionDto;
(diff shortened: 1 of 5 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/question/ValidationRuleCreator.java
```diff
@@ -207,6 +208,12 @@ private UniqueValueRule createUniqueValueRule(AIBuilderContext ctx, UniqueValueR
     }
 
     private ComparisonRule createComparisonRule(AIBuilderContext ctx, ComparisonRuleDef ruleDef) {
+        QuestionDto questionDto = TransactionWrapper.withTxn(handle -> handle.attach(QuestionDao.class).getJdbiQuestion()
```
ssettipalli, line 211: not checking if user has instance of the activity with this question !!

Trimax: It doesn't really matter since in `ComparisonRule.validate` there is a check if the answer is present.

Trimax (approved):

PR #1570, opened 2022-05-02
[SB] Bring contents of studybuilder folder into studybuilder-cli folder

MocanaAtBroad:
## Context
This is a change that we have been discussing for a while: bring the contents of study-builder directory, including all the study config files, into studybuilder-cli Maven module directory.
Main motivation is that it allows loading all the assets in the repo into a single IDE project.

The number of "Files changed" is a ridiculous 2,034 files but it only involves moving files.
Only files that had its contents modified are `render.sh` (the script that generates the configuration files). Had change some paths.
And the README.md to indicate how the study builder cli should now be built (as a Maven module).

## Checklist

- [ X] I have labeled the type of changes involved using the `C-*` labels.
- [ X] I have assessed potential risks and labeled using the `R-*` labels.
- [X ] I have considered error handling and alerts, and added `L-*` labels as needed.
- [ X] I have considered security and privacy, and added `I-*` labels as needed
- [X ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [X ] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ X] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

_If unsure or need help with any of the above items, add the `help wanted` label. For items that starts with `If applicable`, if it is not applicable, check it off and add `n/a` in front._

## FUD Score

_Overall, how are you feeling about these changes?_

- [X ] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## Release

- [X ] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #1598, opened 2022-05-06
fix: add necessary `.gitignore` changes after #1570 PR

grigol-brd:
Updated study builder configurations path in `.gitignore` after #1570


grigol-brd:
@MocanaAtBroad @Trimax Please take a look at this change, I believe it's correct but just to double-check.

nsakande (approved):
LGTM

PR #1599, opened 2022-05-06
[SB] Resolved DDP-7954 PS: Child relationship to SVD patient multi-select is broken

leonidrakitin:
Resolved DDP-7954


grigol-brd:
Please run tests to allow us to merge

grigol-brd (approved):

PR #1609, opened 2022-05-11
[DSS] Further Lombok migrations

Trimax:


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/migration/AngioMigrationRun.java
```diff
@@ -1,84 +1,27 @@
 package org.broadinstitute.ddp.model.migration;
```
ssettipalli, line 1: Code under migration package is not in use any more. I think we don't need to invest time to refactor it.
It was written and used to migrated some studies from different system to pepper.

ssettipalli (approved):
Except Auth0Util, all classes in this PR are not in use any more. All used for migration.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/migration/AngioMigrationRun.java
Trimax: Should we delete this code then?

PR #1615, opened 2022-05-12
[DSS] DDP-7713: Tabular block implementation

Trimax:
Local Basil study import was successful.

**tabular_block** table
Снимок экрана 2022-05-12 в 14 50 09

**tabular_block_header** table
Снимок экрана 2022-05-12 в 14 50 16
_(startColumn & endColumn were replaced by columnSpan according to the API definition, I'm just too lazy to update the screenshot)_

**tabular_block_question** table
Снимок экрана 2022-05-12 в 14 50 26

Response trimmed to tabular section (Activity Get Test):
```
{
  "formType": "GENERAL",
  "listStyleHint": "NONE",
  "readonlyHint": null,
  "introduction": null,
  "closing": null,
  "sections": [
    ...
    {
      "name": null,
      "icons": [],
      "blocks": [
        {
          "headers": [
            {
              "columnSpan": 1,
              "labelTemplateId": 21401
            },
            {
              "columnSpan": 1,
              "labelTemplateId": 21402
            }
          ],
          "content": [
            [
              {
                "placeholderText": null,
                "scale": 2,
                "questionType": "DECIMAL",
                "stableId": "QUESTION_LU",
                "prompt": "This is value",
                "textPrompt": "This is value",
                "tooltip": null,
                "readonly": false,
                "additionalInfoHeader": null,
                "additionalInfoFooter": null,
                "answers": [],
                "validations": [],
                "validationFailures": null
              },
              {
                "placeholderText": null,
                "scale": 2,
                "questionType": "DECIMAL",
                "stableId": "QUESTION_RU",
                "prompt": "This is value",
                "textPrompt": "This is value",
                "tooltip": null,
                "readonly": false,
                "additionalInfoHeader": null,
                "additionalInfoFooter": null,
                "answers": [],
                "validations": [],
                "validationFailures": null
              }
            ],
            [
              {
                "placeholderText": null,
                "scale": 2,
                "questionType": "DECIMAL",
                "stableId": "QUESTION_LB",
                "prompt": "This is value",
                "textPrompt": "This is value",
                "tooltip": null,
                "readonly": false,
                "additionalInfoHeader": null,
                "additionalInfoFooter": null,
                "answers": [],
                "validations": [],
                "validationFailures": null
              },
              {
                "placeholderText": null,
                "scale": 2,
                "questionType": "DECIMAL",
                "stableId": "QUESTION_RB",
                "prompt": "This is value",
                "textPrompt": "This is value",
                "tooltip": null,
                "readonly": false,
                "additionalInfoHeader": null,
                "additionalInfoFooter": null,
                "answers": [],
                "validations": [],
                "validationFailures": null
              }
            ]
          ],
          "columnsCount": 2,
          "blockType": "TABULAR",
          "blockGuid": "7O99MZ4SEW",
          "shown": true,
          "enabled": true
        }
      ]
    }
  ],
  "lastUpdated": null,
  "lastUpdatedText": null,
  "sectionIndex": 0,
  "activityType": "FORMS",
  "guid": "U8P8AOJW4M",
  "parentInstanceGuid": "DJZN2XH1O7",
  "activityCode": "ACT_ROUTE_ACT1652706625819",
  "statusCode": "CREATED",
  "readonly": false,
  "title": null,
  "subtitle": null,
  "canDelete": false,
  "isFollowup": false,
  "isHidden": false
}
```


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
```diff
@@ -730,6 +789,65 @@ default Map<Long, ConditionalBlockDef> collectConditionalBlockDefs(Collection<Fo
...
+
+        Map<Long, TabularBlockDef> blockDefs = new HashMap<>();
+        for (var blockDto : blockDtos) {
+            final int rowCount = 1 + StreamEx.of(questionsByBlocks.get(blockDto.getId()))
+                    .map(BlockTabularQuestionDto::getRow)
+                    .max(Integer::compareTo)
+                    .orElse(0);
+            if (rowCount == 0) {
+                throw new DaoException("Tabular block " + blockDto.getId() + " doesn't have any rows");
+            }
```
bskinner: I'm not following your logic here- in what cases will `rowCount` be equal to 0?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/BlockTabularQuestionDto.java
```diff
@@ -0,0 +1,28 @@
...
+    long questionId;
+
+    @ColumnName("block_id")
+    long blockId;
+
+    @ColumnName("_column")
+    int column;
+
+    @ColumnName("_row")
+    int row;
```
bskinner, line 27: What is the intent of the leading underscore? This is not a pattern we've used prior.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Stream;
+
+@Value
+public class TabularBlockDef extends FormBlockDef {
+    int columnsCount;
+
+    @NotEmpty
+    List<@Valid @NotNull TabularHeaderDef> headers = new ArrayList<>();
```
bskinner: Did you intend to use the `javax.validation` annotations here instead of the Lombok annotations?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+    
+    public QuestionDef get(final int row, final int column) {
+        if ((row < 0) || (row >= rows.size())) {
+            throw new IllegalArgumentException("The row must be a number between 0 and " + rows.size());
+        }
+        
+        final TabularRowDef rowDef = rows.get(row);
+        if ((column < 0) || (column >= rowDef.getQuestions().size())) {
+            throw new IllegalArgumentException("The column must be a number between 0 and " + rowDef.getQuestions().size());
```
bskinner: `java.lang.IndexOutOfBoundException` would be more appropriate here.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/ConsentElection.java
```diff
@@ -3,37 +3,22 @@
...
     @SerializedName("selected")
     private Boolean selected;
 
-    private transient String selectedExpr;
+    private final transient String selectedExpr;
 
     public ConsentElection(String stableId, String selectedExpr) {
         this.stableId = MiscUtil.checkNotBlank(stableId, "stableId");
         this.selectedExpr = MiscUtil.checkNotBlank(selectedExpr, "selectedExpr");
     }
```
bskinner, line 23: This class does not seem to have any relation to the PR- why are these modifications being included here?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/ConsentElectionDef.java
```diff
@@ -3,46 +3,20 @@
...
-    private String stableId;
+    private final String stableId;
 
     @NotBlank
     @SerializedName("selectedExpr")
-    private String selectedExpr;
+    private final String selectedExpr;
 
     private transient Long consentElectionId;
     private transient Long selectedExprId;
```
bskinner, line 21: This class does not have any clearly visible relation to the rest of the modifications here. Why is it being included with this PR?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
```diff
@@ -65,6 +73,33 @@ ComponentBlock createComponentBlock(AIBuilderContext ctx, ComponentBlockDef comp
...
+    TabularBlock createTabularBlock(AIBuilderContext ctx, TabularBlockDef tabularBlockDef) {
+        final List<List<Question>> rows = new ArrayList<>();
+        for (int row = 0; row < tabularBlockDef.getRowsCount(); row++) {
+            final List<Question> questionsRow = new ArrayList<>();
+            for (int column = 0; column < tabularBlockDef.getColumnsCount(); column++) {
+                final QuestionDef questionDef = tabularBlockDef.get(row, column);
+                if (questionDef == null) {
+                    questionsRow.add(null);
+                    continue;
+                }
```
bskinner, line 85: In what cases should we be anticipating `null` values in the row, and are there ways this can be avoided?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -959,6 +960,9 @@ private List<QuestionRecord> createQuestionRecordsForActivity(ActivityDef defini
                             allQuestions.add(questionBlock.getQuestion());
                         }
                     }
+                } else if (formBlock.getBlockType() == BlockType.TABULAR) {
+                    TabularBlockDef tabularBlock = (TabularBlockDef) formBlock;
+                    allQuestions.addAll(tabularBlock.getQuestions().collect(Collectors.toList()));
```
ssettipalli, line 965: DataExporter stuff looks ok to me.
Please make sure you run DataExporterCli for elastic and csv export and make sure questions in the new block & responses/answers  are handled and exported, particularly elastic.

pepper-apis/studybuilder-cli/studies/basil/tabular-activity.conf
```diff
@@ -0,0 +1,66 @@
...
+            },
+            {
+              "questions": [
+                {
+                  include required("../snippets/decimal-question.conf"),
+                  "stableId": "TABULAR_QUESTION_L2",
+                  "promptTemplate": {"templateType": "TEXT", "templateText": "$tabular.question.l2"},
+                  "validations": []
+                },
+                {
```
ssettipalli, line 50: Is it a valid scenario if you have only 1 question in one of the rows rather than 2 making null/no question in one of cells [row, col] ?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
ssettipalli: I would think if one of the cells will have NO question. For ex: 3 columns and one of the rows will have only 2 questions. Not sure if we have this support though. Will let dimitry confirm

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
Trimax: No cases. Forgot to remove this check. Originally it was counting, not maximum. Removed.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/BlockTabularQuestionDto.java
Trimax: Because row & column are the MySQL functions, we can't use them as-is.
https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-C
https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-R

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
Trimax: Yep. This is similar to what we have in other DEF classes. Should be used by the hibernate validator.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/ConsentElection.java
Trimax: It looks like I had some Lombok-related changes and forgot to stash them. If it is a big deal, then I'll revert this change.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/ConsentElectionDef.java
Trimax: It looks like I had some Lombok-related changes and forgot to stash them. If it is a big deal, then I'll revert this change.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
Trimax: Sampath is absolutely right. I assume we might have empty cells.

pepper-apis/studybuilder-cli/studies/basil/tabular-activity.conf
Trimax: I am afraid I didn't get your point. Could you please give an example?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/BlockTabularQuestionDto.java
bskinner: Thank you for linking in the docs!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
bskinner: My understanding is we are only using the Hibernate Validator to check POJOs deserialized by GSON from the endpoints and, IIRC, against some Templates loaded from study builder (See [GsonPojoValidator](https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/GsonPojoValidator.java)). I don't believe HOCON takes advantage of these annotations and we do not have a more general use of the library. Historically, the patterns in the `Def` classes (and elsewhere that is not JSON-adjacent) were typically intended to act as a hint to static analysis tools and for developers, and are effectively NOPs otherwise.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
bskinner: Sounds good- thank you!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
Trimax: Am I correct in assuming that your suggestion is to remove those annotations?

Trimax: I was thinking they are in use when we read the Def file from the study definition.

pepper-apis/studybuilder-cli/studies/basil/tabular-activity.conf
Trimax: Valid scenario:
```
[
  [Q, null, Q],
  [null, Q, Q],
  [Q, Q, null]
]
```

Invalid scenario:
```
[
   [Q, Q, Q],
   [Q],
   [Q, Q]
]
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
bskinner: In this case, let's remove the `@NotEmpty`, and `@Valid`, and use Lombok's `@NotNull`. If the property is guaranteed to both be non-`null` and not empty, please add that as part of the description in a Javadoc comment for the property.

Trimax: Then it is better to remove all the annotations since it makes no sense. Valid annotation suppose to say hibernate to validate nested objects. In case of removal them, NotEmpty is also not required since headers and rows are always being initialized. I'll remove all of them.

Trimax: Most likely you were referring to `NonNull` annotation, right?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
Trimax: Could you please let me know what arguments should I use when running this utility to perform the test you're referring to?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
bskinner: Yup- `lombok.NonNull` is the correct annotation- I misspoke. Unless it's on a route that descends from `org.broadinstitute.ddp.util.ValidatedJsonInputRoute`, we aren't using bean validations.

ssettipalli (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
Trimax: Got it, will do after merging

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public TabularBlockDef(final int columnsCount) {
+        super(BlockType.TABULAR);
+
+        this.columnsCount = columnsCount;
+    }
+    
+    public QuestionDef get(final int row, final int column) {
+        if ((row < 0) || (row >= rows.size())) {
+            throw new IllegalArgumentException("The row must be a number between 0 and " + rows.size());
```
bskinner:
```suggestion
            throw new IndexOutOfBoundsException("The row must be a number between 0 and " + rows.size());
```

bskinner (dismissed):
One minor suggestion, but looks good otherwise.

bskinner (approved):

PR #1616, opened 2022-05-12
[DSS] Equation & Picklist are friends or not?

Trimax:
This small PR contains quite a tricky change.
The picklist might have a vector as an answer (since it allows to pick multiple answers).
On the other hand, the picklist might be a child of a composite question. This means the composite might have several picklist answers. After all, we will have a matrix of the answers.

The current implementation of Equation doesn't support matrix as an input. So this PR makes it possible to work with picklist questions inside the composite question, but ignores multi-valued picklist questions.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
@@ -13,7 +13,10 @@ import org.broadinstitute.ddp.model.activity.types.QuestionType;
 import org.jdbi.v3.core.Handle;
 
 import java.math.BigDecimal;
-import java.util.*;
+import java.util.Map;
+import java.util.List;
+import java.util.HashMap;
+import java.util.Optional;
 
 @Slf4j
 @AllArgsConstructor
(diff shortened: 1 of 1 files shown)


bskinner (approved):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
```diff
@@ -93,11 +95,13 @@ private void addValue(final QuestionType type, final String variable, final List
...
                 this.values.put(variable, new EquationResponse(variable,
-                        StreamEx.of(picklistValues)
-                                .map(Answerable::getValue)
-                                .map(this::toString)
+                        StreamEx.of(values)
+                                .map(List.class::cast)
+                                .map(List::iterator)
+                                .map(Iterator::next)
+                                .map(SelectedPicklistOption.class::cast)
+                                .map(SelectedPicklistOption::getStableId)
```
bskinner, line 104: The `Iterator::next` method immediately reads as unusual to me. Can we add a comment here details what & why the `Iterator::next` is here?

bskinner (approved):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/equation/QuestionEvaluator.java
Trimax: Sure. Maybe it's even better to break down the code into several small methods with natural names.

PR #1617, opened 2022-05-13
[DSS] DDP-8012: Picklist option extended by adding value field

Trimax:
We have MANY to ONE relationship between picklist option title and it’s real value. For instance, the question "How do you feel today?” might have following answers (and corresponding values):
1. Perfect (3)
2. Excellent (3)
3. Good (2)
4. Nothing special (2)
5. Bad (1)
6. Awful (1)

Current database restriction limits the values to be unique so we need to do a small modification:
1. Extend picklist option value table by adding new field - value (text field without any restrictions)
2. Use newly added field as the input for equations in case of pick list
3. Extend study builder by adding the possibility to import value to the database

After running Study Builder locally in the database we have following values:
Снимок экрана 2022-05-13 в 11 03 06


ssettipalli:
Looks good.
Need to think if we want DSM to know of this new picklist attribute so that DSM can display it along with activity definition.
If so need to check activity_definition elastic index export code to make sure this is included in the definition.

ssettipalli (approved):

PR #1618, opened 2022-05-13
DSS-Some additional changes to Tabular API. Move to columnSpan instead of…

MocanaAtBroad:
After spending some time with Aleksei's draft PR for tabular container in Angular, I am suggesting a few minor changes to API here. Please see comments.

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -900,6 +900,7 @@ Block.Tabular.Header:
     columnSpan:
       type: integer
       minimum: 1
+      default: 1
       description: the number of columns this header should occupy in a tabular display
 
 Block.Tabular.ContentRow:
@@ -1630,7 +1631,6 @@ Question:
     columnSpan:
       type: integer
       minimum: 1
-      default: 1
       description: The number of columns this question should occupy in a tabular display
 Question.ActivityInstanceSelect:
   allOf:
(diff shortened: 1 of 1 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1630,6 +1628,11 @@ Question:
       type: array
       items:
         $ref: '#/ValidationFailure'
+    columnSpan:
+      type: integer
+      minimum: 1
+      default: 1
+      description: The number of columns this question should occupy in a tabular display
```
MocanaAtBroad, line 1634: Not clear here because of how github displays diffs, but this is being applied to question blocks.
This becomes useful for composites questions. columnSpan of 1 means child questions are stacked on top of each other.  If columnSpan is not used, composite will use one column per child.

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -893,17 +893,15 @@ Block.Tabular.Header:
...
-    startColumn:
+    columnSpan:
       type: integer
-      minimum: 0
-    endColumn:
-      type: integer
-      minimum: 0
+      minimum: 1
+      default: 1
+      description: the number of columns this header should occupy in a tabular display
```
MocanaAtBroad, line 904: Suggesting using columnSpan instead of start/stop. The one attribute, combined with the order of the headers in the list of headers is enough to figure out positioning. (Also avoids confusion with, say, having a header be the third in the array, but having startColumn=1.

Trimax (approved):
LGTM

PR #1639, opened 2022-05-17
[DSS] Fixed a bug with the calculation of blocks for complete.

leonidrakitin:
Fixed PutFormAnswersRoute bug: the calculation of blocks to confirm completion, it doesn't check blocks that are not enabled.


Trimax (approved):
LGTM

PR #1645, opened 2022-05-19
revert DDP-8000 auth0_domain column deletion from client table

ssettipalli:
revert ddp-8000 . auth0_domain column deletion from client table since column doesn't exist in non dev envs
verified in local, that deleting executed liquibase doesn't cause any issue.


Trimax (approved):
As far as I remember Marco mentioned that there is a script that creates this column and it's in the Liquibase list. In this case, we will face instability.

PR #1646, opened 2022-05-19
DDP-8034. file answer null fix.

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-8034
fix to avoid NPE during data export in scenarios where file upload has answer/file_upload row but no file_answer 

Simulated the prod scenario in Dev (Deleted file_answer for my upload) and tested.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -520,7 +520,7 @@ public Answer reduce(Map<Long, Answer> container, RowView view) {
                                 view.getColumn("fa_file_name", String.class),
                                 view.getColumn("fa_file_size", Long.class));
                     }
-                    ((FileAnswer) answer).getValue().add(info);
+                    ((FileAnswer) answer).getValue().add(info); //todo : check adding null FileInfo if no FileAnswer ?
```
ssettipalli: There is a testcase testPatch_fileAnswer_canCreateNullValue_andClearValueBySendingNull in PatchFormAnswersRouteStandaloneTest which is specifically checking for a null FileInfo and that made me move away from making the fix here and handling it in DataExporter.
Added a todo to revisit and check back later

bskinner: @LKapanad You were the last to modify that test- specifically the `null` check in question. Can you explain what is going on in [PatchFormAnswersRouteStandaloneTest.java:1920](https://github.com/broadinstitute/ddp-study-server/blob/6f1488319fb997a9cb9d1c6c12b5a55a8bfadac5/pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/PatchFormAnswersRouteStandaloneTest.java#L1920)? That line should be receiving a `java.lang.IndexOutOfBoundsException`, but it appears we're inserting `null` values into the array instead.

ssettipalli: updated unit test and fixed the bug

bskinner (approved):

PR #1653, opened 2022-05-19
[DSS] DDP-7931: User registration endpoint

Trimax:
This PR contains a couple of things:
1. Implementation of the new endpoint for creation the user
2. Some code cleanup (unused methods and constants were removed)
3. UserProfile class builder class was replaced with Lombok's implementation


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,65 @@
...
+                throw new RuntimeException("Tenant wasn't found for study + " + payload.getStudyGuid());
+            }
+
+            final var centerDto = handle.attach(CenterProfileDao.class).findById(payload.getCenterId());
+            if (centerDto.isEmpty()) {
+                throw new RuntimeException("Center " + payload.getCenterId() + " does not exist");
+            }
+
+            final var clientId = handle.attach(ClientDao.class).getClientDao().insertClient(
+                    "DON'T KNOW", "DON'T KNOW", tenantDto.getId(), "DON'T KNOW");
```
Trimax: How exactly should I create the user without knowing all this auth0 stuff? Should it be sent together with payload from the client?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,65 @@
...
+
+            final var user = handle.attach(UserDao.class).createUser(clientId, "DON'T KNOW");
```
Trimax: How exactly should I create the user without knowing all this auth0 stuff? Should it be sent together with payload from the client?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7931.xml
```diff
@@ -0,0 +1,42 @@
...
+            <column name="center_study_id" type="bigint">
+                <constraints primaryKey="true" primaryKeyName="center_study_pk"/>
+            </column>
+            <column name="center_id" type="bigint">
+                <constraints references="center_profile(center_id)" foreignKeyName="center_study_center_fk"/>
+            </column>
+            <column name="study_id" type="bigint">
+                <constraints references="umbrella_study(umbrella_study_id)" foreignKeyName="center_study_study_fk"/>
+            </column>
+        </createTable>
```
Trimax, line 26: This table is required in case the creation flow will be different. In case we will use some operator id instead of center id.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
```diff
@@ -77,9 +77,10 @@ public Object handle(Request request, Response response, CreateTemporaryUserPayl
             try {
                 User user = handle.attach(UserDao.class).createTempUser(clientDto.getId());
                 //create profile
-                UserProfileDao profileDao = handle.attach(UserProfileDao.class);
-                profileDao.createProfile(new UserProfile.Builder(user.getId())
-                        .setPreferredLangId(languageId)
+                handle.attach(UserProfileDao.class).createProfile(UserProfile.builder()
+                        .userId(user.getId())
+                        .preferredLangId(languageId)
+                        .preferredLangCode(null)
```
bskinner, line 83: We should be setting one of `preferredLangId` or `preferredLangCode`, but not both (even though it is `null`)- is there a way we can avoid this sort of duplicated call?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: We shouldn't be doing anything with the Auth0 client here. Setting up Auth0 clients is done during study initialization and, regardless, any users created via this API will not be bound to an Auth0 account.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: Users created for a Center do not have an associated Auth0 account.

pepper-apis/dss-database/src/main/resources/changelog-master.xml
```diff
@@ -2,7 +2,7 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
-                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
```
bskinner, line 5: Why is this scheme being changed to `https`, but none of the others are?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -368,6 +367,7 @@ private static void start() throws MalformedURLException {
         }
 
         post(API.REGISTRATION, new UserRegistrationRoute(interpreter, new TaskPubSubPublisher()), responseSerializer);
+        post(API.USER_SPECIFIC, new UserCreationRoute(), responseSerializer);
```
bskinner: If your intention is to add a new post endpoint at `/user`, I don't think this is going to do what you want it do to. The `USER_SPECIFIC` includes a `:userGuid` path parameter and, if this is creating a user, we do not have a GUID yet.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserCreationPayload.java
```diff
@@ -0,0 +1,33 @@
...
+    private String lastName;
+
+    @SerializedName("birthDate")
+    private LocalDate birthDate;
+
+    @SerializedName("sex")
+    private UserProfile.SexType sex;
+
+    @SerializedName("centerId")
+    private Long centerId;
```
bskinner: Why is `centerId` a `Long` here? Typically, all external identifiers (such as user guids, activity codes, instance identifiers, etc) are all opaque strings.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/CenterProfileDao.java
```diff
@@ -0,0 +1,17 @@
...
+import org.jdbi.v3.stringtemplate4.UseStringTemplateSqlLocator;
+
+import java.util.Optional;
+
+public interface CenterProfileDao extends SqlObject {
+    @UseStringTemplateSqlLocator
+    @SqlQuery("SELECT * FROM `center_profile` WHERE center_id = :id")
+    @RegisterConstructorMapper(CenterProfileDto.class)
+    Optional<CenterProfileDto> findById(@Bind("id") final Long id);
+}
```
bskinner, line 17: The Center-related code is out-of-scope for this ticket since none of the management or study-related configuration code is in place. Leave this for future expansion, but it's not currently used.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/JavaHeapDumper.java
```diff
@@ -1,60 +1,161 @@
...
-import com.google.cloud.storage.Blob;
-import com.google.cloud.storage.Storage;
-import com.sun.management.HotSpotDiagnosticMXBean;
-import lombok.extern.slf4j.Slf4j;
 import org.broadinstitute.ddp.exception.DDPException;
 
+import lombok.extern.slf4j.Slf4j;
+
 @Slf4j
 public class JavaHeapDumper {
```
bskinner, line 20: This class was modified in order to fix compilation issues when using non-HotSpot JVMs (this was preventing my builds). This method isn't called during normal operation, and only called manually when debugging server memory issues so this shouldn't have a significant impact on the overall operation of the backend.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    protected int getValidationErrorStatus() {
+        return HttpStatus.SC_UNPROCESSABLE_ENTITY;
+    }
+
+    @Override
+    public Object handle(Request request, Response response, UserCreationPayload payload) throws Exception {
+        var auth = RouteUtil.getDDPAuth(request);
+
+        if (false && auth.isAuthenticated() == false) {
```
bskinner: Known `if` bypass here- currently working on testing out the authentication issues. Expect this line to be updated once I get I reliable source of DSM tokens.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+            var error = new ApiError(ErrorCodes.AUTH_CANNOT_BE_DETERMINED, "Authentication required.");
+            throw ResponseUtil.haltError(HttpStatus.SC_FORBIDDEN, error);
+        } else {
+            /* LIE */
+            auth = new DDPAuth("https://domain.example", "abcTESTCLIENTdef", "OPERATOR", "totallyAValidJWT", null, "en");
+        }
```
bskinner: Another block of code that is going away once testing can be performed using DSM tokens.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+        final var operatorId = auth.getOperator();
+        final var domain = auth.getIssuer();
+        final var studyGuid = payload.getStudyGuid();
+
+        // Not sure if this is the proper check, but this is
+        // just to add some sort of safeguard until the filters are
+        // configured properly.
+        //
+        // Could this be replaced with an "isDSM" check?
+        if (false && auth.hasAdminAccessToStudy(studyGuid) == false) {
```
bskinner: Another block of code that is going away once testing can be performed using DSM tokens.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+            // needed below to set the `created_by_client_id` field in the user.
+            final var internalClientId = handle.attach(JdbiClient.class)
+                    .getClientIdByAuth0ClientAndDomain(requestorClientId, domain);
+            if (internalClientId.isEmpty()) {
+                /* Disabled for testing!
+                var error = new ApiError(ErrorCodes.NOT_FOUND,
+                        String.format("Auth0 client '%s' is not authorized for '%s'.",
+                            requestorClientId, domain));
+                throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, error);
+                */
```
bskinner: Needs to be re-enabled after testing can be performed with real tokens.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/UserProfileSql.java
```diff
@@ -19,7 +19,7 @@ public interface UserProfileSql extends SqlObject {
 
     @GetGeneratedKeys
     @SqlUpdate("insert into user_profile (user_id, first_name, last_name, sex, birth_date,"
-            + "        preferred_language_id, time_zone, do_not_contact, is_deceased, skip_language_popup)"
+            + "        preferred_language_id, time_zone, do_not_contact, is_deceased, skip_language_popup, email)"
             + " values (:userId, :firstName, :lastName, :sex, :birthDate,"
             + " :langId, :tz, :doNotContact, :isDeceased, :skipLanguagePopup)")
```
Trimax: An e-mail was added to the columns list, but the value seems to be missing

Trimax: In the list of arguments as well

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/UserDto.java
```diff
@@ -37,6 +41,14 @@ public class UserDto {
     @ColumnName("expires_at")
     Long expiresAtMillis;
 
+    public Optional<String> getAuth0UserId() {
+        return Optional.ofNullable(auth0UserId);
+    }
+
+    public Optional<String> getEmail() {
+        return Optional.ofNullable(email);
+    }
```
Trimax, line 50: Do you think it is useful? I see that this method is being used only in one place - UserCreationRoute class.

The construction
```
model.getEmail().orElse(null);
```
looks a bit less readable for me that 
```
Optional.ofNullable(model.getEmail()).orElse(null);
```
or simple ternary operator
```
model.getEmail() == null ? null : model.getEmail();
```

Anyway, it's just a matter of taste, so if you believe returning optional is better, then let's leave it as is.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserCreationResponse.java
```diff
@@ -0,0 +1,40 @@
...
+    @SerializedName("profile")
+    private Profile profile;
+
+    @SerializedName("center")
+    private String center;
+    
+    public UserCreationResponse(User user, UserProfile profile) {
+        this.guid = user.getGuid();
+        this.hruid = user.getHruid();
+        this.email = user.getEmail().orElseThrow(() -> new IllegalStateException("participant must have a valid email"));
```
Trimax, line 37: We're saying **valid** here, but we don't check the e-mail to be valid. Just not null. Isn't it a bit misleading?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/AdminLookupInvitationRoute.java
```diff
@@ -63,12 +63,16 @@ public Object handle(Request request, Response response, LookupInvitationPayload
                 }
 
                 String loginEmail = null;
-                if (user.getAuth0UserId() != null) {
+                
+                if (user.getAuth0UserId().isPresent()) {
```
Trimax: Regarding returning optionals.
Not sure, but if we'll define the method `hasAuth0UserId` the code will look like
```
if (user.hasAuth0UserId()) {
  ... = user.getAuth0UserId(); //without additional .get statement.
}
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
Trimax: I was thinking about it. And since we're not using custom builder anymore, I guess the answer is no. But on the other side, if we compare the huge builder class that can be replaced with the one annotation and some places with two calls, I would vote for the 2nd option.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
Trimax: :)
Looks like the work here isn't finished yet.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+            // going to be recognized but handle things just in case. Save the ID since it'll be
+            // needed below to set the `created_by_client_id` field in the user.
+            final var internalClientId = handle.attach(JdbiClient.class)
+                    .getClientIdByAuth0ClientAndDomain(requestorClientId, domain);
+            if (internalClientId.isEmpty()) {
+                /* Disabled for testing!
+                var error = new ApiError(ErrorCodes.NOT_FOUND,
+                        String.format("Auth0 client '%s' is not authorized for '%s'.",
+                            requestorClientId, domain));
+                throw ResponseUtil.haltError(HttpStatus.SC_UNAUTHORIZED, error);
```
Trimax, line 81: Should this be uncommented?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
```diff
@@ -368,12 +368,16 @@ private void deleteElasticSearchData(Handle handle, User user, UserCollectedData
     }
 
     private void deleteAuth0User(User user, Auth0ManagementClient auth0ManagementClient) {
-        if (user.getAuth0UserId() != null && auth0ManagementClient != null) {
-            log.info(LOG_MESSAGE_PREFIX__DELETE_FROM_AUTH, user.getGuid());
-            var result = auth0ManagementClient.deleteAuth0User(user.getAuth0UserId());
-            if (result.hasFailure()) {
-                throw new DDPException(result.hasThrown() ? result.getThrown() : result.getError());
-            }
+        if (user.getAuth0UserId().isPresent() == false) {
```
Trimax: Let's better use isEmpty instead?
`if (user.getAuth0UserId().isEmpty())`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/participants/ParticipantsCreateService.java
```diff
@@ -0,0 +1,182 @@
...
+        /* This is from UserRegistrationRoute.java,
+         * but in that class I can't clearly see where `taskPublisher`
+         * is being set. Thoughts?
+         */
+        /*
+        taskPublisher.publishTask(
+                    TaskPubSubPublisher.TASK_PARTICIPANT_REGISTERED,
+                    payload, studyGuid, participantGuid);
+        */
+    }
```
Trimax, line 181: I don't know as well. Shouldn't it be passed as the constructor argument?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/studies/StudiesService.java
```diff
@@ -0,0 +1,125 @@
...
+            // Let the caller know.
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    private boolean studyExists(String studyGuid) {
+        var studyId = handle.attach(JdbiUmbrellaStudy.class).getIdByGuid(studyGuid);
+        if (studyId.isPresent()) {
```
Trimax: Maybe just
`return studyId.isPresent()`

Or even
`return handle.attach(JdbiUmbrellaStudy.class).getIdByGuid(studyGuid).isPresent();`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/JavaHeapDumper.java
```diff
@@ -1,60 +1,161 @@
 package org.broadinstitute.ddp.util;
```
Trimax, line 1: Is this class expected to be in this PR?
Or do you like my way of concomitant unrelated changes? :)

pepper-apis/dss-database/src/main/resources/changelog-master.xml
Trimax: Looks like I just accepted IntelliJ's suggestion

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -334,9 +335,7 @@ private static void start() throws MalformedURLException {
 
         before("*", new HttpHeaderMDCFilter(X_FORWARDED_FOR));
         before("*", new MDCLogBreadCrumbFilter());
-        before("*", (Request request, Response response) -> {
-            MDC.put(MDC_STUDY, RouteUtil.parseStudyGuid(request.pathInfo()));
-        });
+        before("*", (Request request, Response response) -> MDC.put(MDC_STUDY, RouteUtil.parseStudyGuid(request.pathInfo())));
```
MocanaAtBroad, line 338: Not saying this is not a good idea. It is! But dropped in the middle of a 72 file PR that has a completely different purpose without a comment is not the best way to make this change clear and transparent.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserCreationResponse.java
bskinner: Nope. Valid in this case is "non-null". I'd be open to changing the text of the exception to make it clearer, but it is not this object's responsibility to validate a User's email. If a User was created with an invalid email, that's a bug elsewhere.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/CreateTemporaryUserRoute.java
bskinner: That would be fine if these two value were not related. Being able to set both allows the profile to be put into an invalid state.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: See [this comment](https://github.com/broadinstitute/ddp-study-server/pull/1653/files#r902677146)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/UserDeleteService.java
bskinner: Yup! That's far more straightforward- fixed.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
MocanaAtBroad: I would say we should trust inside of the route that authentication info is A-OK. This work happening in filters/separate concern.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/participants/ParticipantsCreateService.java
bskinner: Either that or delegated out- either way I don't immediately see where the object is being created (hence my comment), and I don't know what it's lifecycle looks like.

MocanaAtBroad (commented):
Still reviewing, but want to submit the comment requesting we remove authentication checks from here. Authorization? I am less sure of that/hope can also be handled in filters.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/JavaHeapDumper.java
bskinner: See [this comment](https://github.com/broadinstitute/ddp-study-server/pull/1653/files#r902673172)

pepper-apis/dss-database/src/main/resources/changelog-master.xml
bskinner: Did some digging into this and the other values are technically opaque identifiers, and not URLs- so this is correct (thanks, XML).

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/studies/StudiesService.java
bskinner: I'd go with `return studyId.isPresent();` (mostly because it's easier to step through with a debugger). This whole block is the result of a logical compartmentalization between the Optional check, and the method's returned value.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: Yup- see [this comment](https://github.com/broadinstitute/ddp-study-server/pull/1653/files#r902675285)

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7931.xml
MocanaAtBroad: Maybe this should be "study_center" instead of "center_study"?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
bskinner: Agreed- this was a change in the PR prior to my work.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/UserProfileSql.java
bskinner: Nice catch! Thank you!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: The plan is to remove a lot of this code once I figure out how we're processing the DSM tokens. From what @ssettipalli is telling me, we aren't creating any DDPAuth objects so this code may need to fall-back to manually decoding the token.

bskinner:
Everyone, please leave the "Update branch" button alone.

MocanaAtBroad (commented):
Still reviewing, but want to submit the comment requesting we remove authentication checks from here. Authorization? I am less sure of that/hope can also be handled in filters.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+            auth = new DDPAuth("https://domain.example", "abcTESTCLIENTdef", "OPERATOR", "totallyAValidJWT", null, "en");
+        }
+
+        /*
+        * Grab the requesting client id out of the token
+        * so it can be referred back to after we create the user
+        */ 
+        final var requestorClientId = auth.getClient();
+        final var operatorId = auth.getOperator();
+        final var domain = auth.getIssuer();
```
MocanaAtBroad, line 44: I see this property is being picked up from token. Do we know if we get this from DSM tokens @ssettipalli ?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+        // Could this be replaced with an "isDSM" check?
+        if (false && auth.hasAdminAccessToStudy(studyGuid) == false) {
+            var error = new ApiError(ErrorCodes.INSUFFICIENT_PRIVILEGES,
+                    String.format("User %s does not have sufficient priviliges for %s.",
+                    operatorId, studyGuid));
+            throw ResponseUtil.haltError(HttpStatus.SC_FORBIDDEN, error);
+        }
+
+        final var email = payload.getEmail();
+        final var emailValidator = EmailValidator.getInstance();
```
MocanaAtBroad, line 48: I believe that if you just use the @Email annotation on the payload POJO, email will be checked for syntax and automatically rejected by ValidatedJsonInputRoute, so we probably don't need to use EmailValidator explicitly.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -0,0 +1,168 @@
...
+        *
+        * Checking here is in order to save a few cycles & connections to the DB if it's
+        *  immediately clear the client gave us bad data.
+        */
+        if (emailValidator.isValid(email) == false) {
+            var error = new ApiError(ErrorCodes.MALFORMED_EMAIL, "the email address is not in a valid format.");
+            throw ResponseUtil.haltError(HttpStatus.SC_UNPROCESSABLE_ENTITY, error);
+        }
+
+        if (StringUtils.isNotBlank(payload.getCenterId())) {
```
MocanaAtBroad, line 63: Same sort of validation here. You probably can use @NotEmpty (Part of the purpose of ValidatedJsonInputRoute is to check validation annotations).

https://javadoc.io/doc/javax.validation/validation-api/latest/index.html

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/participants/ParticipantsCreateService.java
```diff
@@ -0,0 +1,182 @@
...
+
+        Long centerId = null;
+
+        if (StringUtils.isNotBlank(centerGuid)) {
+            /* The "centerGuid" is currently just the internal ID
+             * which is a `long` type so convert this to a long before
+             * checking the database. If/when Centers get an separate external
+             * identifier (such as a GUID), then this will need to be updated.
+             */
+            Long tempCenterGuid = Long.valueOf(centerGuid);
```
MocanaAtBroad, line 112: Let's not do this (will not matter for demo as we are not using centerId) but let's use codes/guids whatherver for centers/avoid

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/participants/ParticipantsCreateService.java
```diff
@@ -0,0 +1,182 @@
...
+    public User createWithAuth0(String userId, String tenantDomain) throws ParticipantCreateError {
+        throw new ParticipantCreateError(Code.NOT_IMPLEMENTED,
+                "Auth0 participant account creation not implemented");
+    }
+
+    private void notifyParticipantWasRegistered(String studyGuid, String operatorGuid, String participantGuid) {
+        var studyDto = handle.attach(JdbiUmbrellaStudy.class).findByStudyGuid(studyGuid);
+        var userDao = handle.attach(JdbiUser.class);
+        var operator = userDao.findByUserGuid(operatorGuid);
+        var participant = userDao.findByUserGuid(participantGuid);
```
MocanaAtBroad, line 160: We can use the  of JdbiUmbrellaStudyCached here.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/studies/StudiesService.java
```diff
@@ -0,0 +1,120 @@
...
+            // If any id has been returned, the participant is registered
+            // Let the caller know.
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    private boolean studyExists(String studyGuid) {
+        var studyId = handle.attach(JdbiUmbrellaStudy.class).getIdByGuid(studyGuid);
```
MocanaAtBroad, line 117: JdbiUmbrellaStudyCached. Look in code for how it is instantiated (handle.attach is not how you instantiate Cached Jdbi classes)

MocanaAtBroad (commented):
Not seeing test code for the UserCreationRoute.

MocanaAtBroad (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/AdminLookupInvitationRoute.java
bskinner: Didn't notice that method! Did a bit of refactoring.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: This isn't validation  of the input- `centerId` being `null` is perfectly ok, we just don't handle the the case where it is not null yet.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
bskinner: @MocanaAtBroad Code removed

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserCreationPayload.java
bskinner: Spoke with @SimoneMaiwald- this will be transitioned to an opaque string guid when Centers are implemented (like what we're doing in other objects).

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/UserDto.java
bskinner: The reasoning for the `Optional<>` here was several-fold:
  * Make it explicit that this was a value that may not be present, and force callers to deal with it (instead of using implicit null handling)
  * Identify places where the method was called, in conjunction with the IDE's "Find References" functionality

While I'd prefer to have this be an Optional, I do agree that this is awkward to use given the current style of our codebase (`model.getEmail().orElse(null);` is also an anti-pattern which should be avoided).

After this latest milestone is reached- let's have a discussion in the process meeting, and figure out what style guidelines we'd like to use in relation to Optionals.

ssettipalli (approved):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
Trimax: It wasn't dropped, it was refactored. I just removed braces

PR #1660, opened 2022-05-19
Add retries that skip cache lookup when a permission lookup fails in …

MocanaAtBroad:
## Context
The observed problem was that when the "Enroll Child" button was pressed:
1. The new user was created, including the governance relationship to the parent(and current system user). 
2. The client after receiving the new user guid, would try to obtain the new users workflow from server.
3. But the server would return a 401 (not authorized).

Note that this was an intermittent problem. I was able to reproduce it consistently by hacking the client to allow me to repeatedly click on the "Enroll child" button.

Problem turned out to be that the cached current users DDPAuth on the server's did not at times contain the latest governance. It appears that this might be caused by concurrent requests.

The fix in this code is that, if when checking permissions in the auth filter we find that the user does not have permissions to access that particular resource (in this case the resource is the child's workflow) then we read all the user permissions info from the database directly (bypassing the cache) and check again with that.

@Trimax had proposed a more sophisticated solution that would have involved modifying the cached user object as new children were being added, but the "refresh from database if not found" strategy I used was easier to implement and understand given the time constraints. We might want to recheck at some point.

## FUD Score

## Testing

I deployed my branch to develop environment and tested the changes with my hacked Angular application. I was not able to generate the 401 getting the user workflow.

## Release


.circleci/main-config.yml
```diff
@@ -125,6 +125,9 @@ commands:
             ?(*_)develop)
               DEPLOY_ENV=dev
             ;;
+            DDP-?(*))
+              DEPLOY_ENV=dev
+            ;;
```
MocanaAtBroad, line 130: All the circleci changes are temporary for this branch. Just wanted to test my branch on the server for a realistic test.

pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -9,6 +9,9 @@ manual_scaling:
 {{else if eq $environment "staging"}}
 manual_scaling:
   instances: 1
+{{else if eq $environment "dev"}}
+manual_scaling:
+  instances: 3
```
MocanaAtBroad, line 14: Similarly I hard coded 3 instances for dev in deployment configuration as this is what it looks like in production. We suspect that the cache  incoherence problem would be easier to arise when having multiple servers running and multiple create users requests at the same time.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/TokenConverterFilter.java
```diff
@@ -27,7 +27,7 @@ public class TokenConverterFilter implements Filter {
     @Override
     public void handle(Request request, Response response) {
         try {
-            DDPAuth ddpAuth = jwtConverter.convertJWTFromHeader(request.headers(AUTHORIZATION));
+            DDPAuth ddpAuth = jwtConverter.convertJWTFromHeader(request.headers(AUTHORIZATION), true);
```
MocanaAtBroad, line 30: true argument says it is is OK to use cached version (this is the default and behavior in this filter has not changed at all). Just a refactors that added a new argument.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -64,6 +68,35 @@ public void handle(Request request, Response response) {
...
+                    } catch (Exception e) {
+                        log.error("Error while converting token for request", e);
+                        halt(401);
+                    }
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            } while (!canAccess && --retries > 0);
+        }
+
```
MocanaAtBroad, line 83: This is retry loop is the meat of the fix. Filter first tries to check access using the previous existing code that can use a cached user info. If access is denied, then we are going to get the user info (DDPAuth) using the computedDDPAuth . This method avoids using the cached information.

Note here that an assumption is that the VAST majority of requests will be approved. And approved in the first pass, so this new code should not execute very often.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
```diff
@@ -64,6 +68,35 @@ public void handle(Request request, Response response) {
...
+                    TimeUnit.SECONDS.sleep(1);
+                    log.warn("Retrying checking for access with non-cached permission data");
+                    try {
+                        canAccess = canAccess(request, RouteUtil.computeDDPAuth(request));
+                    } catch (TokenExpiredException e) {
+                        log.error("Found expired token for request", e);
+                        halt(401);
+                    } catch (Exception e) {
+                        log.error("Error while converting token for request", e);
+                        halt(401);
```
MocanaAtBroad: Found that I needed to return 401 when exceptions occurred to allow existing tests to pass.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/security/JWTConverter.java
```diff
@@ -221,11 +223,11 @@ private DDPAuth convertJWT(String jwt) {
      * header, validates the JWT, and converts it into
      * {@link DDPAuth a ddp auth object}.
      */
-    public DDPAuth convertJWTFromHeader(String authHeader) {
+    public DDPAuth convertJWTFromHeader(String authHeader, boolean useCachedIfAvailable) {
```
MocanaAtBroad, line 226: That same new argument we encountered at the top of the PR. Means it is OK (or not if false) to use a cached DDPAuth.

.circleci/config.yml
```diff
@@ -14,8 +14,22 @@ parameters:
...
+  build-test-deploy-dev-workflow:
+    when:
+      and:
+        - << pipeline.parameters.on_demand >>
+        - equal: ["dss", << pipeline.parameters.project >>]
+    jobs:
+      - continuation/continue:
+          configuration_path: .circleci/main-config.yml
+          parameters: '{"action":"build-test-deploy","dss": true,"api-docs": false}'
+
```
MocanaAtBroad, line 32: Added this to CircleCi to allow me to deploy the PR branch to develop. It is difficult to test this problem locally.
(This is likely to stay as it can be very helpful).

.circleci/main-config.yml
MocanaAtBroad: Or we keep it! Welcome opinions here.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/RouteUtil.java
```diff
@@ -58,6 +60,10 @@ public static DDPAuth getDDPAuth(Request req) {
         return ddpAuth;
     }
 
+    public static DDPAuth computeDDPAuth(Request req) {
+        return new JWTConverter().convertJWTFromHeader(req.headers(AUTHORIZATION), false);
+    }
```
MocanaAtBroad, line 65: computedDDPAuth (as opposed to cached). Not the `false` argument used.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/security/JWTConverterTest.java
```diff
@@ -62,13 +62,13 @@ public void testConvertJWTFromHeaderWithCache() {
...
-        DDPAuth ddpAuth = jwtConverter.convertJWTFromHeader("Bearer " + token);
+        DDPAuth ddpAuth = jwtConverter.convertJWTFromHeader("Bearer " + token, true);
 
         assertTrue(ddpAuth.isActive());
         assertEquals(expectedClient, ddpAuth.getClient());
         assertEquals(expectedUserGuid, ddpAuth.getOperator());
 
-        DDPAuth cachedDdpAuth = jwtConverter.convertJWTFromHeader("Bearer " + token);
+        DDPAuth cachedDdpAuth = jwtConverter.convertJWTFromHeader("Bearer " + token, true);
         assertTrue(cachedDdpAuth.isActive());
```
MocanaAtBroad, line 72: Refactoring

.circleci/main-config.yml
Trimax: I don't think keeping it is a good idea since in case the PR branch has liquibase script changes they might be applied in the wrong order.

.circleci/main-config.yml
```diff
@@ -125,6 +125,9 @@ commands:
             ?(*_)develop)
               DEPLOY_ENV=dev
             ;;
+            DDP-?(*))
```
Trimax, line 128: Could you please explain what it means?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
Trimax: Why do you need 3 attempts?

Trimax (dismissed):

.circleci/main-config.yml
MocanaAtBroad: It is matching the branch name (e.g., DDP-1235_Whatever) This is saying that if the branch name has the DDP- prefix, we will target the dev environment.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
MocanaAtBroad: You know what. I think this is probably not necessary. I was worrying about request scenarios that are not realistic.
I am going to just keep it to 1 try and test it out.

.circleci/main-config.yml
MocanaAtBroad: Ah. Liquibase! That is a good point! Hmm. 
I think the fact that I went through the pain of adding all this stuff points to a need though.
Maybe what we need is a simple way to rollback Liquibase changes if applied?
Or maybe we can figure out way to prevent PRs with Liquibase changes to be deployed in this manner? Maybe.
I am still inclined to leave it in. I think the benefit outweighs the potential problems. But would definitely be good if there is way to avoid the Liquibase problem though.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/filter/UserAuthCheckFilter.java
MocanaAtBroad: Removed 902feb3fcdb4a31434dd77a66356c08113a6564f

.circleci/main-config.yml
bskinner: I second @Trimax on this- this violates the contract that what's on `develop` is what's on `dev`. Let's make things easier to deploy to "local" environment instead of taking shortcuts for testing.

ssettipalli (approved):
looks ok to me

bskinner (approved):

PR #1665, opened 2022-05-20
Osteo Prequal updates

archilsharashenidzebroad:


leonidrakitin (changes requested):
I think that it would be better use UpdateTemplatesInPlace.traverseActivity()
You can find an example in this task - BrainConsentVersion2

pepper-apis/studybuilder-cli/studies/osteo/patches/prequal-updates.conf
```diff
@@ -18,6 +19,13 @@
             <p>$intro3</p>
           """,
           "variables": [
+            {
+              "name": "title",
+              "translations": [
+                {"language": "en", "text": "Let's Get Started"},
```
leonidrakitin, line 25: It also would be better if you will add variable at i18n/en.conf and use it here.

nsakande (approved):
LGTM

leonidrakitin (approved):

PR #1666, opened 2022-05-20
[DSS] DDP-8043: Comparison Rule: question extraction fix

Trimax:


ssettipalli (approved):
Please make sure your squash merge comment is updated to include reference to jira ticket these changes should be associated with.

PR #1679, opened 2022-05-23
[DSS] DDP-7711: Fix property name

alextanlan:
Fixed inconsistent property names in the spec.

### DIFF
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -876,7 +876,7 @@ Block.Tabular:
           default: TABULAR
           enum:
             - TABULAR
-        numberOfColumns:
+        columnsCount:
           type: integer
           minimum: 1
           description: The total number of the tabular component columns.


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -876,7 +876,7 @@ Block.Tabular:
           default: TABULAR
           enum:
             - TABULAR
-        numberOfColumns:
+        columnsCount:
```
alextanlan, line 879: Changed the property name according to the line 870.

bskinner:
~This PR only seems to include changes to the API definition. This is a breaking change and requires modifications to both the frontend and DSS backend, as well.~

Looks like the backend is already using `columnsCount`- so moot point!

bskinner (approved):

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: Oh! I kind of like numberOfColumns better! I think I introduced the mistake. If you feel really strongly about keeping columnsCount, the fine. Otherwise then let's fix the `required` entries instead.

Trimax (approved):

PR #1690, opened 2022-05-25
[SB] Removed title and subtitle from component question.

leonidrakitin:
Removed title and subtitle from component question.


nsakande (approved):
LGTM

PR #1697, opened 2022-05-26
[DSS] Update tabular block definition

alextanlan:
Small changes of tabular block definition according to a new way of UI/Angular implementation.


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -887,7 +887,7 @@ Block.Tabular:
         content:
           type: array
           items:
-            $ref: '#/Block.Tabular.ContentRow'
+            $ref: '#/Question'
```
alextanlan: We are replacing 2-dimensional array with just a flat one here (it is possible to do Angular side easier this way).

A question: 
perhaps, we need to put `Block.Question` here ? And add a `columnSpan` property to it (instead of the current place in `Question`)

MocanaAtBroad: I think it should be Block.Question. Looking at similar things to tabular container in our api, like conditional block, they refer to blocks. example below:
https://github.com/broadinstitute/ddp-study-server/blob/674b2216522fd3d83bd1f074d0b74eaccb3ad38e/pepper-apis/docs/specification/src/components/schemas.yml#L773-L800

And if we can have the columnSpan on the Block.Question instead of the Question. Or, if we want to be ambitious, even on Block? Nah! Block.Question would be good for a long while.

alextanlan: Fixed in 21eaf9b59ee2589a40edde101d0d7abea3df6517

MocanaAtBroad: Great! Let's not forget to remove columnCount from Question now that Block.Question has that property:
https://github.com/broadinstitute/ddp-study-server/blob/674b2216522fd3d83bd1f074d0b74eaccb3ad38e/pepper-apis/docs/specification/src/components/schemas.yml#L1626-L1629

pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -846,6 +846,10 @@ Block.Question:
           minimum: 1
           nullable: true
           description: the number to label this block, `null` if it shouldn't be numbered
+        columnSpan:
+          type: integer
+          minimum: 1
+          description: The number of columns this question should occupy in a tabular display
```
Trimax, line 852: To me, it sounds like a violation of the clean code principle. Why should Block.Question know anything about tabular structure? This is an easy way of column-span implementation, but Block.Question is used not only in tabular blocks, right?

pepper-apis/docs/specification/src/components/schemas.yml
alextanlan: You might mean `columnSpan` ? 
If yes, it is removed in 21eaf9b59ee2589a40edde101d0d7abea3df6517

pepper-apis/docs/specification/src/components/schemas.yml
MocanaAtBroad: We have been using the blocks as structural display elements. In some cases they also serve as wrappers for other elements. Current Question block definition below.
I don't see any engineering principles being violated here, but if you have an idea that would be cleaner without overcomplicating things, do suggest. Requirement here is that we need to know how many columns in the question has to display its content.
Screen Shot 2022-05-27 at 8 29 52 AM

MocanaAtBroad (approved):

PR #1698, opened 2022-05-26
[DSM] Improvement: elastic.sh

gmakharat:
## Context

`DSM` started exporting all its internal database data to `ElasticSearch` to make search/sort work properly, thus updating `participants_structured.json` file became frequent operation, as we know every time we were updating something in `participants_structured.json` we had to reindex the `ES` by deleting index and rerunning data exporter script on both side `DSM` and `DSS`, to avoid this unpleasant routine every time new functionality is added, which includes changes in database structure, thus changes in `ES`, this "naive" approach will help to only update the mapping of all of the studies. 

**Things have done:**

- `participants_structured_reader.py` to read `participants_structured.json` and get mapping as a payload which needs to be updated.
- `participants_structured_extractor.py` to extract indices from the response returned by request on `_aliases`.
- `participants_structured_filter.py` to filter indices by `analyzer`, sort analyzer is used on every `text` type fields and when we update mapping this analyzer should be present in `settings` object of template, thus we filter indices by analyzer to avoid exception from `ES` that if we would like to use analyzer on field it should be present.
- Script contains one of the most "heavy" part when we send request in for loop for each index setting, it became necessary because we cannot get settings for all studies at once, getting settings is necessary to extract analyzer from it and filter indices by that analyzer.

P.S I am not good at writing code in shell and python, so please if there are things which may get better, point them out.

## Checklist

## FUD Score

## Testing

## Release


elasticsearch/participants_structured_extractor.py
```diff
@@ -0,0 +1,12 @@
+import sys
+import json
+
+def extract():
+    aliases = sys.argv[1]
+    participants_structured = json.loads(aliases)
+    filtered_indexes = list(filter(lambda x: x.startswith('participants_structured.'), participants_structured))
+    for item in filtered_indexes:
```
ScalaChamp, line 8: *minor* / *optional*
 
although not necessary here list unpacking operator can be used to make the code shorter:

separated by new line:
print(*filtered_indexes, sep="\n")

separated by space:
print(*filtered_indexes)

ScalaChamp (approved):

PR #1706, opened 2022-05-31
[DSM] DDP-7556_mercury_pubusb

pegahtah:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/participant/ParticipantDao.java
```diff
@@ -64,4 +68,25 @@ public Optional<ParticipantDto> get(long id) {
         return Optional.empty();
     }
 
+    public Optional<String> getParticipantFromCollaboratorParticipantId(String participantId) {
```
gmakharat, line 71: The parameter name should be `collaboratorParticipantId` instead of `participantid` I believe, to avoid confusion by what we are fetching the record.
Usually I like such naming on method:
`public Optional<String> getParticinaptIdBy(String collaboratorParticipantId)` method name and parameter name fills each other and we get very straightforward combination for understanding what actually method does.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
```diff
@@ -0,0 +1,52 @@
...
+    String mercuryPdoId;
+
+    public MercuryOrderDto(String ddpParticipantId, String creatorId, String barcode, int kitTypeId) {
+        this.ddpParticipantId = ddpParticipantId;
+        this.creatorId = creatorId;
+        this.barcode = barcode;
+        this.kitTypeId = kitTypeId;
+    }
+
+    public static List<MercuryOrderDto> createAllOrders(String[] barcodes, String ddpParticipantId, String orderId)
```
gmakharat, line 32: In general `DTO` classes are classes which only hold the data and nothing more, usually there are so called `use case` classes which are responsible to do business logic, this method here is out of scope of `DTO`, I would have  class `MercuryOrderUseCase` or something like that, which will be responsible for business logic.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
```diff
@@ -0,0 +1,52 @@
...
+            throws NoSuchElementException {
+        List<MercuryOrderDto> orders = new ArrayList<>();
+        MercuryOrderDao mercuryOrderDao = new MercuryOrderDao();
+        for (String barcode : barcodes) {
+            Optional<MercuryOrderDto> maybeOrder = mercuryOrderDao.getMercuryOrderFromSMIdOrKitLabel(barcode, ddpParticipantId);
+            maybeOrder.orElseThrow(() -> {
```
gmakharat, line 38: `orElseThrow` will return actual value if it is not null, you can keep it in local variable and do all manipulation to that local variable like you do in `ifPresent` below.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderPublisher.java
```diff
@@ -0,0 +1,141 @@
...
+        if (StringUtils.isNotBlank(json)) {
+            orderOnMercury(projectId, topicId, json, barcodes, ddpParticipantId, mercuryOrderId);
+        }
+
+    }
+
+    private void orderOnMercury(String projectId, String topicId, String json, String[] barcodes,
+                                String ddpParticipantId, String orderId) {
+        List<MercuryOrderDto> newOrders = MercuryOrderDto.createAllOrders(barcodes, ddpParticipantId, orderId);
+        SimpleResult results = inTransaction((conn) -> {
```
gmakharat, line 118: `Dao` related things are going here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -65,6 +68,9 @@ private static void updateTissue(String tissueId, String collaboratorSampleId) {
     public Object handle(Request request, Response response) {
```
gmakharat, line 68: This method is getting too messy already, after some time, if you continue putting everything here and not separating business logic into separate classes, it will get such messy that it will be so hard to understand and debug, will become perfect place for hidden bugs and then someone will have to refactor it, so please, avoid writing everything in `Route` classes, create small, only one responsibility holder classes and use them in routes.
`S` from `SOLID` principles:
`A class should have one and only one reason to change, meaning that a class should have only one job.`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -74,23 +80,42 @@ public Object handle(Request request, Response response) {
...
+            if (StringUtils.isBlank(participantId)) {
+                int tries = 0;
+                ddpParticipantId = new BSPDummyKitDao().getRandomParticipantForStudy(ddpInstance);
+                maybeParticipantByParticipantId =
+                        ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
+                if (maybeParticipantByParticipantId.isEmpty()) {
+                    throw new RuntimeException("PT not found " + ddpParticipantId);
+                }
+                while (!maybeParticipantByParticipantId.get().getStatus().get().equals("ENROLLED") && tries < 10) {
+                    tries++;
```
gmakharat, line 98: what is the point of `tries` here? before this whole operation starts new participant is added to `ES` and it may take some time and we need to kinda wait while it will be written to `ES`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -74,23 +80,42 @@ public Object handle(Request request, Response response) {
...
-                    ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
+            if (StringUtils.isBlank(participantId)) {
+                int tries = 0;
+                ddpParticipantId = new BSPDummyKitDao().getRandomParticipantForStudy(ddpInstance);
+                maybeParticipantByParticipantId =
+                        ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
+                if (maybeParticipantByParticipantId.isEmpty()) {
+                    throw new RuntimeException("PT not found " + ddpParticipantId);
+                }
+                while (!maybeParticipantByParticipantId.get().getStatus().get().equals("ENROLLED") && tries < 10) {
```
gmakharat, line 97: `!maybeParticipantByParticipantId.get().getStatus().get().equals("ENROLLED")` could be ->
`public boolean isNotEnrolled(ElasticSearchParticipantDto esParticipantDto)` for quick understanding after sometime when you will need to get back to this code.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -74,23 +80,42 @@ public Object handle(Request request, Response response) {
...
+                            ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
+                }
+                if (tries == 10) {
+                    throw new RuntimeException("No enrolled participants found");
+                }
+            } else {
+                Optional<String> maybeParticipantId =
+                        participantDao.getParticipantFromCollaboratorParticipantId(participantId);
+                maybeParticipantId.orElseThrow();
+                ddpParticipantId = maybeParticipantId.get();
```
gmakharat, line 110: `orElseThrow` above already returns actual value, thus having `.get()` here does not make sense.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderRoute.java
```diff
@@ -0,0 +1,57 @@
...
+    private static String projectId;
+    private static String topicId;
+    private static MercuryOrderPublisher mercuryOrderPublisher = new MercuryOrderPublisher(new MercuryOrderDao());
+
+    public PostMercuryOrderRoute(String projectId, String topicId) {
+        this.projectId = projectId;
+        this.topicId = topicId;
+    }
+
+    public static void publishMessage(MercuryOrderRequest mercuryOrderRequest, DDPInstance ddpInstance) {
```
gmakharat, line 28: having method as `static` is not necessary here at all, removing static on the method will help to remove static on fields `projectId`, `topicId`, `mercuryOrderPublisher`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderRoute.java
```diff
@@ -0,0 +1,57 @@
...
+        log.info("Publishing message to Mercury");
+        mercuryOrderPublisher
+                .createAndPublishMessage(mercuryOrderRequest.getKitLabels(), projectId, topicId, ddpInstance,
+                        mercuryOrderRequest.getDdpParticipantId());
+    }
+
+
+    public Object handle(Request request, Response response) throws Exception {
+        if (RoutePath.RequestMethod.POST.toString().equals(request.requestMethod())) {
```
gmakharat, line 37: object of `PostMercuryOrderRoute` is already passed to `post` method in `DSMServer`, what is the point of having this boolean expression?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderRoute.java
```diff
@@ -0,0 +1,57 @@
...
+            String requestBody = request.body();
+            MercuryOrderRequest mercuryOrderRequest = new Gson().fromJson(requestBody, MercuryOrderRequest.class);
+            if (StringUtils.isNotBlank(mercuryOrderRequest.getDdpParticipantId()) && mercuryOrderRequest.getKitLabels() != null
+                    && mercuryOrderRequest.getKitLabels().length > 0 && StringUtils.isNotBlank(mercuryOrderRequest.getRealm())) {
+                DDPInstance ddpInstance = DDPInstance.getDDPInstance(mercuryOrderRequest.getRealm());
```
gmakharat, line 42: main point of creation of `DDPInstanceDao` was to eliminate statically built classes like `DDPInstance`, so please, whenever it is possible to avoid this class avoid it and use `DDPInstanceDao`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderRoute.java
```diff
@@ -0,0 +1,57 @@
...
+                        mercuryOrderRequest.getDdpParticipantId());
+    }
+
+
+    public Object handle(Request request, Response response) throws Exception {
+        if (RoutePath.RequestMethod.POST.toString().equals(request.requestMethod())) {
+            String requestBody = request.body();
+            MercuryOrderRequest mercuryOrderRequest = new Gson().fromJson(requestBody, MercuryOrderRequest.class);
+            if (StringUtils.isNotBlank(mercuryOrderRequest.getDdpParticipantId()) && mercuryOrderRequest.getKitLabels() != null
+                    && mercuryOrderRequest.getKitLabels().length > 0 && StringUtils.isNotBlank(mercuryOrderRequest.getRealm())) {
```
gmakharat, line 41: This expression is way tooo long that deserves to have its own method with name like `isOrderReadyToPublish` or something like that.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderRoute.java
```diff
@@ -0,0 +1,57 @@
...
+                    return new Result(200);
+                } else {
+                    log.error("Realm was empty");
+                    return new Result(500, UserErrorMessages.CONTACT_DEVELOPER);
+                }
+            }
+            log.error("Request method not known");
+            return new Result(500, UserErrorMessages.CONTACT_DEVELOPER);
+        }
+        log.error("realm was empty");
```
gmakharat, line 54: https://refactoring.guru/replace-nested-conditional-with-guard-clauses one of the strongest tool called `guard conditions` to avoid deep nested if statements.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/participant/ParticipantDao.java
```diff
@@ -19,6 +19,10 @@ public class ParticipantDao implements Dao<ParticipantDto> {
                     + "assignee_id_mr, assignee_id_tissue, last_changed, changed_by) VALUES (?,?,?,?,?,?,?,?,?) "
                     + "ON DUPLICATE KEY UPDATE last_changed = ?, changed_by = ?";
 
+    private static final String SQL_SELECT_PARTICIPANT_FROM_COLLABORATOR_ID = "SELECT p.ddp_participant_id from ddp_participant p "
+            +
+            "left join ddp_kit_request req on (req.ddp_participant_id = p.ddp_participant_id) where req.bsp_collaborator_participant_id = ? ";
```
SimoneMaiwald, line 24: please add ddp_instance_id check as well

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -0,0 +1,133 @@
...
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.lddp.db.SimpleResult;
+
+@Slf4j
+public class MercuryOrderDao implements Dao<MercuryOrderDto> {
+    public static String SQL_INSERT_MERCURY_ORDER = "insert into mercury_sequencing (order_id, order_date, ddp_participant_id, "
+            + "kit_type_id, barcode) values (?, ?, ?, ?, ?)";
+
+    public static String SQL_GET_KIT_FROM_BARCODE =
+            "SELECT p.ddp_participant_id, accession_number, ddp.instance_name, t.collaborator_sample_id, settings.mercury_order_creator, "
```
SimoneMaiwald, line 23: mercury_order_creator should go int ddp_instance like bsp_group. I would like to have all GP related info in one table, easier for setting things up

SimoneMaiwald:
mercury_order_creator should be in the same table as research_project

SimoneMaiwald:
Going to merge PR in, so that GP is not blocked! Changes need to get solved from dev ASAP @pegahtah

SimoneMaiwald (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
pegahtah: we are getting random participants from db for creating dummy test cases. I am trying 10 times for finding an enrolled one, there should be a limit on the number of tries in case there isn't any

PR #1722, opened 2022-06-02
DSS Ddp 8126 fix consul templates for m1

MocanaAtBroad:
## Context
Got new Macbook M1 computer and api-build.sh stopped working for me with a strange go error.
After much research, turns out that when we run api-build.sh we execute the program consul-template that resides in a Docker image.
But the executable on that Docker image will not run in an M1 Processor.
I have rebuilt the Docker images to use a newer version of consul-template that will run in M1 and already uploaded to Docker hub:
https://hub.docker.com/r/broadinstitute/study-server-build/tags
And because this new build is a newer version of consul-template I had to modify every template that uses the "vault" command.


.circleci/main-config.yml
```diff
@@ -9,7 +9,7 @@ references:
   tcell_config_file: &tcell_config_file
                        tcell_agent.config
   ci_support_image: &ci_support_image
-                      broadinstitute/study-server-build:2020-04-30A
+                      broadinstitute/study-server-build:2022-06-01A
   root_path: &root_path
```
MocanaAtBroad, line 13: Using new image in CircleCi

.circleci/main-config.yml
```diff
@@ -48,7 +48,7 @@ references:
 executors:
   test-executor:
     docker:
-      - image: broadinstitute/study-server-build:java-2020-04-20A
+      - image: broadinstitute/study-server-build:java-2022-06-01A
       # start pubsub emulator
```
MocanaAtBroad, line 52: Same

pepper-apis/Dockerfile-build
```diff
@@ -1,50 +0,0 @@
-# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
```
MocanaAtBroad, line 1: Not deleted. Moved! It is in the PR.

pepper-apis/Dockerfile-java-build
```diff
@@ -1,42 +0,0 @@
-# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
```
MocanaAtBroad, line 1: Not deleted. Moved. In this PR.

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+ENV CONSUL_TEMPLATE_VERSION=0.27.2
+ENV VAULT_VERSION=1.4.3
+
+# Set by build
+ARG TARGETARCH
+
+RUN case "${TARGETARCH:-amd64}" in \
+           arm64) export HASHICORP_ARCH='arm64' ;; \
+           amd64) export HASHICORP_ARCH='amd64' ;; \
+    esac; \
```
MocanaAtBroad, line 28: This is the conditional that tells us which architecture this build is for. TARGETARCH variable is generated by Docker for us.

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+    && wget https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/vault_${VAULT_VERSION}_linux.zip \
+    && wget --no-check-certificate https://releases.hashicorp.com/consul-template/${CONSUL_TEMPLATE_VERSION}/consul-template_${CONSUL_TEMPLATE_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip \
+    && cd /tmp  \
+    && unzip vault_${VAULT_VERSION}_linux.zip  \
+    && unzip consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip  \
+    && mv consul-template vault /usr/local/bin/  \
+    && chmod 0755 /usr/local/bin/consul-template \
+    && chmod 0755 /usr/local/bin/vault \
+    && rm -rf /tmp/*.zip
+
```
MocanaAtBroad, line 54: This section above gets the zip files with the builds from Hashicorp's website, unzip's them and put them in the right spot. As far as I know, builds are not in any of the main repos (this approach was mostly copied from DSDE images that  we have been using).

pepper-apis/build-utils/build-support-images/Dockerfile-java-build
```diff
@@ -0,0 +1,59 @@
...
+    && wget https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/vault_${VAULT_VERSION}_linux.zip \
+    && wget --no-check-certificate https://releases.hashicorp.com/consul-template/${CONSUL_TEMPLATE_VERSION}/consul-template_${CONSUL_TEMPLATE_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip \
+    && cd /tmp  \
+    && unzip vault_${VAULT_VERSION}_linux.zip  \
+    && unzip consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip  \
+    && mv consul-template vault /usr/local/bin/  \
+    && chmod 0755 /usr/local/bin/consul-template \
+    && chmod 0755 /usr/local/bin/vault \
+    && rm -rf /tmp/*.zip
+
```
MocanaAtBroad, line 53: Changes to this Dockerfile are similar to the other one. If there are suggestions on how to refactor to make them DRY, I would be grateful.

pepper-apis/build-utils/build-support-images/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+``docker buildx build --platform linux/amd64,linux/arm64 -t broadinstitute/study-server-build:java-2022-06-01A  --build-arg GIT_SHA=`git rev-parse --verify HEAD` -f ..Dockerfile-java-build --push .``
+
+## Additional Notes
+We are using multi-platform builds (amd64 and arm64) to create separate versions of images that can run on Intel 
+processor architectures and on Mac M1 architecture.
+
+Note that in the tag part of the image name we have a specific date. Be aware that the full name of the image, including the tag is referenced
+in the CircleCi YAML configuration files and in the `configure.rb` used to generate configuration files, so make sure you sync everything up if
+you make changes to name.
```
MocanaAtBroad, line 34: Should have created these notes before!

pepper-apis/config/application.conf.ctmpl
```diff
@@ -3,13 +3,13 @@
...
+{{with $conf := secret (printf "secret/pepper/%s/%s/conf" $environment $version)}}
+{{with $dsm_conf := secret (printf "secret/pepper/%s/%s/dsm/config" $environment $version)}}
 {{with $auth0 := index $conf.Data "auth0"}}
-{{with $testingConf := vault (printf "secret/pepper/%s/%s/testing" $environment $version)}}
+{{with $testingConf := secret (printf "secret/pepper/%s/%s/testing" $environment $version)}}
 {{with $testingAuth := index $testingConf.Data "auth0"}}
 {{with $slack := index $conf.Data "slack"}}
-{{with $elasticsearchConf := vault (printf "secret/pepper/%s/%s/elasticsearch" $environment $version)}}
+{{with $elasticsearchConf := secret (printf "secret/pepper/%s/%s/elasticsearch" $environment $version)}}
 {
```
MocanaAtBroad, line 13: All the template changes (.ctmpl files) are of the same kind: Newer consul-template versions (we were using a really old one) use the function `secret` instead of `vault`. 
Was forced to do this as only newer `consul-template` versions include an `arm64` build that will work on Apple M1s, so this is needed to keep `api-build.sh` working.

pepper-apis/config/ddp.conf.ctmpl
```diff
@@ -1,6 +1,6 @@
 {{with $environment := env "ENVIRONMENT"}}
 {{with $version := env "VERSION"}}
-{{with $response := vault (printf "secret/pepper/%s/%s/conf" $environment $version)}}
+{{with $response := secret (printf "secret/pepper/%s/%s/conf" $environment $version)}}
```
MocanaAtBroad, line 3: From `vault` to `secret` again. Get used to it!

pepper-apis/config/sqlproxy.env.ctmpl
```diff
@@ -1,6 +1,11 @@
...
+{{with $version := env "VERSION"}}
+{{with $response := secret (printf "secret/pepper/%s/%s/conf" $environment $version)}}
+{{with $conf := $response.Data}}
 
 CLOUDSQL_CONNECTION_LIST={{$conf.googleProjectId}}:us-central1:{{$conf.dbInstanceId}}=tcp:0.0.0.0:3307
+
+{{end}}
+{{end}}
+{{end}}
+{{end}}
```
MocanaAtBroad, line 11: The one template that I had to do more work on then changing `vault` to `secret`.
Had to use `with` and `end` for new ` consul-template` to accept it. Every other template was already good.

pepper-apis/configure.rb
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/ruby -w
 # Docker Image used to call out to generate config files
-$dsde_toolbox_image_name = "broadinstitute/dsde-toolbox:latest"
-
+$dsde_toolbox_image_name = "broadinstitute/study-server-build:java-2022-06-01A"
```
MocanaAtBroad, line 3: This `configure.rb` script actually calls `consult-template` for us using Docker. Updating to use one of our new images (both of our build images have `consul-template` but chose this one because it is smaller and presumably will be quicker to download.

pepper-apis/build-utils/build-support-images/Dockerfile-java-build
```diff
@@ -0,0 +1,59 @@
...
+    openssh-client \
+    git \
+    openjdk11 \
+    # to support configure.rb
+    ruby \
+    ruby-json  \
+    && rm -rf /var/cache/apk/* \
+    # Get the architecture specific builds of vault and consul template
+    && wget https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/vault_${VAULT_VERSION}_linux.zip \
+    && wget --no-check-certificate https://releases.hashicorp.com/consul-template/${CONSUL_TEMPLATE_VERSION}/consul-template_${CONSUL_TEMPLATE_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip \
```
bskinner: See the `--no-check-certificate` comment above in `Dockerfile-build`

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+    # Get the architecture specific builds of vault and consul template
+    && wget https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/vault_${VAULT_VERSION}_linux.zip \
+    && wget --no-check-certificate https://releases.hashicorp.com/consul-template/${CONSUL_TEMPLATE_VERSION}/consul-template_${CONSUL_TEMPLATE_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip \
+    && cd /tmp  \
+    && unzip vault_${VAULT_VERSION}_linux.zip  \
+    && unzip consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip  \
+    && mv consul-template vault /usr/local/bin/  \
+    && chmod 0755 /usr/local/bin/consul-template \
+    && chmod 0755 /usr/local/bin/vault \
+    && rm -rf /tmp/*.zip
```
bskinner: Would using the existing `vault` and `consult-template` packages present in the apk `community` repository be an option here? The `google/cloud-sdk` base image is build off of Alpine v3.15, and the community repo has version `0.27.2-r2` of `consul-template` and `1.9.6-r1` of `vault` available.

pepper-apis/build-utils/build-support-images/Dockerfile-java-build
```diff
@@ -0,0 +1,59 @@
+# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
+FROM broadinstitute/dsde-toolbox:DDO-1934 AS dsdetoolbox
+
+FROM maven:3.6.3-jdk-11 AS maven
+
+FROM alpine:3.11.5
```
bskinner: Alpine Linux 3.11 is no longer supported, or receiving updates, as of 2021-11-01.

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+    # to support configure.rb
+    ruby \
+    ruby-json  \
+    # to support sheetappend script & OpenAPI build/deploy
+    nodejs \
+    npm \
+    && rm -rf /var/cache/apk/* \
+    # Get the architecture specific builds of vault and consul template
+    && wget https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/vault_${VAULT_VERSION}_linux.zip \
+    && wget --no-check-certificate https://releases.hashicorp.com/consul-template/${CONSUL_TEMPLATE_VERSION}/consul-template_${CONSUL_TEMPLATE_VERSION}_linux_${HASHICORP_ARCH}.zip -O /tmp/consul_template_${CONSUL_TEMPLATE_VERSION}_linux.zip \
```
bskinner: `--no-check-certificate` is an insecure option and we should be avoiding unless there are no other options. Have you tested adding the `ca-certificates` package and seeing if that resolves the certificate validation errors you're running into?

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+FROM broadinstitute/dsde-toolbox:DDO-1934 AS dsdetoolbox
+
+FROM maven:3.6.3-jdk-11 AS maven
+
+FROM google/cloud-sdk:alpine
+
+#Essentials for Maven
+COPY --from=maven /usr/share/maven /usr/share/maven
+RUN ln -s /usr/share/maven/bin/mvn /usr/bin/mvn
+COPY --from=maven /usr/local/bin/mvn-entrypoint.sh /usr/local/bin/mvn-entrypoint.sh
```
bskinner: Is this necessary? It looks like with was pulled over from the previous iteration of the dockerfile, but It doesn't look like anything is calling the entrypoint script, and the dockerfile is not declaring an `ENTRYPOINT`

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
+# To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
+FROM broadinstitute/dsde-toolbox:DDO-1934 AS dsdetoolbox
+
+FROM maven:3.6.3-jdk-11 AS maven
+
+FROM google/cloud-sdk:alpine
```
bskinner, line 6: This is the final `FROM` so this will define the base image, but `google/cloud-sdk:alpine` only offers `amd64` builds- have you had success running these images using the `--platform=linux/arm64` flags?

(_As a potential alternative, Ubuntu offers builds for both architetures, and Google's APT repository includes both `arm64` and `amd64` builds of the Cloud SDK toolkit, and it would avoid the problems java has with the musl libc library._)

pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -0,0 +1,67 @@
...
+# Copying just configuration used by official DSDE
+COPY --from=dsdetoolbox /etc/consul-template/config/config.json /etc/consul-template/config/config.json
+ENV VAULT_ADDR https://clotho.broadinstitute.org:8200
+ENV CONSUL_TEMPLATE_VERSION=0.27.2
+ENV VAULT_VERSION=1.4.3
+
+# Set by build
+ARG TARGETARCH
+
+RUN case "${TARGETARCH:-amd64}" in \
```
bskinner, line 25: Can this RUN directive be broken up into smaller statements in order to make the build process a bit more granular? This is a pretty expensive phase if something changes.

pepper-apis/build-utils/build-support-images/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Procedure
+
+``docker buildx build --platform linux/amd64,linux/arm64 -t broadinstitute/study-server-build:2022-06-01A  --build-arg GIT_SHA=`git rev-parse --verify HEAD` -f ./Dockerfile-build --push .
+``
+
+``docker buildx build --platform linux/amd64,linux/arm64 -t broadinstitute/study-server-build:java-2022-06-01A  --build-arg GIT_SHA=`git rev-parse --verify HEAD` -f ..Dockerfile-java-build --push .``
+
+## Additional Notes
+We are using multi-platform builds (amd64 and arm64) to create separate versions of images that can run on Intel 
+processor architectures and on Mac M1 architecture.
```
bskinner: Referencing M1 directly may be misleading since there are multiple variants, and M2 is on the way. `Apple Silicon` is a more overarching term to refer to Apple's ARM lineup.

```suggestion
processor architectures and on Apple Silicon.
```

bskinner (commented):
Nice to see the tweaks here!

pepper-apis/config/sqlproxy.env.ctmpl
bskinner: A fun trick with these now that we're using a newer version- you can use `{{-<SPACE>`, and `<SPACE>-}}` (where `<SPACE>` is the literal space character (ASCII 0x20)) to suppress leading and trailing newline generation.

pepper-apis/build-utils/build-support-images/Dockerfile-build
MocanaAtBroad: This is a very good catch, @bskinner . Will say, overall, that in retrospect I might have gone overboard with this PR. 

Not being able to generate configurations locally/in my M1 macbook because of a Docker incompatibility drew me to look at these Docker files, but the reality is that these are really meant to just be driven by CircleCi for our builds and deploys.
Originally, when I created these I had illusions that they could also be used locally to do testing, but I never do it and I don't think anybody else does either.

Regardless, I think since I am at it, might as well refresh these as we have not done so in a couple of years. And I still like the idea that I can test stuff on my Mac if I feel like doing so.

On using the google cloud alpine image as base: while working on the CircleCi stuff originally, for a lot of operations, the slowest thing was actually CircleCi loading the Docker image which irritated the hell out of me, so I became obsessed with making the images as small as possible to reduce the annoying image loading overhead time.

Having said all this I tested all the gcloud operations that we use in CircleCi from within the image on my M1 and can confirm that they all work fine.

(this is one of the things that I found confusing with Docker on the M1 situation: most stuff worked. consult-template was the one thing that was not working)

I agree that an arm64 would be the safest bet for Apple Silicon. But upon the reflection of me going overboard that I shared above, M1 is at best a secondary consideration. But it works! So I am inclined to stick with it as it appears to be the best choice for CircleCi because of its size.

pepper-apis/build-utils/build-support-images/Dockerfile-build
MocanaAtBroad: Yeah. I added the entrypoint originally for the notion I referred above about using these for local builds and had ENTRYPOINT in a early version of this.
Dropped it. Not needed.

pepper-apis/build-utils/build-support-images/Dockerfile-build
MocanaAtBroad: Advice from Docker is to minimize number of layers to keep the image as small as possible. And since RUN command creates additional layers:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers
Having said that, people that know more about Docker than me suggest that it might make sense in some cases to just go with multiple layers. But here I am embracing  "more layers bad for performance". Not worried about build time. We will do that very sparingly, like now.

pepper-apis/build-utils/build-support-images/Dockerfile-build
MocanaAtBroad: Good point about the flag. I copied this from the original DevOps image.
(And as it turns out, the download works fine without the flag. Removing it).

pepper-apis/build-utils/build-support-images/Dockerfile-build
MocanaAtBroad: Got excited to see these in the repo and tried it. Alas, vault installation seems to be broken/vault does not execute locally on my M1 or in presumably amd/64 on CircleCI. And consul-template just hangs.
Going to guess that this is probably why DevOps was wget'ting builds rather than using apk: they ran into the same problem I just did.
Going with what work here.

pepper-apis/build-utils/build-support-images/Dockerfile-java-build
MocanaAtBroad: Updating to 3.15

pepper-apis/config/sqlproxy.env.ctmpl
MocanaAtBroad: Ha! Need trick.
Even better: a lot of our templates start with {{with foo=bunchofstuff}}
You can add that - character at the end to not use up whitespace.
(I find the leading blank lines even more annoying than the trailing ones!).
I would like to fix a bunch of templates, but I am going to restrain myself and do that in a separate PR, if you don't mind.

bskinner: That's reasonable- there's a lot here already!

pepper-apis/build-utils/build-support-images/Dockerfile-java-build
bskinner: Given how we're constructing things, the first thing that comes to mind is putting these build commands into their own script, and executing it in place of these commands during the build process.

pepper-apis/build-utils/build-support-images/Dockerfile-build
bskinner: If things are working at the moment, let's go with it. At the very least, this is a good set of first steps- I'm sure there will be plenty of opportunities to adjust this later as more folks switch over to Apple Silicon.

bskinner (approved):
👍

PR #1725, opened 2022-06-02
Composite child visibility API proposal [DDP-8154]

bskinner:
DDP-8154

## Context

This is a proposal for the changes necessary to support hide/show and enable/disable behavior for the Composite child questions.

### Authors
@bskinner 
@ssettipalli

### DIFF
--- a/pepper-apis/docs/specification/src/components/responses.yml
+++ b/pepper-apis/docs/specification/src/components/responses.yml
@@ -61,7 +61,7 @@ ActivityAnswersPatchResponse:
             description: |
               list of visibility updates for composite questions
             items:
-              $ref: '../pepper.yml#/components/schema/Composite.Visibility'
+              $ref: '../pepper.yml#/components/schemas/Composite.Visibility'
           validationFailures:
             type: array
             items:
--- a/pepper-apis/docs/specification/src/components/responses.yml
+++ b/pepper-apis/docs/specification/src/components/responses.yml
@@ -55,6 +55,13 @@ ActivityAnswersPatchResponse:
             description: list of updated block visibilities
             items:
               $ref: '../pepper.yml#/components/schemas/Block.Visibility'
+          compositeVisibility:
+            type: array
+            nullable: true
+            description: |
+              list of visibility updates for composite questions
+            items:
+              $ref: '../pepper.yml#/components/schema/Composite.Visibility'
           validationFailures:
             type: array
             items:
--- a/pepper-apis/docs/specification/src/components/schemas.yml
+++ b/pepper-apis/docs/specification/src/components/schemas.yml
@@ -1754,32 +1754,33 @@ Question.Composite:
       additionalItemText:
         type: string
         description: 'The label to show above entries beyond the initial one'
+      childVisibility:
+        type: array
+        description: |
+          The set of visibility flags for the child questions. There will always be at least one set of flags in this array, even if no answers are present, in order to represent the default appearance state of the questions.
+        items:
+          type: array
+          items:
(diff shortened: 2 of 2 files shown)


pepper-apis/docs/specification/src/components/schemas.yml
```diff
@@ -1724,12 +1754,31 @@ Question.Composite:
       additionalItemText:
         type: string
         description: 'The label to show above entries beyond the initial one'
+      childVisibility:
```
alextanlan, line 1757: That `childVisibility` array should have the same length and order of items like the `children` list, right?

alextanlan:
lgtm

pepper-apis/docs/specification/src/components/schemas.yml
bskinner: The format for the array is based off of how we're handling the `CompositeAnswer` values. The top-level array will either have 1 entry (the case where the questions need to be presented, but the user has not interacted with the composite question yet), or a number of elements equal to the number of answers. The lower-level array will have the same length and order as the child questions.

I'm open to suggestions if you have an alternate representation that may offer other tradeoffs.

## Examples
_any values for shown/enabled below are placeholders_

### 2 questions, no answers
```json5
[
  [
    { // Question 0
      "shown": true,
      "enabled": true
    },
    { // Question 1
      "shown": true,
      "enabled": true
    }
  ]
]
```

### 2 questions, 3 answers
```json5
[
  [
    { // Question 0, answer 0
      "shown": true,
      "enabled": true
    },
    { // Question 1, answer 0
      "shown": true,
      "enabled": true
    }
  ],
  [
    { // Question 0, answer 1
      "shown": false,
      "enabled": true
    },
    { // Question 1, answer 1
      "shown": true,
      "enabled": false
    }
  ],
  [
    { // Question 0, answer 2
      "shown": true,
      "enabled": false
    },
    { // Question 1, answer 2
      "shown": false,
      "enabled": true
    }
  ]
]
```

alextanlan: That looks good for me. A benefit is the backward compatibility as well.

pepper-apis/docs/specification/src/components/responses.yml
```diff
@@ -55,6 +55,13 @@ ActivityAnswersPatchResponse:
             description: list of updated block visibilities
             items:
               $ref: '../pepper.yml#/components/schemas/Block.Visibility'
+          compositeVisibility:
+            type: array
+            nullable: true
+            description: |
+              list of visibility updates for composite questions
+            items:
+              $ref: '../pepper.yml#/components/schema/Composite.Visibility'
```
MocanaAtBroad: Did we consider introducing blocks for composite children? Good part would be that we would not require a separate structure to  keep track of composite children in JSON; it would just be another block with an id, so `compositeVisibility` could go away and we could just stick with `blockVisibility`. The bad part would be that we would have to have blocks for children. Not sure how hard or how much work this would be though. Did we look at blocks for composite children? @bskinner @ssettipalli

bskinner: We did- this is a major api change and far outside our normal use since it inverts the hierarchy (as it would be a question containing blocks). It still does not solve the identification issue since just the block id would not uniquely identify a row in a composite response, and blockVisibility would need to be modified to support the potential for multiple "rows", which would be unusual in the context of the rest of the API.

MocanaAtBroad (dismissed):

alextanlan (approved):

PR #1736, opened 2022-06-06
[DSM] DDP-8125: Kit Upload

SimoneMaiwald:
changing the kit upload duplicate check to check for bsp_collaborator_participant_id instead of ddp_participant_id 
RGP has all individuals of one family with the same proband guid in ddp_kit_request table but the bsp_collaborator_participant_id will have the subjects id


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -289,17 +288,15 @@ private void uploadKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitTyp
...
-        } else if (checkIfKitAlreadyExists(conn, participantLegacyAltPid, ddpInstance.getDdpInstanceId(), kitTypeId)) {
-            isKitExsist = true;
-            kit.setParticipantId(participantLegacyAltPid);
+    private boolean checkAndSetParticipantIdIfKitExists(DDPInstance ddpInstance, Connection conn, String bspCollaboratorParticipantId,
+                                                        int kitTypeId) {
+        if (checkIfKitAlreadyExists(conn, bspCollaboratorParticipantId, ddpInstance.getDdpInstanceId(), kitTypeId)) {
+            return true;
         }
-        return isKitExsist;
+        if (checkIfKitAlreadyExists(conn, bspCollaboratorParticipantId, ddpInstance.getDdpInstanceId(), kitTypeId)) {
```
pegahtah: what is the difference between this and L293?

SimoneMaiwald: 😂

pegahtah (approved):

PR #1737, opened 2022-06-06
[DSS]Ensure needed sheetappend dependencies are installed for sheetappend

MocanaAtBroad:
Updated ticket with problem we just encountered.
https://broadinstitute.atlassian.net/browse/DDP-8126

For some reason, in previous build node dependencies were being included in installation the the sheetappend utility and they were not being installed in the new docker images.

### DIFF
--- a/pepper-apis/build-utils/build-support-images/Dockerfile-build
+++ b/pepper-apis/build-utils/build-support-images/Dockerfile-build
@@ -56,6 +56,7 @@ RUN case "${TARGETARCH:-amd64}" in \
 WORKDIR /opt
 RUN apk --no-cache add npm subversion && \
     svn export https://github.com/broadinstitute/ddp-angular/trunk/build-utils \
+        && npm install ./build-utils/sheetappend \
         && npm install -g ./build-utils/sheetappend \
         && chmod 555 build-utils/reportdeploy.sh
 


pepper-apis/build-utils/build-support-images/Dockerfile-build
```diff
@@ -56,6 +56,7 @@ RUN case "${TARGETARCH:-amd64}" in \
 WORKDIR /opt
 RUN apk --no-cache add npm subversion && \
     svn export https://github.com/broadinstitute/ddp-angular/trunk/build-utils \
+        && npm install ./build-utils/sheetappend \
         && npm install -g ./build-utils/sheetappend \
```
bskinner, line 60: Is having both the global and local install flags required here?

MocanaAtBroad: Well, in the original build that we had been using it was not necessary (apparently).
The -g installs sheetappend it so it is available from everywhere.
But for some reason it is not installing the required dependencies.
The first one/without the -g is doing that now.
Tested locally with and without and this does the trick.
(I don't know if this is new behavior added to newer npm or if the dependencies were somehow being added originally as an artifact of the build process/as part of the context?).
This fixes problem though.

bskinner (approved):

PR #1739, opened 2022-06-07
[DSS] Added a check to the announcement search class for government participants.

leonidrakitin:
Added a check to the announcement search class for government participants.

This is necessary when we create new participants and the announcement event is triggered, it is saved for them, not for main user. Well, to display these announcement in the dashboard, we need to check "child" participant as well.


ssettipalli:
Leonid: Is there a ticket to track this  change ?

ssettipalli:
Leonid: Is there a ticket to track this  change ?

leonidrakitin:
> Leonid: Is there a ticket to track this change ?

https://broadinstitute.atlassian.net/browse/DDP-8091 , 2-4 points
Announcements should be displayed in dashboard. ("thanks to finish ur enrollment process"), however, pepper always creates new user for "new participants" and event is triggered for this new user. that's why it's important to check these "child" users.

ssettipalli (approved):

PR #1769, opened 2022-06-11
DDP-8144 tabular updates

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-8144
Response json is as expected by updated APISpec.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1843,13 +1844,42 @@ default Map<Long, QuestionBlockDef> collectBlockDefs(Collection<FormBlockDto> bl
...
             blockDef.setShownExpr(blockDto.getShownExpr());
             blockDef.setEnabledExpr(blockDto.getEnabledExpr());
-
             blockDefs.put(blockDto.getId(), blockDef);
         }
 
         return blockDefs;
     }
 
+    default Map<Long, QuestionBlockDef> collectTabularBlockDefs(Collection<BlockTabularQuestionDto> blockDtos, long timestamp) {
```
ssettipalli, line 1853: This code is pretty much same as above collectBlockDef. May be can be refactored !

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
```diff
@@ -52,6 +52,10 @@
                 <constraints primaryKey="true" primaryKeyName="block_tabular_question_pk"/>
             </column>
 
+            <column name="column_span" type="bigint">
+                <constraints nullable="false" />
+            </column>
+
```
MocanaAtBroad, line 58: Are we going to have problems applying this change to dev? i.e., I think original version of this file was originally merged but we are making changes to it.
(I actually prefer doing what you are doing here as it keeps all the changes for the feature together. But we need to be ready to fight liquibase on dev, right?)

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
```diff
@@ -52,6 +52,10 @@
                 <constraints primaryKey="true" primaryKeyName="block_tabular_question_pk"/>
             </column>
```
MocanaAtBroad, line 54: This comment is for the headers part of the table (github won't let me add comment there).
The headers have an order that we need to enforce. MySQL PROBABLY will give us the headers in the order inserted, but SQL does not make such guarantees.
We need a sort order key for the headers.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/QuestionDao.java
```diff
@@ -1843,13 +1844,42 @@ default Map<Long, QuestionBlockDef> collectBlockDefs(Collection<FormBlockDto> bl
...
+        if (blockDtos == null || blockDtos.isEmpty()) {
+            return new HashMap<>();
+        }
+
+        Set<Long> blockIds = new HashSet<>();
```
Trimax: I prefer streams in such case, but it is the matter of taste:
```
final var blockIds = StreamEx.of(blockDtos).map(BlockDto::getQuestionBlockId).toSet();
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
```diff
@@ -338,13 +338,14 @@ default void insertTabularBlock(long activityId, TabularBlockDef block, long rev
...
-                final var question = block.get(row, column);
-                if (question == null) {
+                final var questionBlockDef = block.get(row, column);
+                if (questionBlockDef == null || questionBlockDef.getQuestion() == null) {
                     continue;
                 }
-
-                getQuestionDao().insertQuestionByType(activityId, question, revisionId);
-                getJdbiBlockTabular().insertQuestion(tabularId, question.getQuestionId(), column, row);
+                Integer columnSpan = questionBlockDef.getColumnSpan();
```
Trimax: Why do you need a separate variable? It could be inlined.

```
getJdbiBlockTabular().insertQuestion(tabularId, questionBlockDef.getQuestion().getQuestionId(),
                        column, row, questionBlockDef.getColumnSpan());
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/QuestionBlockDef.java
```diff
@@ -16,17 +16,31 @@ public final class QuestionBlockDef extends FormBlockDef {
     @SerializedName("question")
     private QuestionDef question;
 
+    @SerializedName("columnSpan")
+    int columnSpan;
```
Trimax: private

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/QuestionBlockDef.java
```diff
@@ -16,17 +16,31 @@ public final class QuestionBlockDef extends FormBlockDef {
...
+    }
+
     public QuestionDef getQuestion() {
         return question;
     }
 
+    public int getColumnSpan() {
+        return columnSpan;
+    }
+
```
Trimax: Let's remove both getters and annotate the class with @Getter annotation

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/TabularBlockDef.java
```diff
@@ -44,6 +45,15 @@ public int getRowsCount() {
...
+        List<TabularRowDef> rowDefs = getRows().stream().collect(Collectors.toList());
+        List<QuestionDef> questionDefList = new ArrayList<>();
+        for (TabularRowDef rowDef : rowDefs) {
+            for (QuestionBlockDef colDef: rowDef.getQuestions()) {
+                if (colDef != null) {
+                    questionDefList.add(colDef.getQuestion());
+                }
+            }
+        }
+        return questionDefList.stream();
```
Trimax: Instead of adding 9 new lines of code, I'd suggest extending the original stream statement with notNull call

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/definition/tabular/TabularRowDef.java
```diff
@@ -2,15 +2,18 @@
...
+import java.util.ArrayList;
 import java.util.List;
 
 @Value
 public class TabularRowDef {
+
     @NotEmpty
     @SerializedName("questions")
-    List<@Valid QuestionDef> questions;
+    private List<@Valid QuestionBlockDef> questions = new ArrayList<>();
```
Trimax: No need to add `private` here since the class is annotated as Value. Lombok will generate it

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/TabularBlock.java
```diff
@@ -20,34 +19,21 @@ public final class TabularBlock extends FormBlock {
     private final List<TabularHeader> headers;
 
     @SerializedName("content")
-    private final List<List<Question>> rows;
+    private final List<QuestionBlock> rows;
 
     private final int columnsCount;
 
-    public TabularBlock(int columnsCount, List<TabularHeader> headers, List<List<Question>> rows) {
+    public TabularBlock(int columnsCount, List<TabularHeader> headers, List<QuestionBlock> rows) {
```
Trimax: Since we're not using 2d array for table I'd rename rows to blocks then

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/activity/instance/TabularBlock.java
```diff
@@ -20,34 +19,21 @@ public final class TabularBlock extends FormBlock {
...
-            throw new IndexOutOfBoundsException("The column must be between 0 and " + cellsRow.size());
-        }
-
-        return cellsRow.get(column);
-    }
-
     @Override
     public Stream<Question> streamQuestions() {
-        return StreamEx.of(rows).flatMap(Collection::stream).filter(Objects::nonNull);
+        return StreamEx.of(rows).nonNull().map(QuestionBlock::getQuestion).collect(Collectors.toList()).stream();
```
Trimax: .collect(Collectors.toList()).stream() might be removed

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
```diff
@@ -4,7 +4,7 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                    https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 
-    <changeSet author="Dmitrii Eliseev" id="2022-05-12-tabular-block">
+    <changeSet author="Dmitrii Eliseev" id="2022-06-08-tabular-block">
```
Trimax, line 7: I assume you need to create a new changeset. It was merged to dev and applied, so no changes could be done here

Trimax (dismissed):
LGTM with some minor changes

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
```diff
@@ -52,6 +52,10 @@
                 <constraints primaryKey="true" primaryKeyName="block_tabular_question_pk"/>
             </column>
 
+            <column name="column_span" type="bigint">
+                <constraints nullable="false" />
+            </column>
+
             <column name="block_tabular_id" type="bigint">
                 <constraints nullable="false" references="block_tabular(block_tabular_id)" foreignKeyName="block_tabular_question_block_fk"/>
             </column>
```
MocanaAtBroad, line 61: Github won't let me comment directly, but don't see the use of _row and _column. I think original idea was that tabular children was a 2 dimensional array (like a table). But our model calls for the child blocks of a tabular container to be a linear list. So we care about the the order of the children, but that is it.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
```diff
@@ -338,13 +338,14 @@ default void insertTabularBlock(long activityId, TabularBlockDef block, long rev
 
         for (int row = 0; row < block.getRows().size(); row++) {
             for (int column = 0; column < block.getRows().get(row).getQuestions().size(); column++) {
-                final var question = block.get(row, column);
-                if (question == null) {
+                final var questionBlockDef = block.get(row, column);
+                if (questionBlockDef == null || questionBlockDef.getQuestion() == null) {
                     continue;
```
MocanaAtBroad: The two-dimensional treatment is not needed. Child blocks are a single linear list or array. They will show up as a table in the UI, but in the backend we just need to make sure we include the in list in the right order.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
ssettipalli: Since this liquibase deploy only made upto dev : My idea is to cleanup these tabes in Dev and pretend this script was never deployed

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
ssettipalli: Since this liquibase deploy only made upto dev : My idea is to cleanup these tabes in Dev manually and pretend this liquibase script was never deployed. Tried it locally and seemed to work

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
ssettipalli: @Trimax :
Do you think we still need _row and _column columns in DB. 
Might trigger decent code changes with this schema change !!

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
ssettipalli: @MocanaAtBroad :
Is the suggestion to add a new DB column to capture order OR query sql sort by insertion order (may be block_tabular_header_id) !
May be order by will do !

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
ssettipalli: This particular code is representation of how we are capturing Tabular question in the study configuration.
I don't know at this stage we want to change that !

          "rows": [
            {
              "questions": [
                {
                  "columnSpan": 1,
                  "question": {
                    include required("../snippets/decimal-question.conf"),
                    "stableId": "TABULAR_QUESTION_L1",
                    "promptTemplate": {"templateType": "TEXT", "templateText": "$tabular.question.l1"},
                    "validations": []
                  },
                  "blockType": "QUESTION",
                  "shownExpr": true,
                },
                {
                  "columnSpan": 1,
                  "question": {
                    include required("../snippets/decimal-question.conf"),
                    "stableId": "TABULAR_QUESTION_R1",
                    "promptTemplate": {"templateType": "TEXT", "templateText": "$tabular.question.r1"},
                    "validations": []
                  },
                  "blockType": "QUESTION",
                  "shownExpr": true,
                }
              ]
            }
            }
          ],
          "columnsCount": 4,
          "blockType": "TABULAR",
          "shownExpr": null
        }

MocanaAtBroad: Looking at GroupBlockDef and ConditionalBlockDef, they both have a "nested" list of type FormBlockDef.
I don't like "nested" for Tabular (a matter of taste, if you like it, go for it), but my sense is that we want to rename rows to "nested" (if we must) or "blocks" and have that contain the question blocks (the property questions goes away).

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7713.xml
ssettipalli: done. removed these two columns

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/SectionBlockDao.java
ssettipalli: Ended up naming as QuestionBlocks...

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
```diff
@@ -74,20 +73,17 @@ ComponentBlock createComponentBlock(AIBuilderContext ctx, ComponentBlockDef comp
...
             }
-
-            rows.add(questionsRow);
+            Question question = ctx.getAIBuilderFactory().getQuestionCreator().createQuestion(ctx, questionBlockDef.getQuestion());
+            QuestionBlock questionBlock = new QuestionBlock(question, questionBlockDef.getColumnSpan());
+            questionBlock.setGuid(questionBlockDef.getBlockGuid());
+            questionBlock.setShownExpr(questionBlockDef.getShownExpr());
+            questionBlock.setEnabledExpr(questionBlockDef.getEnabledExpr());
+            allQuestions.add(questionBlock);
         }
```
MocanaAtBroad: If you were inclined to do so, this thing can be reduced to:

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
```diff
@@ -74,20 +73,17 @@ ComponentBlock createComponentBlock(AIBuilderContext ctx, ComponentBlockDef comp
...
-                final QuestionDef questionDef = tabularBlockDef.get(row, column);
-                if (questionDef == null) {
-                    questionsRow.add(null);
-                    continue;
-                }
-
-                questionsRow.add(ctx.getAIBuilderFactory().getQuestionCreator().createQuestion(ctx, questionDef));
+        final List<QuestionBlock> allQuestions = new ArrayList<>();
+        for (QuestionBlockDef questionBlockDef : tabularBlockDef.getQuestionBlocks()) {
+            if (questionBlockDef == null) {
```
MocanaAtBroad: Why would there be a null questionBlockDef in here?

MocanaAtBroad: Up to you, but this can be reduced to 
questionBlocks.stream().filter(block [check not null]).map(block -> [create question from block]).

MocanaAtBroad: Better: Better: why are we creating question tabular block from question blocks? Couldn't we just add the columnSpan property to questionBlock or even to FormBlock?
I mean: at this time we are focusing on adding questions to tabular, but high-level, we could potentially at some point decide to add other kinds of blocks. As far as I can see, we just need that one additional property. (columnSpan). Will allow you to delete at least one file and some additional code. And will be more general!

pepper-apis/studybuilder-cli/studies/basil/tabular-activity.conf
```diff
@@ -14,53 +14,66 @@
...
               "label": {"templateText": "$tabular.header.column_left"}
             },
             {
-              "columnSpan":1,
+              "columnSpan": 2,
               "label": {"templateText": "$tabular.header.column_right"}
             }
           ],
-          "rows": [
+          "questionBlocks": [
```
MocanaAtBroad: If not major change, I am going to suggest we go with "blocks". Will allow us to be forwards compatible if we ever include other kinds of blocks.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/actvityinstancebuilder/form/block/FormBlockCreatorHelper.java
ssettipalli: @MocanaAtBroad :
TabularBlock contains TabularHeader in addition to columnSpan and QuestionBlocks

ssettipalli: changed to capture and return as FormBlock's but we might need to update DB schema support later

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiBlockTabular.java
```diff
@@ -58,18 +56,27 @@ long insertQuestion(@Bind("tabularBlockId") long tabularBlockId,
             + "  JOIN revision as rev ON rev.revision_id = bt.revision_id"
             + " WHERE bt.block_id = :blockId"
             + "   AND rev.start_date <= :timestamp"
-            + "   AND (rev.end_date is null or :timestamp < rev.end_date)")
+            + "   AND (rev.end_date is null or :timestamp < rev.end_date)"
+            + " order by bth.block_tabular_header_id")
```
MocanaAtBroad, line 60: Let's not change this now, but there is an assumption we are leaning on here that keys are incremental and that insertions will happen in order. A dedicated sort order key/column is the way to go.

(Was looking how we had done it before for conditional block children, and this kind of funny: the blocks have sort key, but the section does not so we are using the primary key!)

Again, let's not mess with this now. We can take another look at some point.

https://github.com/broadinstitute/ddp-study-server/blob/355b603ee1eac09d6c0c3253f590c8b58cc11b2e/pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiFormSectionBlock.sql.stg#L29

MocanaAtBroad:
Do we have any tests that exercise tabular code in any way?

MocanaAtBroad:
Do we have to worry about export? I have been assuming "no", but would welcome confirmation to this question.

MocanaAtBroad (approved):

ssettipalli:
> Do we have to worry about export? I have been assuming "no", but would welcome confirmation to this question.

Will be doing some export testing post merge , deploy basil and add some data to tabular question

ssettipalli:
> Do we have any tests that exercise tabular code in any way?

GetActivityInstanceRouteStandaloneTest has a test case. The test class is in this PR

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiBlockTabular.java
ssettipalli: Makes sense

MocanaAtBroad:
> > Do we have any tests that exercise tabular code in any way?
> 
> GetActivityInstanceRouteStandaloneTest has a test case. The test class is in this PR

Thnx! I had missed it originally.

MocanaAtBroad (approved):

PR #1771, opened 2022-06-13
Housekeeping test

nsakande:
This change is related to housekeeping local setup
It implies that in case Houekeeping class is run with --local-housekeeping flag
In this case 

- pubSubTaskConnectionService will not be created at all.
- housekeepingDbUrl is set to apisDbUrl
- sendgrid proxy is not used.
- jobs are not started


bskinner:
The description of this PR needs some additions- can you add an overview of the changes you've made, and how do they impact the current behavior?

PR #1775, opened 2022-06-13
Housekeeping local setup

nsakande:
This change is related to housekeeping local setup.
Two flags could be passed from command line: 
  --local-housekeeping
  --emulator (it is superior to usePubSubEmulator attribute in application.conf)

In case of both of them  
- housekeepingDbUrl  is set to apisDbUrl
- proxy is set to null
- jobs are not run

in case of --local-housekeeping pubSubTaskConnectionService is not used.
in case of --emulator pub/sub emulator must be started locally.

if both flags are passed --local-housekeeping has priority.

usePubSubEmulator attribute in application.conf is checked only if --emulator is not passed.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
@@ -5,11 +5,11 @@ import static org.broadinstitute.ddp.event.pubsubtask.api.PubSubTaskConnectionSe
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.List;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -40,6 +40,10 @@ import com.google.pubsub.v1.Subscription;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.Options;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.ddp.cache.LanguageStore;
 import org.broadinstitute.ddp.client.GoogleBucketClient;
@@ -203,15 +207,37 @@ public class Housekeeping {
     }
 
     public static void start(String[] args, SendGridSupplier sendGridSupplier) {
+        boolean housekeepingLocalSetup = false;
+        boolean usePubSubEmulator = false;
+        try {
+            Options options = new Options();
+            options.addOption(null, "local-housekeeping", false, "starts housekeeping as local service");
+            options.addOption(null, "emulator", false, "starts housekeeping as local emulator client");
+            CommandLineParser parser = new DefaultParser();
+            CommandLine cmd = parser.parse(options, args);
+            housekeepingLocalSetup = cmd.hasOption("local-housekeeping");
+            usePubSubEmulator = !housekeepingLocalSetup && cmd.hasOption("emulator");
+        } catch (Exception e) {
+            log.info("command line parameters parse error");
+        }
+
         LogbackConfigurationPrinter.printLoggingConfiguration();
         Config cfg = ConfigManager.getInstance().getConfig();
         boolean doLiquibase = cfg.getBoolean(ConfigFile.DO_LIQUIBASE);
         int maxConnections = cfg.getInt(ConfigFile.HOUSEKEEPING_NUM_POOLED_CONNECTIONS);
         String pubSubProject = cfg.getString(ConfigFile.GOOGLE_PROJECT_ID);
 
-        boolean usePubSubEmulator = cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
-        String housekeepingDbUrl = cfg.getString(TransactionWrapper.DB.HOUSEKEEPING.getDbUrlConfigKey());
+        if (!usePubSubEmulator) {
+            usePubSubEmulator = !housekeepingLocalSetup && cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
+        }
         String apisDbUrl = cfg.getString(TransactionWrapper.DB.APIS.getDbUrlConfigKey());
+
+        final String housekeepingDbUrl = (usePubSubEmulator || housekeepingLocalSetup)
+                ? apisDbUrl
+                : cfg.getString(TransactionWrapper.DB.HOUSEKEEPING.getDbUrlConfigKey());
+        final String proxy = (usePubSubEmulator || housekeepingLocalSetup)
+                ? null
+                : ConfigUtil.getStrIfPresent(cfg, ConfigFile.Sendgrid.PROXY);
 
         Config sqlConfig = ConfigFactory.load(ConfigFile.SQL_CONFIG_FILE);
         DBUtils.loadDaoSqlCommands(sqlConfig);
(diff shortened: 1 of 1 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -760,31 +788,7 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
...
-                                                we accidentally write the entire key to the log.
-                                            */
-                                            if (keyParts.length == 3 && keyParts[0].equals("SG")) {
-                                                log.info("creating EmailNotificationHandler using SendGrid key id {}", keyParts[1]);
-                                            } else {
-                                                log.warn("SendGrid API key is in an unexpected format.");
-                                            }
-
-                                            new EmailNotificationHandler(sendGridProvider.get(sendGridKey),
+                                            new EmailNotificationHandler(sendGridProvider.get(notificationMessage.getApiKey()),
```
bskinner, line 792: Why is this code being removed?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -203,16 +207,37 @@ public static void main(String[] args) {
...
+            usePubSubEmulator =!housekeepingLocalSetup && cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
         String apisDbUrl = cfg.getString(TransactionWrapper.DB.APIS.getDbUrlConfigKey());
 
+        final String housekeepingDbUrl = ( usePubSubEmulator || housekeepingLocalSetup )
+                ? apisDbUrl
+                : cfg.getString(TransactionWrapper.DB.HOUSEKEEPING.getDbUrlConfigKey());
+        final String proxy = ( usePubSubEmulator || housekeepingLocalSetup )
+                ? null
+                : ConfigUtil.getStrIfPresent(cfg, ConfigFile.Sendgrid.PROXY);
+
```
bskinner: The `dbUrl` and `housekeepingDbUrl` use two separate databases, and have two separate schemas. The URLs are not interchangeable.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -232,24 +257,27 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
         }
         TransactionWrapper.useTxn(TransactionWrapper.DB.APIS, LanguageStore::init);
 
-        setupScheduler(cfg);
+        setupScheduler(cfg, !housekeepingLocalSetup && !usePubSubEmulator);
```
bskinner, line 261: Instead of adding another parameter here, can the command line flags be used to override the specific values in the configuration so we only need to deal with a single source of configuration values.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -203,16 +207,38 @@ public static void main(String[] args) {
...
-        boolean usePubSubEmulator = cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
-        String housekeepingDbUrl = cfg.getString(TransactionWrapper.DB.HOUSEKEEPING.getDbUrlConfigKey());
+        if (!usePubSubEmulator) {
+            usePubSubEmulator = !housekeepingLocalSetup && cfg.getBoolean(ConfigFile.USE_PUBSUB_EMULATOR);
+        }
         String apisDbUrl = cfg.getString(TransactionWrapper.DB.APIS.getDbUrlConfigKey());
 
+        final String housekeepingDbUrl = (usePubSubEmulator || housekeepingLocalSetup)
+                ? apisDbUrl
+                : cfg.getString(TransactionWrapper.DB.HOUSEKEEPING.getDbUrlConfigKey());
```
bskinner, line 237: What's your reasoning for swapping the url? These two URLs are not interchangeable- the `dbUrl` and `housekeepingDbUrl` point at two different databases with different schemas.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -203,16 +207,38 @@ public static void main(String[] args) {
...
+        final String proxy = (usePubSubEmulator || housekeepingLocalSetup)
+                ? null
+                : ConfigUtil.getStrIfPresent(cfg, ConfigFile.Sendgrid.PROXY);
+
```
bskinner, line 241: Why is this preferable to removing the proxy configuration from the configuration file?

bskinner (changes requested):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: > Why is this preferable to removing the proxy configuration from the configuration file?

we can not access sendgrid via proxy from local environment. Exception is thrown

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: > swapping

We do not have access to development environment housekeeping db.
locally we use mysql db of ddp-study-server for housekeeping needs.
There is a way to setup housekeeping db locally in docker but in this case, I think,  we need some init script for db creation.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: > The `dbUrl` and `housekeepingDbUrl` use two separate databases, and have two separate schemas. The URLs are not interchangeable.

We do not have access to development environment housekeeping db.
locally we use mysql db of ddp-study-server for housekeeping needs.
There is a way to setup housekeeping db locally in docker but in this case, I think, we need some init script for db creation.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: I understand, but that's not what I was asking.

The proxy will only be used if it is the `sendgrid.proxy` key set to a non-null, non-empty value- why does this need a flag? If a flag is desirable in this case, let's override the value in the generated Config object instead.

In addition, the `dev` proxy _is_ accessible from a local environment when connected to the Broad VPN.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: It also appears that method does not have any side effects if the `allowed` parameter is set to `true`. Can we skip the call to `setupScheduler` entirely in this case?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -264,7 +293,7 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
                 Subscription subscription = pubsubConnectionManager
                         .createSubscriptionIfNotExists(projectSubscriptionName, projectTopicName);
                 // in the real world, listen differently
-                setupMessageReceiver(pubsubConnectionManager, projectSubscriptionName, cfg, sendGridSupplier);
+                setupMessageReceiver(pubsubConnectionManager, projectSubscriptionName, cfg, sendGridSupplier, proxy);
```
bskinner, line 296: It would be preferable to update the configuration to reflect the correct value of the proxy (`null` or the empty string, in this case) instead of adding an additional parameter.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: There is a Liquibase schema included with the project. When `housekeeping` is run for the first time against an empty database (or when there are schema updates), the migrations will trigger and setup the necessary schema.

You will need to set the `doLiquibase` flag to `true` in the configuration for this functionality to be enabled.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: You are right.
It is better to overnight ConfigFile.RUN_SCHEDULER configuration parameter. 
This configuration parameter was read in setupScheduler function that's why I did not touch it.
I think is is better to set runScheduler variable somewhere at the  top of start method and pass it to setupScheduler  method as second parameter.
Is this OK?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: > Why is this code being removed?

Very strange. 
This change is not mine.
And I cannot say that it is accidental removal.

It looks like some kind of automation made code optimization.

Anyway I will restore this part from develop branch

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zbokucha: I will go this way, I have tried with empty  string and it did not work.
I hope with null it will work properly.

There is another point.

Application.conf is updated from vault on every configuration rendering.
and this happens almost on every build (at least very frequently).
That's why I prefer not to change anything in application.conf if this is possible

zbokucha:
--In addition, the dev proxy is accessible from a local environment when connected to the Broad VPN.
I will try with VPN and if it work I will rollback proxy related changes back. OK

zbokucha:
I have tried with VPN and got this:

Caused by: java.net.UnknownHostException: No such host is known (gae-egress-proxy-dev101.c.broad-ddp-dev.internal)
	at java.base/java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
	at java.base/java.net.InetAddress$PlatformNameService.lookupAllHostAddr(InetAddress.java:929)
	at java.base/java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1515)
	at java.base/java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:848)
	at java.base/java.net.InetAddress.getAllByName0(InetAddress.java:1505)

bskinner:
> I have tried with VPN and got this:

I'll reach out to you over Slack- I have an idea what's going on here.

zbokucha:
> > I have tried with VPN and got this:
> 
> I'll reach out to you over Slack- I have an idea what's going on here.

Ok. Thanks.
When can we have a call?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
MocanaAtBroad: Agree with Blake. The schemas are not interchangeable.

MocanaAtBroad (changes requested):
Some improvements flagged by @bskinner needed.

PR #1785, opened 2022-06-14
[DSS] DDP-8024. Update FON forms to use new features

Trimax:
According to the [Google Document](https://docs.google.com/spreadsheets/d/1UAmLS5wV8nSud9U3PGgp9QNbEu4H3WJb/edit#gid=651091428) I've updated **numeric** questions to **decimal** questions, specified the scale and updated the validations accordingly.


MocanaAtBroad:
Did you check that we are using equations/derived fields wherever they are needed? I was pleasantly surprised that it was already set in a bunch of places (I was not expecting that!) but at least in promix proxy form we are missing equation for tscore. There might be other places.
Just look for the word DERIVED in that spreadsheet.

Trimax:
> Did you check that we are using equations/derived fields wherever they are needed? I was pleasantly surprised that it was already set in a bunch of places (I was not expecting that!) but at least in promix proxy form we are missing equation for tscore. There might be other places. Just look for the word DERIVED in that spreadsheet.

Done.

ssettipalli (dismissed):

ssettipalli (approved):

PR #1790, opened 2022-06-15
[DSM] TestBoston_ManualOrders

pegahtah:
This class is moved from branch DDP-6484_TestBostonManualOrders in the ddp-study-manager repo, the only change is I moved it into the `test` folder to be able to use the `setupDB` functionality


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/RegisterBirchOrderCLI.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @BeforeClass
+    public static void before() throws Exception {
+        setupDB();
+    }
+
+
+    public static void main(String[] args) {
+        try {
+            if (args == null || args.length == 0 || ("-h".equals(args[0]))) {
```
SimoneMaiwald, line 56: I think I would also add a check that args.length is 4

PR #1815, opened 2022-06-20
[SB] Resolved DDP-8203

leonidrakitin:
Resolved DDP-8203


gskhirel (approved):

LKapanad:
@leonidrakitin will it increase maxFileSize for activity instances which will be created after running patch?
in my opinion only old instances' size will be updated

PR #1834, opened 2022-06-21
DDP-7869: genomic ID

SimoneMaiwald:
Changing the creation of the genomic ID from prequal to registration activity 
everyone who has registration activity completed should get a genomic id on the DSM side


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/ATDefaultValues.java
```diff
@@ -57,18 +52,15 @@ private boolean isParticipantDataInES() {
     }
 
 
-    boolean isSelfOrDependentParticipant() {
+    boolean isParticipantRegistrationComplete() {
         if (elasticSearchParticipantDto.getActivities().isEmpty()) {
             return false;
         }
```
devonbush:
```suggestion
```

I don't think we need these lines -- if getActivities() is empty, `getActivities().stream().anyMatch()` will immediately return false, right?

devonbush (dismissed):
Looks reasonable to my naive reading.  Just one suggestion around removing some unneeded lines

gmakharat (dismissed):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Participant.java
```diff
@@ -43,97 +43,66 @@ public class Participant implements Cloneable {
             + "LEFT JOIN ddp_onc_history o on (o.participant_id = p.participant_id) "
             + "LEFT JOIN ddp_participant_record r on (r.participant_id = p.participant_id) "
             + "LEFT JOIN ddp_participant_exit ex on (p.ddp_participant_id = ex.ddp_participant_id "
-            + "AND p.ddp_instance_id = ex.ddp_instance_id) "
-            + "WHERE realm.instance_name = ? ";
+            + "AND p.ddp_instance_id = ex.ddp_instance_id) " + "WHERE realm.instance_name = ? ";
```
devonbush, line 46:
```suggestion
            + "AND p.ddp_instance_id = ex.ddp_instance_id) WHERE realm.instance_name = ? ";
```
since you consolidated lines, might as well remove the extra concatentation

devonbush (approved):
Looks good.  One super-trivial suggestion that you should totally feel free to ignore.

gmakharat (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Participant.java
SimoneMaiwald: was auto-consolidated by stupidJ. I will leave it as it is for now, because otherwise I need someone to approve it again, but I will try to keep it in mind and change it with the next PR

PR #1852, opened 2022-06-23
[DSM] DDP-7869: at

SimoneMaiwald:
changing ES upsert

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
@@ -136,16 +136,15 @@ public class ParticipantData {
             throw new RuntimeException("Could not insert participant data for : " + this.ddpParticipantId);
         }
         participantData.setParticipantDataId(createdDataKey);
-        logger.info("Successfully inserted data for participant: " + this.ddpParticipantId);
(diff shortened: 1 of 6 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
```diff
@@ -136,16 +136,15 @@ public long insertParticipantData(String userEmail) {
...
 
         DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceId(ddpInstanceId).orElseThrow();
         String participantGuid = Exportable.getParticipantGuid(ddpParticipantId, ddpInstanceDto.getEsParticipantIndex());
 
         try {
             UpsertPainlessFacade.of(DBConstants.DDP_PARTICIPANT_DATA_ALIAS, participantData, ddpInstanceDto,
-                            ESObjectConstants.PARTICIPANT_DATA_ID, ESObjectConstants.DOC_ID,
-                            participantGuid, new PutToNestedScriptBuilder())
-                    .export();
+                    ESObjectConstants.PARTICIPANT_DATA_ID, ESObjectConstants.PARTICIPANT_DATA_ID, participantDataId,
```
SimoneMaiwald, line 146: @gmakharat can you help me to get it to write into ES. 
With this version it tells me that it doesn't find the doc and with the previous version it throws error on the ES side with 409

PR #1867, opened 2022-06-24
[DSM] DDP-7869: at genomic id

SimoneMaiwald:
1) adding a retry to the UpsertPainless
for AT the first default value is not getting written into ES because the last workflow update is causing 409 conflict of document ```version conflict, required seqNo [712], primary term [1]. current document has seqNo [713] and primary term```

2) writing workflow values also to ES under dsm.participantData before they get exported as workflow object


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -46,7 +46,14 @@ public void export() {
                     "created/updated %s records in ES data for %s", getNumberOfUpserted(bulkByScrollResponse), generator.getPropertyName())
             );
         } catch (IOException e) {
-            throw new RuntimeException("Error occurred while exporting data to ES", e);
+            try {
```
devonbush: rather than duplicating this entire block, let's put it in a method or lambda.  so something like:
```
IntPredicate executeExport = (i) -> {
    try {
            BulkByScrollResponse bulkByScrollResponse = clientInstance.updateByQuery(updateByQueryRequest, RequestOptions.DEFAULT);
            logger.info(String.format(
                    "created/updated %s records in ES data for %s", getNumberOfUpserted(bulkByScrollResponse), generator.getPropertyName())
            );
           return true;
        } catch (IOException e) {
            throw new RuntimeException("Error occurred while exporting data to ES, on try number " + i, e);
       }
       return false;
}
for (int tryNum = 1; tryNum < 3; tryNum++) {
   if (executeExport(tryNum)) {
      break;
    }
}
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -146,10 +150,12 @@ private static Optional<WorkflowForES.StudySpecificData> getProbandStudySpecific
     public static int addNewParticipantDataWithStatus(String workflow, String status, String ddpParticipantId, FieldSettingsDto setting) {
         JsonObject dataJsonObject = new JsonObject();
         dataJsonObject.addProperty(workflow, status);
```
devonbush, line 152: can you add a brief comment explaining this?  I'm having trouble following it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -169,6 +175,23 @@ public static void updateProbandStatusInDB(String workflow, String status, Parti
                             .withDdpInstanceId(participantData.getDdpInstanceId())
                             .withFieldTypeId(participantData.getFieldTypeId().orElse("")).withData(dataJsonObject.toString())
                             .withLastChanged(System.currentTimeMillis()).withChangedBy(DSS).build());
+            writeParticipantDataInES(participantData);
+        }
+    }
+
```
devonbush, line 181: can you add a brief comment explaining this method?

devonbush (changes requested):
Overall looks good, but some comments and making the retry loop explicit would make this much more readable to noobs like me.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -46,7 +46,14 @@ public void export() {
...
+                BulkByScrollResponse bulkByScrollResponse = clientInstance.updateByQuery(updateByQueryRequest, RequestOptions.DEFAULT);
+                logger.info(String.format("created/updated %s records in ES data for %s", getNumberOfUpserted(bulkByScrollResponse),
+                        generator.getPropertyName())
+                );
+            } catch (IOException retryE) {
+                throw new RuntimeException("Error occurred while exporting data to ES", retryE);
+            }
```
gmakharat: I think retry on exception is not a good solution for the error mentioned in the description, there are doubts why it happens:
https://stackoverflow.com/questions/56840637/version-conflict-when-using-the-delete-method-of-elasticsearch-dsl
https://www.elastic.co/guide/en/elasticsearch/guide/master/optimistic-concurrency-control.html

maybe joining those two insertion `insertGenomicIdForParticipant() && insertExitStatusForParticipant()` could fix the problem, if I got correctly from the links why it happens.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -40,13 +40,22 @@ public void export() {
         updateByQueryRequest.setScript(painless);
         updateByQueryRequest.setMaxRetries(5);
         updateByQueryRequest.setRefresh(true);
+        for (int tryNum = 1; tryNum < 3; tryNum++) {
+            if (executeExport(clientInstance, updateByQueryRequest, tryNum)) {
+                break;
+            }
+        }
+    }
+
```
devonbush, line 49: nice, this looks a lot cleaner than the lambda approach I suggested :)

devonbush (approved):
Nice

PR #1897, opened 2022-06-30
[DSM] DDP - 8119 : Draft PR for Stool Sample upload functionality

lasha-gamrekelashvili:
## Context

Draft PR regarding the Stool Upload functionality, atm just missing the SQL script to it iterate over the parsed file and update the DB

Update from @Anzori-Ghurtchumelia: 

- Introduced `StoolUploadService.java` which will exercise any business related IO operations;
- Improved `StoolUploadDao.java` by adding `updateKit` method which updates the kit in a single DB operation instead of two or more successive operations;
- Created a new generic abstraction called `AbstractRecordsParser.java` for parsing table-like records indexed by columns in a file to a `List<T>`, examples can be (TSV, CSV, PSV etc..);
- Established a new interface called `HeadersProvider.java` for providing a list of headers;
- Created an implementation of the `HeadersProvider` called `StoolUploadHeadersProvider.java`;
- Added `TSVRecordsParser.java` abstract class for parsing `tab separated` values;
- Made a special instance of `TSVRecordsParser` for this feature called `TSVStoolUploadRecordsParser.java` for parsing TSV records into the `List<StoolUploadObject>` in order to run DB and other IO operations;
- Refined error handling by returning the messages of concrete exceptions which can be caused by incorrect state of the file (incorrect separation of headers or records, missing records, missing headers, empty file etc..);
- Write comprehensive unit tests for new classes.

Notes: 
- Any descendant of `AbstractRecordsParser` should implement a single abstract method: `T transformMapToObject(Map<String, String> map)`
       explanation: this is a generic method which transforms the map into the instance of generic `T`
       implementation: again, have a look at `TSVStoolUploadRecordsParser` to get the idea.
      

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+
+        if (queryParams.value(RoutePath.REALM) != null) {
+            realm = queryParams.get(RoutePath.REALM).value();
+        } else {
+            throw new RuntimeException("No realm query param was sent");
+        }
+
+        String userIdRequest = UserUtil.getUserId(request);
+
+        if (UserUtil.checkUserAccess(realm, userId, "kit_upload", userIdRequest)) {
```
SimoneMaiwald: I wonder if we should add a new role for that @pegahtah what is your feeling about that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+
+            AtomicReference<String> kitUploadReason = new AtomicReference<>();
+            AtomicReference<String> shippingCarrier = new AtomicReference<>();
+            if (queryParams.value(RoutePath.KIT_TYPE) != null) {
+                kitTypeName = queryParams.get(RoutePath.KIT_TYPE).value();
+            } else {
+                throw new RuntimeException("No kitType query param was sent");
+            }
+            if (queryParams.value("reason") != null) {
+                kitUploadReason.set(queryParams.get("reason").value());
```
SimoneMaiwald: not needed for stool upload

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+        if (UserUtil.checkUserAccess(realm, userId, "kit_upload", userIdRequest)) {
+            String kitTypeName;
+
+            AtomicReference<String> kitUploadReason = new AtomicReference<>();
+            AtomicReference<String> shippingCarrier = new AtomicReference<>();
+            if (queryParams.value(RoutePath.KIT_TYPE) != null) {
+                kitTypeName = queryParams.get(RoutePath.KIT_TYPE).value();
+            } else {
+                throw new RuntimeException("No kitType query param was sent");
+            }
```
SimoneMaiwald: not needed for stool upload. kit should always be of type stool

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+            if (queryParams.value("reason") != null) {
+                kitUploadReason.set(queryParams.get("reason").value());
+            }
+            if (queryParams.value("carrier") != null) {
+                shippingCarrier.set(queryParams.get("carrier").value());
+            }
```
SimoneMaiwald: not needed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/StoolUploadObject.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Data;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
```
gmakharat: `@Data` annotation already covers `@Getter/@Setter` automatically.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+        QueryParamsMap queryParams = request.queryMap();
+        String realm;
+
+        if (queryParams.value(RoutePath.REALM) != null) {
+            realm = queryParams.get(RoutePath.REALM).value();
+        } else {
+            throw new RuntimeException("No realm query param was sent");
+        }
+
+        String userIdRequest = UserUtil.getUserId(request);
```
gmakharat: you get `userId` already in `processRequest` thus is not needed to get it from the request.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+            if (queryParams.value("reason") != null) {
+                kitUploadReason.set(queryParams.get("reason").value());
+            }
+            if (queryParams.value("carrier") != null) {
+                shippingCarrier.set(queryParams.get("carrier").value());
+            }
+
+
+            HttpServletRequest rawRequest = request.raw();
+            String content = SystemUtil.getBody(rawRequest);
```
gmakharat: I believe you don't have to play with `HttpServletRequest` to get body, you could get it by `request.body()`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+
+            AtomicBoolean uploadAnyway = new AtomicBoolean(false);
+            if (queryParams.value("uploadAnyway") != null) {
+                uploadAnyway.set(queryParams.get("uploadAnyway").booleanValue());
+            }
+
+            try {
+                List<StoolUploadObject> stoolUploadContent = null;
```
gmakharat: It is better to separate business logic from route class, you can use `CreateCohortTagRoute` as a reference how it is better route to look like.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+                if (uploadAnyway.get()) {
+                    stoolUploadContent = Arrays.asList(new Gson().fromJson(content, StoolUploadObject[].class));
+                } else {
+                    try {
+                        stoolUploadContent = isFileValid(content, realm);
+                    } catch (Exception e) {
+                        return new Result(500, e.getMessage());
```
gmakharat: `Result` is old class, we try to eliminate it from our code base, instead of using it, you can return value directly and use `response.status(500)` to set specific status to the response.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+
+            AtomicBoolean uploadAnyway = new AtomicBoolean(false);
+            if (queryParams.value("uploadAnyway") != null) {
+                uploadAnyway.set(queryParams.get("uploadAnyway").booleanValue());
+            }
+
+            try {
+                List<StoolUploadObject> stoolUploadContent = null;
+                if (uploadAnyway.get()) {
+                    stoolUploadContent = Arrays.asList(new Gson().fromJson(content, StoolUploadObject[].class));
```
gmakharat: I personally prefer to use `jackson` over to `gson`, there is singleton class called `ObjectMapperSingleton` which provides instance of `ObjectMapper` of `jackson` and also some helper functions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+                } else {
+                    try {
+                        stoolUploadContent = isFileValid(content, realm);
```
gmakharat: as far as I know `isFileValid` already exists in `KitUploadRoute` instead of duplicating, you could introduce the class which will wrap all file validation logic and use object of that class in both of the case.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,193 @@
...
+                        return new Result(500, e.getMessage());
+                    }
+                }
+                final List<StoolUploadObject> stoolUploadObjects = stoolUploadContent;
+                if (stoolUploadObjects == null || stoolUploadObjects.isEmpty()) {
+                    return "Text file was empty or couldn't be parsed to the agreed format";
+                }
+
+                stoolUploadObjects.forEach(stoolUploadObject -> {
+                    //TODO sql scripts to iterate over the objects and update the DB accordingly
```
gmakharat: as the reference, we follow `Dao` approach for database related logics.

ScalaChamp (commented):
@lasha-gamrekelashvili let's add some logging info gradually.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                }
+
+                final List<StoolUploadObject> stoolUploadObjects = stoolUploadContent;
+                if (stoolUploadObjects.isEmpty()) {
+                    return "Text file was empty or couldn't be parsed to the agreed format";
+                }
+
+                stoolUploadObjects.forEach(stoolUploadObject -> {
+                    StoolUploadDao stoolUploadDao = new StoolUploadDao();
```
ScalaChamp: Major: Let's avoid creating a new instance of StoolUploadDao for each iteration. We can extract StoolUploadDao as an instance field and inject it from the client or directly create new instance in the processRequest method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/stoolupload/StoolUploadDao.java
```diff
@@ -0,0 +1,95 @@
...
+
+                if (updateAmount == 1) {
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(
+                            DSMConfig.getSqlFromConfig(ApplicationConfigConstants.GET_RECEIVED_KIT_INFORMATION_FOR_NOTIFICATION_EMAIL),
+                            mfBarcode, 1);
+
+                    if (kitDDPNotification != null) {
+                        EventUtil.triggerDDP(conn, kitDDPNotification);
+                    }
+                }
```
ScalaChamp, line 41: Optional: We can add `else` statement in case `updateAmount` is not equal to 1 and throw RuntimeException indicating that there was an error while updating kit, it will help us tracking down the possible errors faster.

P.S We can also add successful logs in case everything goes well. For reference please take a look at `DDPInstitutionDao` line 57.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/stoolupload/StoolUploadDao.java
```diff
@@ -0,0 +1,95 @@
...
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.util.DSMConfig;
+import org.broadinstitute.dsm.util.EventUtil;
+import org.broadinstitute.lddp.db.SimpleResult;
+
+public class StoolUploadDao implements Dao<StoolUploadDto> {
+    private static final String SELECT_KIT_ID = "SELECT dsm_kit_id FROM ddp_kit INNER JOIN ddp_kit_request dkr"
+            + " on ddp_kit.dsm_kit_request_id = dkr.dsm_kit_request_id";
+    private static final String BY_BARCODE = " WHERE kit_label = ?";
+    private static final String BY_PT_ID = " dkr.ddp_participant_id = ?";
```
SimoneMaiwald: looks like you are using these 3 only ones and only together, so you could combine them to one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/stoolupload/StoolUploadDao.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final String UPDATE_KIT = "UPDATE ddp_kit SET receive_date = ?, receive_by = 'HSPH'";
+    private static final String BY_KIT_ID = " WHERE dsm_kit_id = ?";
+    private static final String IF_NULL = " AND receive_date IS NULL";
```
SimoneMaiwald: same here. no need to split them if there are only used here and only ones. 
I would recommend to only split them if we need to reuse them

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/stoolupload/StoolUploadDao.java
```diff
@@ -0,0 +1,95 @@
...
+                int updateAmount = stmt.executeUpdate();
+
+                if (updateAmount == 1) {
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(
+                            DSMConfig.getSqlFromConfig(ApplicationConfigConstants.GET_RECEIVED_KIT_INFORMATION_FOR_NOTIFICATION_EMAIL),
+                            mfBarcode, 1);
+
+                    if (kitDDPNotification != null) {
+                        EventUtil.triggerDDP(conn, kitDDPNotification);
+                    }
```
SimoneMaiwald: I don't like having the triggering of the DDP in the method to updateKitData. 
Could you change that updateKitData returns a boolean and then depending on that you could in the method that called that one do the notification part afterwards

lasha-gamrekelashvili: @SimoneMaiwald  In that case I would need to create a separate transaction just for triggering DDP, is that okay?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,180 @@
...
+
+                    String participantId = stoolUploadObject.getParticipantId();
+                    String mfBarcode = stoolUploadObject.getMfBarcode();
+                    String receiveDate = stoolUploadObject.getReceiveDate();
+
+                    Optional<StoolUploadDto> stoolUploadDto = stoolUploadDao.getStoolUploadDto(participantId, mfBarcode);
+
+                    if (stoolUploadDto.isPresent()) {
+                        logger.info("Successfully generated kits, trying to update the table with provided parameters...");
+                        stoolUploadDao.updateKitData(receiveDate, stoolUploadDto.get().getKitId(), mfBarcode);
```
SimoneMaiwald: I think I would like to change that to having the select in the update. 
I worry that if they are uploading a file with 100 kits we need to call the db 100*2 
If you combine the update and select query we only need to do 1 call per kit.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/stoolupload/StoolUploadDao.java
```diff
@@ -0,0 +1,100 @@
...
+            return dbVals;
+        });
+        if (simpleResult.resultException != null) {
+            throw new RuntimeException("No kit found with the barcode " + mfBarcode,
+                    simpleResult.resultException);
+        }
+        return Optional.of((StoolUploadDto) simpleResult.resultValue);
+    }
+
+    public void updateKitData(String receiveDate, int kitId, String mfBarcode) {
```
SimoneMaiwald: have the method return a boolean and move the EventUtil.triggerDDP out of the StoolUploadDao. Let the class/method that calls updateKitData handle the triggering

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
gmakharat: No need any manipulation with request to get `userId`, `userId` is already passed to `processRequest`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
```diff
@@ -0,0 +1,180 @@
...
+                });
+            } catch (Exception e) {
+                return e.getMessage();
+            }
+        } else {
+            response.status(500);
+            return (UserErrorMessages.NO_RIGHTS);
+        }
+        return null;
+    }
```
gmakharat, line 46: Once again, too complex logics in `Route` class, main purpose of `Route` class is to extract necessary data from the request and pass it to some so called use case classes, which are responsible to hold all complex logics and encapsulate them, for the reference you can get `CreateCohortTagRoute`, moreover I remember that the similar logic is used somewhere already probably for kit upload, instead of duplicating the same code you may introduce some abstraction which will gather the same logic for both of them.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/StoolUploadRoute.java
gmakharat: agree here, instead of doing DB transaction for each item of the list, introduce bulk upload in Dao class, which will insert list of samples at once.

ScalaChamp (approved):

ScalaChamp:
@gmakharat re-review please?

gmakharat (approved):

PR #1894, opened 2022-06-30
[DSS] DDP-7061: Center creation route

Trimax:

PR #1900, opened 2022-06-30
Updates to /user registration endpoint [DDP-7931]

bskinner:
## Context
_See DDP-7931_

This PR includes several bug fixes, changes resulting from code review comments, and some general code reorganization for the user registration endpoint implemented for FON.

* Fixes a case where an incorrectly formatted email address would result in a 500 instead of 422
* Fixes the error message returned to the client when validation fails on the request payload
* Removes the checks for an integer center id 
  * _centers will need another ticket covering integration once #1894 is merged_
* Clarified behaviors identified in javadoc comments

These changes can be demoed using Postman, or cURL. A valid and unexpired DSM token is required to perform the call.

```
curl --location --request POST 'http://127.0.0.1:5556/pepper/v1/user' \
  --header 'Authorization: Bearer MY_DSM_TOKEN' \
  --header 'Content-Type: application/json' \
  --data-raw '{
    "studyGuid": "fon",
    "firstName": "Test",
    "lastName": "User",
    "birthDate": "1980-01-02",
    "sex": "PREFER_NOT_TO_ANSWER",
    "email": "not.existant@email.example"
  }'
```

## Testing


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserCreationRoute.java
```diff
@@ -47,29 +49,8 @@ public Object handle(Request request, Response response, UserCreationPayload pay
...
-            throw ResponseUtil.haltError(HttpStatus.SC_UNPROCESSABLE_ENTITY, error);
-        }
-
-        if (StringUtils.isNotBlank(payload.getCenterId())) {
-            var error = new ApiError(ErrorCodes.NOT_SUPPORTED,
-                    "passing a non-null center id when creating a participant is not supported yet.");
-            throw ResponseUtil.haltError(HttpStatus.SC_UNPROCESSABLE_ENTITY, error);
-        }
+        final var studyGuid = new Guid(payload.getStudyGuid());
+        final var email = new EmailAddress(payload.getEmail());
```
bskinner, line 53: My intent here is to bake the validation done on the email into the type-  Instead of implicitly assuming that the content of the string has been validated further up the call-tree, an object of this type can only be created if the email address is correctly formatted, and that validation is explicitly declared through the type.

This can be trivially removed if folks feel this is over-engineered.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/users/models/EmailAddress.java
```diff
@@ -0,0 +1,21 @@
...
+@Data
+public class EmailAddress {
+
+    @Email
+    private final String value;
+
+    public EmailAddress(String email) {
+        if (EmailValidator.getInstance().isValid(email) == false) {
+            throw new IllegalArgumentException("email format is invalid");
+        }
```
bskinner, line 17: Quick note on the validation here- The Apache and Hibernate validators accept different formats for the email address.

The `javax.validation.constraints.Email` validator allows for `local-part@the-rest` (a `@` is required), while the Apache EmailValidator requires something in the form of `local-part@domain.part` (A `@` and `.` is required). I don't think we're likely to run into issues with this in normal use, but it is a distinction to be aware of.

Suggestions on how to invoke the javax validator outside of this class be serialized/deserialized (currently, the checking is only done on the request payload by the `ValidatedJsonInputRoute` class)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/users/models/Guid.java
```diff
@@ -0,0 +1,12 @@
...
+import lombok.Value;
+
+@Value
+public class Guid {
+    private String value;
+
+    public String toString() {
+        return value;
+    }
+}
```
bskinner, line 12: This class mostly just serves as explicit indicator of the content type. There are multiple types of GUIDs that we use, and they don't have explicit indicators what each one is used for, so there isn't much we can do here in the way of validation.

Trimax (dismissed):
LGTM

MocanaAtBroad (approved):

ssettipalli (approved):

PR #1922, opened 2022-07-06
DSM export upgrades - DDP-7581

devonbush:
## Context
This upgrades the export capability to support both tsv and xlsx files, and support splitting multiselect questions into one column per option.  It also makes some performance improvements and bugfixes in the handling of JSON data.  It also contains fixes/improvements outlined in DDP-7581.  The functionality has been shown and approved by AT study staff.  

I have focused my testing on AT, RGP, and PanCan

These changes require the frontend PR https://github.com/broadinstitute/ddp-angular/pull/1489 to be visible.

TO TEST:
 1. ensure you have the frontend PR checked out too
 2. load the participant list for the AT study
 3. press the 'download' button
 4. leave all options on defaults
 5. hit download
 6. confirm the file is downloaded as a tsv
 7. open the file in a spreadsheet editor
 8. Search for 'MEDICAL_HISTORY.DETERMINED', and observe that the options have been split into one-column-per-option

## Checklist

- [*] I have labeled the type of changes involved using the `C-*` labels.
- [*] I have assessed potential risks and labeled using the `R-*` labels.
- [*] I have considered error handling and alerts, and added `L-*` labels as needed.
- [*] I have considered security and privacy, and added `I-*` labels as needed
- [*] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [*] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [*] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

## FUD Score

- [*] :sweat_smile: There might be some issues here

- [*] They are user-visible in dev as a regular user journey and require no additional instructions.

## Testing

- [*] I have written zero automated tests but have poked around locally to verify proper functionality

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ModuleExportConfig.java
```diff
@@ -0,0 +1,39 @@
...
+    private final String name;
+    private final Alias filterKey;
+    private final List<FilterExportConfig> questions = new ArrayList<>();
+    private boolean isActivity = false;
+    @Setter
+    private int numMaxRepeats = 1;
+
+    public ModuleExportConfig(ParticipantColumn participantColumn) {
+        filterKey = Alias.of(participantColumn);
+        if (filterKey.equals(Alias.ACTIVITIES)) {
```
SimoneMaiwald: can we do it yoda still? 
Alias.ACTIVITIES.equals(filterKey)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -4,91 +4,115 @@
...
-        generator.formatSizes(columnsNumber);
-        generator.writeInResponse(response);
+        return allResults;
+    }
 
-        return response.raw();
+    @Getter
+    @Setter
+    /** on retirement of feature-flag-export-new, this class should be promoted to DownloadParticipantListPayload */
+    private static class DownloadParticipantListPayloadNew {
```
SimoneMaiwald: I don't think we need to call this "New" anymore if we rip out the other Payload class
Maybe use it in a separate class file, please

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -4,91 +4,115 @@
...
+     * Generates a file for download.  When removing the feature-flag-export-new role, processRequestNew should
+     * be renamed to processRequest, and the old 'processRequest' method should be deleted
+     */
+    public Object processRequest(Request request, Response response, String userId) throws Exception {
+        DownloadParticipantListPayloadNew payload =
+                ObjectMapperSingleton.instance().readValue(request.body(), DownloadParticipantListPayloadNew.class);
+        DownloadParticipantListParams params = new DownloadParticipantListParams(request.queryMap());
+
+        String realm = RoutePath.getRealm(request);
+        DDPInstance instance = DDPInstance.getDDPInstanceWithRole(realm, DBConstants.MEDICAL_RECORD_ACTIVATED);
```
SimoneMaiwald, line 51: Before we do any download here or any more calls to db, I think we should check if the user is actually allowed to see the pt list and therefore allowed to download the data.
Have a look at `FilterRoute` how it checks the permissions there.
and I think we should also add an additional permission here `pt_list_export`. seeing that for FON we need that permission separated from `pt_list_view`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/AmbulationValueProvider.java
```diff
@@ -0,0 +1,27 @@
+package org.broadinstitute.dsm.model.elastic.export.tabular.renderer;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.model.elastic.export.tabular.FilterExportConfig;
+
+public class AmbulationValueProvider extends TextValueProvider {
```
SimoneMaiwald, line 10: That seems like a very study (AT) specific thing. I wonder if we need other special behavior classes than that. Could we please add the study name to the class name
My two thoughts come up here:
1) What makes that question special
2) Are there any other studies that might have that problem/question type

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7581-add-feature-flag-export-new.xml
```diff
@@ -0,0 +1,12 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet  author="dbush"  id="add-feature-flag-export-new">
+        <insert  tableName="access_role">
+            <column  name="name"  value="feature_flag_export_new"/>
+        </insert>
+    </changeSet>
+ </databaseChangeLog>
```
SimoneMaiwald: I don't think checkstyle likes if there is no empty line at the end

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7581-add-feature-flag-export-new.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet  author="dbush"  id="add-feature-flag-export-new">
+        <insert  tableName="access_role">
+            <column  name="name"  value="feature_flag_export_new"/>
```
SimoneMaiwald: should we still add the flag to the permissions even if we are not rolling out the change per feature flag?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public static TabularParticipantExporter getExporter(List<ModuleExportConfig> moduleConfigs,
+                                                         List<Map<String, String>> participantValueMaps, String fileFormat) {
+        if (TSV_FORMAT.equals(fileFormat)) {
+            return new TsvParticipantExporter(moduleConfigs, participantValueMaps, fileFormat);
+        } else if (XLSX_FORMAT.equals(fileFormat)) {
+            return new ExcelParticipantExporter(moduleConfigs, participantValueMaps, fileFormat);
+        }
+        throw new RuntimeException("Unrecognized file format");
```
SimoneMaiwald: I wonder if the route should have already checked that. by the time it throws here it already collected all the pt data for nothing…

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ExcelParticipantExporter.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.google.common.net.MediaType;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.poi.ss.usermodel.Row;
+import org.apache.poi.xssf.streaming.SXSSFSheet;
+import org.apache.poi.xssf.streaming.SXSSFWorkbook;
+import spark.Response;
+
+public class ExcelParticipantExporter extends TabularParticipantExporter {
+    private final SXSSFWorkbook workbook = new SXSSFWorkbook(200);
```
gmakharat: I would extract magic number `200` just to make easier for others who don't know the details of constructor of `SXSSFWorkbook`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ExcelParticipantExporter.java
```diff
@@ -0,0 +1,75 @@
...
+            Map<String, String> valueMap = participantValueMaps.get(i);
+            List<String> rowValues = getRowValues(valueMap, headerValues);
+            writeRowToSheet(rowValues, i + 2);
+        });
+
+        try (ServletOutputStream servOut = response.raw().getOutputStream()) {
+            workbook.write(servOut);
+        }
+        workbook.dispose();
+        workbook.close();
```
gmakharat, line 50: having workbook inside `try-with-resource` like `try (ServletOutputStream servOut = response.raw().getOutputStream(); workbook)` would not automatically close the workbook?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
gmakharat: It is just factory method I believe, calculations will only start after `export` is called, so in this case I think having the check here is totally fine.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -0,0 +1,128 @@
...
+                                       int activityRepeatNum,
+                                       int questionRepeatNum,
+                                       Map<String, Object> option) {
+        String activityName = filterConfig.getParent().getName();
+        String questionStableId = filterConfig.getColumn().getName();
+        String activityExportName = activityRepeatNum > 1 ?
+                activityName + COLUMN_REPEAT_DELIMITER + activityRepeatNum : activityName;
+        String columnExportName = questionRepeatNum > 1 ?
+                questionStableId + COLUMN_REPEAT_DELIMITER + questionRepeatNum : questionStableId;
+        if (option != null && filterConfig.isSplitOptionsIntoColumns()) {
```
gmakharat: I personally prefer to extract conditions like this as a method with meaningful name for quick understanding, in this case something like: `hasOption` if I got the idea correctly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -0,0 +1,329 @@
...
+ * List configs = p.generateExportConfigs();
+ * List participantMaps = p.parse(listOfParticipantsFromES);
+ */
+public class TabularParticipantParser {
+    private static final Logger logger = LoggerFactory.getLogger(TabularParticipantParser.class);
+    private static final String COLUMN_SELECTED = "1";
+    private static final String COLUMN_UNSELECTED = "0";
+    private static final String DATA_AS_MAP = "dataAsMap";
+    private final List<Filter> filters;
+    private final DDPInstance ddpInstance;
```
gmakharat, line 42: `DDPInstance` is old class I would say, slowly we try to switch to `Dao/Dto` pairs, there is a class called `DDPInstanceDto` I would use this one if it is possible.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -0,0 +1,329 @@
...
+     * this makes a call to ElasticSearch to load question schemas, used to determine whether a given question
+     * is single or multiselect.
+     *
+     * @return the configs
+     */
+    public List<ModuleExportConfig> generateExportConfigs() {
+        List<ModuleExportConfig> configs = new ArrayList<>();
+        // map of table name => module export config
+        Map<String, ModuleExportConfig> exportInfoMap = new HashMap<>();
+        Map<String, Map<String, Object>> activityDefs = ElasticSearchUtil.getActivityDefinitions(ddpInstance);
```
gmakharat, line 65: `ElasticSearchUtil` is an old class as well I would say, there are `ElasticSearchable` interface and `ElasticSearch` concrete implementation of `ElasticSearchable`, which I usually use for elastic search, `ElasticSearchable` may not have the function which is presented in `ElasticSearchUtil`, in this case I add new method to `ElasticSearchable`, if you plan to write unit tests for this class having `ElasticSearchUtil` directly here will struggle you I think.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -0,0 +1,329 @@
...
+            } else if (rootObject instanceof Map) {
+                // it's a Map, like 'dsm'
+                return Collections.singletonList((Map<String, Object>) rootObject);
+            }
+            // we don't know what the module/question will be getting at, so return the entire map
+            // in case the question config has the logic to handle it.
+            return Collections.singletonList(esDataAsMap);
+        }
+    }
+}
```
gmakharat, line 350: Personally I prefer to apply strategy design pattern to such condition separation to have better structure.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
```diff
@@ -0,0 +1,62 @@
...
+        List<String> subHeaderRowValues = getSubHeaderRow();
+
+        writer.println(getRowString(headerRowValues));
+        writer.println(getRowString(subHeaderRowValues));
+        for (Map<String, String> valueMap : participantValueMaps) {
+            List<String> rowValues = getRowValues(valueMap, headerRowValues);
+            String rowString = getRowString(rowValues);
+            writer.println(rowString);
+        }
+        writer.flush();
```
gmakharat: is it chance to have resource leaking after not closing `writer`? Maybe it is better to wrap it in `try-with-resource`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public void setResponseHeaders(Response response) {
+        response.type(MediaType.TSV_UTF_8.toString());
+        response.header("Access-Control-Expose-Headers", "Content-Disposition");
+        response.header("Content-Disposition", "attachment;filename=" + getExportFilename(fileFormat));
```
gmakharat, line 42: I would apply template method design pattern here to eliminate obvious duplication, since last two lines are exactly the same as for `ExcelParticipantExporter`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    protected String getRowString(List<String> rowValues) {
+        return String.join(DELIMITER, rowValues);
+    }
+
+    protected String sanitizeValue(String value) {
+        if (value == null) {
+            value = StringUtils.EMPTY;
+        }
```
gmakharat, line 52: I would apply template method design pattern here as well.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/AmbulationValueProvider.java
devonbush: good question, I had thought this was a standard code.  I wonder if we need a framework for putting study-specific export behavior inside the packages for that study?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-7581-add-feature-flag-export-new.xml
devonbush: good catch

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ExcelParticipantExporter.java
devonbush: good question, I don't think so, since the workbook is instantiated outside of the block.  I'll double-check

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
devonbush: agree that might make sense.  For now I'm just mirroring the structure of the other controllers that use this pattern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
devonbush: ah, that's helpful, I didn't know that ElasticSearchUtil was in the process of being deprecated.  I'll see if I can easily move over

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
devonbush: agree this method got too long.  I'll at least refactor it into separate methods for now, and then if it gets more complicated we can move to strategy

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
devonbush: I think keeping it very explicit and simple here is worth the cost of duplication.  If we add a third or fourth file type, then definitely let's template what we can

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
devonbush: I'm not sure what you mean here--but I'm relatively new to Java string templates.  Can they do replaces and null checks?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
devonbush: better yet, I'll just default to tsv.  No reason to have this hard-error

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ExcelParticipantExporter.java
devonbush: yup, we need these outside the loop.  The alternative would be to create the entire sheet inside the `try`, but I *think* its more efficient to not hold the output stream open for that whole time.  We can revisit if we need to start piecemeal streaming exports

SimoneMaiwald (approved):

PR #1927, opened 2022-07-07
Update hospitalization.conf

MishaK1204:
Decreased text field character range for question: "Center where consultation was performed" from 1000 to 350


dgiorbel (approved):

MocanaAtBroad:
Can you add the JIRA ticket number to Pull request title and make sure that when merging to develop comment includes the jira ticket number? We are now enforcing Squash and Merge, so only one commit will be merged into develop.

PR #1940, opened 2022-07-08
DDP-8168 Osteo Add Participant type source to all PDFs that include mailing address

MocanaAtBroad:
PDF config include a "sources" option that is used by system to figure out what data to collect to fill out the blanks/fields in the PDF template files.
Turns out that if you have a PDF template that want to include the mailing address of the participant, unless you specify in source to include the source type PARTICIPANT, when you try to generate the PDF, it will not have retrieved the mailing address for the participant, will not fill it out, but will generate a PDF with the mailing address left blank.


pepper-apis/studybuilder-cli/studies/osteo/patches/osteo-pdfs.conf
```diff
@@ -14,6 +14,9 @@
               "type": "ACTIVITY",
               "activityCode": "CONSENT",
               "versionTag": "v2"
+            },
+            {
```
zyme, line 18: Oh wow--so the expressions we put into the pdf templates themselves are restricted here, such that a reference in the template can be silently ignored if the appropriate `source` isn't included in this array?

zyme (approved):

pepper-apis/studybuilder-cli/studies/osteo/patches/osteo-pdfs.conf
MocanaAtBroad: There is an error message in the logs that says the mail address was not found. But it is not found because we never tried to look for it in the first place.

PR #1959, opened 2022-07-12
DDP-7581 - moving pt_list_export role script

devonbush:
Moves the liquibase script to the correct place


SimoneMaiwald:
Is that PR still needed? I think I saw that in another one?

SimoneMaiwald:
is the update xml added to the right master-changelog.log? I don't see any changes for that

devonbush:
closing as script is obsolete

PR #1972, opened 2022-07-13
DDP-8337 returning pepper order id to GP

pegahtah:
and removing some unnecessary logs

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderDummyRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderDummyRoute.java
@@ -2,6 +2,7 @@ package org.broadinstitute.dsm.route.mercury;
 
 import com.google.gson.Gson;
 import lombok.extern.slf4j.Slf4j;
+import net.minidev.json.JSONObject;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.dao.ddp.instance.DDPInstanceDao;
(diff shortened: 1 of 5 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderDummyRoute.java
```diff
@@ -55,9 +56,10 @@ public Object handle(Request request, Response response) {
             log.error("Realm was null for " + mercuryOrderRequest.getRealm());
             return new Result(500, UserErrorMessages.CONTACT_DEVELOPER);
         }
-        publishMessage(mercuryOrderRequest, ddpInstance, userId);
-        return new Result(200);
-
+        String pepperOrderId = publishMessage(mercuryOrderRequest, ddpInstance, userId);
+        JSONObject main = new JSONObject();
+        main.put("PepperOrderId", pepperOrderId);
```
devonbush: I'll channel Simone and say "PepperOrderId" should be a defined constant

devonbush (dismissed):
Looks good in general, but please create a string constant for 'PepperOrderId"

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/PostMercuryOrderDummyRoute.java
pegahtah: Thanks fo mentioning that! Fixed

devonbush (approved):

PR #1980, opened 2022-07-14
button symbol add

llobzhanidze:


llobzhanidze:
there is no any ticket about this branch because me and QA guys was checking all issues about font problem in consents and other static pages and fixing them all with this branch, it was a refactoring of the styling and this is the reason why i have no ticket

gskhirel (approved):

PR #1986, opened 2022-07-14
Constrain the assent, consent, and birth dates to the past or present [DDP-8357]

bskinner:
DDP-8357

## Context

This adds additional validations to the payload for the `UserCreationRoute` payload.

When registering a new user:
* The `assentDate` must be either today, or in the past
* The `consentDate` must either be today, or in the past
* The `birthDate` must be in the past

If any of the conditions are not met when calling the `/users` endpoint, the response will use a `422 UNPROCESSABLE ENTITY` status code and fail the request.


MocanaAtBroad (approved):

MocanaAtBroad:
Those annotations are super-nice! What will the client get back if they send a date from the future?

bskinner:
> Those annotations are super-nice! What will the client get back if they send a date from the future?

It'll be something along the lines of:

```json
{
    "code": "BAD_PAYLOAD",
    "message": "Property at 'assentDate' must be a date in the past or in the present"
}
```

PR #2012, opened 2022-07-19
Improving DSM participant export performance (DDP-7581)

devonbush:
This massively improves the performance of export by bypassing our existing DTO/serializers.  This comes at the cost of having to reimplement much of the serialization logic in an ad-hoc way.  

This also collates all responses to REGISTRATION_X_STATE_PROVINCE questions into a single column, whereas before there was a separate column for each country

TO TEST:
(Collation)
1. Go to PanCan study participant list
2. Select all columns
3. select download, and then downoad with default options
4. Confirm that only a single "REGISTRATION_STATE_PRIVONICE" column appears for a given registration completion, and that it contains the appropriate value for the participant
5. do a search for "mailing" and confirm that the mailing address column appears correctly

(Kit requests)
1. Go to TestBoston study participant list
2. select all columns
3. select download, and then downoad with default options
4. confirm that kit requests, and test results appear correctly in the download

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
@@ -15,7 +15,6 @@ public class FilterExportConfig {
     private boolean splitOptionsIntoColumns = false;
     private List<Map<String, Object>> options = null;
     private String collationSuffix = null;
-    private final boolean isCollated = false;
 
     public FilterExportConfig(ModuleExportConfig parent, Filter filterColumn, boolean splitOptionsIntoColumns,
                               List<Map<String, Object>> options, String collationSuffix) {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
@@ -14,14 +14,17 @@ public class FilterExportConfig {
     private final String type;
     private boolean splitOptionsIntoColumns = false;
     private List<Map<String, Object>> options = null;
+    private String collationSuffix = null;
+    private final boolean isCollated = false;
 
     public FilterExportConfig(ModuleExportConfig parent, Filter filterColumn, boolean splitOptionsIntoColumns,
-                              List<Map<String, Object>> options) {
+                              List<Map<String, Object>> options, String collationSuffix) {
         this.column = filterColumn.getParticipantColumn();
         this.type = filterColumn.getType();
         this.parent = parent;
         this.splitOptionsIntoColumns = splitOptionsIntoColumns;
         this.options = options;
+        this.collationSuffix = collationSuffix;
     }
 }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
@@ -75,6 +75,9 @@ public abstract class TabularParticipantExporter {
         String questionStableId = filterConfig.getColumn().getName();
         String activityExportName = activityRepeatNum > 1 ?
                 activityName + COLUMN_REPEAT_DELIMITER + activityRepeatNum : activityName;
+        if (filterConfig.getColumn().getObject() != null) {
+            activityName = activityName + DBConstants.ALIAS_DELIMITER + filterConfig.getColumn().getObject();
+        }
         String columnExportName = questionRepeatNum > 1 ?
                 questionStableId + COLUMN_REPEAT_DELIMITER + questionRepeatNum : questionStableId;
         if (hasSeparateColumnForOption(option, filterConfig)) {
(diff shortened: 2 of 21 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
```diff
@@ -14,14 +14,17 @@ public class FilterExportConfig {
     private final String type;
     private boolean splitOptionsIntoColumns = false;
     private List<Map<String, Object>> options = null;
+    private String collationSuffix = null;
+    private final boolean isCollated = false;
```
SimoneMaiwald: Is that parameter used?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -75,11 +75,17 @@ public static String getColumnName(FilterExportConfig filterConfig,
...
+        if (filterConfig.getColumn().getObject() != null) {
+            activityName = activityName + DBConstants.ALIAS_DELIMITER + filterConfig.getColumn().getObject();
+        }
         String columnExportName = questionRepeatNum > 1 ?
                 questionStableId + COLUMN_REPEAT_DELIMITER + questionRepeatNum : questionStableId;
         if (hasSeparateColumnForOption(option, filterConfig)) {
             columnExportName = columnExportName + DBConstants.ALIAS_DELIMITER + option.get(ESObjectConstants.OPTION_STABLE_ID);
         }
+        if (columnExportName.endsWith("REGISTRATION_STATE_PROVINCE")) {
+            columnExportName = "REGISTRATION_STATE_PROVINCE";
```
SimoneMaiwald, line 87: I wonder if there is a specific question type we could use instead of a name that might only work for one study…

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/CompositeValueProvider.java
```diff
@@ -1,29 +1,6 @@
...
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.stream.Collectors;
-
-import org.apache.commons.lang3.StringUtils;
-
 /** handles questions with lists of lists for answers */
 public class CompositeValueProvider extends TextValueProvider {
```
SimoneMaiwald: do we still need the CompositeValueProvider if it only extends TextValueProvider

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
```diff
@@ -32,6 +34,10 @@ public class BaseFilter {
     protected int from;
     protected int to;
     protected SortBy sortBy;
+    @Setter
+    @Getter
```
SimoneMaiwald, line 38: you could just use `@Data`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -45,7 +45,7 @@ public Object processRequest(Request request, Response response, String userId)
 
         String realm = RoutePath.getRealm(request);
         String userIdReq = UserUtil.getUserId(request);
-        if (!UserUtil.checkUserAccess(realm, userId, "pt_list_export", userIdReq)) {
+        if (!UserUtil.checkUserAccess(realm, userId, "pt_list_view", userIdReq)) {
```
SimoneMaiwald, line 48: was that just changed for testing? or are we planning to remove the separate permission

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
devonbush: Agree.  I didn't see anything obvious in the activity schemas though--I'll ping Marco to see if he has any thoughts

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/CompositeValueProvider.java
devonbush: Good point, will remove

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
devonbush: did not know that, thanks!

devonbush: looks like that annotation would have to be applied across the entire class -- I'm nervous about making that change so close to release.  Maybe we wait until we make the performance improvements to the UI participant list to make that change?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
devonbush: yes, since IIRC correctly the entire reason for the separate permission was FON, we can stick to the existing one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
devonbush: nope, good catch

SimoneMaiwald (approved):

PR #2022, opened 2022-07-21
[DSS] DDP-8413: Configuration file from secrets manager

Trimax:
DDP-8413

This PR contains the functionality that allows the application to read the configuration file from the Google Secrets Manager service. In case the configuration file is explicitly specified via CL it will be used.

Open questions:

- Where to get the **project name** and **secret name**?


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
```diff
@@ -68,7 +69,11 @@ public static synchronized ConfigManager getInstance() {
      * please use {@link #getInstance()}
      */
     public static Config parseConfig() {
-        return ConfigFactory.parseFile(TYPESAFE_CONFIG_FILE);
+        if (TYPESAFE_CONFIG_FILE != null) {
+            return ConfigFactory.parseFile(TYPESAFE_CONFIG_FILE);
+        }
+
+        return ConfigFactory.parseString(SecretManager.get("542678289221", "dss-configuration").orElseThrow());
```
Trimax: Obviously, we don't want to hardcode the project's and the secret's name. Do we want to have them as environment variables or CL arguments?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/secrets/SecretManager.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static SecretVersion getLatestSecretVersion(final String projectName, final String secretName) throws Exception {
+        try (final var client = SecretManagerServiceClient.create()) {
+            return StreamSupport.stream(
+                            Spliterators.spliteratorUnknownSize(
+                                    client.listSecretVersions(
+                                            SecretName.of(projectName, secretName)).iterateAll().iterator(), 0), false)
+                    .max(SecretManager::compare)
+                    .orElseThrow();
+        }
```
bskinner: This can be removed- `latest` is the shorthand for the most recent secret version.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
bskinner: Let's pass the secret name, the project id, and the version identifier via system properties.

How about `google.secret.project`, `google.secret.name`, and `google.secret.version`? `google.secret.version` should be optional and, if not set, should default to the literal string `latest`.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/secrets/SecretManager.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private static int compare(final SecretVersion a, final SecretVersion b) {
+        return Long.compare(a.getCreateTime().getSeconds(), b.getCreateTime().getSeconds());
+    }
```
bskinner: The comparison isn't necessary. We should either be using `latest`, or an explicitly specified version identifier.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/secrets/SecretManager.java
```diff
@@ -0,0 +1,48 @@
...
+                    .map(SecretPayload::getData)
+                    .map(ByteString::toStringUtf8);
+        } catch (final Exception e) {
+            log.warn("The secret " + secretName + " wasn't found", e);
+        }
+
+        return Optional.empty();
+    }
+
+    private static SecretVersion getLatestSecretVersion(final String projectName, final String secretName) throws Exception {
```
bskinner: Let's try creating a `SecretVersionName` directly using `SecretVersionName.of(String, String, String)` instead of passing things off to this method.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
bskinner: It's also a bit more unwieldy and unusual in our system to use the project's number. The `SecretVersionName` should permit use of the project id (`broad-ddp-dev`, in this case)

bskinner (commented):
Glad to see this is working!

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/secrets/SecretManager.java
Trimax: Oh, that's great. I googled and played with the library because I wouldn't say I liked this piece of the code. Thank you.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
Trimax: SGTM

bskinner (approved):
This looks good to me!

PR #2026, opened 2022-07-21
[DSM] [DRAFT] DDP-8113_invitae

pegahtah:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/InvitaeReportDao.java
```diff
@@ -0,0 +1,101 @@
...
+public class InvitaeReportDao {
+    private static final String SQL_INSERT_NEW_INVITAE =
+            "INSERT INTO invitae_report (participant_id, last_changed, changed_by) values (?, ?, ?) ";
+
+    private static final String SQL_GET_INVITAE_FOR_REALM = "SELECT * from invitae_report inv "
+            + " left join ddp_participant p on (p.participant_id = inv.participant_id) "
+            + " left join ddp_instance realm on (realm.ddp_instance_id = p.ddp_instance_id) "
+            + " where realm.instance_name = ? ";
+
+    public static String createNewInvitaeReport(String participantId, String user) {
```
gmakharat, line 28: let's remove static from here to follow OOP principles.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/generate/PropertyInfo.java
```diff
@@ -21,16 +22,19 @@
...
+                    Map.entry(DBConstants.DDP_TISSUE_ALIAS, new PropertyInfo(Tissue.class, true)),
+                    Map.entry(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS, new PropertyInfo(OncHistoryDetail.class, true)),
+                    Map.entry(DBConstants.DDP_PARTICIPANT_DATA_ALIAS, new PropertyInfo(ParticipantData.class, true)),
+                    Map.entry(DBConstants.DDP_PARTICIPANT_RECORD_ALIAS, new PropertyInfo(Participant.class, false)),
+                    Map.entry(DBConstants.DDP_PARTICIPANT_ALIAS, new PropertyInfo(Participant.class, false)),
+                    Map.entry(DBConstants.DDP_ONC_HISTORY_ALIAS, new PropertyInfo(OncHistory.class, false)),
+                    Map.entry(DBConstants.SM_ID_TABLE_ALIAS, new PropertyInfo(SmId.class, true)),
+                    Map.entry(DBConstants.COHORT_ALIAS, new PropertyInfo(CohortTag.class, true)),
+                    Map.entry(DBConstants.DDP_KIT_REQUEST_ALIAS, new PropertyInfo(KitRequestShipping.class, true)),
+                    Map.entry(DBConstants.INVITAE_ALIAS, new PropertyInfo(InvitaeReport.class, false))));
```
gmakharat, line 36: as I see from mapping `InvitaeReport` is type of nested, in this case I believe we need to pass true instead of false here to indicate that it is collection, collection property in this class actually means `nested` if we talk about elastic search.

gmakharat: I think this PR needs to be updated from `develop`, because there is an enum class called `Alias` where you need to add alias for `invitae` as well.

SimoneMaiwald:
why was that tab not added per dynamic fields?

PR #2027, opened 2022-07-22
[DSS] DDP-8428: Change uploaded file name format

Trimax:
So, now instead of having a structure like this:
```
study GUID
  |-- upload date
        |-- user GUID
              |-- activity instance code
                    |-- original file name <-- file
```
 
We will have the following:
```
study GUID
  |-- activity instance code
        |-- user GUID + upload time + original file name <-- file
```


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/CreateUserActivityUploadRoute.java
```diff
@@ -148,7 +147,7 @@ private long bytesToMbs(long maxFileSize) {
         return maxFileSize / MB_IN_BYTES;
     }
 
-    private String getCurrentDate() {
-        return new SimpleDateFormat("yyyy_MM_dd").format(new Date());
+    private static String getCurrentTimestamp() {
+        return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss").format(new Date());
```
MocanaAtBroad, line 151: You know what: let's add milliseconds to just virtually eliminate chance of name collision. They will be able to sort stuff like they want in the same way.

Trimax: Let's use nanoseconds then :)

MocanaAtBroad (approved):

ssettipalli:
Looks like the requirement is to update the folder structure and name of the uploaded files what the staff sees.
Files are initially uploaded to  *-file-uploads, then scanned and the files are moved to *file-scanned bucket.
I believe staff has direct access to *file-scanned bucket and not *-file-uploads .
Changes in the PR fixes filename format but want to make sure files are moved to file-scanned bucket as per updated file structure. 
Blake has lot more understanding of file scanning. 
cc @bskinner

PR #2055, opened 2022-07-25
[DSM] DDP-6584 downloading participant files

pegahtah:
An endpoint `DownloadParticipantFileRout`  is added to handle the downloading requests. 
The user should have `file_download` access role to be able to do the download
ElasticSearchUtil has also been changed so users are able to search for files and sort on the time


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
```diff
@@ -0,0 +1,73 @@
...
+        QueryParamsMap queryParams = request.queryMap();
+        String realm = null;
+        if (queryParams.value(RoutePath.REALM) != null) {
+            realm = queryParams.get(RoutePath.REALM).value();
+        }
+        if (StringUtils.isBlank(realm)) {
+            response.status(500);
+            throw new RuntimeException("Realm was missing from the request");
+        }
+        String userIdR = UserUtil.getUserId(request);
```
gmakharat: I think we already have `userId` in the arguments of `processRequest`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
```diff
@@ -0,0 +1,73 @@
...
+            String objectName = queryParams.value("fileName");
+            String blobName = queryParams.value("blob");
+            Optional<byte[]> downloadObject =
+                    downloadObject(googleProjectName, bucketName, objectName, blobName);
+            byte[] downloadObjectArray = downloadObject.orElseThrow();
+            HttpServletResponse rawResponse = response.raw();
+            rawResponse.getOutputStream().write(downloadObjectArray);
+            rawResponse.setStatus(200);
+            rawResponse.getOutputStream().flush();
+            rawResponse.getOutputStream().close();
```
gmakharat: why code from line 50 to line 54 is necessary? Just returning `downloadObjectArray` is not enough?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
```diff
@@ -0,0 +1,73 @@
...
+            return new Result(200);
```
gmakharat: As we agreed, step by step we will try to remove returning `Result` object from routes, let's just return empty string if route is void.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+            return new Result(401, SecurityUtil.ResultType.AUTHENTICATION_ERROR.toString());
```
gmakharat: response.status(401);
return SecurityUtil.ResultType.AUTHENTICATION_ERROR.toString();

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
pegahtah: We check that these two match in `UserUtil.checkUserAccess`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
pegahtah: True, I will fix that, thanks for the reminder!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantFiles/DownloadParticipantFileRoute.java
pegahtah: It is enough. fixed

gmakharat (approved):

PR #2061, opened 2022-07-26
[DSS] DDP-8428: Uploaded file name change

Trimax:
The fix that changes blob name (instead of random GUID we now use file name constructed according to the requested template)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -154,11 +154,10 @@ public AuthorizeResult authorizeUpload(Handle handle, long studyId, long operato
         mimeType = mimeType != null ? mimeType : DEFAULT_MIME_TYPE;
 
         HttpMethod method = resumable ? HttpMethod.POST : HttpMethod.PUT;
-        String uploadGuid = GuidUtils.randomFileUploadGuid();
-        String blobName = blobPrefix + uploadGuid;
+        final var blobName = blobPrefix + fileName;
```
Trimax, line 157: This one is essential. Instead of using a random id, we're now using a file name that is constructed according to the discussed template in terms of the previous PR.

MocanaAtBroad (approved):

PR #2062, opened 2022-07-26
[DSS] DDP-8248: Uploaded filename fix

Trimax:
Two changes:
1. We need to keep the original file name for the client.
2. We would like to keep all files in our bucket not several.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/CreateUserActivityUploadRoute.java
```diff
@@ -117,9 +117,10 @@ public Object handle(Request request, Response response, CreateUserActivityUploa
                     operatorUser.getId(),
                     instanceDto.getParticipantId(),
                     fileQuestionDef,
-                    String.format("%s/%s", studyGuid, instanceDto.getActivityCode()),
+                    String.format("%s/%s_%s_%s_%s_%d", studyGuid, instanceDto.getActivityCode(), userGuid,
```
Trimax, line 120: This one changes ddp-upload/STUDY_NAME/ACTIVITY_NAME/FILE_NAME to ddp-upload/STUDY_NAME/ACTIVITY_NAME_FILE_NAME

bskinner: What are your thoughts on moving this to a private `getBlobName()` method call? Using the String.format call here feels unwieldy and breaks context for me.

bskinner (approved):
Looks good- just had suggestion for the `String.format` parameter in the `authorizeUpload()` call.

PR #2064, opened 2022-07-26
[DSS] DDP-8428: Method extraction (clean code)

Trimax:
Just extracted the blob path formatting to the separate method


bskinner (approved):
Thank you!

PR #2083, opened 2022-07-28
DDP-8428: File upload fix

Trimax:
Since the filename format has changed (previously was GUID and currently a combination of different values) the housekeeping tool needs to obtain GUID from it instead of just using it.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/FileScanResultReceiver.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/FileScanResultReceiver.java
@@ -98,9 +98,10 @@ public class FileScanResultReceiver implements MessageReceiver {
         }
     }
 
-    private String parseFileUploadGuid(String fileName) {
+    private String parseFileUploadGuid(String blobName) {
         // For authorized uploads, the base file name should start with the upload guid.
-        return Path.of(fileName).getFileName().toString().substring(0, fileName.indexOf("_"));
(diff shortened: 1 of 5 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/FileScanResultReceiver.java
```diff
@@ -98,17 +98,24 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer reply) {
...
 
-    private String parseFileUploadGuid(String fileName) {
-        // For authorized uploads, the base file name should be the upload guid.
-        return Path.of(fileName).getFileName().toString();
+    private String parseFileUploadGuid(String blobName) {
+        // For authorized uploads, the base file name should start with the upload guid.
+        final var fileName = Path.of(blobName).getFileName().toString();
+        if (!fileName.contains("_")) {
+            log.error("The blob name {} doesn't have any underscores in it. It must have at least one", blobName);
+            return null;
```
bskinner: What are your thoughts on throwing a `DDPException` here? This should likely be considered a fatal error at this point, and the `null` going into the `findAndLockByGuid()` method is going to result in a SQL error instead of a clear indicator as to what's going on.

bskinner (dismissed):
A minor comment regarding error handling in the case of a malformed filename, but looks good to me otherwise.

bskinner (dismissed):

bskinner (approved):

PR #2087, opened 2022-07-31
DSP-7581 export fixes for AT

devonbush:
This improves handling of composite questions, and multiple choice questions that allow specifying option details.  Now all values for those are given separate columns.  Likewise, if a participant gives multiple answers to a question (such as listing medications), each answer is given its own column (e.g. MEDICATIONS_1.NAME, MEDICATIONS_2.NAME, etc...)

TO TEST:
1.  Go to the AT study participant list in development
2.  Use the "Select all" button to select all columns
3.  Press the download button, and download using the default confiugration
4. For participant PKG8PA, confirm that the columns `MEDICAL_HISTORY.MEDICATION_CATEGORY.MEDICATION_NAME`, `MEDICAL_HISTORY.MEDICATION_CATEGORY.BEGAN_TAKING_AT_AGE` are filled out,  as are `MEDICAL_HISTORY.MEDICATION_CATEGORY_2.MEDICATION_NAME`, `MEDICAL_HISTORY.MEDICATION_CATEGORY_2.BEGAN_TAKING_AT_AGE`
5. For participant PKJD7H, confirm the column `MEDICAL_HISTORY_2.DETERMINED_OTHER_DETAILS` has a value of "Other diagnosis test"
6. generally confirm that the export has the correct number of participants, columns look correct, etc..


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
```diff
@@ -2,28 +2,114 @@
...
+        return false;
+    }
+
+    private List<Map<String, Object>> getOptionsForQuestion(Map<String, Object> questionDef) {
+        List<Map<String,Object>> options = (List<Map<String, Object>>) questionDef.get(ESObjectConstants.OPTIONS);
+        if (questionDef.containsKey(ESObjectConstants.OPTION_GROUPS)) {
+            Object groups = questionDef.get(ESObjectConstants.OPTION_GROUPS);
+            if (groups instanceof List) {
+                for (Map<String, Object> group : (List<Map<String, Object>>) groups) {
+                    if (group.containsKey(ESObjectConstants.OPTIONS) ) {
```
gmakharat, line 98: I personally prefer not to use nested if statements, since it makes code hard to maintain and hard to read, in general I replace nested if statements with guard conditions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
```diff
@@ -2,28 +2,114 @@
...
+                        options.addAll((List<Map<String, Object>>) group.get(ESObjectConstants.OPTIONS));
+                    }
+                }
+            }
+        }
+        return options;
+    }
+
+    public List<Map<String, Object>> getChildQuestions() {
+        return questionDef != null ? (List<Map<String, Object>>) questionDef.get(ESObjectConstants.CHILD_QUESTIONS) : null;
```
gmakharat, line 108: I think it is better technique to return empty list rather than `null` from the method which has to return list, it will help to avoid ugly if check in the place where this method is called.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,29 +50,74 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
 
+    /**
+     * Gets all the column names that need to be rendered for a given config, including repeats, details, and/or
+     *  splitting options across multiple column.  The column name is based off the stable name of the associated item
+     *  and needs to be globally unique
+     */
+
     public static List<String> getAllColumnNames(FilterExportConfig filterConfig, int formRepeatNum, int questionRepeatNum) {
+        List<String> allColNames = new ArrayList<>();
+        List<String> optionStableIds = Collections.singletonList(null);
```
gmakharat: `Collections.emptyList()` is better alternative maybe?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,29 +50,74 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
+        if (filterConfig.isSplitOptionsIntoColumns()) {
+            optionStableIds = filterConfig.getOptions().stream().map(opt -> (String) opt.get(ESObjectConstants.OPTION_STABLE_ID))
+                    .collect(Collectors.toList());
+        }
+        allColNames = optionStableIds.stream().map(opt -> {
+                    List<String> colNames = new ArrayList<>();
```
gmakharat, line 75: I think this `map` lambda function deservers to have its own method with meaningful name for quick understanding in the future.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,29 +50,74 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
+     */
+
     public static List<String> getAllColumnNames(FilterExportConfig filterConfig, int formRepeatNum, int questionRepeatNum) {
+        List<String> allColNames = new ArrayList<>();
+        List<String> optionStableIds = Collections.singletonList(null);
+        if (filterConfig.isSplitOptionsIntoColumns()) {
+            optionStableIds = filterConfig.getOptions().stream().map(opt -> (String) opt.get(ESObjectConstants.OPTION_STABLE_ID))
+                    .collect(Collectors.toList());
+        }
+        allColNames = optionStableIds.stream().map(opt -> {
```
gmakharat: I think this `map` lambda function deservers to have its own method with meaningful name for quick understanding in the future.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,29 +50,74 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
+    }
+
+
+    /**
+     * Gets all the column texts to be rendered.  This is a human-readable counterpart to the column name.  e.g.
+     *  while the column name might be "MEDICIAL_HISTORY_2.ANALYSIS_TYPE" the text might be "Type of sample analysis"
+     */
+    public List<String> getAllColumnTexts(FilterExportConfig filterConfig) {
+        List<String> allColTexts = new ArrayList<>();
+        List<Map<String, Object>> splitOptions = Collections.singletonList(null);
```
gmakharat: the same about `Collections.emtpyList();`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,29 +50,74 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
+            splitOptions = filterConfig.getOptions();
         }
-        return Collections.singletonList(getColumnName(
-                filterConfig,
-                formRepeatNum,
-                questionRepeatNum,
-                null
-        ));
+        allColTexts = splitOptions.stream().map(opt -> {
+            List<String> colTexts = new ArrayList<>();
```
gmakharat: the same about method with meaningful name for `lambda`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -81,75 +134,100 @@ public static List<String> getAllColumnNames(FilterExportConfig filterConfig, in
...
     }
 
+    /** gets the header row -- this iterates through all the configs and calls "getAllColumnNames" for each */
     protected List<String> getHeaderRow() {
         List<String> headers = new ArrayList<>();
         for (ModuleExportConfig moduleConfig : moduleConfigs) {
             for (int formRepeatNum = 1; formRepeatNum <= moduleConfig.getNumMaxRepeats(); formRepeatNum++) {
                 for (FilterExportConfig filterConfig : moduleConfig.getQuestions()) {
-                    headers.addAll(getAllColumnNames(filterConfig, formRepeatNum, 1));
+                    for (int questionRepeatNum = 1; questionRepeatNum <= filterConfig.getMaxRepeats(); questionRepeatNum++) {
```
gmakharat, line 205: Just wondering, having such many nested for loops how badly will affect the performance?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -152,11 +144,16 @@ private Map<String, Object> getDefForQuestion(ParticipantColumn column, Map<Stri
...
+        // find the question with the matching stableId, and get the index
+        List<Map<String, Object>> questionList = ((List<Map<String, Object>>) activityDef.get(ESObjectConstants.QUESTIONS));
+        for (int i = 0; i < questionList.size(); i++) {
+            Map<String, Object> question = questionList.get(i);
+            if (question.get(ESObjectConstants.STABLE_ID).equals((column.getName()))) {
+                question.put("index", i);
+                return question;
+            }
+        }
+        return null;
```
gmakharat, line 156: What do you think about having `Optional` pattern to avoid null returns and null checks? Method which has `List` type return value to return empty list and method which returns not `List` type value we return `Optional` type value, to think about this pattern and follow in general?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/CompositeValueProvider.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * value provider for COMPOSITE-type questions.  This supports providing a separate string value for each
+ * child question, and delegating each child question to a different appropriate value provider.
+ */
+public class CompositeValueProvider extends TextValueProvider {
+    public List<String> formatRawValues(List<?> rawValues, FilterExportConfig filterConfig, Map<String, Object> formMap) {
+        if (CollectionUtils.isNotEmpty(filterConfig.getChildConfigs())) {
+            List<String> answerVals = new ArrayList<>();
+            if (rawValues != null && rawValues.size() == 1 && rawValues.get(0) instanceof List) {
```
gmakharat: I prefer to have method for this condition with good name rather than having comment, because usually comments become out of date very easily and developers are not motivated to update comments.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -8,38 +8,89 @@
...
+        Map<String, Object> answerObject = getRelevantAnswer(moduleMap, filterConfig);
+        if (answerObject != null) {
+            Object optionDetails = answerObject.get(ESObjectConstants.OPTIONDETAILS);
+            // Option details are stored in ES as an array of maps
+            // e.g. [{option: "TELANGIECTASIA_EYES", details: "12"}, {option: "TELANGIECTASIA_SKIN", details: "34"}]
+            if (optionDetails instanceof List) {
+                return ((List<?>) optionDetails).stream().filter(detail ->
+                                StringUtils.equals((String) ((Map) detail).get(ESObjectConstants.OPTION), optionStableId))
+                        .map(detail -> (String) ((Map<String, Object>) detail).get(ESObjectConstants.DETAILS))
+                        .findAny().orElse(StringUtils.EMPTY);
```
gmakharat: I would extract this piece of code in method with name `extractOptionDetails` which is way more better in my opinion than having comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
devonbush: yeah, `for` loops are AFAIK the fastest way to iterate, and since we have to output everything, there's no escaping the loops here.  Fortunately, we only execute this method once for the entire export operation, so it's not a bottleneck.  Running the profiler on the export operation shows that the large majority of the time spent is in navigating the structures returned by elasticSearch, which happens before this method is invoked

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
devonbush: To be honest, I'm not fluent in the optional pattern.  My sense here is that it won't save us on the logic checks, since we need totally different behavior if the question doesn't exist.  But again, not sure

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
devonbush: This is a bit unintuitive, but I actually want one iteration even if there are no options.  I'll see if I can refactor the code around it -- you're right that it's not clear at all

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
devonbush: same, will see if I can refactor the surroundings

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
devonbush: I can see that -- for now I'm mimicking the structure of the data returned from ElasticSearch, so leaving `null` where there are no childQuestions

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
devonbush: can you give me an example of how you'd rewrite this?  I agree that I'd like it to be cleaner, but I'm not sure I know the best way in Java to do it

gmakharat: `if (!questionDef.containsKey(ESObjectConstants.OPTION_GROUPS) || !(questionDef.get(ESObjectConstants.OPTION_GROUPS) instanceof List)) return options`

I would extract boolean expression as a method for better clarification:
`if (hasOptionGroups(questionDef)) return options;`

then for loop would go as before

PR #2088, opened 2022-08-01
DDP-8504: Configuration files merge

Trimax:
This PR introduces functionality that allows us to specify either cloud secret name or local configuration file or even both.
In case both are specified, two configurations will be merged with the priority for the values taken from the local configuration file.

For example,
```
  configCloud = {"a":"a", "common":"a"}
  configLocal = {"b":"b", "common":"b"}

  configMerged = {"a":"a", "b":"b", "common":"b"}
```


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
```diff
@@ -74,29 +77,49 @@ public static synchronized ConfigManager getInstance() {
...
+
+        Optional<Config> configLocal = Optional.empty();
+        if (isLocalConfigurationFileSpecified()) {
+            log.info("The config file name was specified. Trying to load the configuration from the local file");
+            configLocal = Optional.of(ConfigFactory.parseFile(TYPESAFE_CONFIG_FILE));
+        }
+
+        if (configCloud.isEmpty() && configLocal.isEmpty()) {
+            log.error("Can't load configuration. Neither local config not cloud one was specified");
+            throw new DDPException("Can't load configuration. Neither local config not cloud one was specified");
```
bskinner: Suggested update to the message to tell the user exactly what needs to be fixed.

```suggestion
        if (configCloud.isEmpty() && configLocal.isEmpty()) {
            log.error("no configuration was specified. Use properties '{}' to use a local file or '{}' and '{}' to use a cloud secret",
                    TYPESAFE_CONFIG_SYSTEM_VAR,
                    GOOGLE_SECRET_PROJECT,
                    GOOGLE_SECRET_NAME);
            throw new DDPException("no configuration was specified.");
```

bskinner (dismissed):
A small suggested change, but this looks good to me

bskinner (approved):
👍

PR #2095, opened 2022-08-02
DSP-7581: Merging RC export work back into develop

devonbush:
This merges all of the previously approved changes on the RC branch relating to export work.  It also includes an update to correctly handle file uploads in the export

TO TEST:
1. go to participant list for Singular
2. click download, and then download with the default options
3. confirm download does not error
4. review downloaded spreadsheet, confirm that the field "singularNotes" is populated for the participant PPW7EY with the text "enrolled at last"


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantExporter.java
```diff
@@ -42,90 +49,197 @@ public static TabularParticipantExporter getExporter(List<ModuleExportConfig> mo
...
+        return colTexts;
+    }
+
+
+
+    /** some tables are stored off the same data object as another (e.g. proxies are stored off of "profile")
+     * this map lets us add distinguishing names to columns from those objects.  e.g. profile.email will not
+     * collide with profile.PROXY.email
+     */
+    private static final Map<String, String> TABLE_ALIAS_NAME_MAP = Map.of(
```
gmakharat, line 139: constant fields are recommended to be declared on top of the class

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -1,101 +1,224 @@
...
+                    return StringUtils.EMPTY;
+                }).collect(Collectors.toList());
+                answerMap.put(fieldName, answersArray);
+                targetMap = answerMap;
+
+            } else if (formObject instanceof Map) {
+                targetMap = (Map<String, Object>) formObject;
+            }
         }
+        if ((targetMap == null && objectName != null) || ESObjectConstants.ADDITIONAL_VALUE.equals(filterConfig.getType())) {
```
gmakharat, line 125: having method with name `isDynamicFields` is more straightforward than having comment in my opinion

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -1,101 +1,224 @@
...
 
-        return answer;
+                if (dynamicFieldMap.get(camelCasedFieldName) != null) {
+                    targetMap = dynamicFieldMap;
+                    fieldName = camelCasedFieldName;
+                }
+            }
+        }
+        // if this isn't any of the special cases above, just use the passed-in map
+        if (targetMap == null) {
```
gmakharat, line 144: I would introduce command design pattern here with its factory class, this would help to separate logics in its own classes, like `ListValueHandler`, `DynamicFieldValueHandler`, `BasicValueHandler`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -1,101 +1,224 @@
...
+        if (moduleMap != null) {
+            List<Map<String, Object>> allAnswers =
+                    (List<Map<String, Object>>) moduleMap.get(ElasticSearchUtil.QUESTIONS_ANSWER);
+            if (allAnswers != null) {
+                return allAnswers.stream()
+                        .filter(ans -> filterConfig.getColumn().getName().equals(ans.get(ESObjectConstants.STABLE_ID)))
+                        .findAny().orElse(null);
+            }
+        }
+        return null;
```
gmakharat, line 176: I think introducing null object design pattern for this exporting feature could be useful.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -1,101 +1,224 @@
...
+            // flatten any nested lists
+            for (Object item : objList) {
+                if (item instanceof Collection) {
+                    allValues.addAll(flatten((Collection) item));
+                } else {
+                    allValues.add(item);
+                }
+            }
+            // replace any nulls with empty string
+            return allValues.stream().map(val -> val == null ? StringUtils.EMPTY : val).collect(Collectors.toList());
```
gmakharat, line 205: having method with name `replaceNullsWithEmptyString` which would wrap this line would be way more cleaner than having comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
```diff
@@ -1,101 +1,224 @@
...
         }
     }
+
+    /** flatten an arbitrarily nested collection */
+    private Collection<?> flatten(Collection<?> collection) {
+        List<Object> allValues = new ArrayList<>();
+
+        for (Object item : collection) {
+            if (item instanceof Collection) {
+                allValues.addAll(flatten((Collection) item));
```
gmakharat, line 217: is it safe to have recursion without base case in the method?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
```diff
@@ -97,14 +97,20 @@ public Optional<ElasticSearchParticipantDto> parseSourceMap(Map<String, Object>
         return deserializedSourceMap.isPresent() ? deserializedSourceMap : Optional.of(new ElasticSearchParticipantDto.Builder().build());
     }
 
-    public List<ElasticSearchParticipantDto> parseSourceMaps(SearchHit[] searchHits) {
+    public List<ElasticSearchParticipantDto> parseSourceMaps(SearchHit[] searchHits, boolean parseToParticipantDto) {
```
gmakharat, line 100: I am still against of this extra parameter, this is very bad technique of code smells, automatically means that method does more than one job

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
```diff
@@ -277,7 +284,7 @@ public ElasticSearch getAllParticipantsDataByInstanceIndex(String esParticipants
         } catch (Exception e) {
             throw new RuntimeException("Couldn't get participants from ES for instance " + esParticipantsIndex, e);
         }
-        List<ElasticSearchParticipantDto> elasticSearchParticipantDtos = parseSourceMaps(searchResponse.getHits().getHits());
+        List<ElasticSearchParticipantDto> elasticSearchParticipantDtos = parseSourceMaps(searchResponse.getHits().getHits(), true);
```
gmakharat, line 287: because of this extra parameter it forced to add hierarchically everywhere

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -31,11 +33,20 @@ public class ElasticSearchParticipantDto {
     private List<Map<String, Object>> workflows;
     private String status;
     private ESDsm dsm;
+    @Getter
     private String ddp;
 
+    @Getter
+    private Map<String, Object> dataAsMap;
```
gmakharat, line 40: why it became necessary to add this property? I believe `ElasticSearchParticipantDto` already holds extracted data from source map.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -86,8 +89,10 @@ public static Alias of(SortBy sortBy) {
 
     public static Alias of(ParticipantColumn column) {
         Alias esAlias;
-        if (Objects.nonNull(column.getObject())) {
+        if (Objects.nonNull(column.getObject()) && Alias.ofOrNull(column.getObject()) != null) {
             esAlias = Alias.of(column.getObject());
+        } else if (ElasticSearchUtil.QUESTIONS_ANSWER.equals(column.getObject())) {
```
gmakharat, line 94: is this else if really necessary? else block calls `Alias.of` which automatically returns `ACTIVITIES` alias if there is no such enum

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
```diff
@@ -32,6 +34,10 @@ public class BaseFilter {
     protected int from;
     protected int to;
     protected SortBy sortBy;
+    @Setter
+    @Getter
+    // whether the filter method should handle parsing returned objects into DTO
+    private boolean parseDtos = true;
```
gmakharat, line 40: the same concern about additional boolean property, did not get why it became necessary not to parse?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -59,7 +61,7 @@ public ParticipantWrapper(ParticipantWrapperPayload participantWrapperPayload, E
         });
     }
 
-    public ParticipantWrapperResult getFilteredList() {
+    public ParticipantWrapperResult getFilteredList(boolean parseParticipantDtos) {
```
gmakharat, line 64: ugly extra parameter everywhere

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -189,13 +211,27 @@ void fillParticipantWrapperDtosWithProxies(List<ParticipantWrapperDto> result, L
...
+
+            if (parseParticipantDtos) {
+                proxyEsData = proxiesByIds.getEsParticipants().stream()
+                        .filter(elasticSearchParticipantDto -> participantProxyGuids.contains(elasticSearchParticipantDto.getParticipantId()))
+                        .collect(Collectors.toList());
+                participantWrapperDto.setProxyData(proxyEsData);
+            } else {
+                proxyEsData = proxiesByIds.getEsParticipants().stream()
+                        .filter(elasticSearchParticipantDto -> {
+                            Map<String, Object> profile = (Map<String, Object>) elasticSearchParticipantDto.getDataAsMap().get(ESObjectConstants.PROFILE);
```
gmakharat, line 227: what is the idea behind of filtering by `profile`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -39,21 +39,24 @@ public class DownloadParticipantListRoute extends RequestHandler {
...
         DownloadParticipantListPayload payload =
                 ObjectMapperSingleton.instance().readValue(request.body(), DownloadParticipantListPayload.class);
         DownloadParticipantListParams params = new DownloadParticipantListParams(request.queryMap());
 
         String realm = RoutePath.getRealm(request);
+
+        String userIdReq = UserUtil.getUserId(request);
+        if (!UserUtil.checkUserAccess(realm, userId, "pt_list_view", userIdReq)) {
+            response.status(500);
+            return new Result(500, UserErrorMessages.NO_RIGHTS);
```
gmakharat, line 51: slowly we are trying to remove technique of returning object of `Result`, just returning setting `response.status` and return `UserErrorMessages.NO_RIGHTS` is pretty enough

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
devonbush: the base case is the other half of this `if` statement -- if the thing isn't a collection, we just pass it right back

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
devonbush: I don't know if this is complex enough yet for that, but I agree the method has gotten too long -- will pull out submethods as you suggest

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
devonbush: This will also benefit from the Objects::nonNull you pointed me to

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
devonbush: correct, it holds the *extracted* data.  But I need a place to store the raw data, so that I can skip the extraction step.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
devonbush: I think this was necessary to guard against the edge case where a tableAlias might be duplicative of an activity name.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
devonbush: Parsing took too long -- we only have 10 minutes to return data before AppEngine will time requests out.   Running profiling on the export code (when it was taking almost 5 minutes just to export 100 participants) showed that the vast majority of the time was taken in transforming the ES result to our Java object, and then parsing bits and pieces back to maps as we were reading unstructured Map<String, Object>s.  So I eliminated the back-and-forth parsing

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
devonbush: profile is where we find the participants GUID

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
devonbush: gotcha, thanks for the heads up

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/renderer/TextValueProvider.java
gmakharat: in my opinion it is not good idea to wait it may become complex or not, having several if statements to make different things already means that it is enough complex to separate them into different classes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
gmakharat: could you please provide more details why it became necessary to keep raw data separately?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
gmakharat: if something is passed to this method which does not present in `Alias` enum class it will return `ACTIVITIES` by default and `QUESTIONS_ANSWER` does not exist as enum, that's why I got this question

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
gmakharat: was it really parsing time? we use `jackson` library mainly to convert into Java object, I understand your point but to sacrifice the design because of performance and make code messy I don't like this idea, maybe introduce child class of BaseFilter which will be focused on what you need instead of putting it directly here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
gmakharat: that's my main concern of having raw data separately in `ElasticSearchParticipantDto`, it forces to make many casting and verbose code.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/BaseFilter.java
devonbush: Totally agree -- everything would be better if we didn't have to worry about the performance.  :).  I've created an Unparsed subclass as you suggest to better isolate this code

devonbush:
obseleted by later update

PR #2110, opened 2022-08-03
[DSM] DDP-7855 | Refactoring 

gmakharat:
## Context

While ago when I and @Anzori-Ghurtchumelia did huge refactoring we kept already existing filtering functionality for `ES` alias, currently in DSM we have several aliases like: `t`(tissue), `oD`(onc history) and etc, when filter request is sent from frontend side it comes with alias, if we are filtering for tissue related fields we get `t` alias as payload and etc, but if we filter by DSS related fields like answer of question in activity and etc we get alias `ES`, as it turned out old filtering functionality for `ES` alias had several bugs, thus it forced to make refactoring and re-use the same code as it is used for so called "under dsm object" aliases.

**Thins have done:**

- Removed strict date formats from `participants_structured.json`, since it was making harder to do normal filtering, it was requiring exact date time in milliseconds.
- Created classes for `Invitations`, `Files` and etc to represent how they are saved in ES.
- Created `CamelCaseConverter` class which implements `Converter` interface, class encapsulates all necessary details for converting PASCAL_CASE to camelCase, created `NullObjectCamelCaseConverter` class as well, which follows null object design pattern, since in some cases we don't need to convert PASCAL_CASE to camelCase.
- Add more properties to `PropertyInfo` class to match necessary aliases.
- Created new filter separator called `NonDsmFilterSeparator`, which handles other aliases like `profile`, `invitations` and etc.
- Separated `BaseAbstractQueryBuilder` and `DsmAbstractQueryBuilder` to build different path.
- Created `BaseActivitiesStrategy` and its two children: `ActivityStrategy` and `QuestionsAnswersStrategy`, since filtering by questions answers needs different strategy.
- Created `CompositeNestedQueryStrategy` class, which follows composite design pattern to handle filtering by several criteria and they belong to the same nested object.
- Add corresponding unit tests.
- Did refactoring.

## Checklist

## FUD Score

## Testing

## Release


devonbush (commented):
Let's set up some time to talk tomorrow.  I think a lot of these converters might be useful for the parsing I'm doing for the export, but I'm not sure.  Comments at the top of your classes explaining intent would be a huge help!

devonbush:
I'm still perplexed by the lack of comments at the top of classes explaining their purpose.  Is this something we should have a team discussion about to agree on a standard?  Maybe what the classes do is obvious to people more experienced than me on this codebase, but there are a lot that I would have to do a lot of digging to find out

gmakharat:
> I'm still perplexed by the lack of comments at the top of classes explaining their purpose. Is this something we should have a team discussion about to agree on a standard? Maybe what the classes do is obvious to people more experienced than me on this codebase, but there are a lot that I would have to do a lot of digging to find out

@devonbush to be honest, I am not fan of comments, they easily get rot, I prefer to give maximally descriptive names to the classes and methods, I agree that it is good idea to have team discussion about it.

devonbush:
Sounds good, we can start a slack thread about it Monday when Simone is back

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/filter/query/BaseAbstractQueryBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    protected SplitterStrategy splitter;
+    protected AndOrFilterSeparator filterSeparator;
+    private BaseQueryBuilder baseQueryBuilder;
+
+    protected BaseAbstractQueryBuilder() {
+        boolQueryBuilder = new BoolQueryBuilder();
+        parser = new FilterParser();
+    }
+
+    public void setEsIndex(String index) {
```
SimoneMaiwald, line 33: we could also remove the ES here like we did for the object classes. there are only ES indexes…

SimoneMaiwald:
Please run checkstyle on the branch again. Thank you

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/elastic/filter/query/SingleQueryBuilderTest.java
```diff
@@ -1,33 +1,109 @@
...
+
+        BoolQueryBuilder expectedQuery = new BoolQueryBuilder();
+        expectedQuery.must(new MatchQueryBuilder("dsm.diagnosisYear", 2014L));
+
+        Assert.assertEquals(expectedQuery, getNonActivityQueryBuilder(filter));
+
+    }
+
+    @Test
+    public void twoDifferentValue() {
```
SimoneMaiwald, line 90: could we also add a test for two different values for or
``` AND  dsm.diagnosisYear = 2015 OR dsm.diagnosisYear = 2014```

File: pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/elastic/filter/query/ActivitiesCollectionQueryBuilderTest.java [hunk 3]
```diff
@@ -0,0 +1,108 @@
...
+        nestedBoolQuery.must(new MatchQueryBuilder("activities.activityCode", "ANGIORELEASE"));
+        nestedBoolQuery.must(new BoolQueryBuilder().must(new ExistsQueryBuilder("activities.completedAt")));
+        NestedQueryBuilder expectedNestedQuery = new NestedQueryBuilder("activities", nestedBoolQuery, ScoreMode.Avg);
+        expected.must(expectedNestedQuery);
+        Assert.assertEquals(expected, actual);
+    }
+
+    @Test
+    public void activitiesCreatedAtExactDate() {
+        String filter = " AND ANGIORELEASE.completedAt = '2022-07-28'";
```
SimoneMaiwald, line 37: @gmakharat how much work would it be to add a way to filter for 
` AND ANGIORELEASE.activityVersion = 'v2' `
to only show pts that have a new version of an activity filled out

gmakharat: That should not be a big deal I think, probably the best way is to create separate test class for the similar tests but for studies which may have two version.

SimoneMaiwald: osteo would be a study that has two versions :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/filter/query/BaseAbstractQueryBuilder.java
gmakharat: do you mean to remove this method?

ScalaChamp (approved):

PR #2121, opened 2022-08-05
DSM-7581: Adding data dictionary capability to RC

devonbush:
This adds data dictionary support and also cleans up yet more potential null pointer edge cases.  The data dictionary is a complete listing of all study variables, useful for researchers doing work on the data set.  Pressing "download" will now download a .zip file of both the exported data, and a data dictionary excel document.  See the file below for an example data dictionary generated off of development AT

[DataDictionary.xlsx](https://github.com/broadinstitute/ddp-study-server/files/9271277/DataDictionary.xlsx)

TO TEST:
1. load a study participant list
2. select "download"
3. download with the default options
4. observe that a zip file is downloaded, containing both a PArticipant data export, and a "DataDictionary.xslx" file


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+        String previousParentQuestionId = null;
+
+        public void apply(FilterExportConfig filterConfig,
+                          int activityRepeatNum,
+                          int questionRepeatNum,
+                          Map<String, Object> option,
+                          String detailName,
+                          FilterExportConfig parentConfig) {
+            // we don't make separate entries for each iteration of a variable
+            if (activityRepeatNum > 1 || questionRepeatNum > 1) {
```
gmakharat: I would extract this boolean conditions as method with name `isAlreadyIterated` or something like that if I got the idea correctly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+                return;
+            }
+
+            String moduleName = getModuleColumnPrefix(filterConfig, 1);
+            if (!moduleName.equals(previousModuleName)) {
+                addModuleHeaderRows(moduleName, filterConfig.getParent());
+            }
+            String questionId = null;
+            String parentQuestionId = null;
+            if (filterConfig.getQuestionDef() != null) {
```
gmakharat, line 105: #Optional
one a little bit more cleaner alternative of `null` checks in Java is `Objects.nonNull`, `Objects.isNull`, etc in my opinion.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+                if (parentConfig != null) {
+                    parentQuestionId = (String) parentConfig.getQuestionDef().get(ESObjectConstants.STABLE_ID);
+                }
+                questionId = (String) filterConfig.getQuestionDef().get(ESObjectConstants.STABLE_ID);
+            }
+
+            if (!StringUtils.equals(previousParentQuestionId, parentQuestionId) && parentQuestionId != null) {
```
gmakharat: `isCompositeQustion` could be good name for extracted method for this boolean condition.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+                addCompositeQuestionHeaderRow(filterConfig, activityRepeatNum, questionRepeatNum, option, detailName, parentConfig);
+            }
+            if (!StringUtils.equals(questionId, previousQuestionId) && filterConfig.isSplitOptionsIntoColumns()) {
```
gmakharat: similar for this condition

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+        sheet.setColumnWidth(2, 12 * 256);
+        sheet.setColumnWidth(3, 60 * 256);
+        sheet.setColumnWidth(4, 40 * 256);
+
+        applyToEveryColumn(new DictionaryRowWriter());
+
+        writeAndCloseSheet(os);
+    }
+
+    protected class DictionaryRowWriter implements ColumnProcessor {
```
gmakharat, line 83: maybe this class deserves to have its own file due to its size?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+        if (moduleExportConfig.getNumMaxRepeats() > 1) {
+            SXSSFRow moduleRepeatRow = addRowToSheet();
+            String repeatString = "Up to " + moduleExportConfig.getNumMaxRepeats() +
+                    " entries for this module exist for each participant.\n";
+            repeatString += "Entries are indicated in reverse chronological order.\n";
+            repeatString += "The most recent entry has no suffix, the next-most-recent is suffixed with _2," +
+                    " the next-most-recent is suffixed with _3, etc...\n ";
+            repeatString += "e.g. " + moduleName + ".[QUESTION] is the most recent completion, and " + moduleName +
+                    "_2.QUESTION is the next-most recent completion.";
+            moduleNameRow.createCell(1).setCellValue(repeatString);
```
gmakharat: I think having body of this if statement in separated method with meaningful name would be beneficial for quick understanding in future.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+        addRowToSheet(variableName, dataType, questionType, descriptionText, optionText);
+
+    }
+
+    protected void addAdditionalDetailRow(FilterExportConfig filterConfig,
+                                          int activityRepeatNum,
+                                          int questionRepeatNum,
+                                          Map<String, Object> option,
+                                          String detailName,
+                                          FilterExportConfig parentConfig) {
```
gmakharat, line 260: I would say that such parameters list is data clump, because they are used in different methods exactly the same number of arguments with the same types, I would use parameter object design pattern: https://github.com/iluwatar/java-design-patterns/tree/master/parameter-object

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,293 @@
...
+        boldUnderlineStyle.setFont(boldUnderlineFont);
+    }
+
+    /** writes the dictionary */
+    public void export(OutputStream os) throws IOException {
+        sheet.setColumnWidth(0, 40 * 256);
+        sheet.setColumnWidth(1, 10 * 256);
+        sheet.setColumnWidth(2, 12 * 256);
+        sheet.setColumnWidth(3, 60 * 256);
+        sheet.setColumnWidth(4, 40 * 256);
```
gmakharat, line 76: is it fixed size columns number? if so I would extract 0, 1, 2, 3, 4, 5 so called magic numbers as constants with meaningful name like 0 -> variable name, 1 - data type, etc

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/ExcelParticipantExporter.java
```diff
@@ -44,21 +46,16 @@ public void export(Response response) throws IOException {
...
     }
 
-    private void setResponseHeaders(Response response) {
-        response.type(MediaType.OCTET_STREAM.toString());
-        response.header("Access-Control-Expose-Headers", "Content-Disposition");
-        response.header("Content-Disposition", "attachment;filename=" + getExportFilename(fileFormat));
+    protected void writeAndCloseSheet(OutputStream os) throws IOException {
+        workbook.write(os);
+        workbook.dispose();
+        workbook.close();
```
gmakharat: #Optional
You could use so called `try-with-resource` technique which automatically closes any auto closable and catches exceptions
`try (workbook) {
   workbook.write(os);
   workbook.dispose();
}`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/FilterExportConfig.java
```diff
@@ -19,10 +20,7 @@ public class FilterExportConfig {
     private final ModuleExportConfig parent;
     private final String type;
     private boolean splitOptionsIntoColumns = false;
-    // whether this question has details associated with it -- this property is allowed to be dynamic since we don't know if a
-    // question has details enabled until we start parsing participant responses
-    @Setter
-    private boolean hasDetails = false;
+    private HashSet<String> optionIdsWithDetails = new HashSet<String>();
```
gmakharat: let's use more generic types for variables like `Set<String>`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
```diff
@@ -20,28 +20,19 @@ public TsvParticipantExporter(List<ModuleExportConfig> moduleConfigs,
...
+            String rowString = getRowString(rowValues);
+            printWriter.println(rowString);
         }
-    }
-
-    public void setResponseHeaders(Response response) {
-        response.type(MEDIA_TYPE);
-        response.header("Access-Control-Expose-Headers", "Content-Disposition");
-        response.header("Content-Disposition", "attachment;filename=" + getExportFilename(fileFormat));
+        printWriter.flush();
```
gmakharat, line 34: I guess `printWriter` is not closed anywhere.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
devonbush: Yeah, I could see that going either way.  Here, not encasing the parameters in an object was very useful to me as I was writing the code, because then I could see which parameters I was/wasn't using in the method logic

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TsvParticipantExporter.java
devonbush: That's correct, I'll add a comment to that effect.  Closing PrintWriter would close the underlying OutpuStream, which would then prevent the rest of the ZipFile from being written

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
devonbush: Thanks for the tip, I haven't used those much.  Per https://stackoverflow.com/questions/37972859/java-util-objects-isnull-vs-object-null, those methods are mainly intended for usage within lambdas (the underlying implementation is just `== null`)  So I'll leave this as-is here, but check for any places in lambdas where I should be using Objects.nonNull

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
devonbush: Yeah, I could see that going either way.  Maybe it should get pulled out if the DataDictionaryExporter class gets larger too.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
gmakharat: Yes, but having such many parameters in a method is not considered as a good technique, later when someone will touch this code for further changes he/she will probably add method which will be needed to called from this particular method and pass every parameter and it will become huge mess

gmakharat (approved):

ScalaChamp (approved):

PR #2129, opened 2022-08-08
[DSM | Singular] DDP-8494 Downloading from a signed URL

pegahtah:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
```diff
@@ -0,0 +1,56 @@
...
+    private String bucketName;
+    private String blobName;
+    private String ddpParticipantId;
+    private String fileGuid;
+    FileDownloadService fileDownloadService;
+    Optional<DDPInstanceDto> ddpInstanceDto;
+
+
+    public boolean isFileClean() {
+        ddpInstanceDto.orElseThrow();
```
Trimax: No need to put it in a separate line I assume. You can just do
```
String participantIndex = ddpInstanceDto.orElseThrow().getEsParticipantIndex();
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
```diff
@@ -0,0 +1,56 @@
...
+            throw new RuntimeException("Participant ES Data is not found for " + ddpParticipantId);
+        }
+        Optional<ESFile> maybeFile =
+                maybeParticipantESDataByParticipantId.get().getFiles().stream().filter(file -> file.guid.equals(fileGuid)).findAny();
+        ESFile file = maybeFile.orElseThrow();
+        return isFileClean(file);
+    }
+
+    private boolean isFileClean(ESFile file) {
+        return StringUtils.isNotBlank(file.scannedAt) && CLEAN.equals(file.scanResult);
```
Trimax: I would recommend add to add `isClean` method to ESFile model

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
pegahtah: Yes good catch!! I'll make the fix

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/ESFile.java
```diff
@@ -0,0 +1,19 @@
+package org.broadinstitute.dsm.model.elastic;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class ESFile {
```
Trimax: It's up to you, but don't you want mark it with:
```
@Value
@SuperBuilder(toBuilder = true)
```

Since we're not using java 17 with record class, we might want to follow incapsulation principle and making fields private is the first step to achieve that.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -159,7 +160,7 @@ public Builder withProfile(ESProfile profile) {
             return this;
         }
 
-        public Builder withFiles(List<Object> files) {
+        public Builder withFiles(List<ESFile> files) {
```
Trimax: We could use Lombok's superbuilder

pegahtah: Thanks, I wasn't familiar with that, I just read about it and (if I understood correctly ) if we want to change to that then  lots of classes in DSM should also change to using that ( for both getting it to work correctly as well as keeping the structure the same across the whole code). I suggest we don't make changes in other classes right now, but I'd appreciate your thoughts on this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/ESFile.java
gmakharat: @pegahtah I have created such file as well when I did filtering refactoring which is still in PR, https://github.com/broadinstitute/ddp-study-server/blob/c763ecd6088954d1dde14f02135d99cbe1799e39/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/Files.java could you please take a look at it and change this file accordingly to avoid conflicts in the future?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
gmakharat: I am not really fan of Lombok's builder, I prefer to do it manually since `with` prefix is more cleaner in my opinion which Lombok does not provide

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
```diff
@@ -0,0 +1,51 @@
...
+    private String ddpParticipantId;
+    private String fileGuid;
+    FileDownloadService fileDownloadService;
+    Optional<DDPInstanceDto> ddpInstanceDto;
+
+
+    public boolean isFileClean() {
+        String participantIndex = ddpInstanceDto.orElseThrow().getEsParticipantIndex();
+        Optional<ElasticSearchParticipantDto> maybeParticipantESDataByParticipantId =
+                ElasticSearchUtil.getParticipantESDataByParticipantId(participantIndex, ddpParticipantId);
```
gmakharat: let's slowly eliminate usage of `ElasticSearchUtil` and move to `ElasticSearchable` interface and its one of the implementation called `ElasticSearch` for more flexibility.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
```diff
@@ -0,0 +1,51 @@
...
+@AllArgsConstructor
+public class ParticipantFilesUseCase {
+
+    private static String CLEAN = "CLEAN";
+    private String bucketName;
+    private String blobName;
+    private String ddpParticipantId;
+    private String fileGuid;
+    FileDownloadService fileDownloadService;
+    Optional<DDPInstanceDto> ddpInstanceDto;
```
gmakharat: in general it is not considered as good technique to pass `Optional` type as parameter to the constructor or to the method, as I see this class is used in route class and you could use `orElseThrow` in route class and pass not optional `DDPInstanceDto` here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantfiles/DownloadParticipantFileRoute.java
```diff
@@ -43,27 +49,28 @@ protected Object processRequest(Request request, spark.Response response, String
...
-            Optional<byte[]> downloadObject =
-                    downloadObject(googleProjectName, bucketName, objectName, blobName);
-            byte[] downloadObjectArray = downloadObject.orElseThrow();
-            response.status(200);
-            return downloadObjectArray;
+            String blobName = queryParams.value(BLOB_NAME);
+            String fileName = queryParams.value(FILE_NAME);
+            String fileGuid = queryParams.value(FILE_GUID);
+            String ddpParticipantId = queryParams.value(DDP_PARTICIPANT_ID);
+            Optional<DDPInstanceDto> ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(realm);
```
gmakharat: `new DDPInstanceDao().getDDPInstanceByInstanceName(realm).orElseThrow()` is way more cleaner since if study does not exist we should not continue any process

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -919,7 +920,9 @@ private void setupPubSubPublisherRoutes(Config config) {
             PostMercuryOrderDummyRoute postMercuryOrderDummyRoute = new PostMercuryOrderDummyRoute(projectId, mercuryTopicId);
             post(API_ROOT + RoutePath.SUBMIT_MERCURY_ORDER, postMercuryOrderDummyRoute, new JsonTransformer());
         }
-        get(UI_ROOT + RoutePath.DOWNLOAD_PARTICIPANT_FILE, new DownloadParticipantFileRoute(projectId));
+
+        FileDownloadService fileDownloadService = FileDownloadService.fromConfig(config);
+        get(UI_ROOT + RoutePath.DOWNLOAD_PARTICIPANT_FILE, new DownloadParticipantFileRoute(projectId, fileDownloadService));
```
gmakharat: I think you need to provide `new JsonTransformer()` as 3rd argument of `get` which will help you to remove `gson` converter in route.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/participantfiles/DownloadParticipantFileRoute.java
```diff
@@ -43,27 +49,28 @@ protected Object processRequest(Request request, spark.Response response, String
...
+            ParticipantFilesUseCase participantFilesUseCase =
+                    new ParticipantFilesUseCase(bucketName, blobName, ddpParticipantId, fileGuid, fileDownloadService, ddpInstanceDto);
+
+            try {
+                SignedUrlResponse url = participantFilesUseCase.createSignedURLForDownload();
+                log.info("Signed URL generated for file download for participant " + ddpParticipantId + " file");
+                response.status(200);
+                return new Gson().toJson(url);
```
gmakharat: once you add `new JsonTransformer()` while creating route in `DSMServer` you won't have to add `new Gson().toJson`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/FileDownloadService.java
```diff
@@ -0,0 +1,79 @@
...
+    private final ServiceAccountSigner signer;
+    private final GoogleBucketClient storageClient;
+    private final int maxSignedUrlMins;
+
+
+    public static FileDownloadService fromConfig(Config cfg) {
+        String signerJson = ConfigUtil.toJson(cfg.getConfig(ApplicationConfigConstants.FILE_DOWNLOAD_CREDENTIALS));
+        InputStream signerStream = new ByteArrayInputStream(signerJson.getBytes(StandardCharsets.UTF_8));
+        ServiceAccountCredentials signerCredentials;
+        try {
```
gmakharat, line 34: `signerStream` is `InputStream` type and I believe we need to close it after its usage to avoid memory leakage, you could use `try-with-resource` to do close of streams automatically like: 
`try (signerStream) {
            signerCredentials = ServiceAccountCredentials.fromStream(signerStream);
        } catch (IOException e) {
            throw new DDPException("Could not get signer credentials", e);
        }`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participantfiles/ParticipantFilesUseCase.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+    public boolean isFileClean() {
```
ScalaChamp, line 30: Optional: we can use more java-8 like features to make code shorter but it's not necessary:

implementation suggestion, method `isFileClean`:
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```
public boolean isFileClean() {
        return Optional.ofNullable(ddpInstanceDto.orElseThrow().getEsParticipantIndex())
                .map(this::buildElasticSearchParticipantDtoFromESIndex)
                .map(this::filterFileFromElasticSearchParticipantDto)
                .map(this::isFilteredFileClean)
                .orElse(false);
    }

    private ElasticSearchParticipantDto buildElasticSearchParticipantDtoFromESIndex(String esIndex) {
        return ElasticSearchUtil.getParticipantESDataByParticipantId(esIndex, ddpParticipantId)
                .orElseThrow(() -> new RuntimeException("Participant ES Data is not found for " + ddpParticipantId));
    }

    private ESFile filterFileFromElasticSearchParticipantDto(ElasticSearchParticipantDto esPtDto) {
        return esPtDto.getFiles().stream().filter(this::filterFileByFileGuid).findAny()
                .orElseThrow(() -> new RuntimeException("Could not match any files with " + fileGuid));
    }

    private boolean filterFileByFileGuid(ESFile file) {
        return file.guid.equals(fileGuid);
    }

    private boolean isFilteredFileClean(ESFile esFile) {
        return esFile.isFileClean();
    }
```
    
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
Trimax: ```@Builder(setterPrefix = "with")``` will do exactly what you're talking about :)
Anyway, my idea is to get rid of boilerplate code wherever it is possible.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: Good catch!! Thanks!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/FileDownloadService.java
pegahtah: I believe `signerCredentials = ServiceAccountCredentials.fromStream(signerStream);` closes it automatically

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/ESFile.java
pegahtah: I matched the naming of my changes to your file

gmakharat (approved):

PR #2132, opened 2022-08-09
DSP-7581: adding data dictionary to export and parse refactor

devonbush:
This includes some work based on https://github.com/broadinstitute/ddp-study-server/pull/2095 and responding to the comments there.  This is opened as a fresh PR since it's now different enough from what's on the RC that it counts as a fresh feature.  this also includes work in response to comments on https://github.com/broadinstitute/ddp-study-server/pull/2121

This adds data dictionary support and also cleans up yet more potential null pointer edge cases. The data dictionary is a complete listing of all study variables, useful for researchers doing work on the data set. Pressing "download" will now download a .zip file of both the exported data, and a data dictionary excel document. See the file below for an example data dictionary generated off of development AT

[DataDictionary.xlsx](https://github.com/broadinstitute/ddp-study-server/files/9271277/DataDictionary.xlsx)

TO TEST:

load a study participant list
select "download"
download with the default options
observe that a zip file is downloaded, containing both a PArticipant data export, and a "DataDictionary.xslx" file


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/DataDictionaryExporter.java
```diff
@@ -0,0 +1,312 @@
...
+        boldUnderlineStyle.setFont(boldUnderlineFont);
+    }
+
+    /** writes the dictionary */
+    public void export(OutputStream os) throws IOException {
+        sheet.setColumnWidth(VARIABLE_NAME_COL_NUMBER, 40 * 256);
+        sheet.setColumnWidth(1, 10 * 256);
+        sheet.setColumnWidth(2, 12 * 256);
+        sheet.setColumnWidth(3, 60 * 256);
+        sheet.setColumnWidth(4, 40 * 256);
```
gmakharat: probably you forgot to use constants here for 1,2,3 and 4

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -86,8 +89,10 @@ public static Alias of(SortBy sortBy) {
 
     public static Alias of(ParticipantColumn column) {
         Alias esAlias;
-        if (Objects.nonNull(column.getObject())) {
+        if (Objects.nonNull(column.getObject()) && Alias.ofOrNull(column.getObject()) != null) {
```
gmakharat, line 92: I was thinking that this method is pretty related to `ParticipantColumn` class, I believe no one will create object of `ParticipantColumn` to get alias, everyone will use `private static Alias of(String alias)` in most of the cases, `public static Alias of(ParticipantColumn column)` is specific case only for export feature, maybe we could move that method to `ParticipantColumn` and I believe it will become more cleaner that you will directly call for example `getAlias` method on the instance of `ParticipantColumn`, instead of calling `Alias.of` and passing instance of `ParticipantColumn` to it?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -114,52 +106,65 @@ private List<ParticipantWrapperDto> collectData(DDPInstanceDto ddpInstanceDto) {
         List<String> proxyGuids = new ArrayList<>();
 
         for (ElasticSearchParticipantDto elasticSearchParticipantDto : esData.getEsParticipants()) {
+            if (elasticSearchParticipantDto instanceof UnparsedESParticipantDto) {
+                addWrapperToList((UnparsedESParticipantDto) elasticSearchParticipantDto, result);
+            } else {
+                addWrapperToList(elasticSearchParticipantDto, result);
```
gmakharat: I would create child class of `ParticipantWrapper` for export feature as well to avoid such implicit if statements, would be more clear in my opinion, but it's personal choice I would say

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -114,52 +106,65 @@ private List<ParticipantWrapperDto> collectData(DDPInstanceDto ddpInstanceDto) {
...
+    }
 
-                esDsm.getOncHistory().ifPresent(oncHistory -> {
-                    participant.setCreated(oncHistory.getCreated());
-                    participant.setReviewed(oncHistory.getReviewed());
-                });
+    private void addWrapperToList(ElasticSearchParticipantDto elasticSearchParticipantDto, List<ParticipantWrapperDto> result) {
 
-                StudyPreFilter.fromPayload(StudyPreFilterPayload.of(elasticSearchParticipantDto, ddpInstanceDto))
-                        .ifPresent(StudyPreFilter::filter);
```
gmakharat, line 128: I think this is necessary for studies which have different versions, @Anzori-Ghurtchumelia can help about this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
devonbush: Yeah, once a decision gets made on whether we also want to move the participant list to using unparsed DTOs, we can make a good decision about how best to refactor this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
devonbush: ah, great catch, I didn't mean to remove this -- it must have gotten accidentally deleted in a merge

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
devonbush: Those are really good points.  the downside of moving it is that it relies on a lot of currently private constructors, ad making those public would lose some of the nice encapsulation Alias gives now.  I think leaving it in will help us better to say "Alias" handles all logic related to mapping filters to tables.

pegahtah (dismissed):

pegahtah (approved):

PR #2134, opened 2022-08-09
DDP-8599: exporting option details in question defs

devonbush:
This adds option details text and whether a question has option details to the ElasticSearch export.  I've written an automated test for this, but I'm not sure how to test this on a running system.  Also, what would the process need to be to have all the question definitions currently in production get re-exported?

TO TEST: 
1. Trigger export of study in dev
2.  confirm updated definitions were written for option details
3. ,confirm updated text is used by the DSM study export feature


ssettipalli:
Devon:
If we don't have a jira ticket to track these changes, can we please create one.

reg testing/prod deploy
I am assuming changes effect only activity_definition index.
We need to trigger export of activity_definition index for each study and that should load updated definitions to elastic.
If we don't see new changes , probably need to delete the index and re-trigger the export. This can be tested in Dev.
We also have a nightly export of all indices which probably should export updated definitions. We should be able to see & test the behavior in Dev once changes are merged.

devonbush:
thanks, just created a separate ticket for this.  let me know what else you need for a review

ssettipalli (approved):
looks ok to me.
Devon: Pls make sure you add jira ticketID to comments when you do squash merge to develop so that it helps back track changes to the ticket.
Thanks.

PR #2144, opened 2022-08-12
DDP-8589 extend copy configuration to handle copying from parent activities

LKapanad:
DDP-8589 in the scope of this ticket we need parent's typed country and state values in agedup child to check whether CONSENT_ADDENDUM should be created or not and since proxy relationship is revoked after ageup we can't get this data. to resolve this problem I added a new feature to COPY_CONFIGURATION to be able to transfer parent's typed in data into child's activity data
- added a new config property "user" into copyConfigPairs "source" and "target" fields. possible values "USER" or "OPERATOR". default value is USER
- add new columns "user" in COPY_ANSWER_LOCATION table
- Handle operator copy case, leave the default functionality as it is


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
```diff
@@ -24,6 +26,8 @@
 import org.broadinstitute.ddp.model.copy.CopyConfiguration;
 import org.broadinstitute.ddp.model.copy.CopyLocation;
 import org.broadinstitute.ddp.model.copy.CopyLocationType;
+import org.broadinstitute.ddp.model.governance.Governance;
+import org.broadinstitute.ddp.pex.UserType;
```
bskinner: This is a class from the PEX packages, is not intended for long-term storage in the database, and limits future changes in the PEX package. If there isn't a pre-existing type which already fits, please use a dedicated type in the `org.broadinstitute.ddp.model.copy` package.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
```diff
@@ -56,13 +60,32 @@ public void execute(Handle handle, long operatorId, long participantId, CopyConf
...
 
             if (source.getType() == CopyLocationType.ANSWER && target.getType() == CopyLocationType.ANSWER) {
+                QuestionDto sourceQuestion;
+                FormResponse sourceInstance;
                 String sourceStableId = ((CopyAnswerLocation) source).getQuestionStableId();
-                QuestionDto sourceQuestion = questionDtosByStableId.get(sourceStableId);
-                if (sourceQuestion == null) {
-                    continue; // Question might have been removed from activity, so we skip it.
+                String user = ((CopyAnswerLocation) source).getUser();
+                if (UserType.OPERATOR.equalsIgnoreCase(user)) {
```
ssettipalli: @LKapanad :
Trying to understand scenario where proxy/operator has question answers that need to be copied to aged-up user ? Does lms consent have question answers by operator that need to be copied to aged-up user ?
In existing studies we are creating new consent for aged-up users, trying to understand what is it different for lms !

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
```diff
@@ -56,13 +60,32 @@ public void execute(Handle handle, long operatorId, long participantId, CopyConf
...
+                    Stream<Governance> governances = handle.attach(UserGovernanceDao.class)
+                            .findActiveGovernancesByParticipantAndStudyIds(participantId, config.getStudyId());
+                    Optional<Governance> optionalGovernance = governances.findFirst();
+                    if (optionalGovernance.isEmpty()) {
+                        throw new DDPException(String.format("Governance not found for participant %s", participantId));
+                    }
+                    Governance governance = optionalGovernance.get();
```
bskinner: Given that `governances` is not used later in this scope, the intermediate variables can be removed and the code simplified to something like the suggested change.
```suggestion
                    var governance = handle.attach(UserGovernanceDao.class)
                            .findActiveGovernancesByParticipantAndStudyIds(participantId, config.getStudyId())
                            .findFirst()
                            .orElseThrow(() -> {
                                return new DDPException(String.format("Governance not found for participant %s", participantId));
                            });
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
```diff
@@ -56,13 +60,32 @@ public void execute(Handle handle, long operatorId, long participantId, CopyConf
...
             if (source.getType() == CopyLocationType.ANSWER && target.getType() == CopyLocationType.ANSWER) {
+                QuestionDto sourceQuestion;
+                FormResponse sourceInstance;
                 String sourceStableId = ((CopyAnswerLocation) source).getQuestionStableId();
-                QuestionDto sourceQuestion = questionDtosByStableId.get(sourceStableId);
-                if (sourceQuestion == null) {
-                    continue; // Question might have been removed from activity, so we skip it.
+                String user = ((CopyAnswerLocation) source).getUser();
+                if (UserType.OPERATOR.equalsIgnoreCase(user)) {
+                    Stream<Governance> governances = handle.attach(UserGovernanceDao.class)
```
bskinner: Let's move the `attach()` call to get a `UserGovernanceDao` instance to outside of the scope of the parent `for` loop. This is an expensive operation, and it's not necessary to keep re-creating it.

bskinner (changes requested):
Couple of comments, but the highest priority is the out-of-scope use of `org.broadinstitute.ddp.pex.UserType`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
```diff
@@ -56,13 +60,32 @@ public void execute(Handle handle, long operatorId, long participantId, CopyConf
...
+                    }
+                    Governance governance = optionalGovernance.get();
+                    Map<Long, FormResponse> proxyActivities = retrieveActivityData(handle, governance.getProxyUserId(),
+                            List.copyOf(questionDtosByStableId.values()));
+                    sourceQuestion = questionDtosByStableId.get(sourceStableId);
+                    if (sourceQuestion == null) {
+                        continue; // Question might have been removed from activity, so we skip it.
+                    }
+                    sourceInstance = proxyActivities.get(sourceQuestion.getActivityId());
+                } else {
```
ssettipalli, line 77: This code is also widely used in copying answers in non-ageup scenarios. We need to test and make sure those scenarios are not effected with new changes

LKapanad: @ssettipalli yes, all the existing copy configurations will have the same behavior like before. Just one if statement is added in case of operator which will be present only for lms and osteo.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/copy/CopyExecutor.java
LKapanad: @ssettipalli the difference is that the parent fills the child's country and state in PREQUAL activity and this data gets lost for child after ageup. Yes consent is generated for ageup participant too but there's no question about country. which is needed in the post-consent activity consent_addendum

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8484-copy-configuration-operator.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="lkapanad" id="copy-configuration-operator">
+        <addColumn tableName="copy_answer_location">
+            <column name="user" type="varchar(30)" defaultValue="USER">
```
bskinner: Naming the column `user` could be misleading. How about we name this `user_type` instead?

bskinner:
Minor tweak after my last review, but I'm ok with approving this assuming @ssettipalli doesn't have anything else he'd like addressed.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8484-copy-configuration-operator.xml
ssettipalli: user_type suits better for sure.

ssettipalli: Rest all looks fine to me

ssettipalli (approved):

bskinner (approved):

PR #2149, opened 2022-08-15
Ddp 8511 update medical provider table

gskhirel:
https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/836?modal=detail&selectedIssue=DDP-8511
1.After autocompletion for physician component is done, Patch request for medical provider should be updated
2. Updated Medical provider Dtos and table by adding country field


pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8511-physician-country-add.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="gskhirel" id="physician_country_add">
+        <addColumn tableName="user_medical_provider">
+            <column name="country" type="varchar(100)">
```
ssettipalli: wondering if it should be country or country_id (reference to country) like we have in mailing_address .
I think country should be fine though. Everything else looks to me.
@MocanaAtBroad : any thoughts on this ?

MocanaAtBroad: I would encourage you to use country_address_info table. Has all the countries in the world and we have translations for them. You can also do autocompletion.
Screen Shot 2022-08-16 at 8 51 26 AM

ssettipalli:
Most important thing to watchout is:
This component is widely used by ALL studies. 
We might not be deploying angular/front-end for ALL existing studies and changes should be backward compatible and not break any existing studies if front-end/angular is not updated.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PatchMedicalProviderRoute.java
```diff
@@ -66,6 +66,7 @@ public Object handle(
                     String physicianName;
                     String city;
                     String state;
+                    String country;
                     try {
                         JsonElement data = new JsonParser().parse(request.body());
                         if (!data.isJsonObject() || data.getAsJsonObject().entrySet().size() == 0) {
```
bskinner: Any API changes must also have their associated documentation updated in `pepper-apis/docs/specification`.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DataLoader.java
```diff
@@ -732,6 +738,10 @@ String getMedicalProviderGuid(Handle handle) {
                 MedicalProviderTable.TABLE_NAME, MedicalProviderTable.MEDICAL_PROVIDER_GUID);
     }
 
+    long getCountryAdressId(Handle handle, String countryName) {
```
bskinner: There's a typo here- this should be `Address`, not `Adress`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DataLoader.java
```diff
@@ -688,6 +692,7 @@ public void loadReleaseSurveyData(Handle handle, ParticipantData participantData
                 || releaseSurvey.getInitialBiopsyCity() != null
                 || releaseSurvey.getInitialBiopsyState() != null) {
             String guid = getMedicalProviderGuid(handle);
+            long countryAdressId = getCountryAdressId(handle, releaseSurvey.getCountry());
```
bskinner: Should be `Address`, not `Adress`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiMedicalProvider.java
```diff
@@ -211,6 +217,9 @@ List<MedicalProviderDto> getAllByUserGuidStudyGuid(
             @Bind("studyGuid") String studyGuid
     );
 
+    @SqlQuery("SELECT country_address_info_id from country_address_info where name =: name")
+    long getCountryAddressInfoId(@Bind("name")String countryName);
```
bskinner: There should be a space between `("name")` and `String`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DataLoader.java
```diff
@@ -666,6 +668,7 @@ public void loadReleaseSurveyData(Handle handle, ParticipantData participantData
             for (Institution institution : institutionList) {
 
                 String guid = getMedicalProviderGuid(handle);
+                long countryAdressId = getCountryAdressId(handle, institution.getCountry());
```
bskinner: The `attach()` call made by `getCountryAddressId()` is expensive- try to avoid calling it multiple times withing the scope of a single method.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8511-physician-country-add.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="gskhirel" id="physician_country_add">
+        <addColumn tableName="user_medical_provider">
+            <column name="country_address_info_id" type="bigint" >
+            </column>
```
bskinner: Let's use `<column .../>` instead of `<column...></column>` here

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/PostMedicalProviderRouteTest.java
```diff
@@ -267,7 +271,7 @@ private static final class TestMedicalProviderData {
         public static final String PHYSICIAN_NAME = "House MD";
         public static final String CITY = "West Windsor Township";
         public static final String STATE = "New Jersey";
-
+        public static final String COUNTRY = "United States";
```
bskinner, line 274: Using the 2 letter codes would be preferred to using the readable name in the request payload. Our 2-letter country code columns try to adhere to the `ISO 3166-1 alpha-2` standard.

gskhirel: After autocompletion, suggested countries are fully written, populated data of country names in country_address_info table is also written as a full country name. When post request is being sent country names are written fully and we do search for corresponding ids in  country_address_info via full name, therefore if we use ISO 3166-1 alpha-2 standard, we might mess up this test.

bskinner: Where are the autocompletion details coming from?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/migration/Physician.java
```diff
@@ -20,6 +20,9 @@ public final class Physician {
     @SerializedName("state")
     private String state;
 
+    @SerializedName("country")
+    private String country;
+
```
bskinner, line 25: What is the expected behavior if a client passes `null` for this field? Given there are a number of clients using this API, and they may not all be updated, this should be backwards compatible with any currently deployed sites.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/PostMedicalProviderRouteTest.java
gskhirel: It comes from : institution, city, country_subnational_division, country_address_info tables.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/migration/Physician.java
gskhirel: Actually we discussed the scenario, when an user manually enters the country name which is not present in country_address_info, if we consider the current state of user_medical_provider table that kind of data will be lost. So maybe it will be better if we revert the column type back to varchar and put country names directly without relations?

gskhirel: This issue is also described in https://broadinstitute.atlassian.net/browse/DDP-8631

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PatchMedicalProviderRoute.java
```diff
@@ -107,6 +107,7 @@ public Object handle(
                             physicianName,
                             city,
                             state,
+                            existingMedicalProviderDto.getCountryId(),
```
bskinner, line 110: This is following a different pattern from the rest of the parameters included with the payload. Is there a particular reason for the deviation?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/migration/Physician.java
bskinner: I talked this over with @MocanaAtBroad- let's revert this back to just storing the bare string in the database, and not worrying about making sure it's a proper foreign reference.

PR #2157, opened 2022-08-16
[SB] Exclude PanCanValidationRules patch from pancan rootpatch

LKapanad:
Exclude PanCanValidationRules patch from pancan rootpatch, since already applied on prod


ssettipalli:
Please add PancanUpdates to pancan/patch-log.conf

ssettipalli (approved):

PR #2159, opened 2022-08-16
DDP-8539: Block DTO identifiers for existing Expression objects

bskinner:
DDP-8539

## Context

This PR exposes the complete expression objects used in the various Block types.

* Updates SQL statements to include the ID and GUID of shown and enabled PEX expressions
* Exposes nested DTO Expression objects
* Updates necessary methods to pass this information forward to the various `Def` classes.
* Fixes a bug where a pre-existing and valid id is being used for all newly instantiated Expression objects.


bskinner:
Reorganized the branch to ease PR reviews.

PR #2160, opened 2022-08-16
DDP-8539: Identifiers for existing Expressions in Block DTOs

bskinner:
DDP-8539

## Context

This PR exposes the complete expression objects used in the various Block types.

* Fixes a bug where a pre-existing and valid id is being used for all newly instantiated Expression objects using the `Expression(String)` constructor.
* Updates SQL statements to include the ID and GUID of the `shown` and `enabled` PEX expressions associated with a block
* Exposes nested Expression objects on Block DTOs.
* Updates necessary methods to pass this information forward to the various `Def` classes.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFormSectionBlock.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFormSectionBlock.java
@@ -1,14 +1,11 @@
 package org.broadinstitute.ddp.db.dao;
 
-import org.broadinstitute.ddp.constants.SqlConstants;
 import org.broadinstitute.ddp.db.dto.FormBlockDto;
 import org.broadinstitute.ddp.db.dto.SectionBlockMembershipDto;
 import org.broadinstitute.ddp.model.activity.instance.ActivityInstance;
 import org.broadinstitute.ddp.service.actvityinstancebuilder.ActivityInstanceFromDefinitionBuilder;
-import org.jdbi.v3.core.statement.Query;
 import org.jdbi.v3.sqlobject.SqlObject;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
-import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindList;
 import org.jdbi.v3.sqlobject.customizer.BindList.EmptyHandling;
@@ -17,13 +14,13 @@ import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
-import org.jdbi.v3.stringtemplate4.StringTemplateSqlLocator;
+import org.jdbi.v3.stringtemplate4.UseStringTemplateSqlLocator;
 
 import java.util.ArrayList;
 import java.util.Map;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import java.util.Collections;
 
 public interface JdbiFormSectionBlock extends SqlObject {
@@ -59,7 +56,13 @@ public interface JdbiFormSectionBlock extends SqlObject {
             + " where form_section__block_id = :id")
     int[] bulkUpdateRevisionIdsByIds(@Bind("id") List<Long> membershipIds,
                                      @Bind("revisionId") long[] revisionIds);
-
+    
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryOrderedFormBlockDtosBySectionIdsAndInstanceGuid")
+    @RegisterConstructorMapper(value = FormBlockDto.class)
+    List<FormBlockDto> findOrderedFormBlocksForSectionsAndInstance(
+                @BindList(value = "sectionIds", onEmpty = EmptyHandling.NULL_VALUE) List<Long> sectionIds,
+                @Bind("instanceGuid") String instanceGuid);
 
     /**
      * Query for all the block data for given section ids, ordered by their display order.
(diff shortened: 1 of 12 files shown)


pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiBlockNesting.sql.stg
```diff
@@ -1,30 +1,72 @@
...
+WHERE bn.parent_block_id = :blockId
+    AND rev.start_date \<= ai.created_at
+    AND (rev.end_date IS null OR ai.created_at \< rev.end_date)
+    AND (bev.block__expression_id IS null
+        OR (bev_rev.start_date \<= ai.created_at
+            AND (bev_rev.end_date IS null OR ai.created_at \< bev_rev.end_date)))
+    AND (bee.block_enabled_expression_id IS null
+        OR (bee_rev.start_date \<= ai.created_at
+            AND (bee_rev.end_date IS null OR ai.created_at \< bee_rev.end_date)))
+ORDER BY bn.display_order asc
```
bskinner, line 36: The only substantiative changes to this statement is in the `SELECT` part to support the nested `shownExpression`, `enabledExpression` and `blockDto` objects in the `FormBlockDto` class. The rest is just formatting changes.

pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiBlockNesting.sql.stg
```diff
@@ -1,30 +1,72 @@
...
+    bn.parent_block_id AS parent_block_id,
+    bt.block_type_code AS block_meta_block_type_code,
+    nested.block_id AS block_meta_block_id,
+    nested.block_guid AS block_meta_block_guid,
+    ev.expression_id AS sh_expr_expression_id,
+    ev.expression_guid AS sh_expr_expression_guid,
+    ev.expression_text AS sh_expr_expression_text,
+    ee.expression_id AS en_expr_expression_id,
+    ee.expression_guid AS en_expr_expression_guid,
+    ee.expression_text AS en_expr_expression_text
```
bskinner, line 51: Same as [this comment](https://github.com/broadinstitute/ddp-study-server/pull/2160/files#r946770772)

ssettipalli (approved):
looks good to me

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/JdbiFormSectionBlock.java
```diff
@@ -59,7 +56,13 @@ long[] bulkInsert(@BindMethods("dto") List<SectionBlockMembershipDto> membership
             + " where form_section__block_id = :id")
     int[] bulkUpdateRevisionIdsByIds(@Bind("id") List<Long> membershipIds,
                                      @Bind("revisionId") long[] revisionIds);
-
+    
+    @UseStringTemplateSqlLocator
+    @SqlQuery("queryOrderedFormBlockDtosBySectionIdsAndInstanceGuid")
+    @RegisterConstructorMapper(value = FormBlockDto.class)
```
Trimax, line 62: `value = ` might be avoided since FormBlockDto.class is the only one argument here

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/BlockTabularQuestionDto.java
```diff
@@ -29,10 +32,18 @@ public class BlockTabularQuestionDto {
...
-    @ColumnName("shown_expr")
-    String shownExpr;
+    @Nested("sh_expr")
+    Expression shownExpression;
+
+    @Nested("en_expr")
+    Expression enabledExpression;
+
+    public String getShownExpr() {
+        return shownExpression.getText();
```
Trimax, line 42: Is it possible that `shownExpression` or `enabledExpression` reference will be set to NULL? If yes, I'd suggest modifying the getter expression to avoid NPE.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/FormBlockDto.java
```diff
@@ -2,26 +2,34 @@
...
+import org.jdbi.v3.core.mapper.Nested;
+import org.jdbi.v3.core.mapper.reflect.ColumnName;
+import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
 
 /**
  * A wrapper data object around a block and its optional conditional expression.
  */
 @Value
-@AllArgsConstructor
+@AllArgsConstructor(onConstructor = @__({@JdbiConstructor}))
```
Trimax, line 16: `{` and `}` might be avoided here since `JdbiConstructor` is  the only one propagated annotation

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/pex/Expression.java
```diff
@@ -18,6 +18,6 @@ public class Expression {
     String text;
 
     public Expression(final String text) {
```
Trimax, line 20: I assume this constructor might be used for test purposes only. If yes, I suggest annotating the class with the `SuperBuilder` annotation and using the builder pattern in tests. I did it for UserDto and StudyDto partially. It helped in cases when a new column was added (or an existing one was removed) to the database's table. Once you update DTO correspondingly, AllArgsConstructor will have a different set of arguments which means that you'll have to update all the tests that are using this constructor. Using the builder pattern in the test on the other hand will not require you to review all of them and fix compilation errors.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/BlockTabularQuestionDto.java
bskinner: Yup- it's not a required field- thanks for the catch!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/pex/Expression.java
bskinner: You are correct, that constructor is only used in tests, but it is also the only required field for that type. As for the use of a Builder pattern here- I disagree. Having to review all of the uses and fix compilation errors is a positive, not a negative. The lombok builders are prone to runtime errors and are incapable of identifying required fields at compile time without heavy customization of the codegen. Having to fix compilation errors when something changes is a good thing.

PR #2164, opened 2022-08-16
DDP-8539: Update PEX expressions in the ABOUT_HEALTHY survey

bskinner:
DDP-8539

## Context

This PR fixes a bug in the PEX expressions for the ABOUT_HEALTHY survey where certain questions will not be visible when progressing through the adult dependent workflow. The issue impacts the `WHAT_IS_RELATIONSHIP` and `CHD_OTHER_MYSELF` question- in the current implementation, these two questions are only visible in the `self` workflow, when they should be visible in both the `self` and `adult dependent` (but not in the `child`).

This PR updates the PEX shown expression for the parent blocks of the two impacted questions to check for the `CHILD` workflow and negate the result instead of only checking for the `SELF` state.

_screenshots of the changes will be added- currently standing up a localhost environment to visually confirm the fix_ 

## Testing
This can be validated by standing up complete local development environment with the Angular client, database, and DSS instance. The patch can be successfully run against the current `develop` branch but the `SingularPatientSurveyUpdate` and `SingularUpdates` patches will need to be commented out of the `singular/patch-log.conf`

Alternately for dev deploy: we can just run new task/patch: SingularAboutHealthyPexTask
--run-task SingularAboutHealthyPexTask


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dto/FormBlockDto.java
```diff
@@ -44,10 +44,18 @@ public String getGuid() {
...
     public String getShownExpr() {
+        if (shownExpression == null) {
+            return null;
+        }
+
         return shownExpression.getText();
     }
 
     public String getEnabledExpr() {
+        if (shownExpression == null) {
```
ssettipalli: This should be enabledExpression

bskinner: Thank you!

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularAboutHealthyPexTask.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /** 
+     * Finds the set of parent blocks in an activity for a given set of question stable ids.
+     * 
+     * <p>This will only return blocks for which the question is an immediate child. If
+     * the question is embedded in any other structure, it will not be found
+     * 
+     */
+    private List<QuestionBlockDef> findBlocksForQuestions(FormActivityDef activityDefinition, Set<String> stableIds) {
```
ssettipalli, line 203: Wondered if this can be added some Dao so that it can be reused. May be it can be part of bigger scope where we can try to come up with some generic patches which can update study config (patches like UpdateTemplatesInPlace)

ssettipalli (dismissed):
looks good

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularAboutHealthyPexTask.java
bskinner: That piece of code is highly tailored to this specific situation. If there's a need for something like that, this may be useful inspiration, but I would not consider this safe for general use, and the performance characteristics if it would need a careful evaluation.

ssettipalli (approved):

PR #2168, opened 2022-08-17
DDP-8606. File upload notifications template

Trimax:
## Changes in this PR

1. Date uploaded reformatted using the US format `MM/dd/yyyy`
2. File size is now provided in human-readable format
3. Simple mail templates with substitution possibilities implemented
4. Hard-coded mail template for the file upload notification was moved to the database
5. Tests that use StudyDto constructor were refactored to use StudyDto builder instead

## Release notes
When releasing, the `umbrella_study.mail_template_id` column must be set to the value (most likely 1) that corresponds to the singular file upload notification mail template (if not set, the notification will not be sent)

## Possible improvements
Define mail template somewhere in study definition and import it together with the study


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadNotificationEmailFactory.java
```diff
@@ -7,85 +7,45 @@
...
+    private static final SimpleDateFormat dateFormatter = new SimpleDateFormat("MM/dd/yyyy");
     private static final SimpleDateFormat timeFormatter = new SimpleDateFormat("HH:mm:ss");
 
     public static Mail create(final StudyDto study, final String userGuid, final List<FileUpload> fileUploads) {
+        if (study.getNotificationMailTemplateId() == null) {
+            throw new DDPException("Can't create e-mail. Study " + study.getGuid() + " doesn't have mail template");
+        }
+
+        final var mailTemplate = MailTemplateService.getTemplate(study.getNotificationMailTemplateId());
+        mailTemplate.setSubstitutions("PROJECT_SINGULAR_FILE_UPLOAD_ROW", StreamEx.of(fileUploads)
```
ssettipalli: Need to find a way to make this (PROJECT_SINGULAR_FILE_UPLOAD_ROW ) non study specific since the class is not specific to just SINGULAR.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
```diff
@@ -0,0 +1,121 @@
...
+                <constraints nullable="false" />
+            </column>
+        </createTable>
+
+        <addUniqueConstraint tableName="mail_template_repeatable_element"
+                             columnNames="mail_template_id, name"
+                             constraintName="mail_template_repeatable_element_unique" />
+    </changeSet>
+
+    <changeSet author="Dmitrii Eliseev" id="2022-08-17-mail-template-for-singular">
```
ssettipalli: Ideally this changes that inserts data should be in separate file under db-changes/seed/* rather than schema but we have this all over any way.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplateRepeatableElement.java
```diff
@@ -0,0 +1,27 @@
...
+public class MailTemplateRepeatableElement {
+    String name;
+    String content;
+    Map<String, String> substitutions;
+
+    public String render() {
+        var renderedContent = String.valueOf(content);
+        for (final var substitution : substitutions.entrySet()) {
+            renderedContent = renderedContent.replace(toKey(substitution.getKey()), substitution.getValue());
+        }
```
bskinner, line 19: There's a potential for an NPE here from the `String::replace` method which will not give a clear message to the developer as to what is wrong- let's guard around a key in the template with an explicit `null` value.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplate.java
```diff
@@ -0,0 +1,62 @@
...
+@Value
+@AllArgsConstructor
+public class MailTemplate {
+    String name;
+    String contentType;
+    String subject;
+    String body;
+    List<MailTemplateRepeatableElement> repeatableElements;
+
+    Map<String, List<MailTemplateRepeatableElement>> substitutions = new HashMap<>();
```
bskinner: Do we need another templating engine? Is there a reason Velocity can't be used for this purpose instead of rolling our own?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplateSubstitution.java
```diff
@@ -0,0 +1,19 @@
...
+@Value
+@AllArgsConstructor
+public class MailTemplateSubstitution {
+    Map<String, String> values = new HashMap<>();
+
+    public MailTemplateSubstitution withValue(final String name, final String value) {
+        values.put(name, value);
+
+        return this;
+    }
```
bskinner: `HashMap` allows for keys to have an explicit `null` value. This will lead to weird NPEs down the line when applying the template and should be guarded against (this relates to my comment in [MailTemplateRepeatableElement.java](https://github.com/broadinstitute/ddp-study-server/pull/2168/files#diff-5660ee13d3903b43839ba2454eec5082e0b0ce8a015b7c2177b9926dfabd207d))

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
bskinner: I'd argue that this data doesn't even belong in the `seed` directory- that directory is intended for inserts which are for general use by the platform. This is _highly_ study specific, and the ideal spot for this is in the `singular` study configuration

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadNotificationEmailFactory.java
```diff
@@ -7,85 +7,45 @@
...
+                .map(fileUpload -> toSubstitution(userGuid, fileUpload))
+                .toList());
```
bskinner: References to specific studies shouldn't be in the core codebase. This should be refactored for general use, or moved outside of the shared code.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/mail/MailTemplateService.java
```diff
@@ -0,0 +1,46 @@
...
+import org.broadinstitute.ddp.exception.DDPException;
+import org.broadinstitute.ddp.model.mail.MailTemplate;
+import org.broadinstitute.ddp.model.mail.MailTemplateRepeatableElement;
+import org.jdbi.v3.core.Handle;
+
+import java.util.HashMap;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class MailTemplateService {
+    public static MailTemplate getTemplate(final Long id) {
```
bskinner: This looks like it should be `long`, not `Long`- a `null` here seems like it should be treated as an error before this point.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/MailTemplateDao.java
```diff
@@ -0,0 +1,26 @@
...
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+import java.util.Optional;
+
+public interface MailTemplateDao extends SqlObject {
+    @SqlQuery("SELECT * FROM `mail_template` WHERE mail_template_id = :id")
+    @RegisterConstructorMapper(MailTemplateDto.class)
+    Optional<MailTemplateDto> findById(@Bind("id") final Long id);
```
bskinner: Should this be `long` instead of `Long`? I don't see any cases where `findById(null)` would return a useful result.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
Trimax: Yeah, as I mentioned in the ticket description, the right way is to move this to study configuration. But it is quite a big chunk of work that is not so important at the moment I believe.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadNotificationEmailFactory.java
Trimax: I agree. Let's rename the constant to something neutral: FILE_UPLOAD_RECORD

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadNotificationEmailFactory.java
Trimax: I agree. Will rename the constant to something study-neutral.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplateRepeatableElement.java
Trimax: Extended `setSubstitutions` method's strem with following step:
`.removeBy(MailTemplateRepeatableElement::getName, null)`

It will remove all such repeatable elements that have null instead of name

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplate.java
Trimax: Probably not. I need to investigate Velocity then. Have never worked with it.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplateSubstitution.java
Trimax: Added not null check

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplate.java
Trimax: Do we want to work on that in terms of this PR?

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
bskinner: How about this-
* leave the `2022-08-17-mail-template` changeset where it is- it's independent, and we can always shift things around
* remove the `2022-08-17-mail-template-for-singular` changeset and, instead, use its content to create a patch for singular.

Not a perfect solution, but it works with the pieces we have, and keeps study-specific changes in an appropriate place.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/mail/MailTemplate.java
bskinner: Let's hold off- this works well, and we have some scaffolding for bringing in Velocity (after some investigation).

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
Trimax: Sounds good Blake.

Trimax:
Mail template was moved out from the Liquibase script. Now we have separate studybuilder task that inserts mailtemplate and updates umbrella study table accordingly.

pepper-apis/dss-database/src/main/resources/db-changes/schema/DDP-8606.xml
```diff
@@ -0,0 +1,55 @@
...
+                             constraintName="mail_template_repeatable_element_unique" />
+    </changeSet>
+
+    <changeSet author="Dmitrii Eliseev" id="2022-08-17-mail-template-for-singular-reference">
+        <addColumn tableName="umbrella_study">
+            <column name="notification_mail_template_id" type="bigint" defaultValue="null">
+                <constraints nullable="true" references="mail_template(mail_template_id)" foreignKeyName="umbrella_study_file_upload_notification_mail_template_fk"/>
+            </column>
+        </addColumn>
+    </changeSet>
```
bskinner: What's going on with this changeset?

Trimax: Hm. We add the reference from the study to the file upload mail template. We need it to create the message

```
  if (study.getNotificationMailTemplateId() == null) {
            throw new DDPException("Can't create e-mail. Study " + study.getGuid() + " doesn't have mail template");
        }

        final var mailTemplate = MailTemplateService.getTemplate(study.getNotificationMailTemplateId());
```

Trimax: Discussed. Changeset name changed

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularUpdates.java
```diff
@@ -13,6 +13,7 @@ public class SingularUpdates implements CustomTask {
 
     @Override
     public void init(Path cfgPath, Config studyCfg, Config varsCfg) {
+        tasks.add(new SingularFileUploadNotifications());
```
bskinner, line 16: @ssettipalli Are new tasks supposed to be added here?

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularFileUploadNotifications.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public void run(final Handle handle) {
+        log.info("TASK:: {}", SingularFileUploadNotifications.class.getSimpleName());
+
+        final var mailTemplateId = insertMailTemplate(handle);
+        insertMailTemplateRepeatableElement(handle, mailTemplateId);
+        log.info("Mail template inserted");
+
+        handle.attach(SqlHelper.class).updateStudyMailTemplate(cfg.getString("study.guid"), mailTemplateId);
+        log.info("Study {} mail template was updated", cfg.getString("study.guid"));
```
bskinner: I would suggest returning early and without errors if the insertion has already been performed. ~Right now, this patch will just keep attempting to insert the template, even if it's already been done.~

edit:
Walked through the code again and it looks like your unique constraints on the `mail_template` will catch multiple insertions, it's still going to result in a sql error.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularUpdates.java
ssettipalli: It is ok but would be better to add to patch-log.conf or just run the task separately.

bskinner (dismissed):

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularFileUploadNotifications.java
Trimax: Well, reviewed once again. I decided to remove the template's name. We don't use it at all. So, now each time we run the patch it will create a new mail template instance and update the study's reference to the newly created one. I believe it corresponds to the logic we use for study reimport.

bskinner: This is not a study reimport, though- this is being run against a study that has previously been configured. This is now leaving unreferenced values in the database.

Trimax: So, what is your suggestion?
1. Unlink study & template
2. Delete template
3. Insert template

Trimax: Or! We can insert new template and then delete everything unlinked.

bskinner: My perspective on this patch is this is being run prior to any study configuration being implemented for handling templates in the HOCON files, so when this patch is run, I would expect `notification_mail_template_id` to be `null`.

If `notification_mail_template_id` is not `null`, it's likely the patch has already been run, or this patch has been superseded by study configuration, in which case I would suggest just logging that a template has already been added, and successfully returning early so we don't overwrite the potentially newer changes.

Trimax: Anyway, I implemented the deletion. But to be honest, I don't think it's an important thing in our case. How many patches are we going to apply? 10? 100? It will not affect the performance but will keep the changes history.

bskinner: I never mentioned anything about performance. It is making sure the patch doesn't leave a mess for others to clean up later.

Trimax: So, the current implementation should not leave any garbage.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularFileUploadNotifications.java
```diff
@@ -0,0 +1,106 @@
...
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Task to update existing singular study by specifying newly added mail template
+ */
```
bskinner: Let's update this comment to mention the fact that if a pre-existing template is set for the study, it will be deleted.

bskinner (approved):
👍

PR #2175, opened 2022-08-18
DDP-8628: supporting human-readable export format

devonbush:
We would like to give the users an explicit choice between human readable and analysis friendly formatting.  This adds a single new variable "humanReadable" which consolidates a choice between splitting multiselect questions into multiple columns, and rendering option stable ids vs. the display text.  This also merges some bugfixes and test improvements from develop

This is paired with  https://github.com/broadinstitute/ddp-angular/pull/1658

TO TEST:
1. open the participant list for the AT study
2. customize view to include the Medical History "telangiectasia" question
3. export the participant list, selecting "analysis-friendly" 
4. confirm the telangiectasia responses are split into 4 columns - 1 each for eyes and skin, and 1 for each of the supporting details
5. export the participant list again, selecting "human-readable"
6. confirm the telangiectasia responses are shown in two columns -- one showing the main responses as dispaly text, and the other showing the consolidated details

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
@@ -13,7 +13,13 @@ import org.broadinstitute.dsm.util.ElasticSearchUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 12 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -22,6 +13,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.*;
```
SimoneMaiwald: checkstyle would not like that…

devonbush: ugh, IntelliJ is insisting on doing that

PR #2193, opened 2022-08-24
DDP-8616 singular email updates

ssettipalli:
singular DSM_NOTIFICATION email content fix and expression updates to handle aged-up user notifications.
https://broadinstitute.atlassian.net/browse/DDP-8616?focusedCommentId=1335433


Trimax (approved):
If I understood the idea correctly, then it LGTM.

PR #2202, opened 2022-08-25
DDP-8587: checkAnswer NPE if no answer is selected

bskinner:
DDP-8587

## Context

This PR fixes a bug in the `RenderValueProvider.checkAnswer(String,String,String,String)` method (used by the `$ddp.checkAnswer` macro) where a `NullPointerException` would be thrown in the event that an answer for the checked Stable ID was not present in the `FormResponse` context.

Existing behavior was to consider the option as selected, and return the `stringIfMatches` value, if there was no response recorded for the activity (`formResponse == null`). This existing behavior has been extended to also apply to the previously undefined case of a missing answer.

## Testing
See [DDP-8587](https://broadinstitute.atlassian.net/browse/DDP-8587) for reproduction instructions.

## Additional Notes
The below is the appearance of the activity names on the Singular Dashboard after the fix. The value of the titles are controlled by checking which option is selected in the `PATIENT_SURVEY_WHO_ENROLLING_HIDDEN` question (the `en` template is given below)
```
 "name": "About $ddp.checkAnswer(\"ABOUT_PATIENT_WHO_ENROLLING_HIDDEN\",\"CHILD\", \"My Child\", \"Me\")",
```
Screen Shot 2022-08-25 at 16 34 16

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/RenderValueProvider.java
@@ -215,13 +215,20 @@ public class RenderValueProvider {
     }
 
     /**
-     * checkAnswer
-     *
-     * @param questionStableId the question stable id
-     * @param optionStableId    the option stable Id
-     * @param stringIfMatches    string if question is answered with option stable Id
-     * @param stringOtherwise    string if question is not answered with option stable Id
-     * @return string stringIfMatches or stringOtherwise
+     * Returns a string value depending on whether a specified PICKLIST question's option
+     * is selected or not.
+     * 
+     * <p>If {@link useDefaultsForDdpMethods} is true, this method will always return
+     * a value of the form "stringIfMatches/stringOtherwise".
+     * 
+     * <p>If an activity does not have a response,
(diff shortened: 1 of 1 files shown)


MocanaAtBroad (dismissed):
Fix makes sense, but if you can rename that method to something reasonable, even better. I find myself being confused when I see the method name and in contrast to what it actually does.

bskinner:
> Fix makes sense, but if you can rename that method to something reasonable, even better. I find myself being confused when I see the method name and in contrast to what it actually does.

I agree- the name is unclear as to what the method is actually doing. However, renaming it at this point is going to be a time-intensive task as that macro is used 85 times in Singular, and replacing the method name would require each one of the impacted templates to be patched. As a stop-gap, I could move the logic to a more appropriately named method- we would need to be vigilant watching for uses of the older method in future studies, and redirect folks.

MocanaAtBroad:
Maybe update the javadoc comment for method alluding to the fact that it is awkwardly named?

bskinner:
> Maybe update the javadoc comment for method alluding to the fact that it is awkwardly named?

I like that idea- the javadoc would likely also be a better spot for my comment about the unusual default behavior.

MocanaAtBroad (approved):

PR #2210, opened 2022-08-26
[DSM] DDP-8642 Singular default value

pegahtah:
Singular default values should be in the Participant object but we were inserting a `participantData`  instead of `participant.dynamicFields` for the default value. I created a `ParticipantRecord` class to take care of creating and upserting new values


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantRecord.java
```diff
@@ -0,0 +1,101 @@
...
+                        .build();
+        return participantRecordDao.create(participantRecordDto);
+    }
+
+    public boolean insertDefaultValues(Map<String, String> data, int participantId) {
+        String additionalValues = new Gson().toJson(data);
+        participantRecordDao.insertDefaultAdditionalValues(participantId, additionalValues);
+        participant = new DefaultParticipantExporter(participantExportPayload).buildParticipantFromPayload(participantExportPayload);
+        participant.setAdditionalValuesJson(additionalValues);
+        String participantGuid = Exportable.getParticipantGuid(ddpParticipantId, ddpInstanceDto.getEsParticipantIndex());
```
ScalaChamp: I believe `ddpParticipantId` is the `guid` so we don't need to read it, or am I missing something?

gmakharat: @pegahtah you already have `guid` in `SingularDefaultValues`, you may pass it to this class to avoid extra ES call here.

pegahtah: Thanks folks ! I changed it!

ScalaChamp (approved):

PR #2211, opened 2022-08-26
[DSM] DDP-7606: received by

SimoneMaiwald:
adding received by to the response for kitrequests endpoint to be able to display "research sample" or "clinical sample" for LMS/ OSTEO

1) go to sample > receive or overview page for lms
2) select saliva kit

Screen Shot 2022-08-26 at 4 42 39 PM
2


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -301,6 +302,7 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
     private String collectionDate;
     @ColumnName(DBConstants.SEQUENCING_RESTRICTION)
     private String sequencingRestriction;
+    private String receivedBy;
```
pegahtah, line 318: If we are adding it to the participant list, or to the index, I think it should also get added to the mapping and have a `@ColumnName` on top of the field name here

SimoneMaiwald: I wasn't planning on adding it to the index yet. I only wanted it on the sample pages for now

SimoneMaiwald: yeah, let's write it in and see if we need it…

devonbush (dismissed):
looks good, confirmed works on my machine

PR #2212, opened 2022-08-27
[DSM] DDP-8561: external report

SimoneMaiwald:
External users which are allowed to see Osteo2 and LMS are allowed to see everything except the cmi mr/oncHistory tracker
Therefore this info should not get responded to the frontend for these users

adding a 'hideMedicalRecordRequestTracking' to not respond to frontend with the info external users should not see for CMI studies (if they are missing 'mr_view' permission)

to test: see frontend PR https://github.com/broadinstitute/ddp-angular/pull/1684


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -140,43 +140,46 @@ private void addWrapperToList(UnparsedESParticipantDto elasticSearchParticipantD
...
             participantWrapperDto.setEsData(elasticSearchParticipantDto);
             participantWrapperDto.setParticipant(participant);
-            participantWrapperDto.setMedicalRecords(medicalRecord);
-            participantWrapperDto.setOncHistoryDetails(oncHistoryDetails);
-            participantWrapperDto.setKits(kitRequestShipping);
             participantWrapperDto.setParticipantData(participantData);
-            participantWrapperDto.setAbstractionActivities(Collections.emptyList());
-            participantWrapperDto.setAbstractionSummary(Collections.emptyList());
+
+            if (hideMedicalRecordRequestTracking) {
```
pegahtah: Shouldn't this be `if (!hideMedicalRecordRequestTracking) {`??

SimoneMaiwald: Good catch. changed name to make it easier readable

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
```diff
@@ -46,6 +50,11 @@ public BaseFilterParticipantList(String jsonBody) {
     @Override
     public ParticipantWrapperResult filter(QueryParamsMap queryParamsMap, Deserializer deserializer) {
         this.deserializer = deserializer;
+
+        //no need to check if realm and userID is set because it was already checked in FilterRoute
```
devonbush, line 54: but this is a public method -- how do we know FilterRoute was already called?

devonbush (approved):
One question which is probably just for my information

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
SimoneMaiwald: FilterRoute is entry to the BaseFilterParticipantList

SimoneMaiwald: other class who is calling it would be `DownloadParticipantListRoute` and this one is also checking the permissions

PR #2214, opened 2022-08-28
DDP-8051 . Handle pex expr evaluation error in DataExporter

ssettipalli:
look at https://broadinstitute.atlassian.net/browse/DDP-8051
Handle pex expr evaluation error and continue export.
Error will fail to add AOM content for dsm in export but will let export continue.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -675,11 +675,17 @@ void enrichWithDSMEventDates(Handle handle,
...
-                        .orElse(null);
+                AgeOfMajorityRule aomRule = null;
+                try {
+                    aomRule = governancePolicy.getApplicableAgeOfMajorityRule(handle,
+                            pexInterpreter,
+                            participantGuid,
+                            operatorGuid)
+                            .orElse(null);
+                } catch (Exception e) {
+                    log.warn("Error while evaluating age-of-majority rules for participant {} operator {} and studyId {}, ignoring.",
```
ssettipalli: wondered if log.error is better but ended up using warn to avoid unnecessary alerts in prod.
This just misses adding dateOfMajority in elastic export for dsm to display.
Actual AgeOfMajority process is not impacted with this change.

Trimax (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
MocanaAtBroad: I would encourage you make this a log.error rather than a warning. When this happens this is something that needs to be fixed.
Also consider adding more information to the log message that is likely to allow someone from just looking at log to figure out what went wrong.
I am not sure what is the right thing to do to expand the message, but perhaps ensuring that the error you are trying to fix is debuggable from just looking a the log would be a good guide on what to do.

ssettipalli: I tried to avoid stacktrace and added e.getMessage(). I will add stack trace.

MocanaAtBroad (approved):

PR #2254, opened 2022-09-07
DDP-8554. Border fixes

Trimax:


bskinner (commented):
Does this patch successfully apply against an install which has already had the previous version run?

MocanaAtBroad:
@bskinner I am pretty sure this is a modification to @Trimax's previous patch. Has been freakin' tricky to get this right!

MocanaAtBroad:
@Trimax Left you comment on my own PR. Would welcome suggested changes (or a PR on my PR) that would add the same checks you have been doing for AOM to the new medical release v2.
https://github.com/broadinstitute/ddp-study-server/pull/2256

MocanaAtBroad (approved):

Trimax:
> Does this patch successfully apply against an install which has already had the previous version run?

Yep, it is.

PR #2265, opened 2022-09-08
final scan feedback for pe-cgs

SimoneMaiwald:
adding user feedback to final scan page so that study staff sees the hruid after the final scan was successful. 
only for pe-cgs because for other studies 
```
StringUtils.isEmpty(kitRequestShipping.getKitLabel()) ? null :
                                        kitRequestShipping.getBspCollaboratorParticipantId());
``` 
will be null


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
```diff
@@ -55,7 +57,9 @@ protected Optional<ScanError> process(ScanPayload scanPayload) {
             } else {
                 if (StringUtils.isNotEmpty(kitRequestShipping.getKitLabel()) && kitLabel.equals(kitRequestShipping.getKitLabel())
                         || StringUtils.isEmpty(kitRequestShipping.getKitLabel())) {
-                    result = updateKitRequest(kitLabel, ddpLabel, kitRequestShipping.getBspCollaboratorParticipantId());
+                    result = updateKitRequest(kitLabel, ddpLabel,
+                            StringUtils.isEmpty(kitRequestShipping.getKitLabel()) ? null :
```
gmakharat: it is duplication, the same code is used on 46 line, extract it as a method and use that method in both place.

pegahtah (approved):

PR #2283, opened 2022-09-12
DDP-8730: Attempt to implement a feature that will provide activity i…

Trimax:
A new feature was introduced - a variable substituted with the `firstCompletedAt` date.

Снимок экрана 2022-09-12 в 16 04 01
Снимок экрана 2022-09-12 в 15 42 22


MocanaAtBroad:
This looks really really close, but one important issue: we are not looking for instance creation date. We are looking for the date when the activity instance reaches COMPLETE status.Date of creation of the instance can potentially be months before!

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -4,9 +4,7 @@
 import static org.broadinstitute.ddp.content.VelocityUtil.VARIABLE_PREFIX;
 
 import java.io.StringWriter;
-import java.time.LocalDate;
-import java.time.ZoneId;
-import java.time.ZoneOffset;
+import java.time.*;
```
bskinner: Wildcard imports should not be used.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -100,6 +106,14 @@ public static RenderValueProvider.Builder newValueProviderBuilder(Handle handle,
         builder.setParticipantTimeZone(zone);
         builder.setDate(LocalDate.now(zone));
 
+        builder.setActivityInstanceCreationDate(handle.attach(ActivityInstanceDao.class)
+                .findBaseResponseByInstanceId(activityInstanceId)
+                .map(ActivityResponse::getFirstCompletedAt)
+                .map(Instant::ofEpochMilli)
+                .map(x -> x.atZone(ZoneId.systemDefault()))
```
bskinner: ~There's a resolved ZoneId for the participant above- is there a reason that being ignored and the system's default zone is being used instead?~

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
Trimax: I know. It is still in Draft. Experimenting with submissionDate. Checkstyle is the last step before making this PR ready to review.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
bskinner: Ah, I see what's going on here. 👍

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
```diff
@@ -486,8 +486,10 @@ public void renderInstanceSummaries(Handle handle, long userId, String operatorG
             substitutions = substitutionStream
                     .collect(Collectors.toMap(SubstitutionsWrapper::getActivityInstanceId, SubstitutionsWrapper::unwrap));
         }
+
+        //TODO: Not sure if it is correct behaviour
         var sharedSnapshot = I18nContentRenderer
-                .newValueProviderBuilder(handle, userId, operatorGuid, studyGuid)
+                .newValueProviderBuilder(handle, instanceIds.iterator().next(), userId, operatorGuid, studyGuid)
```
Trimax: This is the weak place in my code. What instance ID should we use when the code works for multiple instances and creates a shared snapshot? I believe that any of them?

pepper-apis/studybuilder-cli/studies/singular/activities/consent-self.conf
```diff
@@ -2289,7 +2289,7 @@
           "bodyTemplate": {
             "templateType": "HTML",
             "templateText": """
-              <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
+              <p class="paragraph">$ddp.submissionDate("MM / dd / yyyy", $ddp.date("MM / dd / yyyy"))</p>
```
MocanaAtBroad: Let's not change this file. We will need to make the change to consent form via a patch.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nTemplateConstants.java
```diff
@@ -11,6 +11,7 @@ public static final class Snapshot {
         public static final String PARTICIPANT_LAST_NAME = "DDP_PARTICIPANT_LAST_NAME";
         public static final String PARTICIPANT_BIRTH_DATE = "DDP_PARTICIPANT_BIRTH_DATE";
         public static final String PARTICIPANT_TIME_ZONE = "DDP_PARTICIPANT_TIME_ZONE";
+        public static final String DDP_SUBMISSION_DATE = "DDP_SUBMISSION_DATE";
```
MocanaAtBroad: Let's rename to DDP_DATE_FIRST_COMPLETED . More consistent with what this really is.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/ActivityInstanceService.java
bskinner: I'm looking through the code now, but I do not believe this is the correct behavior as adding an instance-specific id changes the context of the `I18nContentRenderer`, and this method may generate summaries for activities unrelated to the instance ID used (before this, `I18nContentRenderer` was invariant over all activities, but now it is not). My suggestion at this point would be to pass `null` for the instance ID and make sure that `I18nContentRenderer` properly handles the value (AFAIK, none of the activity summaries are making use of this value)

MocanaAtBroad: Can we implement the needed functionality without saving data as snapshot? firstCompletedAt value for activity should always be available (once activity is completed).

bskinner: Should be easy enough, but it may have a performance impact since the `I18nContentRenderer` will be regenerated for _each_ activity summary (since there is no guarantee that all the activities here are related). I do not know if the current process of caching the snapshot is a case of premature optimization or not.

pepper-apis/studybuilder-cli/studies/singular/activities/consent-self.conf
bskinner: Fixed in 61a2b9c100a9c06ee6f8727d30d1e1c5b26ca9ac

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nTemplateConstants.java
bskinner: Should this change be applied to all `submissionDate` references?

pepper-apis/studybuilder-cli/studies/singular/activities/consent-self.conf
MocanaAtBroad: Code looks reasonable and appears to be working. Could we scrutinize further regarding how these changes would affect studies that are using the snapshot feature , i.e.,. that have activities with the snapshotSubstitutionsOnSubmit=true flag .
I BELIEVE it is OK, but would welcome a closer look with the users of the snapshot functionality in mind.

MocanaAtBroad (approved):

PR #2289, opened 2022-09-13
Rc onc history fix

SimoneMaiwald:
adding logic to patch oncHistoryDetail of parentId is participantId and participantId is null but ddpParticipantId is set

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
@@ -1,12 +1,9 @@
 package org.broadinstitute.dsm.model.patch;
 
 import java.util.Objects;
-import java.util.Optional;
 
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.OncHistoryDetail;
-import org.broadinstitute.dsm.db.dao.ddp.participant.ParticipantDao;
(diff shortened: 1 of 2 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -29,6 +29,9 @@ public static BasePatch makePatch(Patch patch, NotificationUtil notificationUtil
             } else if (isParticipantIdForRecord(patch)) {
                 patcher = new ParticipantRecordPatch(patch);
             }
+        } else if (isParentParticipantId(patch) && !isMedicalRecordAbstractionFieldId(patch)
```
gmakharat, line 32: @SimoneMaiwald does this boolean condition apply only for onc history detail patch? I am afraid that we may get into this if statement for other cases as well and we may face some exceptions

SimoneMaiwald: all other patches should have already been caught by all the logic before

SimoneMaiwald: all other patches at this point would throw a NullPatch. so either the patch works because it is a oncHistoryDetail field and the process -> update db will work because the column is the correct on or it will throw an error… but it would have thrown an NullPatch error next anyways…

PR #2291, opened 2022-09-13
PS: DDP-8730 - Add patch Singular to use ddp.firstCompletedDate wherever ddp.date is used

MocanaAtBroad:
Code is pretty simple: look for all the places where we use ddp.date in template_text and put in a new Velocity html template fragment that includes the new ddp.firstCompletedDate

If you do the following query:
```sql
SELECT template_text 
FROM
    study_activity sa
        JOIN
    umbrella_study us ON sa.study_id = us.umbrella_study_id
        JOIN
    form_activity fa ON fa.study_activity_id = sa.study_activity_id
        JOIN
    form_activity__form_section fafs ON fafs.form_activity_id = fa.study_activity_id
        JOIN
    form_section fs ON fs.form_section_id = fafs.form_section_id
        JOIN
    form_section__block fsb ON fsb.form_section_id = fs.form_section_id
        JOIN
    block_content bc ON fsb.block_id = bc.block_id
        JOIN
    template t ON bc.body_template_id = t.template_id
WHERE
    us.guid = 'singular'
AND t.template_text LIKE '%$ddp.date%'
```

You get following results:
Screen Shot 2022-09-12 at 11 39 52 PM

Following command also renders the same thing, which gives me more assurance that this is correct:
```bash
find pepper-apis/studybuilder-cli/studies/singular -name "*.conf" | grep -v newFirstCreatedAt | xargs grep -Eh '\$ddp\.date'
```
```
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
            <p class="paragraph">$ddp.date("MM / dd / yyyy")</p>
```

Tested locally by:
- updating the Singular study configuration to what is currently in v1.32 (original ps-init-version plus patches in v1.32), 
- signed up locally, created a few activities, completing some but not all
- applied patch in this PR
- set the created_at and first_completed_at times in all activity instances in Singular back 24 hours
- set every revision start_date and revision end_date back 24 hours (if instances are older than the revision timestamp, nothing works)
- built and ran backend using DDP-8730
Below is screenshot of user consent that I had created earlier, but had set the timestamps back
Note that while it is 9/12 today it show 9/11
I created new activites and they display todays date as expected.
Screen Shot 2022-09-12 at 11 11 20 PM

### DIFF
--- a/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularUpdateActivitiesToUseDateFirstCompleted.java
+++ b/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/SingularUpdateActivitiesToUseDateFirstCompleted.java
@@ -1,0 +1,62 @@
+package org.broadinstitute.ddp.studybuilder.task;
+
+import java.io.File;
+import java.nio.file.Path;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.ddp.exception.DDPException;
+import org.jdbi.v3.core.Handle;
+
+@Slf4j
+public class SingularUpdateActivitiesToUseDateFirstCompleted implements CustomTask {
+    private static final String STUDY_GUID = "singular";
+    private static final String EXISTING_DATE_TEMPLATE_PATTERN = "%$ddp.date%";
+    private static final String DATA_FILE = "patches/newFirstCreatedAtDateTemplateString.conf";
+
+    private static final int EXPECTED_NUM_ROWS_TO_BE_MODIFIED = 12;
+    private static final String QUESTION_TEMPLATE_VARIABLE_SUBSTITUTION_VALUE = "UPDATE template t"
+            + " JOIN block_content bc ON bc.body_template_id=t.template_id"
+            + " JOIN form_section__block fsb ON fsb.block_id=bc.block_id"
+            + " JOIN form_section fs ON fsb.form_section_id=fs.form_section_id"
+            + " JOIN form_activity__form_section fafs ON fs.form_section_id=fafs.form_section_id"
+            + " JOIN form_activity fa ON fafs.form_activity_id = fa.study_activity_id"
+            + " JOIN study_activity sa ON fa.study_activity_id = sa.study_activity_id"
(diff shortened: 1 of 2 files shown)


bskinner (approved):
👍 

I would appreciate it if others would double check the count on the changed values. My local environment returned `11`, but it looks like both `dev` and `test` are returning `12`, so I'm writing this off as my local environment being out of date. A second set of eyes would be welcomed, though, just in case.

Trimax (approved):

Trimax:
> 👍
> 
> I would appreciate it if others would double check the count on the changed values. My local environment returned `11`, but it looks like both `dev` and `test` are returning `12`, so I'm writing this off as my local environment being out of date. A second set of eyes would be welcomed, though, just in case.

I have nearly the same (10 vs 12). But before running the patch I reimported singular study locally.
```
10:51:48.558   C: S: [main] INFO  o.b.d.s.t.SingularUpdateActivitiesToUseDateFirstCompleted Modified number of rows: 10
Exception in thread "main" org.broadinstitute.ddp.exception.DDPException: Was expecting exactly: 12 Rolling back
```

MocanaAtBroad:
> > 👍
> > I would appreciate it if others would double check the count on the changed values. My local environment returned `11`, but it looks like both `dev` and `test` are returning `12`, so I'm writing this off as my local environment being out of date. A second set of eyes would be welcomed, though, just in case.
> 
> I have nearly the same (10 vs 12). But before running the patch I reimported singular study locally.
> 
> ```
> 10:51:48.558   C: S: [main] INFO  o.b.d.s.t.SingularUpdateActivitiesToUseDateFirstCompleted Modified number of rows: 10
> Exception in thread "main" org.broadinstitute.ddp.exception.DDPException: Was expecting exactly: 12 Rolling back
> ```

The query returns 12 rows for me locally, on dev, test, and staging.

PR #2297, opened 2022-09-13
Ddp 8726 osteo about you fix

LKapanad:
DDP-8726 osteo about-you pediatric fix
Created a patch
Updated dss code to handle numbering of blocks correctly: old numbering code didn't consider whether the block is hidden or not. since ABOUTYOU activity is dynamic for both child and adult, hidden block was still numbered and the first question had displayNumber=2.

### DIFF
--- a/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/osteo/OsteoAboutYouChildV2.java
+++ b/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/osteo/OsteoAboutYouChildV2.java
@@ -1,0 +1,102 @@
+package org.broadinstitute.ddp.studybuilder.task.osteo;
+
+import com.google.gson.Gson;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import lombok.extern.slf4j.Slf4j;
+import org.broadinstitute.ddp.db.dao.ActivityDao;
+import org.broadinstitute.ddp.db.dao.JdbiExpression;
+import org.broadinstitute.ddp.db.dao.JdbiUmbrellaStudy;
+import org.broadinstitute.ddp.db.dao.SectionBlockDao;
+import org.broadinstitute.ddp.db.dto.ActivityVersionDto;
+import org.broadinstitute.ddp.exception.DDPException;
+import org.broadinstitute.ddp.model.activity.definition.FormBlockDef;
+import org.broadinstitute.ddp.studybuilder.ActivityBuilder;
+import org.broadinstitute.ddp.studybuilder.task.CustomTask;
+import org.broadinstitute.ddp.util.ConfigUtil;
+import org.broadinstitute.ddp.util.GsonUtil;
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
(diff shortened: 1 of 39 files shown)


pepper-apis/studybuilder-cli/studies/osteo/patches/about-child-v2.conf
```diff
@@ -0,0 +1,73 @@
+{
+  "searchSectionByQuestionCode":"WHO_IS_FILLING_ABOUTYOU",
+  "shownExpr":"""!user.studies["CMI-OSTEO"].isGovernedParticipant()""",
+  "who_filling_q_child": {
+    "question": {
+      include required("../../snippets/picklist-question-single-list.conf"),
+      "renderMode": "CHECKBOX_LIST",
+      "hideNumber": true,
+      "stableId": "WHO_IS_FILLING_ABOUTYOU_CHILD",
```
MocanaAtBroad, line 10: I just added these 2 options.
So now there is a checkbox instead of a radio button (which allows user to unselect).
And the number besides the question prompt goes away with hide number.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/osteo/OsteoAboutYouChildV2.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        long revisionId = activeVersion.get().getRevId();
+        long stableId = helper.getQuestionStableIdByCode(dataCfg.getString("searchSectionByQuestionCode"), studyDto.getId());
+        long blockId = helper.getBlockIdByStableId(stableId);
+        long sectionId = helper.getSectionIdByBlockId(blockId);
+        String guid = handle.attach(JdbiExpression.class).generateUniqueGuid();
+        long shownExpr = helper.insertExpression(guid, dataCfg.getString("shownExpr"));
+        helper.insertBlockExpression(blockId, shownExpr, revisionId);
+        FormBlockDef blockDefChild = gson.fromJson(ConfigUtil.toJson(dataCfg.getConfig("who_filling_q_child")), FormBlockDef.class);
+        sectionBlockDao.insertBlockForSection(
```
MocanaAtBroad, line 88: After I removed your changes to ActivityService (too dangerous to try that on every single one of our 20 studies at this point!) and modified the .conf file (see my comments there) the page is looking like this:
Screen Shot 2022-09-13 at 9 35 23 PM

MocanaAtBroad: I think the reason for that **1.** is that you are inserting a new block to hold the question and the old block is still there.
I BELIEVE that if before calling `insertBlockForSection` you delete blocks in the section, that 1 will go away.
I am not really setup to try this stuff/did not have time to setup Osteo locally, but I would encourage you to try clean this last thing.

pepper-apis/studybuilder-cli/studies/osteo/patches/about-child-v2.conf
```diff
@@ -0,0 +1,74 @@
...
+                    "language": "en",
+                    "text": "My child has been diagnosed with osteosarcoma, and I am filling out this survey on their behalf."
+                  }
+                ]
+              }
+            ]
+          },
+          "detailLabelTemplate": null,
+          "allowDetails": false,
+          "exclusive": false
```
zyme, line 67: Now that we are hiding this option for adult, please return these to radio buttons for child consent so that only one can be selected (not sure if that means setting `exclusive` as well--you know better than I do).

MocanaAtBroad (approved):

PR #2298, opened 2022-09-13
[DDP-8756] File upload notification email proxy & error handling

bskinner:
DDP-8756

## Context

* Fixes an issue where the SendGrid client used for the file upload notification issues would not respect the proxy settings, if configured.
* Errors from the SendGrid client were not being completely logged

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
@@ -84,8 +84,10 @@ public class FileUploadService {
         long removalExpireTime = cfg.getLong(ConfigFile.FileUploads.REMOVAL_EXPIRE_TIME);
         TimeUnit removalExpireUnit = TimeUnit.valueOf(cfg.getString(ConfigFile.FileUploads.REMOVAL_EXPIRE_UNIT));
 
+        var apiKey = cfg.getString(ConfigFile.SENDGRID_API_KEY);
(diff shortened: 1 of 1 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
```diff
@@ -84,8 +84,10 @@ public static FileUploadService fromConfig(Config cfg) {
         long removalExpireTime = cfg.getLong(ConfigFile.FileUploads.REMOVAL_EXPIRE_TIME);
         TimeUnit removalExpireUnit = TimeUnit.valueOf(cfg.getString(ConfigFile.FileUploads.REMOVAL_EXPIRE_UNIT));
 
+        var apiKey = cfg.getString(ConfigFile.SENDGRID_API_KEY);
+        var connectProxy = ConfigUtil.getStrIfPresent(cfg, ConfigFile.Sendgrid.PROXY);
         return new FileUploadService(
-                new SendGridClient(cfg.getString(ConfigFile.SENDGRID_API_KEY)),
+                new SendGridClient(apiKey, connectProxy),
```
MocanaAtBroad, line 90: I checked vault for all our environments. We proxy address in dev, staging, and prod but not in test.
Does not to be a problem though.

MocanaAtBroad (approved):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/FileUploadService.java
bskinner: In that case, this will just fallback to existing behavior (which is fine- test does not need the proxy)

PR #2318, opened 2022-09-16
DDP-8824: changed query after configuration changes made on prod for OS2

SimoneMaiwald:
query for viewFilters would display filters for 14 when searching for 4


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -277,15 +277,16 @@ public static List<ViewFilter> getAllFilters(@NonNull String userId, @NonNull Ma
...
             }
-            query = query.replaceAll("(%#%)", "%" + realm + ",%");
 
             try (PreparedStatement stmt = conn.prepareStatement(query)) {
                 stmt.setString(1, userId);
                 stmt.setString(2, ddpGroupId);
                 stmt.setString(3, ddpGroupId);
                 stmt.setString(4, ddpGroupId);
+                stmt.setString(5, "%," + realm + ",%");
+                stmt.setString(6, "%[" + realm + ",%");
```
zyme, line 287: Did we mean to put `%[` here instead of just `%`?  I'm not familiar with that syntax.

PR #2320, opened 2022-09-19
DDP-8742. By default, use US/Central timezone instead of UTC.

Trimax:
When rendering the date, by default we use US/Central timezone. If the user profile has a timezone specified, it is used instead.


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
```diff
@@ -91,7 +91,7 @@ public static RenderValueProvider.Builder newValueProviderBuilder(Handle handle,
                 .findProfileByUserId(participantUserId)
                 .orElse(null);
 
-        ZoneId zone = ZoneOffset.UTC;
+        ZoneId zone = TimeZone.getTimeZone("US/Central").toZoneId();
```
bskinner, line 94: Currently we are defining the default timezone for the server in the configuration file under the `defaultTimezone` key, and the `UTC` timezone is hardcoded elsewhere in the server (these are distinct from the GAE system default timezone, which we cannot change). Are there any cases where this could potentially lead to two dates being rendered in separate timezones?

bskinner (approved):
Looks good & fixes the immediate issue- it would be nice to move this sort of information to the session, and pass it into the renderer in the future.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/content/I18nContentRenderer.java
Trimax: Well, I searched through the project by `.UTC` and by `ZoneOffset`. Looks like we use UTC everywhere.

Trimax: Also, we have export to the CSV file functionality in `CustomExportCoordinator` class. It uses the UTC timezone. Might be inconsistent in this way.

Trimax: `DataExporter` does the same.

Actually, that's why I absolutely don't like the idea to play with the dates on the backend side. We have everything in UTC and now we would like to change one place to a different TZ. Due to our code coupling and complexity, I can't guarantee we will not have any side effects in other places. Besides, it is really difficult to test. When I tested it, I used Japan time instead (the furthest point from my location with a significant TZ difference) - it worked. But you raised a really good question here. The front end, on the other hand, most likely has a special component for the date rendering. And assuming that the date always comes in the UTC TZ it is one line change there to convert it to whatever TZ we want and for all dates we use. @alextanlan, could you please confirm?
@MocanaAtBroad, @bskinner, what do you think?

PR #2334, opened 2022-09-21
[DSM] DDP-8810_text_field_size

pegahtah:
I added a `details` variable to FieldSettings Object and the column to the table. 
We still have to set it for the fields when they are added per DDP-8808
frontend PR: [here](https://github.com/broadinstitute/ddp-angular/pull/1770)


pegahtah:
I would like to merge this PR and it's corresponding [fronted PR](https://github.com/broadinstitute/ddp-angular/pull/1770) before the DDP-8808 pull requests to make changes in that one based on the new code

pepper-apis/dsm-core/src/main/resources/liquibase/singular/DDP-8810_feild_settings_details_column.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="DDP-8810_field_settings_details_columns" author="pegah">
+        <addColumn tableName="field_settings">
+            <column name="details" type="VARCHAR(500)" afterColumn="actions"></column>
```
gmakharat, line 9: what is `details` for?

pegahtah: So the ticket asks for a way to dynamically change some text area sizes and limits, we will add the new "details" of each field in this `details` column. ( So for the fields added per DDP-8808 it will be like `{size:100}`. we can use this column for other features of the fields in future too ( color, font,  readonly, etc)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/FieldSettings.java
```diff
@@ -106,6 +115,10 @@ public static Map<String, Collection<FieldSettings>> getFieldSettings(@NonNull S
                             fieldSettingsList.put(type, subList);
                         }
                     }
+                } catch (JsonMappingException e) {
+                    logger.error("Unable to get the FieldSetting ", e);
+                } catch (JsonProcessingException e) {
+                    logger.error("Unable to parse the FieldSetting json ", e);
```
SimoneMaiwald: please add the field_settings_id or column_name here as well

pegahtah: @SimoneMaiwald fixed! Thank you

SimoneMaiwald (approved):

PR #2351, opened 2022-10-06
[DDP-8508] Track uploaded at date for file uploads

bskinner:
DDP-8508

## Context

This PR updates the `FileAnswer` handling such that when a client persists an uploaded file reference, the `uploaded_at` date of the `FileUpload` value is set to the update time of the answer. The justification used for this is, once a client has created & uploaded a file, the API expects clients to make a `PATCH` request to the `../answers` to finalize the upload.

### DIFF
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
@@ -102,8 +102,7 @@ public class ConfigUtil {
      * Get sql from config file
      * @param queryName name of query
      * @return the SQL for query
-     * @deprecated
-     * please use JDBI and daos instead of looking up
+     * @deprecated please use JDBI and daos instead of looking up
      */
     @Deprecated
     public static String getSqlFromConfig(@NonNull String queryName) {
@@ -112,8 +111,7 @@ public class ConfigUtil {
 
     /**
      * Get sql from config file
-     * @deprecated
-     * please use JDBI and daos instead of looking up SQL snippets from config files
+     * @deprecated please use JDBI and daos instead of looking up SQL snippets from config files
      */
     @Deprecated
     public static String getSqlFromConfig(String sourceName, @NonNull String queryName) {
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.ddp.util;
 
 import java.time.Instant;
-import java.util.Optional;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigRenderOptions;
@@ -52,18 +51,6 @@ public class ConfigUtil {
      */
     public static String getStringOrElse(Config cfg, String key, String defaultValue) {
         return cfg.hasPath(key) ? cfg.getString(key) : defaultValue;
-    }
-
-    public static Optional<String> getString(Config cfg, String key) {
-        if (cfg.hasPathOrNull(key)) {
-            if (cfg.getIsNull(key)) {
-                return Optional.empty();
-            } else {
-                return Optional.of(cfg.getString(key));
-            }
-        } else {
-            return Optional.empty();
-        }
     }
 
     /**
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.ddp.util;
 
 import java.time.Instant;
+import java.util.Optional;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigRenderOptions;
@@ -51,6 +52,18 @@ public class ConfigUtil {
      */
     public static String getStringOrElse(Config cfg, String key, String defaultValue) {
         return cfg.hasPath(key) ? cfg.getString(key) : defaultValue;
+    }
+
+    public static Optional<String> getString(Config cfg, String key) {
+        if (cfg.hasPathOrNull(key)) {
+            if (cfg.getIsNull(key)) {
+                return Optional.empty();
+            } else {
+                return Optional.of(cfg.getString(key));
+            }
+        } else {
+            return Optional.empty();
+        }
     }
 
     /**
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.ddp.db.dao;
 
 import java.math.BigDecimal;
+import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -175,6 +176,7 @@ public interface AnswerDao extends SqlObject {
         final var uploadedAt = getAnswerSql()
                 .findDtoById(answerId)
                 .map(AnswerDto::getLastUpdatedAt)
+                .map(timestamp -> Instant.ofEpochMilli((long)timestamp))
                 .orElse(Instant.now());
         final var fileUploadDao = getHandle().attach(FileUploadDao.class);
         fileUploadDao.findByIds(uploadIds.stream().mapToLong(id -> id).toArray())
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -165,6 +165,11 @@ public interface AnswerDao extends SqlObject {
             throw new DaoException(message);
         }
 
+        if (uploadIds.isEmpty()) {
+            // No file uploads specified- nothing more to do here.
+            return;
+        }
+
         /*
          * The answer dto needs to be pulled directly here as the `Answer` class does not
          *  include the lastUpdatedAt time. It could be added- the ideal spot would likely be
@@ -178,6 +183,7 @@ public interface AnswerDao extends SqlObject {
                 .map(timestamp -> Instant.ofEpochMilli((long)timestamp))
                 .orElse(Instant.now());
         final var fileUploadDao = getHandle().attach(FileUploadDao.class);
+
         fileUploadDao.findByIds(uploadIds.stream().mapToLong(id -> id).toArray())
                 .forEach(record -> {
                     fileUploadDao.updateStatus(record.getId(),
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -158,11 +158,10 @@ public interface AnswerDao extends SqlObject {
         final List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
         final int[] inserted = getAnswerSql().bulkInsertFileValue(answerId, uploadIds);
         if (inserted.length != uploadIds.size()) {
-            final var message = String.format("Failed to insert uploads ids for [answer:%s] (%s inserted, %s expected)",
+            throw new DaoException(String.format("Failed to insert uploads ids for [answer:%s] (%s inserted, %s expected)",
                     answerId,
                     inserted.length,
-                    uploadIds.size());
-            throw new DaoException(message);
+                    uploadIds.size()));
         }
 
         if (uploadIds.isEmpty()) {
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -481,7 +481,7 @@ public interface AnswerDao extends SqlObject {
             boolean isChildAnswer = view.getColumn("is_child_answer", Boolean.class);
             String actInstanceGuid = view.getColumn("activity_instance_guid", String.class);
 
-            Answer answer;
+            final Answer answer;
             switch (type) {
                 case AGREEMENT:
                     answer = new AgreementAnswer(answerId, questionStableId, answerGuid, view.getColumn("aa_value", Boolean.class),
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.ddp.db.dao;
 
 import java.math.BigDecimal;
-import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
@@ -154,11 +154,36 @@ public interface AnswerDao extends SqlObject {
         if (answer.getValue() == null) {
             return;
         }
-        List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
-        int[] inserted = getAnswerSql().bulkInsertFileValue(answerId, uploadIds);
+
+        final List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
+        final int[] inserted = getAnswerSql().bulkInsertFileValue(answerId, uploadIds);
         if (inserted.length != uploadIds.size()) {
-            throw new DaoException("Not all file uploads were assigned to answer " + answerId);
+            final var message = String.format("Failed to insert uploads ids for [answer:%s] (%s inserted, %s expected)",
+                    answerId,
+                    inserted.length,
+                    uploadIds.size());
+            throw new DaoException(message);
         }
+
+        /*
+         * The answer dto needs to be pulled directly here as the `Answer` class does not
+         *  include the lastUpdatedAt time. It could be added- the ideal spot would likely be
+         *  in {@link AnswerWithValueReducer}- but that would involve altering the ctors for a
+         *  very widely used base class. This should work for the time being.
+         *  (bskinner - 20221007)
+         */
+        final var uploadedAt = getAnswerSql()
+                .findDtoById(answerId)
+                .map(AnswerDto::getLastUpdatedAt)
+                .orElse(Instant.now());
+        final var fileUploadDao = getHandle().attach(FileUploadDao.class);
+        fileUploadDao.findByIds(uploadIds.stream().mapToLong(id -> id).toArray())
+                .forEach(record -> {
+                    fileUploadDao.updateStatus(record.getId(),
+                            uploadedAt,
+                            record.getScannedAt(),
+                            record.getScanResult());
+                });
     }
 
     private void createAnswerCompositeValue(long operatorId, long instanceId, long answerId, CompositeAnswer answer) {
--- a/pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiCompositeAnswer.sql.stg
+++ b/pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiCompositeAnswer.sql.stg
@@ -3,53 +3,55 @@ group JdbiCompositeAnswer;
 //Note the 2 left joins
 getCompositeAnswerSummary() ::= <<
 SELECT
-	pa.answer_id AS parent_answer_id,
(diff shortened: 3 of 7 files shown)


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
```diff
@@ -76,4 +87,73 @@ public void deleteBlob(Blob blob) {
...
+             * This does _not_ preclude other notifications with different ETags,
+             * and other attributes.
+             * 
+             * None of the additional attributes are being used as of this time.
+             */
+            return true;
+        }
+
+        var createdNotification = storage.createNotification(bucket, notificationInfo);
+        return (createdNotification != null);
```
bskinner: The intent of this block is to allow for automated setup of the necessary Cloud Storage notifications. Since there is no identifier for these notifications being persisted, heuristics needs to be used to determine if the existing subscriptions (if there are any) are needed or not in an effort to prevent creating multiple notifications.

I'm on the fence about this code as, while it does ease setup, it introduces risk in managing the notifications (including potential privilege limitations).

See the description of the PR for the manual process for setting up the required notification with `gsutil`.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
```diff
@@ -202,17 +202,35 @@ public static class Auth0LogEvents {
...
+        public static final String MAX_FILE_SIZE_BYTES = prefix + MAX_FILE_SIZE_BYTES_KEY;
+        public static final String MAX_SIGNED_URL_MINS = prefix + MAX_SIGNED_URL_MINS_KEY;
+        public static final String REMOVAL_EXPIRE_TIME = prefix + REMOVAL_EXPIRE_TIME_KEY;
+        public static final String REMOVAL_EXPIRE_UNIT = prefix + REMOVAL_EXPIRE_UNIT_KEY;
+        public static final String REMOVAL_BATCH_SIZE = prefix + REMOVAL_BATCH_SIZE_KEY;
+        public static final String ENABLE_SCAN_RESULT_HANDLER = prefix + ENABLE_SCAN_RESULT_HANDLER_KEY;
+        public static final String SCAN_RESULT_SUBSCRIPTION = prefix + SCAN_RESULT_SUBSCRIPTION_KEY;
+        public static final String UPLOADS_NOTIFICATION_TOPIC = prefix + UPLOADS_NOTIFICATION_TOPIC_KEY;
+        public static final String UPLOADS_NOTIFICATION_SUBSCRIPTION = prefix + UPLOADS_NOTIFICATION_SUBSCRIPTION_KEY;
+        public static final String SIGNER_SERVICE_ACCOUNT = prefix + SIGNER_SERVICE_ACCOUNT_KEY;
```
bskinner: In the app we are using an inconsistent mix of absolute and relative keypaths when accessing configuration values. The `_KEY` suffixed values were added to the above so that only the subconfig was required for the setup of the file upload notifications, and any required values from the global configuration were passed in via method parameters.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
```diff
@@ -53,6 +54,18 @@ public static String getStringOrElse(Config cfg, String key, String defaultValue
...
+            if (cfg.getIsNull(key)) {
+                return Optional.empty();
+            } else {
+                return Optional.of(cfg.getString(key));
+            }
+        } else {
+            return Optional.empty();
+        }
+    }
+
```
bskinner: Added as the existing `getStringOrElse` and `getStrIfPresent` methods will throw an error if the specified key is explicitly set to `null`.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
```diff
@@ -112,8 +124,7 @@ public static String getSqlFromConfig(@NonNull String queryName) {
 
     /**
      * Get sql from config file
-     * @deprecated
-     * please use JDBI and daos instead of looking up SQL snippets from config files
+     * @deprecated please use JDBI and daos instead of looking up SQL snippets from config files
```
bskinner, line 114: No logic changes here- Just fixes to keep checkstyle happy.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -458,26 +475,33 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
                 log.info("Housekeeping interrupted during sleep", e);
             }
         }
-        pubsubConnectionManager.close();
+        
+        log.info("Housekeeping is shutting down");
```
bskinner: Log that Housekeeping is shutting down _before_ beginning to shut down.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -538,22 +562,87 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
+        if (!enabled) {
+            log.info("File upload handler is disabled ({} == false)", ConfigFile.FileUploads.ENABLE_SCAN_RESULT_HANDLER_KEY);
+            return;
+        }
+
+        final var configSubscriptionNamePath = ConfigFile.FileUploads.UPLOADS_NOTIFICATION_SUBSCRIPTION_KEY;
+
+        if (!fileUploadConfig.hasPath(projectId)) {
+            log.info("A value for the key {} was not specified, file upload notifications will not be tracked.",
+                    ConfigFile.FileUploads.UPLOADS_NOTIFICATION_SUBSCRIPTION);
```
bskinner: The use of the absolute keypath here is intentional, but I have no strong ties to it. If folks feel this would improve consistency, I'm on-board for changing this to the `_KEY` suffixed value

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -538,22 +562,87 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
+        //
+        // If the credentialed account does not have sufficient privileges for this operation,
+        // an error will be thrown.
+        final var configTopicNamePath = ConfigFile.FileUploads.UPLOADS_NOTIFICATION_TOPIC_KEY;
+        ConfigUtil.getString(fileUploadConfig, configTopicNamePath)
+                .map(name -> ProjectTopicName.of(projectId, name))
+                .ifPresentOrElse(topicName -> {
+                    pubsubManager.createTopicIfNotExists(topicName);
+                    storageClient.enableNotifications(uploadsBucketName, topicName, uploadsNotificationsReceiver.acceptedEventTypes());
+
```
bskinner: If the housekeeping credentials do not have sufficient privileges to create the topics and enable notifications, this will result in an authentication error from google. This should be considered expected behavior. If topic & notification creation is not desired, the `fileUploads.uploadsNotificationTopic` key in the configuration may be omitted or `null`ed out.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -538,22 +562,87 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
+                .setThreadFactory(threadFactory)
+                .build();
+
+        final var subscriberBuilder = Subscriber.newBuilder(configSubscriptionName, uploadsNotificationsReceiver)
+                .setExecutorProvider(executorProvider);
+            
+        startNewSubscriberWithRecovery(configSubscriptionName,
+                subscriberBuilder,
+                executorProvider,
+                MoreExecutors.directExecutor(),
```
bskinner: Ensure the callback occurs on the dispatching thread.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -538,22 +562,87 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
+                    log.info("Started file uploads subscriber for subscription {}", configSubscriptionName);
+                    uploadsNotificationSubscriber = subscriber;
+                });
+    }
+
+    private static void setupFileScanResultReceiver(Config cfg, GoogleCredentials credentials, String projectId) {
         boolean enabled = cfg.getBoolean(ConfigFile.FileUploads.ENABLE_SCAN_RESULT_HANDLER);
         if (!enabled) {
-            log.warn("File scan result handler is not enabled");
+            log.warn("File scan result handler is disabled ({} == false)", ConfigFile.FileUploads.ENABLE_SCAN_RESULT_HANDLER);
```
bskinner: Adds a hint where to look if the disabled message is being printed.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -538,22 +562,87 @@ private static void setupTaskReceiver(Config cfg, String projectId) {
...
 
         var subName = ProjectSubscriptionName.of(projectId,
                 cfg.getString(ConfigFile.FileUploads.SCAN_RESULT_SUBSCRIPTION));
 
-        boolean ensureDefault = cfg.getBoolean(ConfigFile.REQUIRE_DEFAULT_GCP_CREDENTIALS);
-        GoogleCredentials credentials = GoogleCredentialUtil.initCredentials(ensureDefault);
-        if (credentials == null) {
-            throw new DDPException("Could not get bucket credentials");
-        }
-
```
bskinner: This was moved to the enclosing method, and the credentials are passed in via a method parameter.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -589,7 +678,7 @@ private static void setupFileScanResultReceiver(Config cfg, String projectId) {
     private static void startNewSubscriberWithRecovery(ProjectSubscriptionName subscription,
                                                        Subscriber.Builder builder,
                                                        ExecutorProvider executorProvider,
-                                                       ExecutorService callbackExecutor,
+                                                       Executor callbackExecutor,
```
bskinner: A specific `ExecutorService` is not required here- an `Executor` instance suffices for the needs of this method.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/UploadsEventReceiver.java
```diff
@@ -0,0 +1,138 @@
...
+     *  any {@code file_upload} records matching the object's {@code blob_name}.
+     * 
+     * @param message the incoming message from Cloud Storage
+     */
+    private void handleFinalizeNotification(PubsubMessage message) {
+        final var storage = StorageOptions.getDefaultInstance().getService();
+        assert storage != null;
+
+        final var blobName = message.getAttributesOrThrow(Attribute.OBJECT_ID.key);
+        final var publishedAt = message.getPublishTime();
```
bskinner: There are multiple values for the time in the message. This is using the publish time of the event, but there is another time available under the EVENT_TIME attribute in the message. The expected difference between these two values is in the range of nanos/millis, but this can be updated to use the more applicable EVENT_TIME field.

pepper-apis/pom.xml
```diff
@@ -19,6 +19,7 @@
         <itext.licensekey.version>3.0.3</itext.licensekey.version>
         <liquibase.version>3.6.3</liquibase.version>
         <elasticsearch.version>6.8.2</elasticsearch.version>
+        <google-cloud-storage.version>2.11.3</google-cloud-storage.version>
```
bskinner: This also fixes the open issue DDP-7843

MocanaAtBroad:
Need to ask: was using the value of `last_updated_at` in `file_answer table` not acceptable as the upload time? I believe client is supposed to PATCH upon completion of the upload.

bskinner:
> Need to ask: was using the value of `last_updated_at` in `file_answer table` not acceptable as the upload time? I believe client is supposed to PATCH upon completion of the upload.

That wasn't considered as that is when the answer was updated, and not when the upload was technically completed (and this matches the triggering behavior of the malware scanner).

I don't see any reason why that approach wouldn't work. There is the risk of the client completing the upload, but the answer update call not occurring (for whatever reason), and that the malware scan may complete before the client updates their answer (leading to `uploaded_at` being updated after `scanned_at`). AFAIK, we aren't offering any ordering guarantees, and this feature is intended to catch file which have been uploaded and then get stuck in the scan process, so the issues may be minimal.

MocanaAtBroad:
I don't _BELIEVE_ that the file will show up in DSM unless that PATCH that follows the upload goes through. I think what you are doing here with the bucket notification is sound engineering-wise, though a little more complicated. Will defer to you on which approach you want to stick with.

bskinner:
> I don't _BELIEVE_ that the file will show up in DSM unless that PATCH that follows the upload goes through. I think what you are doing here with the bucket notification is sound engineering-wise, though a little more complicated. Will defer to you on which approach you want to stick with.

Let me take a look at what's needed there. You are absolutely correct about the complexity, and if using the `last_updated_at` value meets our needs without adding more services to manage for Housekeeping , I think it would be a better approach.

bskinner:
@MocanaAtBroad 

I've updated the implementation with the answer's `last_updated_at` approach. The overall implementation is far simpler, but there is one edge case which is encountered when updating existing `FileUpload` records (see below). Depending on how this API is being called, the code can be modified to update/delete/insert as needed, but is may be a moot point- I'm not sure what behavior the frontend expects.

@alextanlan @juanseba21 

Can you tell me under what circumstances a `FileAnswer` would be updated by the front-end? At the moment, when an existing `FileAnswer` is updated, the backend is first removing all of the references `FileUpload` records, and then creating new records. This results in the loss of the `scanned_at` and `uploaded_at` values if one of the referenced files has already been uploaded and/or scanned (and just updating this object will not trigger another scan).

The relevant code for the backend handling of the HTTP call is in [AnswerDao#updateFileAnswerValue](https://github.com/broadinstitute/ddp-study-server/blob/4b6aae6fb1d16b4ec58cf8def60c1efb13385cf3/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java#L278-L281).

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
bskinner: No longer applicable.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: No longer applicable.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: No longer applicable.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/UploadsEventReceiver.java
bskinner: No longer applicable.

pepper-apis/pom.xml
bskinner: Moved to [ddp-study-server/ddp-7843-cloud-storage-update](/broadinstitute/ddp-study-server/tree/ddp-7843-cloud-storage-update)

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/client/GoogleBucketClient.java
bskinner: No longer applicable

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: No longer applicable

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
bskinner: No longer applicable

pepper-apis/dss-core/src/main/resources/org/broadinstitute/ddp/db/dao/JdbiCompositeAnswer.sql.stg
```diff
@@ -3,53 +3,55 @@ group JdbiCompositeAnswer;
...
+    FROM
+        composite_answer_item i
+    LEFT JOIN
+        answer ca on ca.answer_id=i.child_answer_id) AS ar ON ar.parent_answer_id=pa.answer_id AND ar.question_id=cqq.child_question_id AND qr.response_order=ar.response_order
   WHERE
-	pa.answer_id=:parentAnswerId
+    pa.answer_id=:parentAnswerId
   ORDER BY
-	qr.response_order,
+    qr.response_order,
```
bskinner, line 54: No SQL changes here- normalizing the whitespace to all spaces (ASCII 0x20)

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
Trimax: Isn't it overcomplicated? Config also has the method `hasPath` that covers the external switch statement.

I think this will do the same:
```
public static Optional<String> getString(Config cfg, String key) {
    if (!cfg.hasPath(key)) {
        return Optional.empty();
    }

    return Optional.of(cfg.getString(key));
}
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -154,11 +154,43 @@ private void createAnswerFileValue(long answerId, FileAnswer answer) {
...
             return;
         }
-        List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
-        int[] inserted = getAnswerSql().bulkInsertFileValue(answerId, uploadIds);
+
+        final List<Long> uploadIds = answer.getValue().stream().map(FileInfo::getUploadId).collect(Collectors.toList());
+        final int[] inserted = getAnswerSql().bulkInsertFileValue(answerId, uploadIds);
         if (inserted.length != uploadIds.size()) {
-            throw new DaoException("Not all file uploads were assigned to answer " + answerId);
+            final var message = String.format("Failed to insert uploads ids for [answer:%s] (%s inserted, %s expected)",
```
Trimax: Not a comment, just want to know your opinion: why do you extracted the message? It's a matter of taste I would say, but do you think this is less readable:

```
  throw new DaoException(String.format("Failed to insert uploads ids for [answer:%s] (%s inserted, %s expected)",
                    answerId,
                    inserted.length,
                    uploadIds.size()));
```

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
```diff
@@ -154,11 +154,43 @@ private void createAnswerFileValue(long answerId, FileAnswer answer) {
...
+         * The answer dto needs to be pulled directly here as the `Answer` class does not
+         *  include the lastUpdatedAt time. It could be added- the ideal spot would likely be
+         *  in {@link AnswerWithValueReducer}- but that would involve altering the ctors for a
+         *  very widely used base class. This should work for the time being.
+         *  (bskinner - 20221007)
+         */
+        final var uploadedAt = getAnswerSql()
+                .findDtoById(answerId)
+                .map(AnswerDto::getLastUpdatedAt)
+                .map(timestamp -> Instant.ofEpochMilli((long)timestamp))
```
Trimax, line 182: Could be replaced with
```
  .mapToLong(Long::longValue)
  .map(Instant::ofEpochMilli)
```

Trimax (dismissed):
LGTM. Left a couple of minor comments

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigUtil.java
bskinner: I was wrong- Just double checked the docs and `hasPath` will return `false` in the event of a missing key _or_ `null` value, so this can be simplified. This isn't used anywhere in this PR anymore so it can be removed entirely as I don't think this adds anything over `getStrIfPresent()`

_Previous comment_
~~No, it does not- `cfg.getString()` will throw [ConfigException.Missing](https://lightbend.github.io/config/latest/api/com/typesafe/config/ConfigException.Missing.html) in case of `{ myKey = null }`~~
~~In this case, `hasPath()` returns `true`, but `getString()` will throw due to the null value.~~

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
bskinner: The embedded `String.format` is more readable- that is an artifact from a prior change which extended the line past the 140c limit, and was just never collapsed back.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/AnswerDao.java
bskinner: Not in the current form- `mapToLong()` is a method on the `Stream` class, not on the `Optional<T>` class.

MocanaAtBroad (approved):

MocanaAtBroad (approved):

PR #2373, opened 2022-10-27
[DSM] PEPPER-104 Care_Of in shipping label

pegahtah:
* liquibase added for a column named has_care_of in ddp_kit_request_settings. 
* The ElasticSearchUtil.java class is changed to get the proxy info if there is any when making a DDPParticipant object. 
* In EasyPostUtil in createAddress, I am checking if the kit settings has the flag set for has_care_of and if it does, I add the proxy info as the C/O in the company field in the address 
image


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -184,7 +186,8 @@ public Shipment buyShipment(@NonNull String carrier, String carrierId, String se
         }
     }
 
-    public Address createAddress(DDPParticipant ddpParticipant, @NonNull String phone) throws EasyPostException {
+    public Address createAddress(DDPParticipant ddpParticipant, @NonNull String phone, KitRequestSettings kitRequestSettings)
```
SimoneMaiwald: No need to pass the whole `KitRequestSettings` object in. Just the boolean value would be enough

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -776,12 +776,23 @@ public static Map<String, Object> getObjectsMap(String index, String id, String
     }
 
     public static DDPParticipant getParticipantAsDDPParticipant(@NonNull Map<String, Map<String, Object>> participantsESData,
```
SimoneMaiwald: Please rollback these changes! and check it in the `KitRequestShipping` class.
1) if the kit setting has care_of check the `ddpParticipant` if it has proxy guid
2) if proxy guid is not empty get the first-/lastname and add to ddpParticipant

pegahtah: Honestly I think it makes more sense for it being here, we are getting the participant info from ES in this method and we know about the proxy guid here, in `KitRequestShipping` we only call this and get the ddpParticipant object that was already made

pegahtah: Changes applied

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1347,7 +1347,7 @@ public static String createRandom(int len) {
     }
 
     public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequestSettings kitRequestSettings, String addressId,
-                                         DDPParticipant participant) throws Exception {
+                                         DDPParticipant participant, Integer ddpInstanceId) throws Exception {
```
SimoneMaiwald: do we not already have the instance in DDPParticipant or KitRequestSettings?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -30,6 +30,7 @@
 import org.apache.http.impl.nio.reactor.IOReactorConfig;
 import org.apache.lucene.search.join.ScoreMode;
 import org.broadinstitute.dsm.db.DDPInstance;
+import org.broadinstitute.dsm.db.dto.ddp.instance.DDPInstanceDto;
```
SimoneMaiwald, line 33: is that import used?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1357,8 +1357,44 @@ public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequ
...
+                    return toAddress;
+                }
+                toAddress = getAddressForStudiesWithCareOfField(easyPostUtil, kitRequestSettings, participant, ddpInstanceDto);
+            }
+        }
+        return toAddress;
+    }
+
+    private static Address getAddressForStudiesWithCareOfField(@NonNull EasyPostUtil easyPostUtil, KitRequestSettings kitRequestSettings,
+                                                               DDPParticipant participant, DDPInstanceDto ddpInstanceDto)
```
pegahtah, line 1372: created a new method for this one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1357,8 +1357,44 @@ public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequ
...
+                proxyFirstName = (String) proxyProfile.get("firstName");
+                proxyLastName = (String) proxyProfile.get("lastName");
+                String careOf = String.format("C/O %s %s", Objects.toString(proxyFirstName, ""),
+                        Objects.toString(proxyLastName, ""));
+                toAddress = easyPostUtil.createAddress(participant, kitRequestSettings.getPhone(), careOf);
+            } else { // participant doesn't have proxies, proceed with normal label
                 toAddress = easyPostUtil.createAddress(participant, kitRequestSettings.getPhone());
             }
+        } else { // participant is not in ES
+            toAddress = easyPostUtil.createAddress(participant, kitRequestSettings.getPhone());
```
pegahtah: @SimoneMaiwald not sure what should happen here, we don't have pt in ES but we have DDPParticipant

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -206,7 +208,25 @@ public Address createAddress(DDPParticipant ddpParticipant, @NonNull String phon
...
+        return Address.create(toAddressMap);
+    }
+
+    public Address createAddress(DDPParticipant ddpParticipant, @NonNull String phone)
+            throws EasyPostException {
+        if (StringUtils.isEmpty(this.phone)) {
+            throw new RuntimeException("Contact phone number is needed");
+        }
+        Map<String, Object> toAddressMap = createAddressMap(ddpParticipant, phone);
         return Address.create(toAddressMap);
```
pegahtah, line 230: We can only call Address.create once, otherwise the company field won't be added if we add it later. So I created the `createAddressMap` to make the hashmap, and then we can add the company field if there is one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
SimoneMaiwald: the nothing. label creation with info available

pegahtah: ok good so this is correct

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1357,8 +1357,44 @@ public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequ
             if (StringUtils.isNotBlank(addressId)) {
                 toAddress = easyPostUtil.getAddress(addressId);
             } else if (participant != null) {
+                boolean hasCareOf = kitRequestSettings.getHasCareOF() == 1;
```
SimoneMaiwald: if hasCareOf is not used anywhere else, you could put that `kitRequestSettings.getHasCareOF()` directly in the if

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
SimoneMaiwald: wait, that is for the singular way? I guess all pts what get a label in singular should be in ES. So I would say error log and no label

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -354,6 +355,17 @@ public static ElasticSearchParticipantDto getElasticSearchForGivenMatch(String i
 
     public static Map<String, Map<String, Object>> getFilteredDDPParticipantsFromES(@NonNull DDPInstance instance, @NonNull String filter) {
         String index = instance.getParticipantIndexES();
+        return getFilteredDDPParticipantsFromES(index, instance.getName(), filter);
+    }
+
+    public static Map<String, Map<String, Object>> getFilteredDDPParticipantsFromES(@NonNull DDPInstanceDto instance,
```
SimoneMaiwald: please rename instance to instanceDto

pepper-apis/dsm-core/src/main/resources/liquibase/singular/PEPPER-104_care_of.xml
```diff
@@ -0,0 +1,12 @@
...
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="PEPPER-104_care_of" author="pegah">
+        <addColumn tableName="ddp_kit_request_settings">
+            <column name="has_care_of" type="TINYINT(1)"></column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
```
SimoneMaiwald, line 12: if that setting should go out to singular on prod without us waiting for IRB, please also add a update query for ddp_instance_role table

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/tools/CreateLabelTool.java
```diff
@@ -4,6 +4,7 @@
 import com.easypost.model.Address;
 import com.easypost.model.Parcel;
 import com.easypost.model.Shipment;
+import org.broadinstitute.dsm.model.KitRequestSettings;
```
SimoneMaiwald: used?

pepper-apis/dsm-core/src/main/resources/liquibase/singular/PEPPER-104_care_of.xml
pegahtah: It needs IRB Approval unfortunately we have to wait

SimoneMaiwald (dismissed):

SimoneMaiwald (approved):

PR #2381, opened 2022-11-01
PEPPER-29: saved filter number

SimoneMaiwald:
fixing query if string contains a number

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
@@ -606,7 +606,7 @@ public class ViewFilter {
                                 type = Filter.NUMBER;
                                 state = 40;
                             } else {
-                                tempValue = word;
+                                tempValue = word.trim();
                                 if (!longWord) {
                                     if (tempValue.contains(Filter.SINGLE_QUOTE)) {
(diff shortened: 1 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -688,17 +688,43 @@ public static ViewFilter parseFilteringQuery(String str, ViewFilter viewFilter)
...
+                                        tempValue = word;
+                                        if (!longWord) {
+                                            if (tempValue.contains(Filter.SINGLE_QUOTE)) {
+                                                if (tempValue.indexOf(Filter.SINGLE_QUOTE) != tempValue.lastIndexOf(Filter.SINGLE_QUOTE)) {
+                                                    value = trimValue(tempValue);
+                                                    state = 11;
+                                                    break;
+                                                } else {
+                                                    longWord = true;
+                                                    value += trimValue(tempValue) + " ";
```
pegahtah, line 709: i think a potential bug here would be searching for something with double spaces ( it could already be in the current code too)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -688,17 +688,43 @@ public static ViewFilter parseFilteringQuery(String str, ViewFilter viewFilter)
...
+                                                }
+                                            } else {
```
pegahtah, line 711: I don't understand when this happens...

pegahtah (approved):
LGtM

PR #2396, opened 2022-11-15
[DSM] PEPPER-239 PECGS: check the normal samples received

pegahtah:
Adding a check in KitDaoImpl and MercurySampleDao to check for at least one kit having a received date


SimoneMaiwald (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/GetMercuryEligibleSamplesRoute.java
```diff
@@ -16,14 +17,17 @@
 @Slf4j
 public class GetMercuryEligibleSamplesRoute extends RequestHandler {
     private MercurySampleDao mercurySampleDao;
+    private KitDaoImpl kitDaoImpl;
     public static String projectId;
     public static String topicId;
```
Trimax, line 22: I know this hasn't been changed in terms of this PR, but I see that those two fields are static ones and we initialize them in the constructor. Is that expected behavior?

Trimax (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/mercury/GetMercuryEligibleSamplesRoute.java
pegahtah: Yeah they are just values that are read from secret, that's why we don't have it at this point to initialize it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -583,4 +587,17 @@ public Optional<ScanError> updateKitLabel(KitRequestShipping kitRequestShipping)
         }
         return result;
     }
+
+    public boolean hasKitReceived(Connection connection, String ddpParticipantId) {
```
Trimax, line 591: This method looks like something very generic. Something like `contains` or `exists`. I would suggest defining it somewhere separately and reusing it wherever we do similar operations.

PR #2398, opened 2022-11-16
PEPPER-280: adding bspCollaboratorParticipantId to ES

SimoneMaiwald:
adding collaborator pt id to ES


pegahtah:
Based on the ticket I think there should be a liquibase for adding the Collaborator Participant ID to the quick filter now that it has been renamed

GiCharkviani (approved):

SimoneMaiwald:
@pegahtah no, because that field was never displayed at the pt list so there can't be any quick filters/saved filters with that column!

pegahtah (approved):

PR #2412, opened 2022-12-09
[DSM] PEPPER-395 Clinical Kit Dummy Endpoint small fixes

pegahtah:
Just small changes for the ClinicalKit creator endpoint to make it more robust

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
@@ -2,6 +2,7 @@ package org.broadinstitute.dsm.route;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.Random;
 
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.DDPInstance;
@@ -52,7 +53,8 @@ public class CreateClinicalDummyKitRoute implements Route {
         String collaboratorParticipantId = KitRequestShipping
                 .getCollaboratorParticipantId(ddpInstance.getBaseUrl(), ddpInstance.getDdpInstanceId(), ddpInstance.isMigratedDDP(),
                         ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, "4");
-        updateTissue(tissueId, collaboratorParticipantId + "_T1");
+        int randomINT = new Random().nextInt(10000);
+        updateTissue(tissueId, collaboratorParticipantId + "_" + randomINT);
 
     }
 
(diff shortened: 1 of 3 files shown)


ngigliotosBroad (dismissed):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -52,7 +53,8 @@ public static void addCollaboratorSampleId(String tissueId, DDPInstance ddpInsta
         String collaboratorParticipantId = KitRequestShipping
                 .getCollaboratorParticipantId(ddpInstance.getBaseUrl(), ddpInstance.getDdpInstanceId(), ddpInstance.isMigratedDDP(),
                         ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, "4");
-        updateTissue(tissueId, collaboratorParticipantId + "_T1");
+        int randomINT = new Random().nextInt(10000);
+        updateTissue(tissueId, collaboratorParticipantId + "_" + randomINT);
```
SimoneMaiwald: do we want to keep the `T` ?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/QueryExtension.java
```diff
@@ -11,6 +11,7 @@ public class QueryExtension {
     public static final String BY_ROLE_NAME_START_WITH = " and role.name regexp \'^%1\' ";
 
     public static final String BY_DDP_PARTICIPANT_ID = " AND p.ddp_participant_id = ?";
+    public static final String PECGS_TEST_ELIGIBLE_ONC_HISTORY = " AND oD.accession_number IS NOT null AND oD.date_px IS NOT null ";
```
SimoneMaiwald: can we change `IS NOT null` to `IS NOT NULL`

SimoneMaiwald (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -52,7 +53,8 @@ public static void addCollaboratorSampleId(String tissueId, DDPInstance ddpInsta
         String collaboratorParticipantId = KitRequestShipping
                 .getCollaboratorParticipantId(ddpInstance.getBaseUrl(), ddpInstance.getDdpInstanceId(), ddpInstance.isMigratedDDP(),
                         ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, "4");
-        updateTissue(tissueId, collaboratorParticipantId + "_T1");
+        int randomINT = new Random().nextInt(100000);
+        updateTissue(tissueId, collaboratorParticipantId + "_T" + randomINT);
```
bskinner, line 57: What are the consequences of a collision here? There's no guarantee that two calls to `nextInt()` will not return the same value.

pegahtah: So the collaborator sample id should be unique, but since we are using it with `collaboratorParticipantId` , it is already different between tissues for 2 different participants, so the only case of that is when 2 dummy tissues being created for  the same participant get the same random int. We can add a check, but I think in that case they can just test again. I don't think this will happen

bskinner (approved):
Resolve the outstanding comments via Slack- 👍

PR #2414, opened 2022-12-12
[DSM] PEPPER-232 PDF Fix

pegahtah:
## Context

I changed the code to wait for the PDF to be generated for 5 tries, each with a wait time for 2 seconds. If after that the PDF is still of 0 size DSM will throw an error 
## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -84,39 +86,60 @@ public DownloadPDF(@NonNull String requestBody) {
...
-                    pdfBytes = generateSinglePDF(requestBody, configName, user, ddpInstance);
+                    maybePdf = generateSinglePDF(requestBody, configName, user, ddpInstance);
+                }
+                if (maybePdf.isPresent()) {
+                    byte[] pdfByte = maybePdf.get();
+                    if (pdfByte.length > 0) {
+                        savePDFinBucket(ddpInstance.getName(), ddpParticipantId, new ByteArrayInputStream(pdfByte),
+                                fileName, user.getId());
+                    } else {
+                        throw new RuntimeException("PDF size was zero!" + fileName);
```
pegahtah: this will happen after the multiple tries with wait times. So at this point which is more than 10 seconds later we are just trying an error

MocanaAtBroad: The error you are observing that you are trying to fix: from getPDFBundle or generateSinglePDF?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -84,39 +86,60 @@ public DownloadPDF(@NonNull String requestBody) {
...
+                    }
                 }
-                pdfBytes.ifPresent(bytes -> {
-                    savePDFinBucket(ddpInstance.getName(), ddpParticipantId, new ByteArrayInputStream(bytes), fileName, user.getId());
-                });
-                pdfBytes.orElseThrow();
+                maybePdf.orElseThrow();
```
MocanaAtBroad: Recommend you specify the exception to be thrown and add some logging here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -84,39 +86,60 @@ public DownloadPDF(@NonNull String requestBody) {
...
+                pdfByte = new MRCoverPDF(this).getMRCoverPDF(requestBody, ddpInstance, user);
+            } else if (IRB.equals(configName)) {
+                String groupId = DDPInstance.getDDPGroupId(ddpInstance.getName());
+                pdfByte = PDFProcessor.getTemplateFromGoogleBucket(groupId + "_IRB_Letter.pdf");
+            } else if (TISSUE.equals(configName) || REQUEST.equals(configName)) {
+                TissueCoverPDF tissueCoverPDF = new TissueCoverPDF(this);
+                pdfByte = tissueCoverPDF.getTissueCoverPDF(ddpInstance, user);
+            } else if (configName != null) {
+                pdfByte = requestPDF(ddpInstance, ddpParticipantId, configName);
+            }
```
MocanaAtBroad, line 130: (My previous comment might have not gone through?) Make sure you have logging that tells you which case is causing the pdfByte to be an empty array.

MocanaAtBroad (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
MocanaAtBroad: Your approach seems to be to retry getting pdfs at the top-level. But at the lower level looks like you have multiple ways that those PDFS are generated (i.e., grabbed from a bucket, generated locally or fetched from DSS). Make sure you know which one is causing you trouble.
If it is the DSS fetching, then wonder if you should have a retry at the level of calling DSS? Would also be interested in seeing the actual error messages. Could very well be that there is a problem on the DSS side.
I mean: in distributed systems you always need to assume that stuff will fail, and retries are one strategy to deal with it. But you would hope that the retries are mostly needed in rare occasions!

SimoneMaiwald: @MocanaAtBroad our suspicion is that we are getting pdfs from DSS but that they are size 0. this is why we are now adding a check of length > 0 and the retry if it was so. 
@pegahtah but I agree that I would like to know when this was the case and for which pdf. Can you please add in the else also a log.error: "[PDF] was size 0", Thank you. Error for now and when we know if identified the problem further we can change it from error to warn later (if it is to chatty…)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
pegahtah: `getPDFBundle` calls `generateSinglePDF` for each of them to generate the bundle.  The bug specifically happens in the `generateSinglePDF`, where a byte[] with size zero is returned from DSS.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
pegahtah: Yes I agree, changed it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -84,39 +86,61 @@ public DownloadPDF(@NonNull String requestBody) {
...
+            if (pdfByte.length > 0) {
+                break;
+            } else {
+                try {
+                    logger.error(String.format("Got %s PDF with size zero in try %d", configName, turn));
+                    Thread.sleep(retryWaitMillis);
```
pegahtah, line 136: more logging is added here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -84,39 +86,60 @@ public DownloadPDF(@NonNull String requestBody) {
 
     public Optional<byte[]> getPDFs(long userIdRequest, String realm, String requestBody) {
         UserDto user = new UserDao().get(userIdRequest).orElseThrow();
-        Optional<byte[]> pdfBytes = Optional.empty();
+        Optional<byte[]> maybePdf = Optional.empty();
         if (StringUtils.isNotBlank(this.ddpParticipantId)) {
             DDPInstance ddpInstance = DDPInstance.getDDPInstance(realm);
             if (ddpInstance != null && StringUtils.isNotBlank(ddpParticipantId)) {
-                String fileName = "";
```
pegahtah, line 91: I found a bug that this removed the name of the file type when saving file to the bucket (already on dev) so I changed it use configName from now

ngigliotosBroad (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/pdf/DownloadPDF.java
```diff
@@ -57,6 +57,8 @@ public class DownloadPDF {
     private String medicalRecordId;
     private List<PDF> pdfs;
 
+    private static long retryWaitMillis = 2 * 1000;
```
SimoneMaiwald, line 60: I think 2 seconds is too long. can we reduce it to 500 and just see how it goes. or maybe put it in the bookmark table so we can change it when we see it happening

SimoneMaiwald (approved):

PR #2415, opened 2022-12-14
[PEPPER-36] Adds additional diagnoses to the Pancan cancer lists

bskinner:
PEPPER-36

## Context
This PR adds a patch named `PancanAddAdditionalDiagnoses`. This patch inserts the additional options named in the description of PEPPER-36 to the Genitourinary (`C_GROUP_GENITOURINARY_GENITOURINARY_CANCERS`) group in all Pancan activities which make use of the shared `cancer-picklist-groups.conf` snippet.

## Running the patch
The patch can be run using the standard `studybuilder-cli` `--run-task` flag using the task name `PancanAddAdditionalDiagnoses`. No addition configuration changes are necessary outside of those required to deploy the `pancan` study via studybuilder.

The list of impacted activities & questions is available in [this comment](https://broadworkbench.atlassian.net/browse/PEPPER-36?focusedCommentId=71686) on the Jira ticket.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
@@ -35,7 +35,7 @@ public interface PicklistQuestionDao extends SqlObject {
 
     Logger LOG = LoggerFactory.getLogger(PicklistQuestionDao.class);
 
-    int DISPLAY_ORDER_GAP = 10;
+    public final int DISPLAY_ORDER_GAP = 10;
 
     @CreateSqlObject
     JdbiPicklistQuestion getJdbiPicklistQuestion();
@@ -178,6 +178,8 @@ public interface PicklistQuestionDao extends SqlObject {
 
     /**
      * Create a new picklist option for given question.
+     * 
+     * <p>The `option` parameter <b>will be modified</b> by this method.
      *
      * @param questionId   the associated picklist question
      * @param option       the picklist option definition
@@ -185,7 +187,7 @@ public interface PicklistQuestionDao extends SqlObject {
      * @param revisionId   the revision to use, will be shared for all created data
      * @return the option id
      */
-    default long insertOption(long questionId, PicklistOptionDef option, int displayOrder, long revisionId) {
+    public default long insertOption(long questionId, PicklistOptionDef option, int displayOrder, long revisionId) {
         JdbiPicklistOption jdbiOption = getJdbiPicklistOption();
         TemplateDao templateDao = getTemplateDao();
 
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
@@ -35,7 +35,7 @@ public interface PicklistQuestionDao extends SqlObject {
 
     Logger LOG = LoggerFactory.getLogger(PicklistQuestionDao.class);
 
-    public final int DISPLAY_ORDER_GAP = 10;
+    int DISPLAY_ORDER_GAP = 10;
 
     @CreateSqlObject
     JdbiPicklistQuestion getJdbiPicklistQuestion();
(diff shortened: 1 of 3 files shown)


pepper-apis/studybuilder-cli/studies/pancan/patches/add-additional-diagnosis.conf
```diff
@@ -0,0 +1,241 @@
...
+        "c_genitourinary_renal_oncocytoma": "Kidney Cancer / Renal oncocytoma",
+        "c_genitourinary_sdha_b_rcc": "Kidney Cancer / SDHA/B RCC",
+        "c_genitourinary_trcc_renal_cell": "Kidney Cancer / Translocation renal cell carcinoma (tRCC)",
+        "c_genitourinary_unclassified_rcc": "Kidney Cancer / Unclassified RCC"
+      }
+    },
+
+    "es": {
+      "cancer": {
+        "c_genitourinary_chrcc_renal_cell": "Cancer de riñon / Carcinoma de células renales cromófobo",
```
MocanaAtBroad: I believe that it should be **Cáncer de riñón** (note the accents). But do check against documentation provided to you (I did not see it in ticket).

MocanaAtBroad (commented):
Need do to check the Spanish accents but code looks 
good. Let's see if we can get changes applied to dev to see in action.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
```diff
@@ -35,7 +35,7 @@ public interface PicklistQuestionDao extends SqlObject {
 
     Logger LOG = LoggerFactory.getLogger(PicklistQuestionDao.class);
 
-    int DISPLAY_ORDER_GAP = 10;
+    public final int DISPLAY_ORDER_GAP = 10;
```
Trimax: I believe it is public by default according to the [specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.6.1)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
```diff
@@ -178,14 +178,16 @@ default void insertNestedOptions(long questionId, PicklistOptionDef option,
...
+     * <p>The `option` parameter <b>will be modified</b> by this method.
      *
      * @param questionId   the associated picklist question
      * @param option       the picklist option definition
      * @param displayOrder the display order number for the option
      * @param revisionId   the revision to use, will be shared for all created data
      * @return the option id
      */
-    default long insertOption(long questionId, PicklistOptionDef option, int displayOrder, long revisionId) {
+    public default long insertOption(long questionId, PicklistOptionDef option, int displayOrder, long revisionId) {
```
Trimax: Same here. Is there any reason why you want to explicitly define the method as public?

Trimax (dismissed):

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
bskinner: Yup- they are- thank you for catch! Given these SqlObjects sit in a weird space, these all were colored as `package-private`. Given it's unnecessary, I'll go through and remove them to match the style in the rest of the definition.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/dao/PicklistQuestionDao.java
bskinner: In general, I feel explicit is better than implicit when it comes to visibility, but in these cases they had been represented in my mind more as classes than interfaces, and got assigned `package-private` as the default visibility and were mechanically updated when I needed them

MocanaAtBroad (approved):

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/PancanAddAdditionalDiagnosis.java
```diff
@@ -0,0 +1,294 @@
...
+                + "  FROM picklist_group AS pg"
+                + "  JOIN picklist_question AS pq ON pq.question_id = pg.picklist_question_id"
+                + "  JOIN question AS q ON q.question_id = pq.question_id"
+                + "  JOIN study_activity AS sa ON q.study_activity_id = sa.study_activity_id"
+                + "  JOIN umbrella_study AS us ON sa.study_id = us.umbrella_study_id"
+                + "  JOIN revision ON revision.revision_id = pg.revision_id"
+                + "  WHERE group_stable_id = :groupStableId"
+                + "    AND us.guid = :studyGuid")
+        @RegisterConstructorMapper(PicklistGroupDto.class)
+        public List<PicklistGroupDto> fetchPicklistGroupsByName(
```
bskinner, line 69: Interface methods are public by default, so these visibility specifiers are unnecessary. Leaving them for now to prevent dismissing existing reviews on the PR.

pegahtah (approved):
with my limited knowledge of how the schema in DSS works the code looks good to me

PR #2427, opened 2023-01-05
Pepper-445: tissue received

SimoneMaiwald:
tissueReceived event trigger for DSS

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
@@ -10,6 +10,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.dao.settings.EventTypeDao;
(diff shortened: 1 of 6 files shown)


pegahtah (dismissed):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
```diff
@@ -208,6 +209,14 @@ Object handleSingleNameValue() {
         if (Patch.patch(patch.getId(), patch.getUser(), patch.getNameValue(), dbElement)) {
             nameValues.addAll(setWorkflowRelatedFields(patch));
             exportToESWithId(patch.getId(), patch.getNameValue());
+            if (patch.getNameValue().getName().equals("oD.tissueReceived")) {
+                if (StringUtils.isBlank(patch.getParentId()) && StringUtils.isNotBlank(patch.getDdpParticipantId())) {
+                    patch.setParentId(patch.getDdpParticipantId());
+                }
+                if (StringUtils.isNotBlank(patch.getParentId())) {
+                    triggerParticipantEvent(ddpInstance, patch, new Value(null, EventTypeDao.EVENT, "TISSUE_RECEIVED"));
+                }
```
MocanaAtBroad, line 220: Would recommend making the literal strings here static variables.
Also, there are some checks here for things being blank that affect logic. Some logging might be helpful if we are trying to figure out what the specific case that we are dealing with is or if there is an unexpected error condition.

MocanaAtBroad (approved):

pegahtah (approved):

PR #2428, opened 2023-01-05
PEPPER-403 fixed the casting bug

pegahtah:
There was a casting error from `Object` to `ScanError`

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
@@ -503,7 +503,7 @@ public class KitDaoImpl implements KitDao {
             return dbVals;
         });
         if (Objects.nonNull(results.resultValue)) {
-            result = (Optional<ScanError>) results.resultValue;
+            result = Optional.ofNullable((ScanError)results.resultValue);
         }
         return result;
     }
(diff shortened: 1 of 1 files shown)


SimoneMaiwald (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -492,18 +492,20 @@ public Optional<ScanError> insertKitTracking(KitRequestShipping kitRequestShippi
...
                         "Kit Label \"" + kitRequestShipping.getKitLabel() + "\" does not exist.\n"
                                 + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER);
+                logger.error("Unable to save kit tracking information for kit label " + kitRequestShipping.getKitLabel(), ex);
             }
             return dbVals;
         });
         if (Objects.nonNull(results.resultValue)) {
-            result = (Optional<ScanError>) results.resultValue;
+            result = Optional.ofNullable((ScanError)results.resultValue);
         }
```
MocanaAtBroad, line 509: My advice is to get rid of `SimpleResult dbVals` and create instead an object of type ScanError . SimpleResults is not doing anything. And you will be able to get rid of the casting.

pegahtah: Thanks for the tip! I tried it. The thing is, I can't create the `ScanError` Object  before the `inTransaction` lambda and set it in the lambda and also I can't create it in the lambda and return it, since it should return `SimpleResult`.

MocanaAtBroad: I see! That SimpleResult thing ought to use generics. At any rate ✅

MocanaAtBroad (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: Thanks!

PR #2433, opened 2023-01-17
PEPPER-127: os2 filters

SimoneMaiwald:
adding ddpInstanceId when querying for osteo kit, mr or oncHistoryDetail related information

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
@@ -35,6 +35,7 @@ public class DBConstants {
     public static final String DDP_ABSTRACTION_ALIAS = "a";
     public static final String DRUG_ALIAS = "d";
     public static final String ALIAS_DELIMITER = ".";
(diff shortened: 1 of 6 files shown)


pegahtah (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
```diff
@@ -35,6 +35,7 @@ public class DBConstants {
     public static final String DDP_ABSTRACTION_ALIAS = "a";
     public static final String DRUG_ALIAS = "d";
     public static final String ALIAS_DELIMITER = ".";
+    public static final String DDP_INSTANCE_ID_CAMEL_CASE = "ddpInstanceId";
```
zyme, line 38: Is this truly a database constant?  Or do we have an ES constant class that we can put this into?

zyme (approved):
LGTM, but I will admit that my knowledge in this area is pretty weak!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
SimoneMaiwald: I will move it to another place with one of my other PRs

PR #2434, opened 2023-01-17
Extends studybuilder-cli to support syncing & rotating client secrets

bskinner:
## Context

Updating the DSS database with updated secrets in the event of a change of the `auth0.encryptionKey` value, or a breach requiring the Auth0 client secrets to be rotated is currently a extremely time-intensive task- each client takes 5-10 minutes to rotate, and there are upwards of 30 clients per environment (not including management clients). This PR adds the `--only-sync-client-secrets` action to `studybuilder-cli`.

When invoked, `--only-sync-client-secrets` will loop through all the available tenants and, using their associated management clients, fetch the client secrets for each client in the tenant, encrypt it using the `auth0.encryptionKey` value and `AesUtil`, and update the database. This operation is non-destructive, and will not change any values in Auth0.

The `--only-sync-client-secrets` action has an additional flag, `--rotate`. When `--rotate` is specified, the clients will have their client secret rotated prior to updating the database. This is a destructive option, and other configuration sources may need to be updated with the new client secrets after the rotation is complete.

When operating in sync-only mode, the management clients for each tenant require the following permissions to be granted for the Auth0 Management API
* read:clients
* read:client_keys

When operating in `rotate` mode, the management clients for each tenant require the following additional permissions to be granted for the Auth0 Management API
*  update:clients
*  update:client_keys

## Limitations
This functionality does not handle rotation of the management clients. If the Auth0 Management Clients require their secrets to be re-encrypted or rotated, this *must* first be done manually before this operation can be successfully used.

## Examples
1. Perform a client secret sync
    ```
    java -Dconfig.file="path/to/dss/application.conf" \
      -jar "target/StudyBuilder.jar" \
      --vars "path/to/vars.conf" \
      --substitutions path/to/subs.conf \
      path/to/study.conf \
      --only-sync-client-secrets
    ```
2. Rotate & update all client secrets
    ```
    java -Dconfig.file="path/to/dss/application.conf" \
      -jar "target/StudyBuilder.jar" \
      --vars "path/to/vars.conf" \
      --substitutions path/to/subs.conf \
      path/to/study.conf \
      --only-sync-client-secrets \
      --rotate
    ```

## Further work
In the future, adding flags to limit the rotation behavior to a specific domain or study would expand the usefulness of this feature, and limit potential impacts.

## Release

No special release procedures are necessary, this functionality must be manually invoked via `studybuilder-cli` if necessary.

### DIFF
--- a/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
+++ b/pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
@@ -15,6 +15,9 @@ import java.util.function.Consumer;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+
+import lombok.extern.slf4j.Slf4j;
+
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
 import org.apache.commons.cli.DefaultParser;
@@ -32,6 +35,7 @@ import org.jdbi.v3.core.Handle;
 /**
  * Main entry point for command-line tool that helps stand up a study.
  */
+@Slf4j
 public class StudyBuilderCli {
 
     private static final String USAGE = "StudyBuilder [-h, --help] [OPTIONS] <STUDY_CONFIG_FILE | STUDY_KEY>";
@@ -50,6 +54,8 @@ public class StudyBuilderCli {
     private static final String OPT_PROCESS_TRANSLATIONS = "process-translations";
     private static final String OPT_I18N_PATH = "i18n-path";
     private static final String OPT_TRANSLATIONS_TO_DB_JSON = "translations-to-db-json";
+    private static final String OPT_ROTATE_CLIENT_SECRETS = "only-rotate-client-secrets";
+    private static final String OPT_DRY_RUN = "dry-run";
 
     private static final String DEFAULT_STUDIES_DIR = "studies";
     private static final String DEFAULT_STUDY_CONF_FILENAME = "study.conf";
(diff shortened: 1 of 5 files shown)


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/StudyBuilderCli.java
```diff
@@ -212,6 +228,25 @@ private void run(String[] args) throws Exception {
...
             return;
+        } else if (cmd.hasOption(OPT_SYNC_CLIENT_SECRETS)) {
+            if (isDryRun) {
+                throw new DDPException("-" + OPT_SYNC_CLIENT_SECRETS + " does not support -" + OPT_DRY_RUN);
+            }
+
+            final var rotateSecrets = cmd.hasOption(OPT_ROTATE_SECRETS);
+
+            if (rotateSecrets) {
+                final var shouldContinue = promptForConfirmation("A client secret rotation can not be undone. Do you wish to proceed?");
```
ssettipalli, line 239: I was going to suggest this confirmation check because its too powerful like invalidate study.
Glad you added it

ssettipalli:
@bskinner 
StudyBuilder is by study, so rotation works per study, right ?
This is with respect to your future suggestions

bskinner:
> @bskinner StudyBuilder is by study, so rotation works per study, right ? This is with respect to your future suggestions

No- right now, this will either sync or rotate _all_ tenants and _all_ clients. Per-study or per-domain is something that could be added, but it was not a priority for me to implement in light of the other tasks I have to wrap up. This distinction also becomes irrelevant if the value of `auth0.encryptionSecret` changes as all encrypted client secrets will become irretrievable.

bskinner:
@ssettipalli If you'd like to pair up either tomorrow or Thursday, I'd be happy work with you and implement study-specific syncing.

PR #2436, opened 2023-01-19
Pepper-445: change trigger

SimoneMaiwald:
DSM will trigger DSS if `saliva or blood and tissue` is received for a pt with the event `REQUIRED_SAMPLES_RECEIVED`


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
```diff
@@ -29,8 +33,24 @@ public class ClinicalKitDao {
                     + "LEFT JOIN ddp_participant as p on (p.participant_id = inst.participant_id) "
                     + "LEFT JOIN ddp_instance as ddp on (ddp.ddp_instance_id = p.ddp_instance_id) "
                     + "LEFT JOIN sm_id_type sit on (sit.sm_id_type_id = sm.sm_id_type_id) "
-                    + "LEFT JOIN kit_type ktype on ( sit.kit_type_id = ktype.kit_type_id) WHERE sm.sm_id_value = ? "
-                    + "AND NOT sm.deleted <=> 1 ";
+                    + "LEFT JOIN kit_type ktype on ( sit.kit_type_id = ktype.kit_type_id) "
```
zyme, line 36: Looks like the value of `SQL_GET_CLINICAL_KIT_BASED_ON_SM_ID_VALUE` has not actually changed here.  If that's the case but we wanted to change the whitespace, please put in a PR comment to that effect so reviewers know--or leave the whitespace as is.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
```diff
@@ -29,8 +33,24 @@ public class ClinicalKitDao {
...
+                    + "WHERE sm.sm_id_value = ? AND NOT sm.deleted <=> 1 ";
+    private static final String SQL_GET_CLINICAL_TISSUE_BY_DDP_PARTICIPANT_ID = "SELECT kit.received_date FROM ddp_kit "
```
zyme: How about naming the variable so it includes the fact that the query requires receipt?  Something like `SQL_GET_RECEIVED_TISSUE_BY_DDP_PARTICIPANT_ID`?  For this and L49,   I'm unsure of what it is about the query that makes it "clinical" vs. "sample has been received".

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
```diff
@@ -209,16 +208,6 @@ Object handleSingleNameValue() {
         if (Patch.patch(patch.getId(), patch.getUser(), patch.getNameValue(), dbElement)) {
             nameValues.addAll(setWorkflowRelatedFields(patch));
             exportToESWithId(patch.getId(), patch.getNameValue());
-            if (patch.getNameValue().getName().equals(DBConstants.OD_TISSUE_RECEIVED)) {
```
zyme, line 212: Some PR comments here about why this is being deleted would be helpful so that reviewers understand where the code has moved to (or why it's no longer needed).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -70,9 +72,9 @@ private ClinicalKitDto getClinicalKit(String kitLabel, Optional<BSPKitDto> optio
             clinicalKit.setCollectionDate(kitInfo.getCollectionDate());
             clinicalKit.setSampleCollection(ClinicalKitDao.PECGS);
             DDPInstance ddpInstance = DDPInstance.getDDPInstance(kitInfo.getRealm());
-            clinicalKit.setNecessaryParticipantDataToClinicalKit(optionalBSPKitDto.get().getDdpParticipantId(), ddpInstance);
+            clinicalKit.setNecessaryParticipantDataToClinicalKit(ddpParticipantId, ddpInstance);
+            new ClinicalKitDao().ifTissueAccessionedTriggerDDP(ddpParticipantId, ddpInstance);
```
zyme: It looks like `ClinicalKitDao` has no state.  If that's the case, then the methods should probably all be static.

pepper-apis/dsm-core/src/main/resources/liquibase/CMI/PEPPER-445_tissueReceivedEvent.xml
```diff
@@ -5,19 +5,19 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 
-    <changeSet author="simone" id="PEPPER-445_tissueReceivedEvent">
+    <changeSet author="simone" id="PEPPER-445_samplesReceivedEvent">
```
zyme, line 8: Just to sanity check since this is a change to an existing liquibase file--is it safe to assume that this can be applied without triggering a liquibase conflict?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
```diff
@@ -107,4 +130,80 @@ private void setAccessionTimeForSMID(String smIdValue) {
...
         }
     }
+
+    public void ifTissueAccessionedTriggerDDP(String ddpParticipantId, DDPInstance ddpInstance) {
+        if (hasTissueAccessioned(ddpParticipantId, ddpInstance)) {
+            triggerParticipantEvent(ddpInstance, ddpParticipantId, DBConstants.REQUIRED_SAMPLES_RECEIVED_EVENT);
+        }
+    }
+
+    public boolean hasTissueAccessioned(String ddpParticipantId, DDPInstance ddpInstance) {
```
zyme: I thought our general pattern was to always pass in `conn` so that callers could better manage transaction scope.  This method and `setAccessionTimeForSMID` both hide the fact that they are making single transaction updates, which means that callers may accidentally spawn multiple transactions and leave the database in an inconsistent state if their intent was to do all-or-nothing updates.

Things will also run slower if a route or CF needs to spawn multiple connections/transactions, so here's a good opportunity to speed things up and reduce the chance of future bugs.

pepper-apis/dsm-core/src/main/resources/liquibase/CMI/PEPPER-445_tissueReceivedEvent.xml
SimoneMaiwald: yes. liquibase is fine with that

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
SimoneMaiwald: only whitespace change because I like having the `WHERE` start in a new line for easier reading of query

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
SimoneMaiwald: normal samples wouldn't have a received date at all. we only know about clinical tissues being received per mercury call

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
SimoneMaiwald: I agree but this is how the bsp routes where historically made and I don't want for that change to refactor the whole handling of the connection now.
We should at one point revisit these things in DSM but I think that would take too much time at the moment

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
```diff
@@ -89,7 +112,7 @@ private void setAccessionTimeForSMID(String smIdValue) {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement stmt = conn.prepareStatement(SQL_SET_ACCESSION_TIME)) {
                 stmt.setLong(1, System.currentTimeMillis());
-                stmt.setString(2, ClinicalKitsRoute.MERCURY);
+                stmt.setString(2, MERCURY);
```
SimoneMaiwald, line 115: moved the constant from the route class into this one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/settings/EventTypeDao.java
```diff
@@ -75,12 +75,12 @@ public List<EventTypeDto> getEventTypeByInstanceName(@NonNull String instanceNam
         return eventTypeList;
     }
 
-    public Optional<EventTypeDto> getEventTypeByEventTypeAndInstanceId(@NotNull String eventType, @NonNull String instanceId) {
+    public Optional<EventTypeDto> getEventTypeByEventNameAndInstanceId(@NotNull String eventName, @NonNull String instanceId) {
```
SimoneMaiwald, line 78: just some renaming to make it easier

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/filter/query/BaseAbstractQueryBuilder.java
```diff
@@ -85,17 +85,17 @@ protected void buildUpQuery(List<String> filterValues, FilterStrategy filterStra
         if (ddpInstanceId != null && esIndex != null && DBConstants.OSTEO_INDEX.equals(esIndex)) {
             if (nestedQueryBuilderLists.containsKey(ESObjectConstants.DSM + DBConstants.ALIAS_DELIMITER + ESObjectConstants.KIT_REQUEST_SHIPPING)) {
                 creatingQueryBuilder(DBConstants.DDP_KIT_REQUEST_ALIAS + DBConstants.ALIAS_DELIMITER
-                                + DBConstants.DDP_INSTANCE_ID_CAMEL_CASE + Filter.EQUALS + ddpInstanceId,
+                                + ESObjectConstants.DDP_INSTANCE_ID_CAMEL_CASE + Filter.EQUALS + ddpInstanceId,
```
SimoneMaiwald: moving things around

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/BasePatch.java
```diff
@@ -213,7 +213,7 @@ protected void triggerParticipantEvent(DDPInstance ddpInstance, Patch patch, Val
         final EventDao eventDao = new EventDao();
         final EventTypeDao eventTypeDao = new EventTypeDao();
         Optional<EventTypeDto> eventType =
-                eventTypeDao.getEventTypeByEventTypeAndInstanceId(action.getName(), ddpInstance.getDdpInstanceId());
+                eventTypeDao.getEventTypeByEventNameAndInstanceId(action.getName(), ddpInstance.getDdpInstanceId());
```
SimoneMaiwald, line 216: just renaming

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/ExistingRecordPatch.java
SimoneMaiwald: was deleted because we use the sample accessioning now and not the tissue_received date anymore. change from cmi

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
SimoneMaiwald: :doh: I agree! not sure what I did here 😱

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
```diff
@@ -542,7 +541,7 @@ public class DBConstants {
     public static final String TRACKING = "tracking";
 
     //EVENTS
-    public static final String TISSUE_RECEIVED_EVENT = "TISSUE_RECEIVED";
+    public static final String REQUIRED_SAMPLES_RECEIVED_EVENT = "REQUIRED_SAMPLES_RECEIVED";
```
SimoneMaiwald, line 544: renaming the event because of the changed logic of when DSM is going to trigger DSS

pepper-apis/dsm-core/src/main/resources/liquibase/CMI/PEPPER-445_tissueReceivedEvent.xml
pegahtah: I think because the id is changed the hashes are not compared

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/ClinicalKitDao.java
```diff
@@ -29,8 +33,24 @@ public class ClinicalKitDao {
                     + "LEFT JOIN ddp_participant as p on (p.participant_id = inst.participant_id) "
                     + "LEFT JOIN ddp_instance as ddp on (ddp.ddp_instance_id = p.ddp_instance_id) "
                     + "LEFT JOIN sm_id_type sit on (sit.sm_id_type_id = sm.sm_id_type_id) "
-                    + "LEFT JOIN kit_type ktype on ( sit.kit_type_id = ktype.kit_type_id) WHERE sm.sm_id_value = ? "
-                    + "AND NOT sm.deleted <=> 1 ";
+                    + "LEFT JOIN kit_type ktype on ( sit.kit_type_id = ktype.kit_type_id) "
+                    + "WHERE sm.sm_id_value = ? AND NOT sm.deleted <=> 1 ";
+    private static final String SQL_GET_RECEIVED_CLINICAL_TISSUE_BY_DDP_PARTICIPANT_ID = "SELECT kit.received_date FROM ddp_kit "
```
pegahtah: I think the select should be on sm_id instead of kit for this one

pegahtah: ```SELECT sm.received_date from sm_id sm LEFT JOIN ddp_tissue t on (t.tissue_id = sm.tissue_id) "```

SimoneMaiwald: good catch! Thank you!!!

pepper-apis/dsm-core/src/main/resources/liquibase/CMI/PEPPER-445_tissueReceivedEvent.xml
SimoneMaiwald: 💯

pegahtah (approved):

ssettipalli (approved):

PR #2438, opened 2023-01-20
[DSS] PEPPER-429

pegahtah:
Draft PR for the PEPPER-429
Changes to the lms consent PDF is uploaded as a version 2 consent PDF, a new patch is written and a task to make the changes in the db


pepper-apis/studybuilder-cli/studies/lms/patches/consent-version2.conf
```diff
@@ -0,0 +1,98 @@
+{
```
ssettipalli: Can we rename this file as  consent-pdf-version2.conf so that we know its  pdf config.

ssettipalli (approved):

PR #2442, opened 2023-01-24
Pepper 176 lms dsm events

ssettipalli:
look at https://broadworkbench.atlassian.net/browse/PEPPER-176


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/LmsInsertDsmEvents.java
```diff
@@ -0,0 +1,58 @@
...
+                        .findByStudyGuid(studyGuid).getId()))
+                .filterBy(EventConfiguration::getEventActionType, EventActionType.ACTIVITY_INSTANCE_CREATION)
+                .filterBy(EventConfiguration::getEventTriggerType, EventTriggerType.DSM_NOTIFICATION)
+                .filter(this::hasDSMNotificationTrigger)
+                .filter(this::isExpectedTrigger)
+                .map(EventConfiguration::getEventConfigurationId)
+                .map(id -> deactivateEventConfiguration(handle, id))
+                .count();
+
+        log.info("Successfully deactivated {} DSM Notification events that create new activities of {}.", count, studyGuid);
```
zyme, line 39: Should there be explicit checks on the count so we make sure that we're removing the right number of things instead of just writing something out to the log?

zyme (approved):
LGTM!

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/LmsInsertDsmEvents.java
ssettipalli: Since the query is not checking already disabled/inactive events, usually count is different in Dev.
Moreover , we are just disabling and not actually removing. I should have called the method as disableExistingEvents.  I usually check those displayed events ids by querying in DB.

PR #2454, opened 2023-02-03
removing the event. needs to get added manually to db when IRB approved

SimoneMaiwald:
remove the new event for lms and os2 till it is irb approved. then it should get added manually to the db. info in ticket


SimoneMaiwald:
sql insert here (https://broadworkbench.atlassian.net/browse/PEPPER-445?focusedCommentId=73866)

pepper-apis/dsm-core/src/main/resources/liquibase/CMI/PEPPER-445_tissueReceivedEvent.xml
```diff
@@ -1,25 +0,0 @@
...
-        <insert tableName="event_type">
-            <column name="ddp_instance_id"
-                    valueComputed="(SELECT ddp_instance_id FROM ddp_instance WHERE instance_name='osteo2')"></column>
-            <column name="event_name" value="REQUIRED_SAMPLES_RECEIVED"/>
-            <column name="event_description" value="Samples received"/>
-            <column name="event_type" value="RECEIVED"/>
-        </insert>
-    </changeSet>
-
-</databaseChangeLog>
```
pegahtah, line 25: if we have to run it again at some point maybe we can keep this with a new id?

SimoneMaiwald: then we have to do it per release. if we do it per just db sql we don't need to do it pr RC…

pegahtah (approved):

PR #2483, opened 2023-03-09
[DSM] PEPPER-653 Code to compare data between ES and DB

pegahtah:
There is a data msmatch between ES and DB. I have wrote a uniit tests that gets all of the  participants from an index in ES, and compares their medical record data with what is in the database and writes the difference in an output file. We can use this data to detect the pattern and get to the bottom of what is causing this.


Rinchan20 (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/MedicalRecord.java
```diff
@@ -235,7 +251,7 @@ public static MedicalRecord getMedicalRecord(@NonNull ResultSet rs) throws SQLEx
                 rs.getBoolean(DBConstants.FOLLOWUP_REQUIRED), rs.getString(DBConstants.FOLLOWUP_REQUIRED_TEXT),
                 rs.getString(DBConstants.ADDITIONAL_VALUES_JSON), rs.getString(DBConstants.MR_UNABLE_OBTAIN_TEXT),
                 rs.getString(DBConstants.DDP_PARTICIPANT_ID), rs.getLong(DBConstants.DDP_INSTANCE_ID),
-                rs.getBoolean(DBConstants.NO_ACTION_NEEDED));
+                false);
```
pegahtah, line 254: because we don't have this in envs other than dev

pegahtah: this should not be merged

PR #2489, opened 2023-03-21
[DSM] Hotfix for GP Final Scan page

pegahtah:
Changed the ScanError class to write the short id to a different field

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
@@ -193,7 +193,7 @@ public class KitDaoImpl implements KitDao {
                 } else {
                     logger.info("Updated kitRequests w/ ddp_label " + kitRequestShipping.getDdpLabel());
                     if (StringUtils.isNotBlank(kitRequestShipping.getBspCollaboratorParticipantId())) {
-                        dbVals.resultValue = new ScanError(kitRequestShipping.getBspCollaboratorParticipantId(),
+                        dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(), null,
                                 kitRequestShipping.getBspCollaboratorParticipantId());
                     }
                 }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
@@ -51,7 +51,7 @@ public abstract class BaseKitUseCase implements Supplier<List<ScanError>> {
 
     private boolean isScanErrorOnlyBspParticipantId(ScanError scanError, String bspCollaboratorParticipantId) {
         return (StringUtils.isBlank(scanError.getError()) && StringUtils.isBlank(scanError.getKit()))
-                || (scanError.getError().equals(scanError.getKit()) && scanError.getError().equals(bspCollaboratorParticipantId));
+                || (StringUtils.isBlank(scanError.getError()) && scanError.getShortId().equals(bspCollaboratorParticipantId));
     }
 
     protected BaseKitUseCase getDecoratedScanUseCase() {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
@@ -46,13 +46,10 @@ public abstract class BaseKitUseCase implements Supplier<List<ScanError>> {
     protected abstract Optional<ScanError> processRGPFinalScan(ScanPayload scanPayload);
 
     protected boolean isKitUpdateSuccessful(Optional<ScanError> maybeScanError, String bspCollaboratorParticipantId) {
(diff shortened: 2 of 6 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -193,7 +193,7 @@ public Optional<ScanError> updateKitRequest(KitRequestShipping kitRequestShippin
                 } else {
                     logger.info("Updated kitRequests w/ ddp_label " + kitRequestShipping.getDdpLabel());
                     if (StringUtils.isNotBlank(kitRequestShipping.getBspCollaboratorParticipantId())) {
-                        dbVals.resultValue = new ScanError(kitRequestShipping.getBspCollaboratorParticipantId(),
+                        dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(), null,
```
pegahtah, line 196: setting error to null

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/ScanError.java
```diff
@@ -6,9 +6,17 @@
 public class ScanError {
     private String kit;
     private String error;
+    private String shortId;
 
     public ScanError(String kit, String error) {
         this.kit = kit;
         this.error = error;
+        this.shortId = null;
```
zyme: Probably don't need the explicit assignment here since `shortId` will be null already.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: So the underlying issue is that the two-arg version of this constructor was seeing `getBspCollaboratorParticipantId()` as an error?

pegahtah: Yes it was writing it in the error field.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/ScanError.java
pegahtah: true, I just pushed the change

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
```diff
@@ -51,7 +51,7 @@ protected boolean isKitUpdateSuccessful(Optional<ScanError> maybeScanError, Stri
 
     private boolean isScanErrorOnlyBspParticipantId(ScanError scanError, String bspCollaboratorParticipantId) {
```
zyme: `isScanErrorOnlyBspParticipantId` isn't relying on any state, so I'm wondering if this method should move to `ScanError` like so:

```
public boolean isScanErrorOnlyBspParticipantId(String bspCollaboratorParticipantId) {
        return (StringUtils.isBlank(error) && StringUtils.isBlank(kit))
                || (error.equals(kit) && error.equals(bspCollaboratorParticipantId));
    }
  ```
  And _then_ you could write a unit test of `ScanError.isScanErrorOnlyBspParticipantId()` that would be easy to setup without any database baggage.

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java (hunk 1 above)
Comment (zyme) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java:196]: Can you add a test for `kitDaoImpl.updateKitRequest` that covers this case?

Comment (pegahtah) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java:196]: I updated the PR with this change

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java (hunk 3 above)
Comment (pegahtah) [reply pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java]: I updated the PR with this change

File: pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitScanErrorTest.java [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+public class KitScanErrorTest {
+
+    @Test
+    public void testIsScanErrorOnlyBspParticipantId(){
+        String kit = "TEST_KIT_LABEL";
+        String error = null;
+        String shortId = "TEST_SHORT_ID";
+
+        ScanError notRealScanError = new ScanError(kit, error, shortId);
+        Assert.assertTrue(notRealScanError.isScanErrorOnlyBspParticipantId("TEST_SHORT_ID"));
```
zyme, line 17: Nice!  Maybe break these assertions out into three different test methods and then name them with along the lines of `testShortIdIsNotErrror`, `testNullShortIdNotError`, etc.  That way we'll have finer feedback about the nature of what's broken.

zyme (approved):
LGTM!  Just one small comment for breaking up the test methods.

pegahtah:
This is the hotfix for PEPPER-676  [Ticket here](https://broadworkbench.atlassian.net/browse/PEPPER-676). This PR will be closed and directly deployed to prod, another PR is made for the develop

PR #2498, opened 2023-03-30
PEPPER-554 . text update

ssettipalli:
minor text format fix in translation


Rinchan20 (approved):
checked in dev - looks good

PR #2501, opened 2023-04-04
PEPPER-560 . cleanup

ssettipalli:
minor cleanup

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import lombok.extern.slf4j.Slf4j;
 import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.dao.ddp.instance.DDPInstanceDao;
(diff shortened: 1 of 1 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -77,10 +79,7 @@ public static void updateCustomWorkflow(Map<String, String> attributesMap, Strin
 
                 try {
                     if (isATRelatedStatusUpdate(studyGuid)) {
-                        boolean hasGenomicStudyGroup = participantDatas.stream().anyMatch(
-                                participantDataDto -> ATDefaultValues.GENOME_STUDY_FIELD_TYPE.equals(
-                                        participantDataDto.getFieldTypeId().get()));
-                        logger.info("ddpParticipantId: " + ddpParticipantId + " hasGenomicStudyGroup " + hasGenomicStudyGroup);
+                        log.info("uploading AT values for ddpParticipantId: ", ddpParticipantId);
```
Rinchan20, line 82: this might need to be logger instead of log?
Edit - I looked at full file but didn't see log declared (I see logger), so not sure

ssettipalli: log is fine because of @Slf4j in line#30 
Lombok annotation

Rinchan20: no wait - I see Slf4j there

Rinchan20 (approved):

PR #2527, opened 2023-04-24
[DSM] PEPPER-745 GP Patch Access and some PatchRoute cleaning

pegahtah:
## Context

This PR's main focus is to let GP use the PatchRoute to change kit (and only kit) values, I have also added more loggings in the PatchRoute and changed the access check to check the user in the Patch object and match it with the user id in the token.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/pom.xml
+++ b/pepper-apis/dsm-core/pom.xml
@@ -62,6 +62,11 @@
             <artifactId>commons-dbcp2</artifactId>
             <!-- <version>2.1.1</version> -->
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.12.0</version>
+        </dependency>
 
         <!-- job scheduler -->
         <dependency>
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
@@ -50,7 +50,7 @@ public class PatchRoute extends RequestHandler {
         if ((UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_VIEW, userIdRequest, patch)
                 || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_ABSTRACTER, userIdRequest, patch)
                 || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.PT_LIST_VIEW, userIdRequest, patch))
-                || ( UserUtil.checkGPUserAccessForPatch(realm, userId, userIdRequest, patch))) {
+                || ( UserUtil.checkKitShippingAccessForPatch(realm, userId, userIdRequest, patch))) {
             try {
                 BasePatch patcher = PatchFactory.makePatch(patch, notificationUtil);
                 return patcher.doPatch();
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
@@ -50,8 +50,7 @@ public class PatchRoute extends RequestHandler {
         if ((UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_VIEW, userIdRequest, patch)
                 || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_ABSTRACTER, userIdRequest, patch)
                 || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.PT_LIST_VIEW, userIdRequest, patch))
-                || ( UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.KIT_SHIPPING, userIdRequest, patch)
-                && DBConstants.DDP_KIT_ALIAS.equals(patch.getTableAlias()))) {
+                || ( UserUtil.checkGPUserAccessForPatch(realm, userId, userIdRequest, patch))) {
             try {
                 BasePatch patcher = PatchFactory.makePatch(patch, notificationUtil);
                 return patcher.doPatch();
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
@@ -18,7 +18,11 @@ import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
 
-//Class needs to be refactored as soon as possible!!!
+/**
+ * DSM UI's `patch` requests are handled by this endpoint.
+ * The users who have mr_view and pt_list_view and mr_abstracter have access to make changes in the PatchRoute.
+ * Users with kit_shipping access (GP users) can only change kit values
+ * */
 public class PatchRoute extends RequestHandler {
     private static final Logger logger = LoggerFactory.getLogger(RequestHandler.class);
 
@@ -39,12 +43,16 @@ public class PatchRoute extends RequestHandler {
             throw new RuntimeException("ColumnNameMap is null!");
         }
         String userIdRequest = UserUtil.getUserId(request);
-        if (UserUtil.checkUserAccess(null, userId, DBConstants.MR_VIEW, userIdRequest)
-                || UserUtil.checkUserAccess(null, userId, DBConstants.MR_ABSTRACTER, userIdRequest)
-                || UserUtil.checkUserAccess(null, userId, DBConstants.PT_LIST_VIEW, userIdRequest)) {
+        String requestBody = request.body();
+        Patch patch = GSON.fromJson(requestBody, Patch.class);
+        String realm = patch.getRealm();
+        logger.info("Received a patch request, made by " + userId + " in realm "+realm +" for table alias " + patch.getTableAlias());
+        if ((UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_VIEW, userIdRequest, patch)
+                || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_ABSTRACTER, userIdRequest, patch)
+                || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.PT_LIST_VIEW, userIdRequest, patch))
+                || ( UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.KIT_SHIPPING, userIdRequest, patch)
+                && DBConstants.DDP_KIT_ALIAS.equals(patch.getTableAlias()))) {
             try {
-                String requestBody = request.body();
-                Patch patch = GSON.fromJson(requestBody, Patch.class);
                 BasePatch patcher = PatchFactory.makePatch(patch, notificationUtil);
                 return patcher.doPatch();
             } catch (DuplicateException e) {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
@@ -42,29 +42,31 @@ public class PatchRoute extends RequestHandler {
             response.status(500);
             throw new RuntimeException("ColumnNameMap is null!");
         }
-        String userIdRequest = UserUtil.getUserId(request);
-        String requestBody = request.body();
-        Patch patch = GSON.fromJson(requestBody, Patch.class);
-        String realm = patch.getRealm();
-        logger.info("Received a patch request, made by " + userId + " in realm "+realm +" for table alias " + patch.getTableAlias());
-        if ((UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_VIEW, userIdRequest, patch)
-                || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_ABSTRACTER, userIdRequest, patch)
-                || UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.PT_LIST_VIEW, userIdRequest, patch))
-                || ( UserUtil.checkKitShippingAccessForPatch(realm, userId, userIdRequest, patch))) {
-            try {
-                BasePatch patcher = PatchFactory.makePatch(patch, notificationUtil);
-                return patcher.doPatch();
-            } catch (DuplicateException e) {
-                response.status(500);
-                throw new RuntimeException("Duplicate value", e);
-            } catch (Exception e) {
-                response.status(500);
-                throw new RuntimeException("An error occurred while attempting to patch ", e);
+        try {
+            String userIdRequest = UserUtil.getUserId(request);
+            String requestBody = request.body();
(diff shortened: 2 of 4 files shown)


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,44 +1,110 @@
...
-        //        auth0Util.getClaimValue(auth0Token, clientId).ifPresent( claim -> claims.put(clientId, claim.asString()));
-        //        claims.put("https://datadonationplatform.org/uid", "C2JYXJUUBC2KP9R10G1U");
-        //        String jwtToken = new SecurityHelper().createToken(auth0Token, cookieAgeInSeconds +
-        //        (System.currentTimeMillis() / 1000) + (60 * 5),
-        //                claims);
-
-        Map<String, String> authHeaders = new HashMap<>();
-        authHeaders.put("Authorization", "Bearer " + auth0Token);
-        String reqAddress = "https://pepper-dev.datadonationplatform.org/pepper/v1/user/" + userId + "/studies/cmi-pancan/activities";
-        HttpResponse response = TestUtil.performGet(reqAddress, "", authHeaders).returnResponse();
```
pegahtah, line 41: This was a test made for FON only, I removed it because FON was cancelled and also in order to clean the code

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -18,7 +18,11 @@
 import spark.Request;
 import spark.Response;
 
-//Class needs to be refactored as soon as possible!!!
```
pegahtah, line 21: This comment is no longer valid, the class got refactored in 2022

pepper-apis/dsm-core/pom.xml
```diff
@@ -62,6 +62,11 @@
             <artifactId>commons-dbcp2</artifactId>
             <!-- <version>2.1.1</version> -->
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.12.0</version>
+        </dependency>
```
pegahtah, line 69: to use `isNumeric()`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -298,6 +324,11 @@ public static List<String> getUserRolesPerRealm(@NonNull String query, @NonNull
         return roles;
     }
 
+    public static boolean checkGPUserAccessForPatch(String realm, String userId, String userIdRequest, Patch patch) {
+        return UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.KIT_SHIPPING, userIdRequest, patch)
+                && DBConstants.DDP_KIT_ALIAS.equals(patch.getTableAlias());
+    }
```
pegahtah, line 330: GP can only change kit values

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -39,12 +43,15 @@ public Object processRequest(Request request, Response response, String userId)
             throw new RuntimeException("ColumnNameMap is null!");
         }
         String userIdRequest = UserUtil.getUserId(request);
-        if (UserUtil.checkUserAccess(null, userId, DBConstants.MR_VIEW, userIdRequest)
-                || UserUtil.checkUserAccess(null, userId, DBConstants.MR_ABSTRACTER, userIdRequest)
-                || UserUtil.checkUserAccess(null, userId, DBConstants.PT_LIST_VIEW, userIdRequest)) {
+        String requestBody = request.body();
```
denniscunningham: Looks like some exception throwing calls are no longer protected by try?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -39,12 +43,15 @@ public Object processRequest(Request request, Response response, String userId)
...
+        Patch patch = GSON.fromJson(requestBody, Patch.class);
+        String realm = patch.getRealm();
```
denniscunningham: I'm not familiar with Lombok, but can realm be null?

otchet-broad: While you're modifying this, do you want to check the patch to make sure the userId in the patch is the same as the ID from line 45?

otchet-broad: Ah....I need to keep reading.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -268,6 +271,29 @@ public static boolean checkUserAccess(String realm, String userId, String role,
         return false;
     }
 
+    public static boolean checkUserAccessForPatch(String realm, String userId, String role, String userIdRequest, Patch patch) {
+        String userEmailOrIdInPatch = patch.getUser();
+        String userIdFromPatch;
+        if (userEmailOrIdInPatch.contains("@")) {
+            // patches come with both user id or email in them, this is to find the id in case it was an email
+            UserDto user = new UserDao().getUserByEmail(userEmailOrIdInPatch).orElseThrow();
+            userIdFromPatch = user.getId() + "";
```
denniscunningham: Do we typically use this idiom rather than a toString or similar method?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,44 +1,110 @@
...
+    public static final String GP_UNIT_TESTER_EMAIL = "testUsersEmails.gpUnitTesterEmail";
+
+
+    @Before
+    public void first() {
+        setupDB();
+
+    }
+    @Test
+    @Ignore
```
otchet-broad, line 28: Can we get these classes and methods under test without turning off coverage?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -268,6 +271,29 @@ public static boolean checkUserAccess(String realm, String userId, String role,
         return false;
     }
 
+    public static boolean checkUserAccessForPatch(String realm, String userId, String role, String userIdRequest, Patch patch) {
```
denniscunningham, line 274: Can you comment on what we are doing here?
It looks like we are getting an email or ID from the request body but then check that is it also the verified user ID?(We also do this in checkUserAccess, but I don't understand what that accomplishes. Is it because Patch uses the user ID/email from the content at some point?)
That will help me better understand your tests.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -298,6 +324,11 @@ public static List<String> getUserRolesPerRealm(@NonNull String query, @NonNull
         return roles;
     }
 
+    public static boolean checkGPUserAccessForPatch(String realm, String userId, String userIdRequest, Patch patch) {
```
denniscunningham: Does it make sense to give this a more generic title like 'checkKitShippingAccess'?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
pegahtah: I am actually working on that too... changing configs to see if that fixes the errors

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
pegahtah: Yes, and since in the original access check it was actually using null, I really don't want to mess with that. We have to do a big refactoring in the frontend and check all of the times this call is made to see and fix when frontend is sending nulls.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -268,6 +271,29 @@ public static boolean checkUserAccess(String realm, String userId, String role,
         return false;
     }
 
+    public static boolean checkUserAccessForPatch(String realm, String userId, String role, String userIdRequest, Patch patch) {
+        String userEmailOrIdInPatch = patch.getUser();
+        String userIdFromPatch;
```
otchet-broad, line 276: The primitive form of the id from the dto is an int.  Why not keep it this way in this method so that the comparator at line 289 can be an equals check on the native field in the dto?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -39,12 +43,15 @@ public Object processRequest(Request request, Response response, String userId)
...
+        logger.info("Received a patch request, made by " + userId + " in realm "+realm +" for table alias " + patch.getTableAlias());
+        if ((UserUtil.checkUserAccessForPatch(realm, userId, DBConstants.MR_VIEW, userIdRequest, patch)
```
otchet-broad: Great job including the realm!  Thank you!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
otchet-broad: The consequence of this requires GP Users to be granted the role in each realm (as appropriate).  Is that a different PR?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
pegahtah: Of course!
The `userId` that is passed as an input is the userId that is retrieved from the token that was passed in this call. However, the `patch` object also contains a userId in it, which we use as an audit field to know who changed this record in the last edit. This method is supposed to check that what is passed in the `patch` object matches the token.

Patch calls from the frontend are inconsistent, some include the user emails in the `userId` and some the int value of the userId. We wanted to move to use the emails but it is not consistent, and I wanted to make sure we are covering both cases.

The `checkUserAccess` is matching what was passed in the request to what was in the token. So basically in patch calls 3 user identifiers are being passed to the backend, and we want to make sure they all match.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
pegahtah: I believe userId that is retrieved from the token is String, so we either have to convert one to String or the other one to int

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
pegahtah: Not sure what we use, different people have used different forms in the past.  I will change this to use toString and we can all stay consistent with that

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
pegahtah: That's the role they already have, and one of the reasons i am making the change for this role is so they can all change the values if they have access

the only patch value for kit currently is the `collection_date`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
pegahtah: Ah... yes! i like that more

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
pegahtah: I am not sure I understand...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
denniscunningham: Thx!
It seems the main use for the ID in the patch is for auditing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
denniscunningham: Let's say 'fromJson' throws an exception. In the past we would catch that and respond with a 500. Now it looks like we will not catch that? Is there something in Jetty/Spark that will catch that and convert it to a 500?
Or am I reading the code wrong?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
otchet-broad: Right.  The token value is a String, but we're validating on the database value (as we should!) since that's actually the control.  I'd prefer to validate on that value but see what you mean in how we're coercing the USER_ID into and out of the token with String.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
otchet-broad: @denniscunningham - It looks like processRequest throws Exception and it looks like the method caller is set up to handle an exception (spark RouteImpl).  That being said, this code should probably wrap the deserialization and application of the patch in a try/catch so that we can control the response and avoid leaking internals of JSON and DB processing.

otchet-broad: @denniscunningham - TIL there's a bit of "spark" magic that goes on but two other things to note. 1) we have the ability to wire up exception handling when we wire up the route.  It doesn't look like we currently do that.  2) The default exception handling in spark's GeneralError implementation seems to protect against leaks when a CustomErrorPage is not set up.  It simply returns a 500 with the text: "<html><body><h2>500 Internal Server Error</h2></body></html>"

pegahtah: Thank you both, this had skipped my eyes. I've now surrounded the whole method with try-catch.

denniscunningham (approved):

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
otchet-broad: @pegahtah - does this mean you're going to push changes in this PR that turn these tests back on or is that a future item?

PR #2554, opened 2023-05-15
[DSM] PEPPER-754/PEPPER-840 Juniper Kit classes, unit test and shipKit endpoint

pegahtah:
## Context

PEPPER-840
This PR is now updated to show changes for  PEPPER-840, it includes the endpoint and the service needed to create the kit

-----------
PEPPER-754
I created a unit test to mock what Juniper is supposed to send us in JuniperKitIntegration.java test class and created the JuniperShipKitRoute.java class basically from copied code to make sure the kit can be created in the System.

Little bit of tweaks here and there to make the code functional 
 

## Checklist

- x ] I have labeled the type of changes involved using the `C-*` labels.

## FUD Score

- [x] Requires other features before it's human visible. I have documented the blocking issues in jira.

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1184,8 +1184,9 @@ public static void updateKit(String dsmKitId, Shipment participantShipment, Ship
         if (results.resultException != null) {
             logger.error("Error updating kit w/ dsm_kit_id " + dsmKitId, results.resultException);
         } else {
-            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId, results.resultException);
-
+            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId);
```
pegahtah, line 1174: I fixed the error and move the upsert into a check

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
```diff
@@ -264,6 +265,30 @@ public Optional<DDPInstanceDto> getDDPInstanceByInstanceName(@NonNull String ins
         return Optional.ofNullable((DDPInstanceDto) results.resultValue);
     }
 
+    public Optional<DDPInstanceDto> getDDPInstanceByStudyGuid(@NonNull String studyGuid) {
```
pegahtah, line 268: This method was missing in DDPInstanceDao, basically this is to be able to select the ddpInstaceDto based on the stuyd-guid

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,320 @@
...
+import org.broadinstitute.dsm.util.KitUtil;
+import org.broadinstitute.dsm.util.externalshipper.ExternalShipper;
+import org.broadinstitute.lddp.handlers.util.Result;
+import org.broadinstitute.lddp.util.DeliveryAddress;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.Request;
+import spark.Response;
+
+public class JuniperShipKitRoute extends RequestHandler {
```
pegahtah: This route at this point has many assumptions and unanswered questions, basically I have moved code from other parts of DSM and changed them to fit Juniper, but kept a commented block here and there in case the assumptions are changed in the next days

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,320 @@
...
+////                        uploadReason, carrier);
+////            }
+//        } else {
+            handleKit(conn, ddpInstance, kitType, kit, kitRequestSettings, easyPostUtil, userIdRequest, kitTypeName,
+                    collaboratorParticipantId, KitUtil.IGNORE_AUTO_DEACTIVATION, uploadAnyway, duplicateKitList, orderKits,
+                    externalOrderNumber, uploadReason, carrier);
+//        }
+    }
+
+    private void handleKit(@NonNull Connection conn, @NonNull DDPInstance ddpInstance, @NonNull KitType kitType, @NonNull KitRequest kit,
```
pegahtah: When it is more figured out of what is and is not needed for Juniper, we can merge some of these handleKit methods into one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/JuniperKitRequest.java
```diff
@@ -0,0 +1,22 @@
...
+    private String firstName;
+    private String lastName;
+    private String street1;
+    private String street2;
+    private String city;
+    private String state;
+    private String postalCode;
+    private String country;
+    private String phoneNumber;
+    String juniperKitId;
```
zyme: Could re rename `juniper*` to `external*`?  Today it's juniper, but the API could also be used by other systems.

zyme: Gah never mind, I see we have `external` all over the place already...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1184,8 +1184,9 @@ public static void updateKit(String dsmKitId, Shipment participantShipment, Ship
...
+        }
```
denniscunningham, line 1176: So there's a case where the `setDsmKitId` call fails but we still want to continue?
If so, maybe an explanation?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,320 @@
...
+public class JuniperShipKitRoute extends RequestHandler {
+
+    private static final Logger logger = LoggerFactory.getLogger(JuniperShipKitRoute.class);
+
+    @Override
+    protected Object processRequest(Request request, Response response, String userId) throws Exception {
+        return null;
+    }
+
+    public Object createAJuniperKit(List<KitRequest> kitUploadObjects, String studyGuid, String kitTypeName,
```
denniscunningham: To echo Zim's point elsewhere, would createKit() work here instead?
Improves flexibility in case you want to use this more broadly, and I assume there is nothing particular to Juniper in this method?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperKits/JuniperKitIntegration.java
```diff
@@ -0,0 +1,61 @@
...
+import org.broadinstitute.dsm.model.KitRequest;
+import org.broadinstitute.dsm.model.KitUploadObject;
+import org.broadinstitute.dsm.route.JuniperShipKitRoute;
+import org.broadinstitute.dsm.statics.ApplicationConfigConstants;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class JuniperKitIntegration {
```
denniscunningham: Seems like our convention is to include the term `Test` on all tests?
A nit, but I'm not sure the term `Integration` adds much here since we typically don't label tests that way?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Oh I see what you mean, great catch! I will fix this

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/JuniperKitRequest.java
pegahtah: yeah... GBF...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
pegahtah: that makes sense, 
From my view, we are building a way for DSM to handle none-Pepper kits, so maybe we need some kind of a name like `createANonPepperKit()` ?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperKits/JuniperKitIntegration.java
pegahtah: Good point, how about JuniperKitCreationTest?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
denniscunningham: Perhaps a nit, but articles (the, a) are pretty much never used when naming things (at least in the software I've seen). So perhaps `createNonPepperKit`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,268 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(JuniperShipKitRoute.class);
+
+    @Override
+    protected Object processRequest(Request request, Response response, String userId) throws Exception {
+        return null;
+    }
+
+    public Object createNonPepperKit(KitRequest kitUploadObject, String studyGuid, String kitTypeName,
+                                    boolean skipAddressValidation, String userIdRequest, AtomicReference<String> shippingCarrier) {
+        DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByStudyGuid(studyGuid).orElseThrow();
```
denniscunningham: It looks like some other process adds Juniper studies to the DSM instance table?
How will that work?
As I noted in your design document, I am not comfortable having non-Pepper studies in this table since I don't understand the breadth on the implications.
IOW, it we go this route it would be nice to see an explanation in the design doc about how we keep the studies stored in the DDP instance table from polluting or confusing Pepper.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,268 @@
...
+        Map<String, KitRequest> invalidAddressList =
+                checkAddress(kitUploadObject, kitRequestSettings.getPhone(), skipAddressValidation, easyPostUtil);
+
+        //TODO Pegah discuss with Juniper if they need this, and if so, how the response should look like
+        List<KitRequest> duplicateKitList = new ArrayList<>();
+        ArrayList<KitRequest> orderKits = new ArrayList<>();
+
+        //TODO Pegah fix uploadKit to accept DDPInstanceDto
+        //getting the instance with isHasRole being set to true if the instance has role juniper_study
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole(ddpInstanceDto.getInstanceName(), "juniper_study");
```
denniscunningham: Should this go above, where we first establish the DDP instance?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,268 @@
...
+            handleKit(conn, ddpInstance, kitType, kit, kitRequestSettings, easyPostUtil, userIdRequest, kitTypeName,
+                    collaboratorParticipantId, KitUtil.IGNORE_AUTO_DEACTIVATION, uploadAnyway, duplicateKitList, orderKits,
+                    externalOrderNumber, uploadReason, carrier);
+
+        }
+    }
+
+
+
+    private void handleKit(@NonNull Connection conn, @NonNull DDPInstance ddpInstance, @NonNull KitType kitType, @NonNull KitRequest kit,
```
denniscunningham: What connection is this?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,268 @@
...
+        //getting the instance with isHasRole being set to true if the instance has role juniper_study
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole(ddpInstanceDto.getInstanceName(), "juniper_study");
+
+        if(!ddpInstance.isHasRole()){
+            throw new RuntimeException("This is not a Juniper study!");
+        }
+
+        TransactionWrapper.inTransaction(conn -> {
+            createKit(ddpInstance, kitType, kitUploadObject, kitRequestSettings, easyPostUtil, userIdRequest,
+                    kitTypeName, true, invalidAddressList, duplicateKitList, orderKits,
```
denniscunningham: For this code, and the methods below, is there corresponding code for Pepper?
IOW, don't we have to do something just like this for Pepper, or is there something unique about this code?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -568,13 +570,15 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
         get(apiRoot + RoutePath.BSP_KIT_REGISTERED, new BSPKitRegisteredRoute(), new JsonTransformer());
         get(apiRoot + RoutePath.CLINICAL_KIT_ENDPOINT, new ClinicalKitsRoute(notificationUtil), new JsonTransformer());
-        get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT, new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
-                new JsonTransformer());
-        get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT_WITH_PARTICIPANT,
-                new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
-                new JsonTransformer());
+
+        NotPepperKitCreationService notPepperKitCreationService = new NotPepperKitCreationService();
+        get(apiRoot + RoutePath.SHIP_KIT_ENDPOINT, new JuniperShipKitRoute(notPepperKitCreationService), new JsonTransformer());
```
pegahtah: New Route for kit creation

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -568,13 +570,15 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
 
         if (!cfg.getBoolean("ui.production")) {
+            get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT, new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
+                    new JsonTransformer());
+            get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT_WITH_PARTICIPANT, new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
+                    new JsonTransformer());
```
pegahtah, line 581: Moving these 2 dummy routes out of production

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
pegahtah: I now made a method a lot simpler, tuned it to Juniper variables and also combined some methods all together

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
pegahtah: I think we have discussed this in person before, but let me know if there are any concerns. 

I will add the liquibase for the new Juniper study once we/they know more about the study configs

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
denniscunningham: I realized we are trying to avoid using the term 'Juniper' -- but it is all over this code -- and it looks like we rejected 'external', but it would be really helpful if we came up for a term for these types of kits that would work well for future external kit shipment support. 'NotPepper' feels confusing, but of course that is just my opinion.
Does 'remoteKit' work? I'm up for anything that works, and if we absolutely cannot find a term, I'd prefer to use 'juniper' and remove 'NotPepper'.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
denniscunningham: These are temporary (i.e. will be removed in the future), or just not ready for production yet?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -62,17 +32,21 @@
...
-import org.broadinstitute.dsm.util.ElasticSearchUtil;
-import org.broadinstitute.dsm.util.KitUtil;
+import org.broadinstitute.dsm.util.*;
 import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
 import org.broadinstitute.lddp.db.SimpleResult;
 import org.eclipse.jetty.util.StringUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
```
denniscunningham: I believe java includes go earlier in the file?
Are you getting any lint feedback?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -996,19 +963,20 @@ public static String writeRequest(@NonNull String instanceId, @NonNull String dd
...
-                        Exportable.getParticipantGuid(ddpParticipantId, ddpInstance.getParticipantIndexES()),
-                        new PutToNestedScriptBuilder()).export();
-            } catch (Exception e) {
-                //This error will trigger on studies with no participants, this skips
-                //the error log if that is the reason for the upsert failure.
-                if (StringUtils.isNotBlank((ddpInstance.getParticipantIndexES()))) {
-                    logger.error(String.format("Error inserting newly created kit request shipping with dsm kit request id: %s in "
-                            + "ElasticSearch", kitRequestShipping.getDsmKitRequestId()));
-                    e.printStackTrace();
+            if (StringUtils.isNotBlank(ddpInstance.getParticipantIndexES())) {
```
denniscunningham: I'm guessing this is to exclude pushing Juniper kit info to ES?
If so, maybe a comment explaining that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1084,21 +1052,23 @@ public static void updateKit(long dsmKitId, String userId, DDPInstanceDto ddpIns
             return dbVals;
         });
 
+
```
denniscunningham: Maybe remove this extra blank line?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1184,17 +1154,19 @@ public static void updateKit(String dsmKitId, Shipment participantShipment, Ship
         if (results.resultException != null) {
             logger.error("Error updating kit w/ dsm_kit_id " + dsmKitId, results.resultException);
         } else {
-            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId, results.resultException);
+            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId);
+            if (ddpInstanceDto.getEsParticipantIndex() != null) {
```
denniscunningham: This code looks identical (or perhaps very close to the code above). 
Perhaps create a method to do this?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
```diff
@@ -62,8 +62,10 @@ protected Optional<ScanError> process(ScanPayload scanPayload) {
                     kitRequestShipping.setDdpLabel(ddpLabel);
                     kitRequestShipping.setScanDate(System.currentTimeMillis());
                     result = updateKitRequest(kitRequestShipping);
-                    trigerEventsIfSuccessfulKitUpdate(result, ddpLabel, kitRequestShipping);
-                    this.writeSampleSentToES(kitRequestShipping);
+                    if (StringUtils.isNotBlank(kitPayload.getDdpInstanceDto().getEsParticipantIndex())) {
```
denniscunningham: This test is repeated in several places.
Perhaps a separate method with a more expressive name would help code readability?
`is<Juniper>Kit`? Where we substitute a better term for 'Juniper'.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1084,21 +1052,23 @@ public static void updateKit(long dsmKitId, String userId, DDPInstanceDto ddpIns
...
             kitRequestShipping.setLabelDate(labelDate);
-
-            try {
-                UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto, ESObjectConstants.DSM_KIT_ID,
-                        ESObjectConstants.DSM_KIT_ID, dsmKitId, new PutToNestedScriptBuilder()).export();
-            } catch (Exception e) {
-                logger.error(String.format("Error updating label date of kit request shipping with dsm kit id: %s in ElasticSearch",
-                        dsmKitId));
-                e.printStackTrace();
+            if (StringUtils.isNotBlank(ddpInstanceDto.getEsParticipantIndex())) {
```
denniscunningham: See my comment above

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/KitRequest.java
```diff
@@ -35,6 +35,7 @@ public class KitRequest {
     @DbDateConversion(SqlDateConverter.EPOCH)
     private Long externalOrderDate;
 
+
```
denniscunningham: Are the changes in this file necessary?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/KitStatus.java
```diff
@@ -0,0 +1,7 @@
+package org.broadinstitute.dsm.model.notPepperKit;
+
+public class KitStatus extends KitResponse {
+    public KitStatus(String errorMessage, String juniperKitId) {
```
denniscunningham: Why the indirection of KitStatus? IOW, why not just use KitResponse?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+import java.sql.Connection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+@Slf4j
+public class NotPepperKitCreationService {
+
+    public final static String ADDRESS_VALIDATION_ERROR = "UNABLE_TO_VERIFY_ADDRESS";
```
denniscunningham: Are these strings codes that the requester is expecting? (Just wondering why they are done this way.)
If so, a comment explaining that would be helpful.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
```diff
@@ -0,0 +1,49 @@
...
+            response.status(400);
+            return new KitResponse("EMPTY_STUDY_NAME", null);
+        }
+        if (StringUtils.isBlank(shipKitRequest.getKitType())) {
+            response.status(400);
+            return new KitResponse("EMPTY_KIT_TYPE", null);
+        }
+        KitResponse kitResponse = this.kitCreationService.createNonPepperKit(juniperKitRequest, shipKitRequest.getJuniperStudyGUID(),
+                shipKitRequest.getKitType());
+        if (kitResponse instanceof KitStatus) {
```
denniscunningham: I'm not sure this pattern is particularly obvious. Why not just use KitResponse and check for an error message (or add an error flag if the message is not always an error)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+    public final static String JUNIPER_UNDERSCORE ="JUNIPER_";
+
+    public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, String studyGuid, String kitTypeName) {
+        if (StringUtils.isBlank(juniperKitRequest.getJuniperParticipantID())) {
+            return new KitResponse(MISSING_JUNIPER_PARTICIPANT_ID, juniperKitRequest.getJuniperKitId());
+        }
+        if (StringUtils.isBlank(juniperKitRequest.getJuniperKitId())) {
+            return new KitResponse(MISSING_JUNIPER_KIT_ID, null);
+        }
+        Optional<DDPInstanceDto> maybeDdpInstanceDto = new DDPInstanceDao().getDDPInstanceByStudyGuid(studyGuid);
```
denniscunningham: Would it be helpful to include the study ID in KitResponse?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        if (maybeDdpInstanceDto.isEmpty()) {
+            return new KitResponse(UNKNOWN_STUDY, juniperKitRequest.getJuniperKitId());
+        }
+        DDPInstanceDto ddpInstanceDto = maybeDdpInstanceDto.get();
+        //getting the instance with isHasRole being set to true if the instance has role juniper_study
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole(ddpInstanceDto.getInstanceName(), "juniper_study");
+
+        if (!ddpInstance.isHasRole()) {
```
denniscunningham: Are we leaking information by not just saying this is an invalid study ID, as above?
IOW, I'm not sure this provides useful information to a legitimate user, but would provide information to a bad actor.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+            throw new RuntimeException("This is not a Juniper study!");
+        }
+        HashMap<String, KitType> kitTypes = KitType.getKitLookup();
+        String key = kitTypeName + "_" + ddpInstanceDto.getDdpInstanceId();
```
denniscunningham: If this is a common pattern for creating the key it should be part of KitType.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        KitType kitType = kitTypes.get(key);
+        if (kitType == null) {
+            return new KitResponse(UNKNOWN_KIT_TYPE, juniperKitRequest.getJuniperKitId());
```
denniscunningham: As above, should we add kit type to the response object?
Would the requester find that information helpful?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        Map<Integer, KitRequestSettings> kitRequestSettingsMap =
+                KitRequestSettings.getKitRequestSettings(String.valueOf(ddpInstanceDto.getDdpInstanceId()));
+        KitRequestSettings kitRequestSettings = kitRequestSettingsMap.get(kitType.getKitTypeId());
+
+        // if the kit type has sub kits > like for testBoston
```
denniscunningham: Unclear what is going on here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+
+        if (!checkAddress(juniperKitRequest, kitRequestSettings.getPhone(), easyPostUtil)) {
+            return new KitResponse(ADDRESS_VALIDATION_ERROR, juniperKitRequest.getJuniperKitId());
+        }
+
+        ArrayList<KitRequest> orderKits = new ArrayList<>();
+
+
+        TransactionWrapper.inTransaction(conn -> {
+            createKit(ddpInstance, kitType, juniperKitRequest, kitRequestSettings, easyPostUtil, kitTypeName,  orderKits, conn);
```
denniscunningham: Not sure what's going on here. How do we know if the kit was created successfully?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+
+//            only order if external shipper name is set for that kit request, not needed for now
+//            if (StringUtils.isNotBlank(kitRequestSettings.getExternalShipper())) {
+//                return orderExternalKits(kitRequestSettings, orderKits, easyPostUtil, shippingCarrier, conn);
+//            }
+            return null;
+        });
+
+        //send not valid address back to client
```
denniscunningham: Is this comment correct?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * checkAddress tries creating an address in EasyPost. If it is successful,
+     * sets the kit's easypostAddressId and returns true, if not returns false
+     * An address is valid only if participant has shortId, first - and lastName, for Juniper shortId is the juniperParticipantId
+     *
+     * @param juniperKitRequest the JuniperKitRequest with address to check
+     */
+
+    public boolean checkAddress(JuniperKitRequest juniperKitRequest, String phone, EasyPostUtil easyPostUtil) {
+        if ((StringUtils.isBlank(juniperKitRequest.getJuniperParticipantID()) && StringUtils.isBlank(juniperKitRequest.getExternalOrderNumber()))
```
denniscunningham: Can you comment on what this test is doing? If we have a participant ID we are ok with not having this other info?
Looks like we use (at least) last name below.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+        name += juniperKitRequest.getLastName();
+        if (juniperKitRequest.isSkipAddressValidation()) {
+            try {
+                Address address = easyPostUtil.createBroadAddress(name, juniperKitRequest.getStreet1(), juniperKitRequest.getStreet2(), juniperKitRequest.getCity(),
+                        juniperKitRequest.getPostalCode(), juniperKitRequest.getState(), juniperKitRequest.getCountry(), phone);
+                juniperKitRequest.setEasypostAddressId(address.getId());
+                return true;
+            } catch (EasyPostException e) {
+                log.error("Easypost couldn't create an address for " + juniperKitRequest.getShortId());
```
denniscunningham: Is it possible this was just a bad request?
If so we should probably not log at the error level.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        String errorMessage = "";
+        String collaboratorParticipantId = KitRequestShipping
+                .getCollaboratorParticipantId(null, ddpInstance.getDdpInstanceId(), ddpInstance.isMigratedDDP(),
+                        ddpInstance.getCollaboratorIdPrefix(), kit.getJuniperParticipantID(), kit.getJuniperParticipantID(),
+                        kitRequestSettings.getCollaboratorParticipantLengthOverwrite());
+
+        //In case it's needed, subkits handling should be added here
+        String shippingId;
+        String userId;
+        //checking ddpInstance.isHasRole() to know this is a Juniper Kit
```
denniscunningham: It seems like we already did this checking elsewhere.
If so, perhaps something more like an assertion at the beginning of the method is needed (a check that throws a 500 error).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        if (StringUtils.isNotBlank(kit.getJuniperKitId()) || ddpInstance.isHasRole()) {
+            shippingId = JUNIPER_UNDERSCORE + kit.getJuniperKitId();
+            userId = JUNIPER;
+        } else {
+            log.error("Seems like {} is not a JUNIPER study! ", ddpInstance.getName());
```
denniscunningham: If this is a request error we should not log at error level.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+            KitRequestShipping.writeRequest(ddpInstance.getDdpInstanceId(), shippingId, kitTypeId, kit.getParticipantId().trim(),
+                    collaboratorParticipantId, collaboratorSampleId, userId, addressId, errorMessage, externalOrderNumber, false,
+                    null, ddpInstance, bspCollaboratorSampleType, ddpLabel);
+            kit.setDdpLabel(shippingId);
+            kit.setExternalOrderNumber(externalOrderNumber);
+        } else {
+            if (kitRequestSettings.getCollaboratorSampleTypeOverwrite() != null) {
+                bspCollaboratorSampleType = kitRequestSettings.getCollaboratorSampleTypeOverwrite();
+            }
+            if (StringUtils.isNotBlank(collaboratorParticipantId)) {
```
denniscunningham: Perhaps a comment about what the collaborator participant ID indicates?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
```diff
@@ -0,0 +1,238 @@
...
+        log.info("Address is not valid " + juniperKitRequest.getShortId());
+        return false;
+
+    }
+
+
+    /**
+     * This method creates a collaborator participant id for the kit and then inserts in DB
+     */
+    private void createKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitType, JuniperKitRequest kit,
```
denniscunningham: Somewhere here or in the caller(s) there should be some exception handling?

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: They are dummy kits created for GP tests last year, We normally take them out of production after the tests are completed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Yes, will add

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/KitStatus.java
pegahtah: This is for now just a placeholder for `KitStatus`, but the reason why I put it like this is, Juniper is thinking of after calling the shipKit for kitCreation, to either get a 400 with an error and a kit id (hence the `KitResponse`) or in case of success, get a 200 with a `KitStatus` message with every field set to either null or info about creation (like created time, etc). They also want to get a `KitStatus` object when they call the status endpoint for getting updates on the kit. Therefore I made the KitStatus as a child to `KitResponse`. This class will have all of the different variables and details of a kit in it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
pegahtah: I see what you mean... 
The reason I made it like that is, when we are returning a `KitStatus` object we can still have errorMessages in it that are not particularly related to this request, like the fedex was out or the cost object didn't work, stuff like that. However, if we had a bad request, we are going to return just a KitResponse with an error message related to that request. I was trying to somehow differentiate based on that.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Thank you for pointing it out

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
denniscunningham: At this point KitResponse is identical to KitStatus, but are you imagining that KitStatus will contain more information in the future? If so, who will act on that information?
One way to handle this is "You aren't gonna need it" (YAGNI)... add KitStatus later when it actually is different from KitResponse. But in any case it feels like KitStatus "has a" KitResponse (class composition) rather than KitStatus "is a" KitResponse.
What do you think?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
denniscunningham: I thought more about this an have an opinion, FWIW. 
The thing that's bothering me are the terms 'notPepper' and 'nonPepper' since they don't fit with the rest of the DSM code. IOW, we don't use the term Pepper in the code (at least not in any prominent places), so calling something non-pepper does not fit with the code.
Could we change 'notPepper' to 'external' throughout this code, and just leave the references to 'Juniper' as is? Does that make sense?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -996,19 +991,21 @@ public static String writeRequest(@NonNull String instanceId, @NonNull String dd
...
+            if (StringUtils.isNotBlank(ddpInstance.getParticipantIndexES())) {
+                try {
+                    UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto,
+                            ESObjectConstants.DSM_KIT_REQUEST_ID, ESObjectConstants.DOC_ID,
+                            Exportable.getParticipantGuid(ddpParticipantId, ddpInstance.getParticipantIndexES()),
+                            new PutToNestedScriptBuilder()).export();
+                } catch (Exception e) {
+                    //This error will trigger on studies with no participants, this skips
+                    //the error log if that is the reason for the upsert failure.
+                    if (StringUtils.isNotBlank((ddpInstance.getParticipantIndexES()))) {
```
denniscunningham: It seems like we did this check above to enter this code. Has anything changed since then?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1091,15 +1088,7 @@ public static void updateKit(long dsmKitId, String userId, DDPInstanceDto ddpIns
...
-
-            try {
-                UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto, ESObjectConstants.DSM_KIT_ID,
-                        ESObjectConstants.DSM_KIT_ID, dsmKitId, new PutToNestedScriptBuilder()).export();
-            } catch (Exception e) {
-                logger.error(String.format("Error updating label date of kit request shipping with dsm kit id: %s in ElasticSearch",
-                        dsmKitId));
-                e.printStackTrace();
-            }
+            upsertUpdatedKitInfoIntoES(ddpInstanceDto, kitRequestShipping, dsmKitId);
```
denniscunningham, line 1089: Are we calling upsertUpdatedKitInfoIntoES for Juniper kits? If so, why?

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: Nice! Yes will do!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: That's fair! Will make changes now to the Juniper kit creation part. I'd rather not change the original DSM code until we have great test coverage in place.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: This is the check for subkits (multiple tubes and barcode in one package) . This is not a need right now so I commented the code so we don't spend time debugging it but kept it to for the time that we need it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: of course, it checks if we have some sort of a kit id (or external id) at this point, and it also checks that there is a last name. It will fail if there is no last name or if there is no id.
However, I will remove the external order check because they don't use external shippers anyways and we don't want to make the mistake of passing the condition with an id that is not used

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: changed to warn

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/notPepperKit/NotPepperKitCreationService.java
pegahtah: Not sure how or what to add, it is the id of the participant in GP's view... (I'll add that but not sure how much value it adds to the code reader, unfortunately some of these are just values and ids that GP needs and may not make any sense outside of the scope of DSM)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/JuniperShipKitRoute.java
pegahtah: Oh yes, KitStatus will have every tiny detail about the kit. sent date, received date, dsm shipping id, label, who scanned it and when, basically everything. 

Juniper will get it when they call the status endpoint (coming soon :) ) but they want to get the same object for a successful ship kit call as well. I haven't implemented that yet so for the first draft they are ok with just getting a 200 and this needs more consideration and may change in future.

Currently to me, KitStatus _is_ a KitResponse. A KitResponse could be an error, or it could be all of these other informations

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: external in terms of DSM vocabulary has a different meaning, and that is when a kit is shipped through or working in a non-broad GP

I am more than happy to change the names to Juniper and keep thinking about a good name. Since it is a first draft, I think we can push this to dev so Juniper team can use it while we come up with a better name. To me, for now, Juniper will do, especially since I don't think we will make this service for anything else. What do you think?

We can also create a ticket to come up with a better name, put the code in dev so Juniper engineers can use it and change the name later

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: I added the check in the method, so that method has a check if there is an index in ES

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
pegahtah: Yeah I was thinking about that too, not sure where is best place to put such a method. The check is all over the place

Maybe we can make a code improvement ticket for this epic so we can address them later while getting a working code to dev so we don't block them while we discuss refactoring and design?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1184,10 +1171,16 @@ public static void updateKit(String dsmKitId, Shipment participantShipment, Ship
...
             logger.error("Error updating kit w/ dsm_kit_id " + dsmKitId, results.resultException);
         } else {
-            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId, results.resultException);
+            logger.info("Updated kit w/ dsm_kit_id " + dsmKitId);
+            upsertUpdatedKitInfoIntoES(ddpInstanceDto, kitRequestShipping, dsmKitId);
+        }
+    }
 
+    private static void upsertUpdatedKitInfoIntoES(DDPInstanceDto ddpInstanceDto, KitRequestShipping kitRequestShipping, Object dsmKitId) {
+        if (ddpInstanceDto.getEsParticipantIndex() != null) {
```
pegahtah: @denniscunningham this prevents updates for Juniper

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: removed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+        if (kitType == null) {
+            return new KitResponse(UNKNOWN_KIT_TYPE, juniperKitRequest.getJuniperKitId(), kitTypeName);
+        }
+
+        Map<Integer, KitRequestSettings> kitRequestSettingsMap =
+                KitRequestSettings.getKitRequestSettings(String.valueOf(ddpInstance.getDdpInstanceId()));
+        KitRequestSettings kitRequestSettings = kitRequestSettingsMap.get(kitType.getKitTypeId());
+
+        // if the kit type has sub kits > like for testBoston
+        //        boolean kitHasSubKits = kitRequestSettings.getHasSubKits() != 0;
```
denniscunningham, line 67: Is this code still relevant?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+        name += juniperKitRequest.getLastName();
+        if (juniperKitRequest.isSkipAddressValidation()) {
+            try {
+                Address address = easyPostUtil.createBroadAddress(name, juniperKitRequest.getStreet1(), juniperKitRequest.getStreet2(),
+                        juniperKitRequest.getCity(),
+                        juniperKitRequest.getPostalCode(), juniperKitRequest.getState(), juniperKitRequest.getCountry(), phone);
+                juniperKitRequest.setEasypostAddressId(address.getId());
+                return true;
+            } catch (EasyPostException e) {
+                log.warn("Easypost couldn't create an address for " + juniperKitRequest.getShortId());
```
denniscunningham: Unclear what's going on here. Can you add a comment?
Will we get this exception for a bad request, or is this more like an internal error?
Why are we logging at warning level?
The results of the stack trace are going, I assume, to stdout or stderr, which may or may not be logged.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+        EasyPostUtil easyPostUtil = new EasyPostUtil(ddpInstance.getName());
+
+        if (!checkAddress(juniperKitRequest, kitRequestSettings.getPhone(), easyPostUtil)) {
+            return new KitResponse(ADDRESS_VALIDATION_ERROR, juniperKitRequest.getJuniperKitId(), null);
+        }
+
+        ArrayList<KitRequest> orderKits = new ArrayList<>();
+
+
+        SimpleResult result = TransactionWrapper.inTransaction(conn -> {
```
denniscunningham, line 79: Is there a reason this transaction clause is not in `createKit`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+     * @return SimpleResult of the interaction with database
+     */
+    private SimpleResult createKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitType, JuniperKitRequest kit,
+                                   @NonNull KitRequestSettings kitRequestSettings, @NonNull EasyPostUtil easyPostUtil,
+                                   @NonNull String kitTypeName, ArrayList<KitRequest> orderKits, Connection conn,
+                                   SimpleResult transactionResults) {
+        String juniperKitRequestId;
+        String userId;
+        //checking ddpInstance.isHasRole() to know this is a Juniper Kit
+        if (StringUtils.isNotBlank(kit.getJuniperKitId()) || ddpInstance.isHasRole()) {
```
denniscunningham: It seems like we already checked this further up the stack?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+
+        return transactionResults;
+    }
+
+    private SimpleResult addJuniperKitRequest(Connection conn, String kitTypeName, KitRequestSettings kitRequestSettings,
+                                              DDPInstance ddpInstance,
+                                              int kitTypeId, String collaboratorParticipantId, String errorMessage,
+                                              EasyPostUtil easyPostUtil,
+                                              JuniperKitRequest kit, String externalOrderNumber, String juniperKitRequestId,
+                                              String ddpLabel, String userId, SimpleResult transactionResults) {
```
denniscunningham, line 196: I'm not a fan of lots of parameters -- it makes for poor readability. Is there a way to encapsulate these into a class that is acting as a struct? Or more conventionally, create a kit shipping class that has much of this info as instance variables.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+        try {
+            Address address = easyPostUtil.getAddress(kit.getEasypostAddressId());
+            if (address != null) {
+                addressId = address.getId();
+            }
+        } catch (EasyPostException e) {
+            throw new RuntimeException("EasyPost addressId could not be received ", e);
+        }
+
+        if (StringUtils.isNotBlank(kitRequestSettings.getExternalShipper())) {
```
denniscunningham, line 209: Can you add a comment about how we handle external shippers differently?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+                //             Not needed now
+                //            kit.setExternalOrderNumber(externalOrderNumber);
+            } catch (Exception e) {
+                transactionResults.resultException = e;
+            }
+        } else {
+            if (kitRequestSettings.getCollaboratorSampleTypeOverwrite() != null) {
+                bspCollaboratorSampleType = kitRequestSettings.getCollaboratorSampleTypeOverwrite();
+            }
+            if (StringUtils.isNotBlank(collaboratorParticipantId)) {
```
denniscunningham, line 229: Can you comment on what this case is? What is a `collaboratorParticipantId` and what does this code accomplish?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+                collaboratorSampleId =
+                        KitRequestShipping.generateBspSampleID(conn, collaboratorParticipantId, bspCollaboratorSampleType, kitTypeId);
+                if (collaboratorParticipantId == null) {
```
denniscunningham: How can this be null? The test above (isNotBlank) should assure that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+                    errorMessage += "collaboratorParticipantId was too long ";
+                }
+                if (collaboratorSampleId == null) {
+                    errorMessage += "collaboratorSampleId was too long ";
```
denniscunningham, line 233: I don't think this works (this will not update errorMessage for the caller). Do you have a test for this?
Also, it seems like you are attempting to return errors in two different ways, via the return and via this parameter. One way would be less confusing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+
+            String participantID = kit.getJuniperParticipantID();
+            try {
+                String dsmKitRequestId =
+                        KitRequestShipping.writeRequest(ddpInstance.getDdpInstanceId(), juniperKitRequestId, kitTypeId, participantID,
+                                collaboratorParticipantId, collaboratorSampleId, userId, addressId, errorMessage,
+                                kit.getExternalOrderNumber(),
+                                false,
+                                null, ddpInstance, bspCollaboratorSampleType, ddpLabel);
+                kit.setDdpLabel(juniperKitRequestId);
```
denniscunningham: This clause is similar to that above, but here it includes `kit.setDdpLabel`. Is that intentional?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/JuniperKitStatus.java
```diff
@@ -0,0 +1,7 @@
+package org.broadinstitute.dsm.model.nonpepperkit;
+
+public class JuniperKitStatus extends KitResponse {
```
denniscunningham, line 3: I still don't think this provides a clear way to implement response handling. IOW, I do not see what value this technique provides.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
denniscunningham: Right. As we discussed it would be nice to create a more explicit way of expressing that.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -0,0 +1,276 @@
...
+
+    }
+
+
+    /**
+     * This method creates a collaborator participant id for the kit and then inserts in DB
+     *
+     * @return SimpleResult of the interaction with database
+     */
+    private SimpleResult createKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitType, JuniperKitRequest kit,
```
denniscunningham, line 153: The use of SimpleResult as a return type feels clumsy. We are jamming response information into a throwable type that we ultimately will not throw.
Would it make more sense to return a KitResponse?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: No the error message here will be written to the database. It is one of the GP errors if we can't fit a participant id into their desired length for barcodes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/JuniperKitStatus.java
pegahtah: I don't think there is any value one way or the other. IOW it's just a matter of how you look at if JuniperKitStatus is a KitResponse or has a KitResponse

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: I want to keep the subkit code in there so in any case if anyone wants to add them in neaqr future there is a clue of where to begin

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: addressed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: I agree, we can do this in the next versions

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: true! hmmm! the code is like this in the KitUpload and KitRequestShipping and I just moved it over... I will remove this check

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: it is not used anymore so I removed it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: I moved it to DDPInstanceDto and DDPInstance classes, since it is not really showing a Juniper kit , but the ability for the study, wether or not it has the index to update

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -124,8 +124,8 @@ public boolean checkAddress(JuniperKitRequest juniperKitRequest, String phone, E
                 juniperKitRequest.setEasypostAddressId(address.getId());
                 return true;
             } catch (EasyPostException e) {
-                log.warn("Easypost couldn't create an address for " + juniperKitRequest.getShortId());
-                e.printStackTrace();
+                // log the reason for address creation failure and return false. The method will then return the error code
+                log.warn("Easypost couldn't create an address for " + juniperKitRequest.getShortId(), e);
```
denniscunningham, line 128: Is this due to an bad request or an internal error?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
denniscunningham: Oh, ok. So this is a non-fatal error? Does GP view it? If so, how?

I'm unclear on why errorMessage is passed into addJuniperKitRequest. It appears to be empty when passed.

denniscunningham (approved):
I don't think this code is ready for PROD, so my approval is only for release to Dev so Juniper can test. 
We need to figure out how to limit the impact of this code should it not have the desired changes before the next DSM release.

PR #2571, opened 2023-05-31
[DSM] PEPPER-873 

pegahtah:
## Context

Today during a debug session we realized we are not keeping the message and response we sent and received from Mercury, since these kind of debugs are happening often I added 2 more fields to ddp_mercury_sequencing to keep them

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-873_mercury_order_fields.xml
```diff
@@ -0,0 +1,17 @@
...
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="PEPPER-873_mercury_new_fields" author="pegah">
+        <addColumn tableName="ddp_mercury_sequencing">
+            <column name="order_message" type="TEXT"></column>
+            <column name="status_message" type="TEXT"></column>
+        </addColumn>
+    </changeSet>
+    <changeSet id="PEPPER-873_mercury_status_detail_changes" author="pegah">
+        <modifyDataType columnName="status_detail" newDataType="VARCHAR(2000)" tableName="ddp_mercury_sequencing"/>
```
pegahtah, line 15: There are errors in dev that the field is small so I changed the size

otchet-broad: Why not convert the field to text?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -171,10 +173,19 @@ public HashMap<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String
 
             return dbVals;
         });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Error getting the mercury orders", results.resultException);
```
denniscunningham, line 178: What will catch and process this exception?
Is updateOrderStatus always called from a pubsub message processor?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -171,10 +173,19 @@ public HashMap<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String
...
+        }
         log.info(String.format("Found %d possible barcodes for participant %s", map.size(), ddpParticipantId));
         return map;
     }
 
+    @Override
+    public int create(MercuryOrderDto mercuryOrderDto) {
+        return 0;
```
otchet-broad: Should this really throw something like ("Not implemented") instead of returning 0?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
otchet-broad: @denniscunningham - In this pattern, we seem to catch and swallow `SQLException` within the transaction scope at line 168/170.  If I'm understanding the goal of this change, it's to gain more insight into the errors being encountered here that aren't currently bubbling up.

@pegahtah  - Please confirm.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderPublisher.java
```diff
@@ -99,6 +91,15 @@ public void onSuccess(String messageId) {
         }
     }
 
+    private String createMercuryUniqueOrderId() {
+        String orderNumber = NanoIdUtil.getNanoId("1234567890QWERTYUIOPASDFGHJKLZXCVBNM", 20);
```
otchet-broad, line 95: What's the volume of orders here?  
How often will we have ID collision?
Are these for human consumption?  If not, how about a UUID?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
denniscunningham: Thanks Elliot. But where exactly will this exception be caught and reported?

otchet-broad: If I understand our codebase correctly, this will bubble all the way up to and get logged via the RequestHandler path through Spark.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderPublisher.java
pegahtah: I think the volume is ~ 10 orders every month or so

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-873_mercury_order_fields.xml
pegahtah: @otchet-broad if our theory is correct I think we will see a size increase in the messages again

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
denniscunningham: Thx. I read the call stack incorrectly earlier. I see the route now.

Is there translation of RuntimeException to a specific HTTP status code in Spark? 
In this case, is this a request error or an internal error?

otchet-broad: Any that aren't handled directly are turned into a 500.  In this case, it will present as a 500 to the user.

denniscunningham: Is this an internal error or a problem with the request?

otchet-broad: The best answer for this case is: yes - it's either a request error or an internal one.  It's likely a case of a bad request, but there's at least one edge case I can see that bubbles up where it wouldn't be: the database goes down during this part of the request/response lifecycle.
The only thing this addition is doing is surfacing the error instead of returning a possibly incomplete map.
What would you like to see happen here?  We could continue with the pattern as-is, change it for this case, or change it for all of the cases.

denniscunningham: Thx for the explanation Elliot.
I'd like us to return an appropriate response code from all our routes, but I'm starting to realize that's a big job. So for now if we know the code we want to return (for code under change or development) we should try to align exceptions with response codes. (I see this, but I am not sure if there is a better approach: http://sparkjava.com/documentation.html#error-handling. I have not used Spark before -- at least not this Spark.)
If there is not a clean opportunity to quickly fix those issues here, of course go ahead with the merge. I appreciate what Pegah is trying to accomplish with the PR.

otchet-broad (approved):
Thanks Pegah!

PR #2576, opened 2023-06-07
Pepper-890 dss file download

ssettipalli:
## Context

New DSS route to get signed participant download URL

## FUD Score

## Testing

## Release


pepper-apis/config/application.conf.ctmpl
```diff
@@ -67,6 +67,12 @@
         "scanResultSubscription": "{{$conf.Data.fileUploads.scanResultSubscription}}",
         "signerServiceAccount": {{$conf.Data.fileUploads.signerServiceAccount | toJSONPretty}}
     },
+    "fileDownload":{
+        "signerServiceAccount": {{$conf.Data.fileDownload.signerServiceAccount | toJSONPretty}}
+        "maxSignedUrlMins": 5,
+        "defaultDownloadBucket": "{{$conf.Data.fileDownload.defaultDownloadBucket}}"
```
otchet-broad, line 73: I've been thinking about this in my own code and have decided that I don't need this in configuration.  The flow I'm expecting in DSM is:
DSM authorizes upload to bucket `studyA-upload` (where studyA is the notional name of the study)
User uploads to studyA bucket via DSM UI (POST, non-resumable).

- Anti-virus notified about new upload.
- Upload moved to final resting bucket (e.g. `studyA`)
- Message posted to DSM with the original and new bucket/blob names and/or virus scanning outcome.
- DSM records new bucket/blob names and virus scanning outcome.

Only after DSM stores the scanning outcome will the Follow-Up message be possible to generate.  In that message that we pass to DSS will need to be the bucket name, blob name and ddp participant id.

Because the anti-virus is going to determine the final resting spot for the data, I got rid of my local configuration for the final resting spots based on realm.  They need to be in the messages and not in the configuration.  IMHO, any "default" handling should fail so that we don't accidentally return something we're unsure about.

I also realize that you might want this value in for testing.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
```diff
@@ -102,6 +102,8 @@ public static final class API {
                 PathParam.USER_GUID, PathParam.STUDY_GUID, PathParam.INSTANCE_GUID);
         public static final String USER_ACTIVITY_UPLOADS = fmt(BASE + "/user/%s/studies/%s/activities/%s/uploads",
                 PathParam.USER_GUID, PathParam.STUDY_GUID, PathParam.INSTANCE_GUID);
+        public static final String USER_ACTIVITY_DOWNLOADS = fmt(BASE + "/user/%s/studies/%s/activities/%s/questions/%s/downloads",
```
otchet-broad: Should this be **download** instead of **downloads**?  Should there be a one to many relationship implied in the path?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
```diff
@@ -0,0 +1,86 @@
...
+
+@Slf4j
+public class FileDownloadService {
+    private final ServiceAccountSigner signer;
+    private final GoogleBucketClient storageClient;
+    private final int maxSignedUrlMins;
+    private final String defaultDownloadBucket;
+
+    //GCP Service Account for file download
+    public static final String FILE_DOWNLOAD_CREDENTIALS = "fileDownload.signerServiceAccount";
```
otchet-broad, line 34: Nice! I'm planning on using the same configuration for the signerServiceAccount in DSM.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
```diff
@@ -0,0 +1,86 @@
...
+     * Generating a signed URL for file download
+     *
+     * @param fileName   the name of the file
+     * @param bucketName the name of the bucket
+     * @return URL
+     */
+    public URL getSignedURL(String fileName, String bucketName) {
+        HttpMethod method = HttpMethod.GET;
+        return storageClient.generateSignedUrl(
+                signer, StringUtils.isEmpty(bucketName) ? defaultDownloadBucket : bucketName, fileName,
```
otchet-broad: If it's empty, I wouldn't go to a default bucket configuration.  I'd throw an exception at this point because we can't afford to connect a user to the wrong file.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
```diff
@@ -0,0 +1,68 @@
...
+public class GetParticipantFileDownloadRoute implements Route {
+    private static FileDownloadService fileDownloadService;
+
+    public GetParticipantFileDownloadRoute(FileDownloadService service) {
+        fileDownloadService = service;
+    }
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
```
otchet-broad, line 40: What ensures the request is authorized for this user and that the parameters aren't just guessed?
I'd expect to see code that ensures the active user is either the participant or the proxy and return a 403 if not and a 401 if no authentication information were provided.

pepper-apis/config/application.conf.ctmpl
ssettipalli: Its configuration. Depending on DSM setup , we need to add appropriate bucket to the vault/config entry

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
ssettipalli: user might need to do multiple downloads. As of now one at a time only though. Just went with similar to current "uploads" API in DSS.  I am ok to change if you prefer it.

otchet-broad: Does the code expect a one:many relationship?  If so, I'm okay with the plural and by convention it will be 1:1.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
ssettipalli: As of now assuming top level bucket will be same for results/download files and not expecting to pass for each request. Its more of. for any future use.
Assuming DSM will pass the full path that is specific to each participant results_file with each on-demand trigger. we can discuss more if needed

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
ssettipalli: All the User requests that come through DSS server go through UserAuthCheckFilter.java and passed through only if user is Authorized to the study. 
We can discuss if we want to do something more for this API.

otchet-broad: Right.  We want to make sure they're authorized for the study and that they're either the participant or the proxy.  I don't think our current filter covers that case, does it?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/RouteConstants.java
ssettipalli: Each call only one signed url. will change it.

ssettipalli: updated

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
otchet-broad: DSM needs to provide the bucket and blob.  full stop.

That being said, there's nothing to guarantee that the blob will still exist when you go to create the signed URL, because DSM also has a feature to delete one of these documents.  Not sure if it makes sense to check to see if the bucket/blob still exist before returning the signed URL.  Happy to discuss this further.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
ssettipalli: I believe thats what UserAuthCheckFilter does 
 canAccess = ddpAuth.canAccessGovernedUsers(requestedUserGuid);
 canAccess = ddpAuth.canAccessStudyDataForUser(requestedUserGuid, study);

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
ssettipalli: From my understanding for launch we wont handle any deletions. DSS side deletions will be data change requests to hide that appropriate activity instance.
If the file doesn't exist it will eventually fail when the user tries to download.

otchet-broad: PEPPER-67 requires DSM handle deletion at day 1.  This doesn't mean the activity needs to be cleaned up.  We just need to realize the download link may point to an invalid file.  You have the opportunity to check and see if the file exists at this point and could return something so the client can decide to do something different.  Maybe  a getBlob(bucket, blobPath).exists() and return a 404 if it's been deleted?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
```diff
@@ -76,10 +78,15 @@ public FileDownloadService(ServiceAccountSigner signer, GoogleBucketClient stora
      * @param bucketName the name of the bucket
      * @return URL
      */
-    public URL getSignedURL(String fileName, String bucketName) {
+    public URL getSignedURL(String fileName, String bucketName) throws FileNotFoundException {
+        String bucket = StringUtils.isEmpty(bucketName) ? defaultDownloadBucket : bucketName;
```
ssettipalli, line 82: will deal with defaultBucket once we have broader discussion

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
```diff
@@ -76,10 +78,15 @@ public FileDownloadService(ServiceAccountSigner signer, GoogleBucketClient stora
...
         HttpMethod method = HttpMethod.GET;
+        Blob blob = storageClient.getBlob(bucket, fileName);
+        if (blob == null || !blob.exists()) {
+            throw new FileNotFoundException();
```
ssettipalli, line 86: checking for file existence

ssettipalli: updated

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
ssettipalli: Added additional Authentication check to make sure user guid and passed instance Guid are Authorized

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/service/FileDownloadService.java
ssettipalli: Updated code already for File check

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
```diff
@@ -0,0 +1,91 @@
...
+                throw ResponseUtil.haltError(response, 404, new ApiError(ErrorCodes.USER_NOT_FOUND, "User not found"));
+            }
+            Long userId = userOpt.get().getId();
+
+            ActivityInstanceDao instanceDao = handle.attach(ActivityInstanceDao.class);
+            Optional<ActivityInstanceDto> instanceDtoOpt = instanceDao.findByActivityInstanceGuid(instanceGuid);
+            if (!instanceDtoOpt.isPresent()) {
+                throw ResponseUtil.haltError(response, 404, new ApiError(ErrorCodes.ACTIVITY_NOT_FOUND, "Instance not found"));
+            }
+            //make sure the instanceGuid passed belong to the participant guid passed
```
ssettipalli, line 74: Other Authorization scenarios are covered by UserAuthCheckFilter (canAccess)

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetParticipantFileDownloadRoute.java
```diff
@@ -0,0 +1,91 @@
...
+            ActivityInstanceDto activityInstanceDto = instanceDtoOpt.get();
+            if (activityInstanceDto.getParticipantId() != userId) {
```
otchet-broad, line 76: Will this hold true for proxies?

ssettipalli: activityInstance always belong to participant, so this should be good for proxies too.
Proxy value comes as operatorGuid and UserAuthCheckFilter checks and makes sure operator has access to this participant data. UserAuthCheckFilter/DDPAuth/UserPermissions.canAccessStudyDataForUser.

otchet-broad: I'm still missing something.  Where is the currently authenticated user checked to see if they are the participant or a proxy for them?  I see the parameter extracted from the URL path, but not where we're checking the value from the Bearer auth.

otchet-broad (approved):
LGTM!

ssettipalli:
> LGTM!

Thank you for taking time and checking it out thoroughly

PR #2578, opened 2023-06-09
[PEPPER-811, PEPPER-813, PEPPER-885] Initial implementation of somatic file upload.

otchet-broad:
## Context

[PEPPER-811](https://broadworkbench.atlassian.net/browse/PEPPER-811), [PEPPER-813](https://broadworkbench.atlassian.net/browse/PEPPER-813) and [PEPPER-885](https://broadworkbench.atlassian.net/browse/PEPPER-885) 
DSM is the entry point to share somatic results with users.  This pull request implements file upload service layer by gathering metadata and issuing signed urls so that a DSM user can upload any number of files.

Once uploaded, the files are stored in GCS and await an antivirus solution that will scan the files, disposition (delete or move) them to a final resting location, and notify DSM that the file is ready for a follow-up survey to be issued.  The antivirus cloud function will be completed in a follow-up PR.

This PR stores file metadata including how to find the uploaded file in both the relational database and Elasticsearch.

Three endpoint routes have been created to support the frontend in three ways:
1) Obtaining the signed url so the client can upload the file to GCS without DSM needing to process the bits.
2) The ability for the client to request file deletion
3) Obtaining the latest list of file uploads for a given realm/participant

Releasing requires configuration updates for test/stg/prod and needs to be documented in the release plan.  I'll seek assistance from @pegahtah so we can prepare the environment with credentials if needed.

## Checklist

- [X] I have labeled the type of changes involved using the `C-*` labels.
- [X] I have assessed potential risks and labeled using the `R-*` labels.
- [X] I have considered error handling and alerts, and added `L-*` labels as needed.
- [X] I have considered security and privacy, and added `I-*` labels as needed
- [X] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [ ] N/A If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ ] N/A If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

## FUD Score

- [X] :sweat_smile: There might be some issues here

- [X] Requires other features before it's human visible. I have documented the blocking issues in jira.

## Testing

- [X] I have written automated positive tests

## Release

- [X] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

[PEPPER-811]: https://broadworkbench.atlassian.net/browse/PEPPER-811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PEPPER-813]: https://broadworkbench.atlassian.net/browse/PEPPER-813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PEPPER-885]: https://broadworkbench.atlassian.net/browse/PEPPER-885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pepper-apis/config/testing-inmemorydb.conf.ctmpl
```diff
@@ -208,6 +210,12 @@
           }
       {{end}}
     }
+    "somatic": {
+        "realmToBucketMappings": [
+            {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
```
otchet-broad, line 215: Updated to match PEPPER-882 PR.

pepper-apis/dsm-core/pom.xml
```diff
@@ -129,6 +129,11 @@
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
             <!-- <version>6.8.2</version> -->
         </dependency>
+        <dependency>
```
otchet-broad, line 132: Added for file name verification that Greg uses in DUOS.

pepper-apis/config/testing-inmemorydb.conf.ctmpl
```diff
@@ -107,6 +107,7 @@
         "signerServiceAccount": {{$fileUploads.signerServiceAccount | toJSONPretty}}
     },
     "fileDownload":{
+        "credentials": {{$conf.Data.fileDownload.signerServiceAccount | toJSONPretty}},
```
otchet-broad, line 110: Heads up @pegahtah - Between PEPPER-890 and this PR, there are some changes to the fileDownload "testing" configuration.  My addition at 110 ensures the credentials object contains what DSM expects.  PEPPER-890 added what DSS expects.
When I took develop's changes after 890 merged, I got tripped up by the re-rendering.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,408 @@
...
+public class SomaticResultUpload implements HasDdpInstanceId {
+    @ColumnName(ESObjectConstants.SOMATIC_DOCUMENT_ID)
+    private final Long somaticDocumentId;
+
+    @ColumnName(DBConstants.DDP_INSTANCE_ID)
+    private final Long ddpInstanceId;
+
+    private final String ddpParticipantId;
+
+    @ColumnName(DBConstants.PARTICIPANT_ID)
```
denniscunningham, line 49: Why do we need both ddpParticipantId and participantId?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,408 @@
...
+    private final Long participantId;
+
+    @ColumnName(DBConstants.FILE_NAME)
+    private final String fileName;
+
+    @ColumnName(DBConstants.MIME_TYPE)
+    private final String mimeType;
```
denniscunningham, line 56: Why is mimeType needed? Or is it for future reference in case we need to know the type of the file?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,408 @@
...
+    private final Long createdByUserId;
+
+    @ColumnName(DBConstants.CREATED_AT)
+    private final long createdAt;
+
+    @ColumnName(DBConstants.DELETED_BY_USER_ID)
+    private final Long deletedByUserId;
+
+    @ColumnName(DBConstants.DELETED_AT)
+    private final long deletedAt;
```
denniscunningham, line 72: It looks like this becomes a legacy audit record once the file is deleted, but the queries below do not account for that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,408 @@
...
+            + "deleted_at = ? WHERE bucket = ? AND blob_path = ?";
+
+    private static final String SQL_DELETE_DOCUMENT_BY_DOCUMENT_ID = "UPDATE somatic_documents SET deleted_by_user_id = ?, "
+            + "deleted_at = ? WHERE id = ?";
+
+    private static final String SQL_INSERT_SOMATIC_DOCUMENT = "INSERT INTO somatic_documents SET ddp_instance_id = "
+            + "(SELECT ddp_instance_id FROM ddp_instance WHERE instance_name = ?), file_name = ?, "
+            + "mime_type = ?, bucket = ?, blob_path = ?, created_by_user_id = ?, created_at= ?, "
+            + "participant_id = (SELECT p.participant_id from ddp_participant p "
+            + "LEFT JOIN ddp_instance as ddp ON (ddp.ddp_instance_id = p.ddp_instance_id) "
```
denniscunningham, line 110: What happens if a user deletes a file, then creates a new version of the file with the same name?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,408 @@
...
+     * @param mimeType         Mime type of the file
+     * @param bucket           The bucket in GCS
+     * @param blobPath         Path of the file in GCS
+     * @param createdByUserId  User who uploaded the file
+     * @param createdAt        Epoch of when the file was uploaded
+     * @param deletedByUserId  User who deleted the file
+     * @param deletedAt        Epoch of when the file was deleted
+     * @param isVirusFree       True when the file has been through file scanning and moved to study bucket.
+     */
+    public SomaticResultUpload(Long somaticDocumentId, long ddpInstanceId, String ddpParticipantId, long participantId, String fileName,
```
denniscunningham, line 130: Looks like you skipped the DAO, DTO pattern?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/Dsm.java
```diff
@@ -60,6 +61,9 @@ public class Dsm {
     @SerializedName(ESObjectConstants.ONC_HISTORY_DETAIL_RECORDS)
     List<OncHistoryDetail> oncHistoryDetail;
 
+    @SerializedName(ESObjectConstants.SOMATIC_RESULT_UPLOAD)
```
denniscunningham, line 64: Just curious, what is this file used for?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -183,6 +184,7 @@ private void filterOsteoFromOtherOsteoOutForExport(UnparsedESParticipantDto unpa
             filterData((Map<String, List<Map<String, Object>>>) dsmObject, ESObjectConstants.KIT_REQUEST_SHIPPING, ddpInstanceId);
             filterData((Map<String, List<Map<String, Object>>>) dsmObject, ESObjectConstants.MEDICAL_RECORD, ddpInstanceId);
             filterData((Map<String, List<Map<String, Object>>>) dsmObject, ESObjectConstants.ONC_HISTORY_DETAIL, ddpInstanceId);
+            filterData((Map<String, List<Map<String, Object>>>) dsmObject, ESObjectConstants.SOMATIC_RESULT_UPLOAD, ddpInstanceId);
```
denniscunningham, line 187: We can take this offline,  but I am also curious about what this file does.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/somatic/result/SomaticResultMetaData.java
```diff
@@ -0,0 +1,28 @@
+package org.broadinstitute.dsm.model.somatic.result;
+
+import lombok.NonNull;
+
+public class SomaticResultMetaData {
```
denniscunningham, line 9: Can you add a comment saying what this class is for. (It is not yet obvious from a review perspective, but there may be a pattern I am missing.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
```diff
@@ -0,0 +1,186 @@
...
+        ServiceAccountCredentials signerCredentials;
+        try {
+            signerCredentials = ServiceAccountCredentials.fromStream(signerStream);
+        } catch (IOException e) {
+            throw new DDPException("Could not get signer credentials", e);
+        }
+
+        GoogleCredentials bucketCredentials;
+        boolean ensureDefault = false;
+        bucketCredentials = GoogleCredentialUtil.initCredentials(ensureDefault);
```
denniscunningham: Nit, but this could be assigned above?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
```diff
@@ -0,0 +1,186 @@
...
+        if (bucketCredentials == null) {
+            log.error("Could not get bucket credentials, defaulting to signer credentials");
+            bucketCredentials = signerCredentials;
+        }
+
+        String projectId = cfg.getString(ApplicationConfigConstants.FILE_DOWNLOAD_PROJECT_ID);
```
denniscunningham, line 61: Just curious, what does 'project' refer to?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
```diff
@@ -0,0 +1,186 @@
...
+
+        URL signedURL = storageClient.generateSignedUrl(
+                signer, uploadBucket, blobPath,
+                maxSignedUrlMins, TimeUnit.MINUTES,
+                HttpMethod.PUT, headers);
+
+        return new AuthorizeResult(AuthorizeResultType.OK, signedURL, createdUpload, somaticUploadSettings);
+    }
+
+    public SomaticResultUpload deleteUpload(long userId, int documentId) {
```
denniscunningham: Are all exceptions that can occur in this method internal errors? Likewise for `authorizeUpload`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private String getUploadsBucket(String realm) {
+        if (this.realmToUploadBucketMap.isEmpty()) {
+            throw new RuntimeException("Uploading documents not fully configured.  Please contact a DSM developer");
+        }
+        return this.realmToUploadBucketMap.get(realm.toLowerCase());
+    }
+
+    private enum AuthorizeResultType {
```
denniscunningham, line 184: Seems like some of these would be 400 errors?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    protected Object processRequest(Request request, Response response, String userId) throws Exception {
+        Gson gson = new Gson();
+        QueryParamsMap queryParams = request.queryMap();
+        String realm = queryParams.get(RoutePath.REALM).value();
+        String ddpParticipantId = queryParams.get(RoutePath.DDP_PARTICIPANT_ID).value();
+        SomaticResultMetaData somaticResultMetaData = gson.fromJson(request.body(), SomaticResultMetaData.class);
+        if (isValidRequest(realm, ddpParticipantId, userId, somaticResultMetaData)) {
+            return service.authorizeUpload(realm, userId, ddpParticipantId, somaticResultMetaData);
```
denniscunningham, line 38: How will the return of `AuthorizeResult` be processed after it is returned here?

denniscunningham (commented):
Thanks for all the work and some nice code to read!
Left some questions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: It's a future placeholder in case the file type changes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: We need the internal ID when doing DSM related key constraints and the ddpParticipantId to communicate with DSS for PEPPER-903.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: They do not.  We can let the CRCs know that a file has been uploaded and since deleted.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: It saves as expected.  There's a UUID generated on the blobPath but no constraint that the "fileName" has to be unique.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: I did.  I wanted to keep this with the least amount of indirection needed - since the client gets all of the information at points it didn't make sense to me to translate between.  Plenty in DSM already skips the DAO, DTO pattern.  Happy to implement if you'd like to see it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/Dsm.java
otchet-broad: Writing to Elastic.  It's the "thing" that ultimately creates the `dsm` portion of the document in participant_structured.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
otchet-broad: It could.  I was going between two patterns (one in DSS and one in DSM) and ended here.  I'll update.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
otchet-broad: The GCP `project` name.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
otchet-broad: Maybe.
IMHO, the following errors are worthy of 500s.
138 - we marked the database as deleted but the GCS object wasn't.  There's nothing for a user to do - and I would expect this to alert us.
165 - DSM isn't properly configured for uploads - nothing a user can do.

The exception thrown at 144 might be worthy of a different exception and error handling - but it also might be an internal error (e.g. misconfiguration).  I assumed misconfiguration but can see where we might allow that one to proceed without error.  Have any strong feelings about this before I refactor?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
otchet-broad: It will contain the signedUrl (and related record metadata) so the client can post the pdf into GCS and visualize the entry (or errors) in the list of uploaded files in the tab the user is working in.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
otchet-broad: Routes use this "wrapper" to build a collection of data from the participants_structured index.  Happy to discuss more offline.  IMHO, this needs more testing and is the source of some of the OS1/OS2 issues.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: We also need the ddpParticipantId for updating Elastic during the `create` use case.

denniscunningham: I realize there is a pattern in place already, but I'm not a fan of storing the DDP participant ID anywhere by the participant table. It lacks key constraints, and is mirrored by a real iD. 
As I take a second look at the code, it seems like we should verify the DDP participant ID at some point in process, ideally sooner than later, and stash that and the participant ID as needed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: What would we do differently based on the file type?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
denniscunningham: But it also has error information?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
denniscunningham: I was mostly looking at the AuthorizeResultTypes, which are used earlier in the file.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: Maybe there is nothing to do here, but I imagine for many uses of these queries the caller will have to then filter out deleted files. Is that not the case?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: Mostly curious because as you mentioned there are two patterns, and I've just been dealing with code that uses *both* patterns for the same table simultaneously. I don't feel particularly strongly about the pattern, just trying to avoid the situation where both are used for the same table.
As I look at this I don't see a strong need for a DTO, but this long call signature feels like a builder might be in future revisions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
otchet-broad: It does. AuthorizeResult will return a response so the client can tell if the fileName meets naming requirements, if the mimeType is one in the configured list to support, or if the file exceeds the maximum size allocated.  See AuthorizeResultType.  In cases where this happens, we also send back an object that contains the settings the client should be enforcing (e.g. mime type, max file size).  In an error state, the signed URL will be null and so will the SomaticResultUpload since we're rejecting the file.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: I'm storing the `DSM` participant ID and joining to get the DDP Participant ID.  See `SQL_SELECT_DOCUMENTS_BASE`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: Not sure.  Would you like it removed?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
denniscunningham: Ah, so we are using the mimetype. :-)
But these bad requests will all go back with status = 500?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: The caller will be able to answer the question:
What are the somatic results that have been provided to this participant.  Some may have been deleted (by users or by virus scanning) - and the study staff may be fielding phone calls related to the deleted files since a follow-up survey may have been created.
It's trivial to filter out the deleted files.

Happy to implement a change here.  What would you like to see?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: Maybe we can discuss offline? I'm still not sure why we need to store DDP participant ID here.
And it seems like we will be deep into the code before we discover a bad DDP participant ID was used in the request.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: I boiled down to the same opinion.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
otchet-broad: Ah, those are handed back to the client.  It's an existing pattern DSS uses to communicate with their FileUpload pattern there.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
otchet-broad: No.  Change pushed to clarify which the user can respond to and which are internal.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: FWIW, I'm using the value supplied for validation so it is memorializing information provided when the file was uploaded but otherwise doesn't really matter.  It's easy to remove.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/somaticresults/PostSomaticResultUploadRoute.java
otchet-broad: We are.  We validate it against the list of supported mime types and will return a 200 with a payload that indicates the mime type isn't supported.

otchet-broad: This behavior is cloned from DSS' own file upload component.

denniscunningham: Let's talk directly when you get a chance?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: If you feel the main use case is to return deleted entries along with live ones, then no change is needed. In any case, if you've thought about the use cases and are comfortable with this implementation, then so am I. Just wanted to be sure you understood my concern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: Thx for the DM explanation. This was my misunderstanding of the columnname annotation.
We are NOT storing the DDP participant ID in SOMATIC_DOCUMENTS_TABLE, which was my concern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: My understanding from DM is that we are using mimetype to verify file type. Although we only have one type now, this would be useful in a future where we store other types of somatic result files. 
Your call on the likelihood of that future.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
denniscunningham: I see that we don't currently offer 'by filename' query criteria, so this seems fine.
Thx for the explanation.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/SomaticResultUploadService.java
denniscunningham: Discussed in person.

denniscunningham (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -17,7 +17,7 @@
 import org.slf4j.LoggerFactory;
 
 public class MercuryOrderStatusListener {
-    private static final Logger logger = LoggerFactory.getLogger(DSMtasksSubscription.class);
+    private static final Logger logger = LoggerFactory.getLogger(MercuryOrderStatusListener.class);
```
pegahtah, line 20: Thank you

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -0,0 +1,411 @@
...
+                stmt.setString(8, ddpParticipantId);
+                stmt.setString(9, realm);
+                int result = stmt.executeUpdate();
+                if (result == 1) {
+                    try (ResultSet rs = stmt.getGeneratedKeys()) {
+                        if (rs.next()) {
+                            dbVals.resultValue = rs.getInt(1);
+                        }
+                    } catch (Exception e) {
+                        throw new RuntimeException("Error inserting somatic file", e);
```
pegahtah, line 291: Are we handling none unique names with this exception as here?

pegahtah (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
otchet-broad: We're making the files in GCS unique with the blobPath.  This is the metadata out of the file that was uploaded.  See SomaticResultUploadService line 107 for how we're building the blobPath.

otchet-broad: ...and I forgot to mention that there's a UNIQUE constraint on bucketName+blobPath in pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-811_somatic_documents.xml

PR #2582, opened 2023-06-14
PEPPER-824 . somatic-results workflow special case handling

ssettipalli:
## Context

SOMATIC_RESULTS has special use case where same activity should be displayed to user even when current SOMATIC_RESULTS activity is submitted. 
SOMATIC_RESULTS can have multiple instances at any gives point of time.
Current DSS platform pulls latest activity instance if multiple instances exist.
Added a special case for SOMATIC_RESULTS.

## FUD Score

## Testing

- [ x] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [x ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
@@ -43,6 +43,7 @@ import org.broadinstitute.ddp.db.dto.UserActivityInstanceSummary;
 import org.broadinstitute.ddp.exception.DDPException;
 import org.broadinstitute.ddp.json.PutAnswersResponse;
 import org.broadinstitute.ddp.json.errors.ApiError;
+import org.broadinstitute.ddp.json.workflow.WorkflowActivityResponse;
 import org.broadinstitute.ddp.json.workflow.WorkflowResponse;
 import org.broadinstitute.ddp.model.activity.definition.FormActivityDef;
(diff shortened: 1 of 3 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
```diff
@@ -214,6 +215,14 @@ public Object handle(Request request, Response response) {
                     WorkflowResponse workflowResp = workflowService
                             .suggestNextState(handle, operatorGuid, userGuid, studyGuid, fromState)
                             .map(nextState -> {
+                                //special case for SOMATIC_RESULTS ... need to pass same instance in next state and NOT latest
+                                if (instanceDto.getActivityCode().equalsIgnoreCase("SOMATIC_RESULTS")
```
ssettipalli, line 219: To be extra safe added SOMATIC_RESULTS check even though its not clean. Better to be extra safe rather

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/PutFormAnswersRoute.java
```diff
@@ -214,6 +215,14 @@ public Object handle(Request request, Response response) {
...
+                                        && fromState.matches(nextState)) {
```
ssettipalli, line 220: Workflow state is defined to return same state for SOMATIC_RESULTS and hence this will match

otchet-broad (approved):

PR #2581, opened 2023-06-13
pepper-68 basic implementation

denniscunningham:
Support for pepper-68.

The implementation is not yet complete -- there are TODOs, etc. -- but this is intended to support integration with the frontend, and limited manual testing on the Dev instance for Osteo2.
Pls review what is there now, and feel free to comment on things that appear missing but are not captured with a TODO or are related to LMS testing.

## Checklist

- [x] I have labeled the type of changes involved using the `C-*` labels.
- [x] I have assessed potential risks and labeled using the `R-*` labels.
- [x] I have considered error handling and alerts, and added `L-*` labels as needed.
- [x] I have considered security and privacy, and added `I-*` labels as needed
- [ ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [ ] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ ] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

_If unsure or need help with any of the above items, add the `help wanted` label. For items that starts with `If applicable`, if it is not applicable, check it off and add `n/a` in front._

## FUD Score

_Overall, how are you feeling about these changes?_

- [x] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [x] They are user-visible in dev as a regular user journey and require no additional instructions.
- [ ] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [x] I have written automated positive tests
- [x] I have written automated negative tests
- [ ] I have written zero automated tests but have poked around locally to verify proper functionality
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [x] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #2588, opened 2023-06-15
Follow-up for pepper-68

denniscunningham:
pepper-68
This follow-up PR covers a few areas:
- Changes indicated by code review of prior pepper-68 PR (https://github.com/broadinstitute/ddp-study-server/pull/2581).
- Support for appending updates to ES (oncHistoryDetail).
- Fix to file upload processor (remove trailing carriage returns and newlines).
- LMS automated testing

There will be at least one more PR to round out automated testing and set the access role properly, but this PR is necessary now to enable manual testing.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -377,6 +408,31 @@ private static void createFieldSettings() {
...
+                .withDisplayType("OPTIONS")
+                .withPossibleValues("[{\"value\": \"Needs Review\"},"
+                        + "{\"value\": \"Don't Request\"},"
+                        + "{\"value\": \"On Hold\"},"
+                        + "{\"value\": \"Request\"},"
+                        + "{\"value\": \"Sent\"},"
+                        + "{\"value\": \"Received\"},"
+                        + "{\"value\": \"Returned\"},"
+                        + "{\"value\": \"Unable to Obtain\"}]").build();
+        fieldSettingsDao.create(fieldSettings);
```
otchet-broad: Do these need to be cleaned up after the class is finished?

denniscunningham: Now that these are deployed as part of Liquibase (PR merged earlier) I can remove this method.

denniscunningham:
New commit pushed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -41,22 +43,33 @@ public void export() {
         updateByQueryRequest.setMaxRetries(5);
         updateByQueryRequest.setRefresh(true);
         updateByQueryRequest.setAbortOnVersionConflict(false);
+
+        Throwable ioException = null;
         for (int tryNum = 1; tryNum < 3; tryNum++) {
-            if (executeExport(clientInstance, updateByQueryRequest, tryNum)) {
+            ioException = executeExport(clientInstance, updateByQueryRequest, tryNum);
+            if (ioException == null) {
```
otchet-broad, line 50: Some observations around this block:
1) line 43 sets a MaxRetries to 5, but this loop seems to be hard coded to 3.  Is that intended?
2) Is it important to know which attempt succeeded?
3) Is the only type of exception thrown an ioException?  If not, the variable name seems misleading. (After scrolling a bit, I see it now throws Exception too.  I like that as well and recommend the variable name change to make it clear.  Maybe thrownException?)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -41,22 +43,33 @@ public void export() {
...
                 break;
             }
         }
+        if (ioException != null) {
+            throw new DsmInternalError("Unable to connect to ElasticSearch", ioException);
```
otchet-broad, line 56: "Problem updating the document database" might be a little better of an error message to throw here so we don't leak internal details of the stack we're using.  This comment applies to other cases where we refer to ES too.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/CodeStudyColumnsProvider.java
```diff
@@ -22,15 +22,15 @@
...
         } else if (realm.equals("cmi-lms")) {
             return getLmsColumns();
         }
 
-        throw new DSMBadRequestException("Invalid realm " + realm);
+        throw new DSMBadRequestException("Unsupported realm " + realm);
     }
 
-    private Map<String, OncHistoryUploadColumn> getOsteoColumns() {
+    private Map<String, OncHistoryUploadColumn> getOsteo2Columns() {
```
otchet-broad, line 33: Thanks for renaming.  I now know exactly what this means.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryElasticUpdater.java
```diff
@@ -0,0 +1,96 @@
...
+    private static class OncHistoryGenerator implements Generator {
+        private final Map<String, Object> source;
+
+        public OncHistoryGenerator(Map<String, Object> source) {
+            this.source = source;
+        }
+
+        @Override
+        public String getPropertyName() {
+            return "oncHistoryDetail";
```
otchet-broad: Should you use ONC_HISTORY_DETAIL from ESObjectConstants?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryElasticUpdater.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Append values to an object in the elastic document for participant
+     */
+    public void updateAppend(Map<String, Object> source, String participantShortId) {
+
+        OncHistoryGenerator generator = new OncHistoryGenerator(source);
+
+        String scriptText = "if (ctx._source.dsm.oncHistoryDetail == null) {"
```
otchet-broad: Is there a way to leverage the ONC_HISTORY_DETAIL value from ESObjectConstants here so that if the field name changes in that file, the change will work in this code too?

otchet-broad (approved):
Good changes, Dennis.  I left you some comments about the code, but the PR is approved.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
denniscunningham: I think we 'leak' information about ES all over the code, but I get your point. However, the key thing to record here is that we could not connect to ES. For now I will log an error.
I suspect the right way to stop leaking information about internal errors is to catch them at a high level (perhaps in the Spark exception handling), log and strip the content, then send a generic message back to the requester.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
denniscunningham: I had no reason to update this code per se, but did so since it was incorrect: if DSM could not connect to ES the code just behaved as if success. One of those hidden errors where ES gets out of sync with DSM DB and we don't know about it until it impacts a customer.
It *might* be important know if it took multiple attempts to connect, but I think that is best left to the engineer who analyzes and fixes our ES connection issues.
In the existing version of this code `executeExport` only catches IOExceptions for the sole purpose of retrying, which is split into two methods. All other exceptions throw out of this method. What I did was made that clearer, save the IOException message, since it likely has important information, and ensure we recored the error. Since I do not have a good way of creating an automated test for this code I wanted to fix the problem but limit the impact of the change.
The way I understood the retries, was 5 per attempt internally by the ES client, and 3 in the loop here, so 15 in total. I don't know if there was any science behind that so left it as is, again something an engineer exploring  ES timeouts can investigate.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryElasticUpdater.java
denniscunningham: Will do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryElasticUpdater.java
denniscunningham: Will do.

denniscunningham (commented):
Thanks for your review Elliot. Pls see my comments.

PR #2595, opened 2023-06-21
PEPPER-826 lms somatic-results email updates

ssettipalli:
minor updates to lms somatic-results emails & cleanup


Rinchan20 (approved):
seems to have fixed things in dev

PR #2598, opened 2023-06-21
PEPPER-535 removed backend code that sent GA metrics.

zyme:
## Context

PEPPER-535 calls for removing all uses of GA.  In this PR, much code has been preserved, but the actual calls to send metrics data from the backend have been removed.  There will be a separate PR to remove GA from DSM and patient-facing sites.

These changes should be low impact.  Happy side effect: they will likely improve response teams for endpoints like saving forms, as well as startup and shutdown time.

I'm thinking we should probably do a hotfix update for this PR and angular changes so we get things ship-shape before July 1.

## FUD Score

## Testing

To test these changes, just run the usual playwright tests and then pop over to various GA accounts and verify that no metrics are coming in.

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/FrontendAnalyticsRoute.java
```diff
@@ -1,37 +0,0 @@
...
-
-import org.broadinstitute.dsm.analytics.GoogleAnalyticsMetrics;
-import org.broadinstitute.dsm.analytics.GoogleAnalyticsMetricsTracker;
-import org.broadinstitute.dsm.security.RequestHandler;
-import org.broadinstitute.dsm.statics.RoutePath;
-import org.broadinstitute.lddp.handlers.util.Result;
-import spark.QueryParamsMap;
-import spark.Request;
-import spark.Response;
-
```
zyme, line 11: I couldn't find any use of this route in frontend code, and since the route just posts some timing information to GA, I decided to delete it entirely.  CC @pegahtah

ssettipalli (approved):
looks good to me

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -581,8 +581,8 @@ protected void setupCustomRouting(@NonNull Config cfg) {
         if (!cfg.getBoolean("ui.production")) {
             get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT, new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
                     new JsonTransformer());
-            get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT_WITH_PARTICIPANT, new CreateClinicalDummyKitRoute(new OncHistoryDetailDaoImpl()),
-                    new JsonTransformer());
+            get(apiRoot + RoutePath.CREATE_CLINICAL_KIT_ENDPOINT_WITH_PARTICIPANT, new CreateClinicalDummyKitRoute(
+                    new OncHistoryDetailDaoImpl()), new JsonTransformer());
```
pegahtah, line 585: checkstyle was throwing errors for this one, so I just fixed that

pegahtah (approved):

PR #2612, opened 2023-06-28
[DSM] PEPPER-934 CF logging fix

pegahtah:
Fixing the logging from warning to Severe and fixing how we throw the exceptions

### DIFF
--- a/cloud-functions/cf-dsm-file-scanner/src/main/java/org/broadinstitute/ddp/cf/DSMSomaticFileScanner.java
+++ b/cloud-functions/cf-dsm-file-scanner/src/main/java/org/broadinstitute/ddp/cf/DSMSomaticFileScanner.java
@@ -259,8 +259,8 @@ public class DSMSomaticFileScanner implements BackgroundFunction<DSMSomaticFileS
             String json = new Gson().toJson(message);
             publishAntivirusResultMessageToDSM(json);
         } catch (Exception e) {
-            log.atWarning().log(String.format("Unable to move the file %s from bucket %s to %s ", objectName, uploadBucketName,
-                    finalBucketName), e);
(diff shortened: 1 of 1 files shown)


cloud-functions/cf-dsm-file-scanner/src/main/java/org/broadinstitute/ddp/cf/DSMSomaticFileScanner.java
```diff
@@ -259,8 +259,8 @@ private void moveFileIntoFinalBucket(String uploadBucketName, String objectName,
             String json = new Gson().toJson(message);
             publishAntivirusResultMessageToDSM(json);
         } catch (Exception e) {
-            log.atWarning().log(String.format("Unable to move the file %s from bucket %s to %s ", objectName, uploadBucketName,
-                    finalBucketName), e);
+            log.atSevere().withCause(e).log(String.format("Unable to move the file %s from bucket %s to %s ", objectName, uploadBucketName,
```
denniscunningham, line 262: Erf. This is why I was hoping we would move away from the fluent logger.
What logging level does 'severe' map to? Is this 'error'?
I think this is more than semantics since I assume we want to get alerted?

pegahtah: I know and totally agree with moving on from debugger. I tried lombok's 2 or 3 other loggers before but  they didn't get to the cloud run logs in the right way. Maybe we can investigate that later? 
Or do you think this is something for us to look into today?

I followed the `Severe` pattern from the original code here, based on the limited documentations I found, Severe seems to play the role of error here ` severe, warning, config, info, fine, finer, finest`

denniscunningham (approved):

PR #2618, opened 2023-07-05
[DSM] PEPPER-947 changes from RC

pegahtah:
Fixing RGP's RNA kit label not getting saved


denniscunningham (commented):
My main concern is that we are now bringing processRGPFinalScan on-line, but may not have a good way of testing/validating its correctness?
Offhand I don't see any JUnit tests for this?

pegahtah:
> My main concern is that we are now bringing processRGPFinalScan on-line, but may not have a good way of testing/validating its correctness? Offhand I don't see any JUnit tests for this?

Hi Dennis no there are not, we only tested it manually
If you look at the git history this change got dropped around March because two branches that were changing the same part of the code kinda over rode each other.

denniscunningham (approved):

PR #2631, opened 2023-07-26
Edit user info admin service

denniscunningham:
PEPPER-973
PEPPER-974
PEPPER-975

This PR implements editing user information but also includes improvements to adding and removing users.
It also implements creating and deleting UserSettings as part of adding and removing users.
And it also implements consistent use of the access_user.is_active flag. More info [here](https://docs.google.com/document/d/1VMRdOXKXPCBlcoEQh44BaAK0VKxuppR4S50C97mDX9U/edit).

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
```diff
@@ -110,6 +112,26 @@ public int create(UserDto userDto) {
         });
     }
 
+    public static void update(int userId, UserDto userDto) {
+        String email = userDto.getEmailOrThrow();
+        String errorMsg = "Error updating user " + email;
+        int res = inTransaction(conn -> {
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_UPDATE_USER)) {
+                stmt.setString(1, userDto.getName().orElse(null));
```
ssettipalli, line 120: Since its updating existing user, aren't we nullifying the existing name if NOT passed ?
looks like in UserAdminService we are checking for name to be non null though

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
```diff
@@ -110,6 +112,26 @@ public int create(UserDto userDto) {
...
+                stmt.setString(2, userDto.getPhoneNumber().orElse(null));
```
ssettipalli, line 121: Since its updating existing user, aren't we nullifying the existing phone number if NOT passed ?
May be it will handle scenario where we want to remove existing phone number . 
Pointing out just in case

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
```diff
@@ -36,10 +37,20 @@ public Optional<String> getEmail() {
         return Optional.ofNullable(email);
     }
 
+    // TODO: getEmail should always throw on a missing email since it is not nullable
+    // but there are a lot of callers. Feel free to chase them all down and rename this method -DC
+    public String getEmailOrThrow() {
+        return getEmail().orElseThrow(() -> new DsmInternalError("User email cannot be null"));
```
ssettipalli, line 43: Its a DTO and we are not setting the email.
DTO should just return  whatever exists rather than exception and up to caller to decide what to do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserInfo.java
```diff
@@ -3,26 +3,29 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import lombok.AllArgsConstructor;
 import lombok.Data;
+import org.broadinstitute.dsm.db.dto.user.UserDto;
 
+@AllArgsConstructor
 @Data
 public class UserInfo {
```
ssettipalli, line 12: wondering if this can extend UserDto and add  List<UserRole>

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
denniscunningham: Yup, I chose to implement it this way (the caller determines what to update by looking at the existing record) rather than doing a selective update based on the DTO. But I can change that if there is already a convention in place. 

I think you are suggesting that the logic in the caller only include the fields to update, and the logic here would be to only update the non-null fields. One downside is that you cannot set a field to null with the update.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
denniscunningham: See my answer above. Open to whatever makes sense based on our conventions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
denniscunningham: Yup, perhaps that's the downside of using the same data structure for input and output, but in this case you basically expect a non-null email for both (not nullable in DB). Are you thinking that DTOs can be used more widely that just DB input and output?

I added this because we are repeating similar logic elsewhere in the code, and I was just adding a few more instances of the same code. So having a central place for the logic made sense. 

Are you concerned about having the call, or my comment that says we should not have the other call (that does not throw)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserInfo.java
denniscunningham: The UserDto includes an is_active flag, which is only used internally. IOW, we don't want users setting that flag directly.

denniscunningham (commented):
@ssettipalli Thx for your review.
Left responses to your comments. Pls let me know how you want to proceed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
ssettipalli: I was thinking more of like getEmail() and just return whatever was set into the Dto and let the callers handle nulls.
However since the method is getEmailOrThrow() , may be its ok to leave it this way.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
ssettipalli: In that case we can just dostmt.setString(1, userDto.getName());
If caller sets/passes null, update as null.  End result will be same as existing code I guess.

denniscunningham: I think we should decide on the behavior, then consider the implementation.
In this case it is either: assign whatever the DTO has, including NULLs, which is the implementation in this PR; or only assign those values that are not null. The latter seems problematic since it does not provide the caller a way to assign NULL to a field.

What do you think?

This is an unusual DTO since its get methods return optional, which is not our convention. If that seems confusing we can change that, but perhaps we can do that as a separate PR?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
denniscunningham: We already have a getEmail call, which returns an optional. That means all users are forced to process the return in some fashion. This is a convenience method that benefits the great majority of callers who are effectively asserting that the email value is not null.

I can remove this and go back to calling getEmail... it just ends up in a lot of redundant code in the callers.

denniscunningham (commented):
@ssettipalli Thx for your review.
Let me know how you want to proceed. Feel free to DM me if I am not understanding your concerns.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
ssettipalli: assign whatever the DTO has, including NULLs seems good to me.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
ssettipalli: If we already have the getEmail call, lets leave the new convenience method in there. sorry for the confusion.

ssettipalli (approved):

PR #2633, opened 2023-07-26
PEPPER-BACKEND Update function openJdbiWithAuthRetry

aweng98:
Observed a trend in GCP logs, too many `org.broadinstitute.ddp.exception.DDPException: Could not get a connection from the pool`. After some digging, it seems to me there is a bug in `openJdbiWithAuthRetry` function which retries never happen when open DB connection failed because new DDPException is thrown. Judging by the name of function, I believe connection should be retried.

Minor update to fix it:  Retry Jdbi connection to DB up to `PASSWORD_ROTATION_MAX_RETRIES` times.

GCP log screenshot:

Screenshot 2023-07-26 at 2 35 26 PM

GCP log [filter](https://console.cloud.google.com/logs/query;query=%2528resource.type%3D%22gae_app%22%20AND%20resource.labels.project_id%3D%22broad-ddp-dev%22%2529%0Aseverity%3E%3DERROR%20AND%0AtextPayload%3D~%22org.broadinstitute.ddp.db.TransactionWrapper.openJdbiWithAuthRetry%22;summaryFields=:false:32:beginning;cursorTimestamp=2023-07-26T16:48:53.770835Z;duration=P2D?_ga=2.19987832.1777900978.1690290588-1108477672.1690290588&project=broad-ddp-dev).

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
@@ -343,9 +343,8 @@ public class TransactionWrapper {
                         log.error("Sleep before dbpool credential reload has been interrupted.", interrupted);
                     }
                     reloadDbPoolConfiguration(false);
-                } else {
-                    log.info("Open jdbi connection retried {} time.", tryCount + 1);
                 }
+                log.info("Open jdbi connection retried {} time.", tryCount + 1);
             } catch (InvalidConfigurationException e) {
                 lastThrownException = e;
                 log.error("Database connection configuration is invalid. Proceeding with original configuration values.");
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
@@ -329,27 +329,30 @@ public class TransactionWrapper {
      * values so that we can rotate the password without rebooting the app.
      */
     private static Handle openJdbiWithAuthRetry(DB db) {
+        Exception lastThrownException = null;
         for (int tryCount = 0; tryCount < PASSWORD_ROTATION_MAX_RETRIES; tryCount++) {
             try {
                 return getInstance(db).jdbi.open();
             } catch (ConnectionException e) {
+                lastThrownException = e;
                 if (isAuthException(e)) {
-                    log.info("Database pool credentials have been rejected; pausing and reloading creds from config file.", e);
+                    log.info("Database pool credentials have been rejected; pausing and reloading credentials from config file.", e);
                     try {
                         TimeUnit.SECONDS.sleep(PASSWORD_ROTATION_SLEEP);
                     } catch (InterruptedException interrupted) {
-                        log.error("Sleep before dbpool credential reload has been interrupted", interrupted);
+                        log.error("Sleep before dbpool credential reload has been interrupted.", interrupted);
                     }
                     reloadDbPoolConfiguration(false);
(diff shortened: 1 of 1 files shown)


denniscunningham:
Hmmm... hold on. 
I may be misunderstanding, but I do not think this is correct.
First, is there a bug associated with this? I am working on https://broadworkbench.atlassian.net/browse/PEPPER-983 to address the connection exceptions we are seeing in production. Is that the problem you are addressing?

Second, I'm not sure what this PR accomplishes. The retries in the modified method were designed to handle a blip when passwords are rotated, which is rare.

Third, I do not understand this statement: "After some digging, it seems to me there is a bug in openJdbiWithAuthRetry function which retries never happen when open DB connection failed because new DDPException is thrown." Can you provide some details?

denniscunningham (changes requested):
I don't think this is correct. Pls see my comment.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -329,27 +329,30 @@ public static <R, X extends Exception> R withTxn(HandleCallback<R, X> callback)
...
                     try {
                         TimeUnit.SECONDS.sleep(PASSWORD_ROTATION_SLEEP);
                     } catch (InterruptedException interrupted) {
-                        log.error("Sleep before dbpool credential reload has been interrupted", interrupted);
+                        log.error("Sleep before dbpool credential reload has been interrupted.", interrupted);
                     }
                     reloadDbPoolConfiguration(false);
                 } else {
-                    throw new DDPException(COULD_NOT_GET_CONNECTION, e);
+                    log.info("Open jdbi connection retried {} time.", tryCount + 1);
```
aweng98: Should I add `TimeUnit.SECONDS.sleep(PASSWORD_ROTATION_SLEEP)` here before next retry?

aweng98:
> Hmmm... hold on. I may be misunderstanding, but I do not think this is correct. First, is there a bug associated with this? I am working on https://broadworkbench.atlassian.net/browse/PEPPER-983 to address the connection exceptions we are seeing in production. Is that the problem you are addressing?
> 
> Second, I'm not sure what this PR accomplishes. The retries in the modified method were designed to handle a blip when passwords are rotated, which is rare.
> 
> Third, I do not understand this statement: "After some digging, it seems to me there is a bug in openJdbiWithAuthRetry function which retries never happen when open DB connection failed because new DDPException is thrown." Can you provide some details?

I'm not certain, but for me, the code smells.
I don't know about the connection exceptions in Prod as I don't have access to that project.
If retries is designed to only handle a passwords issue, then I should close this PR. Because we don't want to retry other types of connection failures, right?

denniscunningham:
@aweng98 What part of the code smells? I have a branch where I've made some changes to TransactionWrapper, but not to that method. Although I may not have implemented that code the way the author did, I don't see anything wrong with it offhand. What concerns you about that code?

More broadly, while one solution to the connection exceptions may be to retry, doing may mask an underlying issues. If  you look at the stack trace for (most of) the connection exceptions you will see that the failure is due to `SQLNonTransientConnectionException: No operations allowed after connection closed.`, which should not happen based on the configuration for DSM. My objective for PEPPER-983 is to find out why we see this error before deciding on a solution.

Does that make sense?

aweng98:
I'll close PR. It's my misunderstand this func is designed not to retry all connection failures. I assumed retries should have happened in a consistent mannor. Why is `InvalidConfigurationException` not thrown and included in retry without sleep; `AuthException` is retried after sleep; and other exception are simply thrown. Function javadoc could be more detailed as to what this function does exactly.

PR #2635, opened 2023-07-28
PEPPER-130 elastic upgrade

ssettipalli:
Changes to upgrade elasticsearch to 7.17


pepper-apis/parent-pom.xml
```diff
@@ -18,7 +18,7 @@
         <itext.version>7.1.4</itext.version>
         <itext.licensekey.version>3.0.3</itext.licensekey.version>
         <liquibase.version>4.8.0</liquibase.version>
-        <elasticsearch.version>6.8.15</elasticsearch.version>
+        <elasticsearch.version>7.17.10</elasticsearch.version>
```
denniscunningham, line 21: Are we upgrading in steps (i.e. to 7.x then 8.x)?

denniscunningham (approved):

pepper-apis/parent-pom.xml
ssettipalli: yes Dennis. upgrading to 7.17 will bring it into supported version.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -196,7 +196,7 @@ public static RestHighLevelClient getClientForElasticsearchCloud(@NonNull String
                         httpClientBuilder.setDefaultIOReactorConfig(IOReactorConfig.custom().setSoKeepAlive(true).build());
                     }
                     return httpClientBuilder;
-                }).setMaxRetryTimeoutMillis(100000);
```
aweng98, line 199: question: this method is deprecated. for my undertsanding, do we set socket and connect timeouts elsewhere or just relying on default timeout?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -373,6 +371,7 @@ public void exportToElasticsearch(Handle handle, StudyDto studyDto,
                             exportsSoFar, participants.size(), studyDto.getGuid());
                 }
             } catch (Exception e) {
+                e.printStackTrace();
```
aweng98: why add this? is it for debugging on localhost?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/ElasticsearchServiceUtil.java
```diff
@@ -84,18 +84,18 @@ public static synchronized RestHighLevelClient getElasticsearchClient(Config cfg
...
                     if (proxyUrl != null) {
                         httpClientBuilder.setProxy(new HttpHost(proxyUrl.getHost(), proxyUrl.getPort(), proxyUrl.getProtocol()));
                         httpClientBuilder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
                     }
                     return httpClientBuilder;
-                })
-                .setMaxRetryTimeoutMillis(100000);
-
+                });
+        //.setMaxRetryTimeoutMillis(100000);
```
aweng98: can we just remove it because no value to keep it around.

aweng98 (commented):
I think changes are good. 
Last question, do we know if reindex is needed or not? I read [migration doc](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/reindex-upgrade.html), it says upgrade from 6.8.x to 7.x doesn't need to do reindexing. but just want to ask to make it aware.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
ssettipalli: we can move to new Client when we eventually move to 8.* and this RestHighLevelClient seem to work fine with 7.17 even though deprecated in 7.10 
I don't think we are setting else where. Updated new version of this RestClientBuilder doesnt support it and hence removed.

aweng98 (approved):

PR #2640, opened 2023-08-01
DB connection and transaction processing improvements

denniscunningham:
Towards PEPPER-983

This PR is to support the investigation into the DB connection issues we are seeing across Pepper services. It includes the following changes:

- Cleanup logging in TransactionWrapper to not include DB URLs since they may contain secrets.
- Use DDPInternalError to distinguish the errors reported in TransactionWrapper. (Only one place in the code catches DDPException and it is not relevant to TransactionWrapper.)
- Simplify the connection timeout detection code I added to TransactionWrapper a few months ago, and apply it to TransactionWrapper::inTransaction.
- Remove unneeded InvalidConfigurationException catch from TransactionWrapper::openJdbiWithAuthRetry (not thrown in that context).
- Separate the closeable handle and connection blocks in TransactionWrapper::inTransaction. Lint was complaining about it and having them both in the same statement makes the close ambiguous.
- Added a try-catch to DSMServer::main so there is explicit handling and reporting of startup errors.
- Updated Hikari and JDBI libraries. JDBI was required to support my changes, but having up to date libraries presumably eliminates old library bugs as a factor.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -446,6 +431,9 @@ private HikariDataSource createDataSource(int maxConnections, String dbUrl, DB d
         config.setMaxLifetime(TimeUnit.SECONDS.toMillis(14400)); // 4 hours, which is half the default wait_timeout of mysql
         config.setPoolName(db.name());
 
+        log.info("Created data source for {} with maxConnections={}, connectionTimeout={}s, maxLifetime={}s",
```
aweng98, line 434: Out of scope question. Can we have a string identifier in DB logs to make logs searching easier in GCP? For example, most logs for exporting starts with `[export]` like [so](https://github.com/broadinstitute/ddp-study-server/blob/8338187bc8169382039e35bb7382c19668a410ef/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/DataExporterCli.java#L200).

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -366,22 +351,13 @@ public static <X extends Exception> void useTxn(DB db, HandleConsumer<X> callbac
         long startTime = System.currentTimeMillis();
         try (Handle h = openJdbiWithAuthRetry(db)) {
             h.useTransaction(callback);
+        } finally {
             long endTime = System.currentTimeMillis();
             // 30s threshold
             if (endTime - startTime > 30000) {
                 logger.warn("DB transaction open for > 30s\n {}",
```
aweng98, line 358: I have seen this log in GCP before. Why log this at all because there isn't anything we do about it?

aweng98 (approved):
Big improvement! Much easier to read and follow the code now. Thank you.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
denniscunningham: Ah, I've never seen the message, and I don't see it in the logs now, but I am very interested if you can point me to it in the logs. If we find a case we definitely *can* do something about it.
I added this code when we were concerned about abandoned connections, and there's no harm in having it around. This PR extends it to cover more DSM connections.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
denniscunningham: When you create a log message you are free to add any (reasonable) text you'd like. Since log messages include the name of the class that produced them, that combined with the text of a given message is typically enough to find information in the logs.
What problem are you trying to solve?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
aweng98: GCP log search `"DB transaction open for > 30s"` should do it. 
Screenshot 2023-08-02 at 1 50 17 PM

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
aweng98: Just an idea. Reason is that I think a prefix makes it easier for searches. For example, if I want to search for all DB action logs (info, error), I'd type in `[DB]` and get just DB logs. Faster and precise.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
denniscunningham: Thx Alex! Very helpful. My search filter was too narrow and I was missing those entries.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -168,21 +169,19 @@ public static synchronized void init(DbConfiguration... dbConfigs) {
...
-                            + "it with different params " + maxConnections
-                            + " and " + dbUrl);
+                    throw new DDPInternalError("init() has already been called with different parameters: DB=" + db.name());
                 } else {
-                    log.warn("TransactionWrapper has already been initialized.");
+                    log.warn("TransactionWrapper has already been initialized for {}", db.name());
                 }
             }
-            gTxnWrapper.put(dbConfig.getDb(), new TransactionWrapper(dbConfig.getMaxConnections(), dbConfig.getDbUrl(), dbConfig.getDb()));
+            gTxnWrapper.put(db, new TransactionWrapper(maxConnections, dbUrl, db));
```
ssettipalli, line 184: @denniscunningham 
Taking another look at this code (actually existing code though).
We are calling this new TransactionWrapper(maxConnections, dbUrl, db) even if the TransactionWrapper is initialized (we are logging a warning) but still creating DataSources. 
Wondering if the code is called again during the init and are we creating DataSources (connections) again.

ssettipalli: Whenever we restart Housekeeping I see connections created twice . second set within 10seconds of first connections creation. 
Screenshot 2023-08-17 at 1 41 34 PM

denniscunningham: I think I see what you are seeing: there is an initialized flag that is not handled very well. IOW, the flag is set, and there is a way for callers to check it, but nothing that prevents re-entry to the init() method. Is that what you see?

Yup, we can refactor a bit to clean that up. I'd be surprised that it is actually causing an issue, but it should be cleaner.

denniscunningham: Missed your second comment... Well so much for it not happening in practice. On the surface that looks like it called init() twice. (I'd be surprised if it called reset() in between). 
Do you want to fix this code or should I?

ssettipalli: Looks like we are trying to close existing conns and create new set though
if (dataSource != null) {
            dataSource.close();
}

ssettipalli: Screenshot 2023-08-17 at 2 21 00 PM

ssettipalli: Dennis: Please go ahead and do cleanup as part of PEPPER-983

PR #2654, opened 2023-08-14
[DSM] PEPPER-1023 Juniper authentication

pegahtah:
## Context

Moving Juniper endpoints to be separate from GP's and to have a separate authentication, with it's own secret and issuer

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/JWTRouteFilterTest.java
```diff
@@ -101,14 +101,21 @@ public void printTestToken() throws Exception {
...
+        String juniperSigner = cfg.getString("juniper.signer");
         String ddpSecret = cfg.getString("portal.jwtDdpSecret");
         String monitoringSecret = cfg.getString("portal.jwtMonitoringSecret");
 
         System.out.println("monitoring token: " + SecurityHelper.createMonitoringToken(monitoringSecret, SecurityHelper.MONITORING_SYSTEM));
 
-        String bspToken = SecurityHelper.createGpToken(bspSecret, getCurrentUnixUTCTime() + THIRTY_MIN_IN_SECONDS, new HashMap<>());
+        String bspToken = SecurityHelper.createTokenWithSigner(bspSecret, getCurrentUnixUTCTime() + THIRTY_MIN_IN_SECONDS,
+                new HashMap<>(), SecurityHelper.BSP_SIGNER);
         System.out.println("Token for bsp: " + bspToken);
```
ssettipalli, line 113: We need to avoid logging the tokens unless we really need something out of it.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/JWTRouteFilterTest.java
```diff
@@ -101,14 +101,21 @@ public void printTestToken() throws Exception {
...
 
+        String juniperToken = SecurityHelper.createTokenWithSigner(juniperSecret, getCurrentUnixUTCTime() + THIRTY_MIN_IN_SECONDS,
+                new HashMap<>(), juniperSigner);
+        System.out.println("Token for juniper: " + juniperToken);
```
ssettipalli, line 117: Lets try to avoid logging the tokens

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/JWTRouteFilterTest.java
```diff
@@ -101,14 +101,21 @@ public void printTestToken() throws Exception {
...
+
         String ddpToken = SecurityHelper.createToken(ddpSecret, getCurrentUnixUTCTime() + THIRTY_MIN_IN_SECONDS, new HashMap<>());
         System.out.println("Token for appRoute: " + ddpToken);
```
ssettipalli, line 120: Lets try to avoid logging the tokens

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/JWTRouteFilterTest.java
pegahtah: I agree! This is a unit test that generates the token for us (developers) so we can use in our tests

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/SecurityHelper.java
```diff
@@ -69,12 +69,12 @@ public static String createToken(@NonNull String secret, long invalidAfter, @Non
         }
     }
 
-    public static String createGpToken(@NonNull String secret, long invalidAfter, @NonNull Map<String, String> claims) {
+    public static String createTokenWithSigner(@NonNull String secret, long invalidAfter, @NonNull Map<String, String> claims, String signer) {
         try {
```
denniscunningham, line 68: Should the signer also be NonNull?

pegahtah: No unfortunately, GP does not have one set string that they shared with us to check

denniscunningham: Not sure I understand. Do we need a check for a null signer in the code?

pegahtah: What I meant is, GP tokens come in all shapes and forms, we had to drop some conditions and checks between Mercury and BSP so their tokens get validated

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -543,70 +540,83 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
-        before(apiRoot + "*", (req, res) -> {
+        before(API_ROOT + "*", (req, res) -> {
             if (!new JWTRouteFilter(auth0Domain).isAccessAllowed(req, false, bspSecret)) {
                 logger.info("Returning 401 because token was not verified");
                 halt(401);
             }
             res.header(HttpHeaders.CONTENT_TYPE, MediaType.JSON_UTF_8.toString());
         });
 
+        before(DSM_ROOT + "*", (req, res) -> {
```
denniscunningham, line 587: It looks like here and above that the generic DSM_ROOT is dedicated to Juniper.
IOW, will all users of this path be Juniper?
If so, does it make sense to make this more generic now?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -1038,32 +1040,34 @@ private void setupJobs(@NonNull Config cfg, @NonNull KitUtil kitUtil, @NonNull N
...
         setupErrorNotifications(cfg, schedulerName);
     }
 
+    private void startScheduler(Scheduler scheduler) throws DsmInternalError {
+        logger.info("Setup Job Scheduler...");
+        try {
+            scheduler.start();
+            logger.info("Job Scheduler setup complete.");
+        } catch (Exception ex) {
+            throw new DsmInternalError("Unable to setup Job Scheduler.", ex);
```
denniscunningham: It looks like in all cases this throw will be caught (above) and converted to a slightly different exception. 
Does it make sense to just eliminate this catch? It's it unclear what value it provides.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/security/SecurityHelper.java
denniscunningham: It seems like we need to check for an empty/null signer since we are assigning it below.
If @NonNull is not a good choice then we should check for null/empty before assigning the signer.
If a empty/null signer is valid, then all we need is a comment saying so, since that is unusual.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: I _think_ given the current predicted lifetime for DSM (that I know of) this route will be only called by Juniper to create kits in Pepper DSM.
However, the api is designed to work for any non-DSS kit creation. Given that, I'd say if time comes for other users to use this api to trigger a kit creation, we would need to make changes in our authentication methods to make sure an endpoint can work with more than one possible secret (to avoid sharing Juniper's secret with other users)

Since Juniper's secret should definitely be different than GP's and the endpoint should go into a different category than the one GP uses, I thought this much change would be enough for the scope of this work given we don't know how long DSM will live, or even if other future Juniper studies will use Pepper DSM or not. 

What do you think?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: Oh yes, great catch!! Will make the changes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
denniscunningham: Yup, that makes sense.
I think a comment on why we are handling a generic path (dsm/) with a Juniper specific handler would be helpful. It could be as simple as "This path only handles Juniper for now."

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -543,70 +540,83 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
                 logger.info("Returning 401 because token was not verified");
                 halt(401);
             }
             res.header(HttpHeaders.CONTENT_TYPE, MediaType.JSON_UTF_8.toString());
         });
 
+        before(DSM_ROOT + "*", (req, res) -> {
+            if (!new JWTRouteFilter(auth0Domain).isAccessAllowed(req, false, dsmSecretForJuniper)) {
+                logger.info("Returning 401 because token was not verified");
+                halt(401);
```
denniscunningham: It would be helpful to include a message with the 401 (e.g. "Token could not be verified").

denniscunningham (commented):
Looks good. Just a couple comments.

denniscunningham (approved):

PR #2533, opened 2023-04-28
Pepper 575

denniscunningham:
## Context
PEPPER-575
Also see the discussion of the underlying issue possible approaches: [link](https://docs.google.com/document/d/1jyj-tQkryZsE9AMWtkVR4zL6-rJXFAcC35h86l7MyR8/edit?usp=sharing).

At a high level this was straightforward: map a DSS FIND_OUT answer into a DSM REF_SOURCE ID. However, the challenge was finding a way that was not too brittle: likely to break with a future change and be hard to detect. The changes here focus on detecting when things get out of sync.

Note: I really wanted to change the name RgpAutomaticProbandDataCreator to the simpler RgpProbandDataCreator, since the term 'Automatic' adds no value and is likely misleading. Since git history is lost with this I kept the name as is.

Also included in the PR are changes to get the Liquibase Maven plug-in configured correctly.

## Checklist

- [x ] I have labeled the type of changes involved using the `C-*` labels.
- [x ] I have assessed potential risks and labeled using the `R-*` labels.
- [x ] I have considered error handling and alerts, and added `L-*` labels as needed.
- [ x] I have considered security and privacy, and added `I-*` labels as needed
- [x ] I have analyzed my changes for stability, fault tolerance, graceful degradation, performance bottlenecks and written a brief summary in this PR.
- [ ] If applicable, I have discussed the analytics needs at both a platform and study level with Product and instrumented code accordingly.
- [ ] If applicable, my UI/UX changes have passed muster with Product/Design via an over-the-shoulder review, screenshots, etc.

_If unsure or need help with any of the above items, add the `help wanted` label. For items that starts with `If applicable`, if it is not applicable, check it off and add `n/a` in front._

## FUD Score

_Overall, how are you feeling about these changes?_

- [x] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [x ] They are user-visible in dev as a regular user journey and require no additional instructions.
- [ ] Getting dev into a state where this is user-visible requires some tech fiddling. I have documented these steps in the related ticket.
- [ ] Requires other features before it's human visible. I have documented the blocking issues in jira.
- [ ] I have no idea how to demo this. Please help me!

## Testing

- [ x] I have written automated positive tests
[...]

PR #2657, opened 2023-08-16
Update participant data via UPDATE_CUSTOM_WORFLOW

denniscunningham:
PEPPER-575
PEPPER-360

This PR provides a reliable way to get data about newly enrolled RGP participants from DSS and DSM. It also implements a handful of changes that provide better error reporting in that process, across all studies.

Specifically, the existing DSM code handled the PARTICIPANT_REGISTERED message for RGP users, but since that message is sent when a participant first logged in (after getting an acceptance email) any delay in filling out the participant data activity meant DSM would likely never get the participant data. This PR instead collects participant data from ES after the participant completes enrollment, via the UPDATE_CUSTOM_WORKFLOW message. See [this](https://docs.google.com/document/d/1etwifxKS4JoTdUoFRdFagdc8XuvFI4rZGWaNt_xWtyg/edit?usp=sharing) doc for more info.

(This change completes the original work for PEPPER-575, which was to populate the DSM referral source from the participant survey data in DSS. Since that data is not available until enrollment is complete, this change enables the code added in https://github.com/broadinstitute/ddp-study-server/pull/2533.)

Additional details:

- Made the error handling in DSMtasksSubscription more consistent. DSM needs to log at error level if it is unable to process the message, and the existing code was scattered.
- Replaced RuntimeException with DsmInternalErrror, where applicable, to allow better handling of internal errors.
- Updated a couple method names since they no longer accurately described the function performed by the method.
- Improved error handling in StudyBuilderCLI

**Testing note:** 
The tests for PEPPER-575 were part of https://github.com/broadinstitute/ddp-study-server/pull/2533. For PEPPER-360, the key tests are in Playwright (authored by Rinchan20). Integration tests for these changes are unfortunately difficult to craft without something to take on the ES role. We plan to add an ES test container, and at that point this feature would be a good candidate for an integration test to troubleshoot the ES test container. For now, the plan is to push these changes to the Dev environment to run it against the Playwright test that specifically covers the PEPPER-360 workflow, and run it against Playwright tests that cover other workflows that may be impacted by these changes. If the testing plan for the PR seems insufficient, we can probably attempt to mock ES, but that is not straightforward so will take a few days to complete.

**Deployment note:**
The following Study Builder command should be run to deploy this change in each environment:
```java -Dconfig.file=./output-config/application.conf -jar ./target/StudyBuilder.jar --vars ./output-config/vars.conf  --substitutions ./studies/rgp/substitutions.conf --i18n-path ./studies/rgp/i18n/en.conf ./studies/rgp/study.conf --only-labeled-events enrollment_update```

- ## Checklist

## FUD Score

## Testing

## Release

### CODE REVIEW

**Comment by ssettipalli** [comment]
Please add any needed non RC deploy instructions (like StudyBuilder patch) to the ticket so that they can be added to planned release document.

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Study.java [hunk 1]
```diff
@@ -1,48 +1,26 @@
...
-import java.util.NoSuchElementException;
-
-import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.exception.DsmInternalError;
 
 public enum Study {
     ATCP("ATCP"),
     RGP("RGP"),
     SINGULAR("SINGULAR"),
     LMS("LMS"),
```
ssettipalli: If the value is expected to be study guid, it should be "cmi-lms"

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/ESMissingParticipantData.java
```diff
@@ -0,0 +1,15 @@
+package org.broadinstitute.dsm.exception;
+
+public class ESMissingParticipantData extends RuntimeException {
```
pegahtah: maybe we should add `Exception` to the end of this class name (to become `ESMissingParticipantDataException`) ?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Study.java
pegahtah: I understand the concern, but dsm task runs by instance_name  in dsm db

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/SingularDefaultValues.java
```diff
@@ -26,14 +27,14 @@ public class SingularDefaultValues extends BasicDefaultDataMaker {
...
         if (isParticipantDataInES()) {
-            return false;
+            throw new ESMissingParticipantData("Participant does not yet have profile and activities in ES");
         }
         return insertDefaultEnrollmentStatusForParticipant();
     }
 
     private boolean isParticipantDataInES() {
-        logger.info("Participant does not have profile and activities in ES yet...");
         return elasticSearchParticipantDto.getProfile().isEmpty() && elasticSearchParticipantDto.getActivities().isEmpty();
```
pegahtah: Singular default values never made it into production, maybe we can get rid of these classes in general?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/ESMissingParticipantData.java
denniscunningham: Will do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/SingularDefaultValues.java
denniscunningham: If we are sure this is obsolete I can definitely remove the Singular handling for this message.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Study.java
denniscunningham: Actually, I believe Sampath is correct, since we are using Study (in this case) to validate study GUIDs provided in messages from DSS. In that case, I believe cmi-lms is correct (or CMI-LMS in the case of this class). 
The reason I touched this class at all was because a couple things were wrong in the code, and then noticed there was some incorrect data, and I removed that. The change Sampath suggests fits with that theme.

We have a couple options: fix or remove "LMS", or more drastically remove the class and use something else in its place. The downside of the class is that it is not tied to a DB table, so if/when changes are made to ddp_instance they are not reflected in Study unless someone remembers to do that (and it appears that's what led to this problem).
The class is lightly used (also in AddFamilyMember and AddFamilyMemberRoute), but perhaps the best thing to do at this point in Pepper's lifecycle is to just fix Study for now. 
What do you think?

denniscunningham (commented):
@ssettipalli @pegahtah Thank you for your review!
I left a couple comments.

ssettipalli (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Study.java
pegahtah: I agree! I'm sorry if I confused you before, the thing is, when we publish to dsm-tasks to run the export to ES, we call by instance name, maybe can you put a comment in the code so we can know which calls are by guid and which calls are by name? It will make future handling of the code or cases of manual publishing to the topic much easier

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/SingularDefaultValues.java
pegahtah: Yeah, Singular is dead, and if you look at `makeDefaultable` it does not have a case for singular. We added the code and then they changed their mind

pegahtah:
Thank you! 
I am also happy to approve it again if you want to remove Singular unused code.

pegahtah (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/SingularDefaultValues.java
denniscunningham: I removed SingularDefaultValues in a commit yesterday. As you said, it was unused and just missed that in the first commit.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/Study.java
denniscunningham: Let me see if I understand, since I *am* concerned about this code and approach, and appreciate your insights. The use of Study in DSMtasksSubscription is to correctly identify the study referred to in message attributes map. In the particular case addressed in this PR, that message typically comes from DSS, and should match ddp_instance.study_guid (of course, that is just by convention, since we do not share a common source of truth for that, as far as I know).
Are you saying there are other instances where we send a message processed by DSMtasksSubscription that may use a different convention for identifying studies (that is, by instance name)?

denniscunningham: The more I look at this, the messier it is...

After a bit more investigation, how we identify a ddp_instance based on message contents varies by message task type (in DSMtasksSubscription), which is unfortunate. Ideally we would have one convention. Even better, we fix having two different non-key identifiers for a ddp_instance (that is, either instance name or study guid, not both). 

Getting that straightened out is a fairly large task that includes DSM, DSS, and the conventions used when sending messages by hand. We also need to see where we depend on a particular value of instance name or study guid (the Study class is just one example). And we also have to consider where we display instance names to users (rather than the display_name).

For the purposes of this PR, I feel like my instincts were correct in trying to delete the unused portions of Study. If this class has a future where it is used more broadly, it should perhaps be moved to ddp_common, so at least DSM and DSS are using the same source. I will add a comment to Study saying that we should not expand its use. I will also attempt to comment any of the message handlers that seem unclear on whether they are handling a study GUID or instance name, but it seems those are pretty clear -- just painfully inconsistent.

However, there is an inconsistency in BasicDefaultDataMaker that I will fix for this PR.

pegahtah (approved):

PR #2662, opened 2023-08-23
PEPPER-1071 re-enabled `AuthenticationTest`

zyme:
## PEPPER-1071

In order to re-enable `AuthenticationTest`, I did the following:
* made the test extend from `DbTxnBaseTest` so it could run in circleCI
* removed the dependency on pre-existing data and replaced it with dynamically generated groups, permissions, users, and ddp_instances

Along the way, I removed some redundant code and fixed up some assertions that were never asserted as a result of expected exceptions.

Search for `AuthenticationTest` [in this CircleCI log](https://circleci.com/api/v1.1/project/github/broadinstitute/ddp-study-server/38689/output/116/0?file=true&allocation-id=64e6739b8965f36d2bd700ad-0-build%2F4D7F7263) to confirm it passes.

## Checklist

## FUD Score

## Testing

- [ x] I have written automated negative tests

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
```diff
@@ -29,6 +29,10 @@ public int getId() {
         return this.id;
     }
 
+    public String getIdAsString() {
```
zyme: There was a bunch of code doing `Integer.toString(getId())` so I moved that logic here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -637,14 +630,14 @@ protected static int verifyStudyGroup(String studyGroup) {
...
-            throw new DSMBadRequestException("Operator does not have user administrator privileges for study group "
+            throw new DSMBadRequestException("Operator " + operatorId + " does not have user administrator privileges for study group "
                     + studyGroup);
         }
         return adminRoles;
     }
 
     protected static Map<String, RoleInfo> getOperatorAdminRoles(int operatorId, int groupId) {
-        Map<String, RoleInfo> adminRoles = null;
+        Map<String, RoleInfo> adminRoles = Collections.emptyMap();
```
zyme: Not sure why this was set to null--when I was building out the test, I hit an NPE here so I changed it to an empty map by default.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/TestHelper.java
```diff
@@ -112,7 +112,7 @@ public static void setupDB(boolean setupDDPConfigLookup) {
         cfg = cfg.withValue("quartz.enableJobs", ConfigValueFactory.fromAnyRef("false"));
         cfg = cfg.withValue("portal.port", ConfigValueFactory.fromAnyRef("9999"));
         cfg = cfg.withValue("errorAlert.recipientAddress", ConfigValueFactory.fromAnyRef(""));
-        if (!cfg.getString("portal.environment").startsWith("Local")) {
+        if (!cfg.getString("portal.environment").toLowerCase().startsWith("local")) {
```
zyme, line 115: Just trying to leave the campsite a little nicer than how I found it, since I flubbed "Local" more than once in my config.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,182 @@
...
         try {
-            Assert.assertFalse(UserUtil.checkUserAccessForPatch(realm, gpUserId, DBConstants.MR_VIEW, null, patch1));
+            UserUtil.checkUserAccessForPatch(realm, cmiKitShippingOnlyUser.getIdAsString(), DBConstants.MR_VIEW, null, patch1);
+            Assert.fail("Did not throw expected exception");
         } catch (Exception e) {
             Assert.assertTrue(e.getMessage().contains("User id in patch did not match the one in token"));
         }
         try {
-            Assert.assertFalse(UserUtil.checkUserAccessForPatch(realm, gpUserId, DBConstants.MR_ABSTRACTER, null, patch1));
+            UserUtil.checkUserAccessForPatch(realm, cmiKitShippingOnlyUser.getIdAsString(), DBConstants.MR_ABSTRACTER, null, patch1);
```
zyme: Since `checkUserAccessForPatch` is expected to throw an exception here, I remove the asserts--which are never run--and added explicit `fail()` in case the exceptions aren't thrown.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -7,14 +7,7 @@
...
 import java.sql.Types;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
denniscunningham: Not sure this is in our checkstyle but most places I've worked prefer to not use '*'. There may be a Java best practices somewhere, but this is frowned on I believe. (And I don't think we use that pattern in DSM.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -741,7 +738,7 @@ protected static Map<Integer, Set<String>> getRolesForStudyUsers(int groupId) {
         });
     }
 
-    protected static int getRoleId(String role) {
```
denniscunningham: I'm not comfortable with this approach (so this applies to many of the changes in this file). IMHO the right way to do this is to not expose the inner workings of this service -- and thus greatly expand its contract -- but to create a test helper class to provide what tests need (since this is not what the service needs to provide per se).
Does that make sense? Happy to discuss.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -995,8 +992,19 @@ protected static int deleteStudyGroup(int groupId) {
...
+        Map<String, RoleInfo> adminRoles = getOperatorAdminRoles(userId, groupId);
+        if (adminRoles.isEmpty()) {
+            addUserRole(userId, UserAdminService.getRoleId(USER_ADMIN_ROLE), groupId);
+        }
+    }
+
 
-    @Data
+        @Data
     protected static class RoleInfo {
```
denniscunningham: Typo. Are you getting lint warnings for this?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,182 @@
...
-        patch.setTableAlias("kit");
+        String nameAppend = "." + System.currentTimeMillis();
+        String pecgsStudyGroup = "pecgs" + nameAppend;
+        String studyInstanceName = "instance." + nameAppend;
+        pecgsStudyGroupId = UserAdminService.addStudyGroup(pecgsStudyGroup);
+
+        String cmiStudyGroup = "cmi" + nameAppend;
+        cmiStudyGroupId = UserAdminService.addStudyGroup(cmiStudyGroup);
+
+        int ddpInstanceId = UserAdminServiceTest.createTestInstance(studyInstanceName, cmiStudyGroupId);
```
denniscunningham: I think with the right test helper class this will be a lot cleaner.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,182 @@
...
+import static org.broadinstitute.dsm.service.admin.UserAdminService.USER_ADMIN_ROLE;
+import static org.broadinstitute.dsm.statics.DBConstants.KIT_SHIPPING;
+import static org.broadinstitute.dsm.statics.DBConstants.PT_LIST_VIEW;
 
-    @Before
-    public void first() {
-        setupDB();
+public class AuthenticationTest extends DbTxnBaseTest {
 
+    static UserDto dsmAdminUser;
```
denniscunningham: These should be private. Are you seeing lint warnings for this in intelliJ?

zyme: I'm not seeing any warnings.  Will change to private.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
zyme: Agreed.  Do you have a suggestion for what existing helper class these should be added to, or what a new helper class should be and where it should be?  Some of the methods I've made public in this PR will be needed in non-test code, so I'm thinking about a utility method that is available outside of tests.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
zyme: Nope, no warnings.  Begone, errant whitespace!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
denniscunningham: I built one yesterday. Should be available this morning.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
denniscunningham: I'm not sure how I setup code stye in Intellij, but my Settings | Editor | Code Style | Java is pretty complete. (Did I import a Pepper checkstyle file? I can't find any instructions.) 
I also recommend installing https://plugins.jetbrains.com/plugin/7973-sonarlint.

cloud-functions/cf-dsm-file-scanner/checkstyle.xml
```diff
@@ -24,12 +24,6 @@
         <property name="lineSeparator" value="lf"/>
     </module>
 
-    <module name="LineLength">
-        <property name="fileExtensions" value="java"/>
-        <property name="max" value="140"/>
-        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
-    </module>
```
denniscunningham: FWIW, every place I've worked had a line length restriction, so not having one is non-standard, and will probably decrease readability for those accustomed to restrictions. 
Is your concern that having this as part of the check style leads to errors, or do you prefer the readability of long lines over multi-line continuations? Or something else?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
denniscunningham: Did this change or did you decide to keep it as-is?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -35,8 +28,8 @@ public class UserAdminService {
 
     private final String operatorId;
     private final String studyGroup;
-    protected static final String USER_ADMIN_ROLE = "study_user_admin";
-    protected static final String PEPPER_ADMIN_ROLE = "pepper_admin";
+    public static final String USER_ADMIN_ROLE = "study_user_admin";
+    public static final String PEPPER_ADMIN_ROLE = "pepper_admin";
```
denniscunningham: You may need to merge with Develop, since I believe these are now public.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
denniscunningham: It feels like we did not connect well on this? I was assuming you would switch to using src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java which take care of managing test resources for test classes. It has the benefit of not requiring changes to the UserAdminService service contract. (All those UserAdminService methods that were not intended to be public.)

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/user/UserDto.java
```diff
@@ -46,6 +46,7 @@ public String getEmailOrThrow() {
     public Optional<String> getPhoneNumber() {
         return Optional.ofNullable(phoneNumber);
     }
+
```
zyme, line 49: Cleaning up as per checkstyle suggestion

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,129 @@
...
+    }
+
+    /**
+     * Create new groups, a new ddp_instance, associated roles,
+     * a few new users and set their permissions.
+     */
+    @BeforeClass
+    public static void setup() {
+        String nameAppend = "." + System.currentTimeMillis();
+        String studyInstanceName = "instance." + nameAppend;
```
denniscunningham: Do you want two dots here?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,129 @@
...
+        Patch patch = new Patch("0",  "dsmKitRequestId", "0", cmiKitShippingOnlyUserId , new NameValue("kit.collectionDate",  "2023-04-24"), null, "XSZSRS1MS3D4OAEK2DPM") ;
+        patch.setTableAlias("kit");
 
-        String gpUserId = Integer.toString(new UserDao().getUserByEmail(cfg.getString(GP_UNIT_TESTER_EMAIL)).orElse(new UserDto()).getId());
+        Assert.assertTrue(UserUtil.checkKitShippingAccessForPatch(studyInstanceName, cmiKitShippingOnlyUserId, null, patch));
+    }
 
-        Patch patch = new Patch("0", "participantId", "0", gpUserId , new NameValue("oD.locationPx",  "location"), null, "XSZSRS1MS3D4OAEK2DPM") ;
+    @Test
+    public void getKitShipperRoleCannotChangeParticipantInfo(){
```
denniscunningham: Nit. The test name lacks symmetry with the test above.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,129 @@
...
+        Patch patch = new Patch("0", "participantId", "0", cmiKitShippingOnlyUserId, new NameValue("oD.locationPx",  "location"), null, "XSZSRS1MS3D4OAEK2DPM") ;
         patch.setTableAlias("oD");
 
-        Assert.assertFalse(UserUtil.checkKitShippingAccessForPatch(realm, gpUserId, null, patch));
+        Assert.assertFalse(UserUtil.checkKitShippingAccessForPatch(studyInstanceName, cmiKitShippingOnlyUserId, null, patch));
     }
 
+    /**
+     * Verifies that users who don't have permission to make
```
denniscunningham, line 75: This seems a bit different. We get a user ID from AuthO and there is also one in the payload. (Not sure why we do things this way.) This verifies that when the two don't match we return an error?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
```diff
@@ -1,110 +1,129 @@
...
+     * a few new users and set their permissions.
+     */
+    @BeforeClass
+    public static void setup() {
+        String nameAppend = "." + System.currentTimeMillis();
+        String studyInstanceName = "instance." + nameAppend;
+        String cmiStudyGroup = "cmi" + nameAppend;
+
+        cmiAdminUtil.createRealmAndStudyGroup(studyInstanceName, cmiStudyGroup);
+        cmiAdminUtil.setStudyAdminAndRoles(generateUserEmail(), USER_ADMIN_ROLE, Arrays.asList(KIT_SHIPPING, PT_LIST_VIEW));
```
denniscunningham: This returns the admin user ID. The line below is unnecessary (and since it creates a different email address there might be unexpected results if the test is expanded).

.github/pull_request_template.md
```diff
@@ -33,6 +33,7 @@ _How does one observe these changes in a deployed system? Note that **user visib
 
 ## Testing
 
+- [ ] I have added regression test labels to the jira ticket
```
denniscunningham, line 36: Do we also want something like "I added verification steps to the Jira ticket"?

denniscunningham (commented):
Looks great!
A few comments.

pegahtah (approved):
LGTM, thanks a lot!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
zyme: If I set `dsmAdminUserId` to the return from L45, one of the tests fails.  So I think we do need L47.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
zyme: CC @pegahtah since I'm not familiar with the rationale behind this test.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
denniscunningham: Ah, I see the issue. setStudyAdminAndRoles set the study roles but does not assign them to the admin. I could probably fix that (a bit of an assumption), but perhaps I was mostly confused by the terminology: dsmAdminUserId in this case is not a user admin, but a regular user.

denniscunningham (commented):
One new comment. But I noticed this test has lots of checkstyle warnings, which we usually try to cleanup when we add new code. (You can see these in IntelliJ in the 'Problems' tab at the bottom of the page, if you are using IntelliJ.)

Also, it sounds like we are waiting on Pegah's feedback for one comment?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
zyme: Will change the name of var to be more clear.

denniscunningham (approved):

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/AuthenticationTest.java
pegahtah: Yes! The user id in the payload should match the one in the token. So basically _I think_ the idea behind this is that I can't use someone else's token to change a value in DSM.
 The user id in the request payload is going to get used for the audit purposes

PR #2664, opened 2023-08-25
Add user admin test utility

denniscunningham:
PEPPER-1076

This PR adds a test utility to help DSM tests set up and tear down test realms/instances, test study roles, and test users and their roles. See the UserAdminTestUtil class comment for more information on its usage.

The UserAdminServiceTest is modified to use UserAdminTestUtil, but is probably not a great example of usage since it is actually testing the UserAdminService, and thus needs to do some specialized test operations.

There may be reasons to expand UserAdminTestUtil, but ideally there are no reasons to modify UserAdminService to support testing.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
+++ b/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
@@ -31,11 +31,11 @@ public class UserAdminServiceTest extends DbTxnBaseTest {
     private static final String TEST_GROUP = "test_group";
     private static final String TEST_INSTANCE = "test_instance";
     private static String userAdminRole;
-    private static UserAdminTestUsers testUsers;
+    private static UserAdminTestUtil testUsers;
 
     @BeforeClass
     public static void setup() {
-        testUsers = new UserAdminTestUsers();
+        testUsers = new UserAdminTestUtil();
         testUsers.createRealmAndStudyGroup(TEST_INSTANCE, TEST_GROUP);
         userAdminRole = USER_ADMIN_ROLE;
         Assert.assertNotEquals(-1, UserAdminService.getRoleId(PEPPER_ADMIN_ROLE));
@@ -55,7 +55,7 @@ public class UserAdminServiceTest extends DbTxnBaseTest {
 
     public void testGetRoleId() {
         // temporary to understand which roles are handled by liquibase
-        log.info("TEMP: all roles: {}", String.join("\n", UserAdminTestUsers.getAllRoles().keySet()));
+        log.info("TEMP: all roles: {}", String.join("\n", UserAdminTestUtil.getAllRoles().keySet()));
         int roleId = UserAdminService.getRoleId("upload_onc_history");
         Assert.assertTrue(roleId > 0);
     }
--- a/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
+++ b/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
@@ -44,7 +44,7 @@ public class UserAdminServiceTest extends DbTxnBaseTest {
 
     @AfterClass
     public static void tearDown() {
-        testUsers.close();
+        testUsers.deleteGeneratedData();
     }
 
     @After
@@ -68,7 +68,7 @@ public class UserAdminServiceTest extends DbTxnBaseTest {
             userId = testUsers.createTestUser("test_admin1@study.org", List.of(roleName));
         } catch (Exception e) {
             e.printStackTrace();
-            Assert.fail();
+            Assert.fail("Unexpected exception: " + e.toString());
         }
         int groupId = UserAdminService.verifyStudyGroup(TEST_GROUP);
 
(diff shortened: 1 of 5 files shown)


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
```diff
@@ -35,122 +28,48 @@
...
-        int roleId = UserAdminService.getRoleId("dashboard_view");
-        Assert.assertTrue(roleId > 0);
-        int userId = createTestUser("test_admin1@study.org", roleId);
+        String roleName = "dashboard_view";
+        int userId = -1;
+        try {
+            userId = testUsers.createTestUser("test_admin1@study.org", List.of(roleName));
+        } catch (Exception e) {
+            e.printStackTrace();
+            Assert.fail();
```
zyme: Maybe add some text to the `fail()` so that whoever's looking at the test result knows a bit more about what was expected?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -0,0 +1,404 @@
...
+    private static final String SQL_DELETE_DDP_INSTANCE =
+            "DELETE FROM ddp_instance WHERE ddp_instance_id = ?";
+
+
+    public UserAdminTestUtil() {}
+
+    /**
+     * Call this to teardown this class, typically in an @After or @AfterClass method
+     */
+    public void close() {
```
zyme: How about `deleteGeneratedData()` or a different name that conveys what the method is doing?  I worry that callers won't expect from the name or javadoc that this is erasing test data.

zyme: It's awesome that this class keep tracks of the data and will wipe it out for you--that will make teardowns super simple.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -0,0 +1,404 @@
...
+            return dbVals;
+        });
+
+        if (res.resultException != null) {
+            throw new DsmInternalError("Error adding DDP instance group " + instanceName, res.resultException);
+        }
+        return (int) res.resultValue;
+    }
+
+    private static void deleteInstance(int instanceId) {
```
zyme: I think this should be `deleteInstanceGroup`?  I was trying to figure out the difference between this method and the one on L392.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -0,0 +1,404 @@
...
+                    }
+                }
+            } catch (SQLException e) {
+                throw new DsmInternalError("Error getting roles", e);
+            }
+            return roles;
+        });
+    }
+
+    private static int createTestInstance(String instanceName, int studyGroupId) {
```
zyme: This code could be used to create an instance regardless of test context, so should it just be called `createInstance()`?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -0,0 +1,404 @@
...
+                    }
+                }
+            } catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
+
+        if (res.resultException != null) {
+            _deleteInstance(instanceId);
```
zyme: Maybe wrap `_deleteInstance` in a try/catch to preserve the originating `res.resultException` error?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminServiceTest.java
denniscunningham: Will do. FWIW, this (and similar in this file) are catching unlikely internal errors, not testing something per se, thus the printStackTrace. It is nicer to have the stack trace in the output than to try to deduce the error from the Surefire logs.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
denniscunningham: Will do. FWIW, I was attempting to follow the Java close() and closeable() pattern since for Java programmers that's typically how a class releases resources when it goes out of scope. It did not make sense for this class to be autoCloseable() based on how I expect it to be used.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
denniscunningham: Will do. I'm typically not too fussy about private method names, but if it helps readability I'm all for it.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
denniscunningham: It actually does both instance and group. I'll name appropriately.

denniscunningham (commented):
@zyme Thx for your review. Updates to follow.

zyme (approved):

PR #2665, opened 2023-08-25
Fix ES version conflict handling

denniscunningham:
PEPPER-692

This PR fixes issues observed in Dev environment where version conflicts occur in UpsertPainless but are not handled properly. These conflicts do not result in an exception, but rather must be deduced from the request response. With this new code, if DSM detects a conflict it attempts to refresh the index then tries the update again.

(NOTE: Ideally we would lessen the likelihood of version conflicts in this scenario by bundling the updates for a participant into one update. Unfortunately, as the code is currently structured that is a big task which probably is not worth it at this point in Pepper's lifecycle. In any case, UpsertPainless needs to handle the version conflict case.)

Testing note: I was unable to reproduce the version conflicts locally, so the refresh code in UpsertPainless is effectively untested. I propose we get this code on Dev via this PR, where I will observe the version conflict behavior. At this point we do not have attractive testing options since we lack an ES test container, and we do not have a way to run PW tests against a branch.

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
@@ -50,19 +50,20 @@ public class UpsertPainless implements Exportable {
 
         Throwable ioException = null;
         for (int tryNum = 1; tryNum < 3; tryNum++) {
-            ioException = executeExport(clientInstance, updateByQueryRequest);
-            if (ioException == null) {
-                break;
+            try {
+                executeExport(clientInstance, updateByQueryRequest);
+                return;
+            } catch (IOException e) {
+                logger.info("Error occurred exporting data to ES on try number {} (retrying): {}", tryNum, e);
+                ioException = e;
             }
-            logger.info("Error occurred exporting data to ES on try number {} (retrying): {}", tryNum, ioException.toString());
         }
-        if (ioException != null) {
-            logger.error("Unable to connect to ElasticSearch: {}", ioException.getMessage());
-            throw new DsmInternalError("Unable to connect to ElasticSearch", ioException);
-        }
+        // TODO: keeping error logging until we are sure exceptions are not swallowed in the call stack
+        logger.error("Unable to connect to ElasticSearch", ioException);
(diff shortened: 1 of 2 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/painless/UpsertPainless.java
```diff
@@ -46,35 +50,52 @@ public void export() {
...
+            logger.info("Error occurred exporting data to ES on try number {} (retrying): {}", tryNum, ioException.toString());
         }
         if (ioException != null) {
             logger.error("Unable to connect to ElasticSearch: {}", ioException.getMessage());
             throw new DsmInternalError("Unable to connect to ElasticSearch", ioException);
         }
     }
 
-    private Throwable executeExport(RestHighLevelClient clientInstance, UpdateByQueryRequest updateByQueryRequest, int i) {
+    private Throwable executeExport(RestHighLevelClient clientInstance, UpdateByQueryRequest updateByQueryRequest) {
```
ssettipalli: Any advantage in returning Exception. 
This method can throw the Exception and the caller can catch it in try..catch , right ?

denniscunningham: Yes, I like your suggestion better than what I have here.
This approach is historical and I was grounded in it rather than seeing a better way.
Thx!

denniscunningham: Actually, I see now why I did it that way. The loop with a catch is doable, but kind of clumsy and hard to read. The most straightforward implementation of that type is a `for` loop with integer counters, but the code inside the loop has to check for the loop max count to decide whether to throw or try again. Alternatively, the exception could be assigned to a variable outside of the loop, but at that point returning the throwable is easier to understand.

What do you think? Are you ok with this implementation or did you have something specific in mind?

denniscunningham:
FWIW, I was able to reproduce the version conflict locally (it's a timing issue) and trace through the new code and it behaved as expected (the refresh worked).

ssettipalli (approved):

PR #2666, opened 2023-08-28
Better patch route error detection and reporting

denniscunningham:
PEPPER-984

This PR does not contain fixes per se for PEPPER-984, but improves input checking and reporting. Said differently, with these changes we can hopefully understand *why* we are seeing the errors reported in the bug.
 
Specifically: 

- Added input checking improvements to PatchRoute, and updated the exception handling since DSM now has uniform exception handling at the Spark level for a handful of common exceptions (see DsmServer). 
- Added RouteUtil class. There are several common route patterns that should be handled uniformly, so this is just the start of getting those patterns in one place.
- Added input checking in OncHistoryDetailPatch and detangled the code a bit. As I noted in a prior PR, this code is very tangled and deserves a rewrite, but we probably need to do this methodically since there is limited testing support.
- Added exception handling in ElasticUtil.updateRequest. We are seeing instances of ES version conflicts in Prod, which should be reported at error level, since we need to fix these. (That means we are failing to update ES, thus ES is wrong until we do an export.)

## FUD Score

## Testing

## Release


pegahtah (approved):
LGTM, thanks!
My only concern is, I'm not sure if all of the patch requests from frontend have a realm

denniscunningham:
@pegahtah I was thinking about that too (is realm required), but I could not think of.a case where we would not need a realm? Can you think of a scenario?
(In any case, we have an upcoming task to investigate requests without realms, so I removed that check in this PR.)

PR #2671, opened 2023-09-06
PEPPER-1093

ssettipalli:
Add email deny list support to Housekeeping so that play playwright user emails can be skipped

Add  emailPatternDenyList entries to vault conf in Dev and test

This PR does not skip/impact any emails sent by DSM which I believe is very small number.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
@@ -79,8 +79,8 @@ public class EmailNotificationHandler implements HousekeepingMessageHandler<Noti
         this.pdfBucketService = pdfBucketService;
         this.pdfGenerationService = pdfGenerationService;
         Config cfg = ConfigManager.getInstance().getConfig();
-        this.emailDenyPatterns =  cfg.hasPath(ConfigFile.EMAIL_PATTERN_DENY_LIST) ?
-                cfg.getStringList(ConfigFile.EMAIL_PATTERN_DENY_LIST) : Collections.EMPTY_LIST;
+        this.emailDenyPatterns =  cfg.hasPath(ConfigFile.EMAIL_PATTERN_DENY_LIST)
+                ? cfg.getStringList(ConfigFile.EMAIL_PATTERN_DENY_LIST) : Collections.EMPTY_LIST;
     }
 
     /**
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
@@ -17,17 +17,22 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.sendgrid.helpers.mail.objects.Attachments;
 import com.sendgrid.helpers.mail.objects.Email;
 import com.sendgrid.helpers.mail.Mail;
 import com.sendgrid.helpers.mail.objects.Personalization;
+import com.typesafe.config.Config;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -89,6 +99,32 @@ public void handleMessage(NotificationMessage message) {
             }
         }
 
+        //skip emails for PlayWright generated users / Email deny list
+        if (!emailDenyPatterns.isEmpty()) {
+            boolean hasDenyListEmail = false;
+            List<String> nonDenyEmailList = new ArrayList<>();
+            for (String toAddress : message.getDistributionList()) {
```
denniscunningham: If I'm reading this correctly,  if all addresses on the distribution list match the deny pattern, the email is not sent. How many addresses typically appear on the distribution list? Besides the ptp, what other addresses does the list contain?
A comment might help explain some of this.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -89,6 +99,32 @@ public void handleMessage(NotificationMessage message) {
...
+                        isDenyEmail = true;
+                        hasDenyListEmail = true; //to handle email DL with a PW test user and other emails
+                        break;
+                    }
+                }
+                if (!isDenyEmail) {
+                    nonDenyEmailList.add(toEmail.getEmail());
+                }
+            }
+            if (hasDenyListEmail && nonDenyEmailList.isEmpty()) {
```
denniscunningham: Your comment above may be directed at this (I don't quite understand it), but I'm not sure why you need an address list, since you never use the contents. You might also get away with one flag and a break statement if part of this clause was in a separate method? But I may be missing something here.

denniscunningham (commented):
A couple questions in my comments.
Also, does it make sense to break the matcher into a method that can be unit tested against a set of likely email address patterns? In this case, I am wondering how the matcher will handle the dots in the email address, but my regex is rusty.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Usually only ptp email address, however code/platform does support more  than 1 to email hence needed to handle it.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: Refactored code to handle it little differently. added detailed comments explaining logic

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
```diff
@@ -89,6 +101,34 @@ public void handleMessage(NotificationMessage message) {
...
+
+        //skip emails for PlayWright generated users / Email deny list
+        if (!emailDenyPatterns.isEmpty()) {
+            //collect emails from actual distribution list (DL) that matches deny patterns and remove from original email DL.
+            //If no additional valid emails, return... else continue with filtered new DL.
+            Collection<String> denyEmailList = new HashSet<>();
+            for (String toAddress : toEmailList) {
+                for (String emailPattern : emailDenyPatterns) {
+                    Pattern pattern = Pattern.compile(emailPattern);
+                    Matcher matcher = pattern.matcher(toAddress);
```
denniscunningham, line 114: Per my earlier comment (https://github.com/broadinstitute/ddp-study-server/pull/2671#pullrequestreview-1615321477), still wondering if an email pattern like `@test.firecloud.org` will also match `@testXfirecloudXorg`. If so, do we care?

denniscunningham (approved):
Left a comment.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/handler/EmailNotificationHandler.java
ssettipalli: I believe it does. It makes it invalid email to start with though.
I think we can tweak the pattern we setup in conf if needed by adding \\
Not too concerned about it for our use case for now

PR #2673, opened 2023-09-11
[DSM] PEPPER-833 RGP BLOOD & RNA kit name

pegahtah:
## Context

Google doc for better explanation of the implementation sketch is [here](https://docs.google.com/document/d/1Jca5QhLEPo47-bPZyNxgr1zSFraRJHEzYWpnUQxwcY0/edit)

**A little background:** RGP used to send the same BLOOD kit type as CMI, therefore the BLOOD entry in the kit_type table was shared between CMI and RGP. 
Now they are going to start sending a new package containing one tube for BLOOD and one for an RNA sample. Some changes that were made in the past for this new kit was
1. adding 2 new kit type in the `kit_type` table
* One representing the new combined BLOOD and RNA kit.
* Another representing the RNA kit type that will be included in the same package.
2. Creating a new entry in the `ddp_kit_request_settings` to associate RGP with the new BLOOD and RNA kit type.
3. Adding the kit_type_id from BLOOD and the kit_type_id from RNA in the `sub_kits_settings` table to the BLOOD and RNA kit
* this change means that when a "BLOOD and RNA" kit is uploaded for an RGP participant, DSM should insert 2 rows for this participant in the kit tables, one for the BLOOD type and one for the RNA type. 
4. Because of 2 rows being added per 1 kit package, we masked the RNA row from frontend so the user treats the combination of one BLOOD and one RNA uploaded kit as one BLOOD & RNA package
5. We have also added the RGP Final Scan page in frontend and the class in backend. 

 **Now in this Pull Request**:
Since per each upload and kit creation DSM was adding one BLOOD and one RNA kit for the participant and the RNA one was hidden from frontend, the kit still showed up named as `BLOOD`, making it confusing for the study staff and GP users. In this PR I am: 
* Adding a `display_name` column to the `kit_type` table to clearly delineate the kit types.
* Separating the RGP Blood kit from others by inserting a distinct blood kit type for RGP, enhancing clarity and adherence to best practices.
* Ensuring CMI studies remain unaffected by preserving the original display name for them.
* Updating `sub_kits_settings` to reference the newly inserted blood type for RGP's BLOOD and RNA kit, fostering accurate and streamlined kit handling.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -270,7 +271,7 @@ public Integer insertKit(KitRequestShipping kitRequestShipping) {
             return dbVals;
         });
         if (Objects.nonNull(results.resultException)) {
-            throw new RuntimeException("Error inserting kit with dsm_kit_request_id: "
+            throw new DsmInternalError("Error inserting kit with dsm_kit_request_id: "
```
pegahtah: sonar checkstyle fix

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -304,7 +305,7 @@ public Integer insertKitRequest(KitRequestShipping kitRequestShipping) {
             return dbVals;
         });
         if (Objects.nonNull(results.resultException)) {
-            throw new RuntimeException("Error inserting kit request for participant id: "
+            throw new DsmInternalError("Error inserting kit request for participant id: "
```
pegahtah: sonar checkstyle fix

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -323,7 +324,7 @@ public Optional<KitRequestShipping> getKitRequest(Long kitRequestId) {
                         dbVals.resultValue = getKitRequestShipping(rs);
                     }
                     if (numRows > 1) {
-                        throw new RuntimeException("Found " + numRows + " kits for dsm_kit_request_id " + kitRequestId);
+                        throw new DsmInternalError("Found " + numRows + " kits for dsm_kit_request_id " + kitRequestId);
```
pegahtah: sonar checkstyle fix

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -662,7 +662,7 @@ public boolean hasKitReceived(Connection connection, String ddpParticipantId) {
                 return true;
             }
         } catch (Exception ex) {
-            throw new RuntimeException(String.format("Error getting kits for %s", ddpParticipantId));
+            throw new DsmInternalError(String.format("Error getting kits for %s", ddpParticipantId));
```
pegahtah, line 651: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -639,7 +639,7 @@ public Optional<ScanError> updateKitLabel(KitRequestShipping kitRequestShipping)
                     dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(), "dsm_kit_id "
                             + kitRequestShipping.getDsmKitId() + " does not exist or already has a Kit Label");
                 } else {
-                    logger.info("Updated kitRequests for pt w/ shortId " + kitRequestShipping.getHruid());
+                    logger.info(String.format("Updated kitRequests for pt w/ shortId %s", kitRequestShipping.getHruid()));
```
pegahtah: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -619,7 +619,7 @@ public List<KitRequestShipping> getKitsByHruid(String hruid) {
                 dbVals.resultException = ex;
             }
             if (dbVals.resultException != null) {
-                throw new RuntimeException("Error getting kit request with shortID " + hruid, dbVals.resultException);
+                throw new DsmInternalError(String.format("Error getting kit request with shortID %s", hruid), dbVals.resultException);
```
pegahtah, line 608: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -588,17 +588,17 @@ private boolean booleanCheckFoundAsName(String kitLabel, String query) {
...
+                throw new DsmInternalError("Error checking if kit exists in tracking table ");
             }
-            logger.info("Found " + dbVals.resultValue + " kit in tracking table w/ kit_label " + kitLabel);
+            logger.info(String.format("Found %s kit in tracking table w/ kit_label %s ", dbVals.resultValue,  kitLabel));
             return dbVals;
         });
 
         if (results.resultException != null) {
-            logger.error("Error checking if kit exists in tracking table w/ kit_label " + kitLabel, results.resultException);
+            logger.error(String.format("Error checking if kit exists in tracking table w/ kit_label %s", kitLabel), results.resultException);
```
pegahtah, line 587: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -588,17 +588,17 @@ private boolean booleanCheckFoundAsName(String kitLabel, String query) {
...
             if (dbVals.resultException != null) {
-                throw new RuntimeException("Error checking if kit exists in tracking table ", dbVals.resultException);
+                throw new DsmInternalError("Error checking if kit exists in tracking table ", dbVals.resultException);
             }
             if (dbVals.resultValue == null) {
-                throw new RuntimeException("Error checking if kit exists in tracking table ");
+                throw new DsmInternalError("Error checking if kit exists in tracking table ");
             }
-            logger.info("Found " + dbVals.resultValue + " kit in tracking table w/ kit_label " + kitLabel);
+            logger.info(String.format("Found %s kit in tracking table w/ kit_label %s ", dbVals.resultValue,  kitLabel));
```
pegahtah: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -554,16 +554,16 @@ public Optional<ScanError> insertKitTracking(KitRequestShipping kitRequestShippi
...
+                                    + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER, kitRequestShipping.getKitLabel()));
+                    logger.error(String.format("The number of affected rows for kit label %s is not 1", kitRequestShipping.getKitLabel()));
                 } else {
                     logger.info("Added tracking for kit w/ kit_label " + kitRequestShipping.getKitLabel());
                 }
             } catch (Exception ex) {
                 dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(),
-                        "Kit Label \"" + kitRequestShipping.getKitLabel() + "\" does not exist.\n"
-                                + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER);
+                        String.format("Error occured for Kit Label \"%s\"\n"
```
pegahtah, line 554: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -492,7 +491,7 @@ public Optional<List<KitRequestShipping>> getSubkitsByDdpLabel(String ddpLabel,
                 dbVals.resultException = ex;
             }
             if (dbVals.resultException != null) {
-                throw new RuntimeException("Error getting kit request with ddp label " + ddpLabel, dbVals.resultException);
+                throw new DsmInternalError(String.format("Error getting kit request with ddp label %s", ddpLabel), dbVals.resultException);
```
pegahtah, line 483: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -444,7 +443,7 @@ public Optional<KitRequestShipping> getKitByDdpLabel(String ddpLabel, String kit
...
```
pegahtah, line 435: sonar fixes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/KitTypeDto.java
```diff
@@ -17,5 +17,6 @@ public class KitTypeDto {
     Boolean requiresInsertInKitTracking;
     Boolean noReturn;
     String sampleType;
+    String kitTypeDisplayName;
```
denniscunningham: Seems like this could just be `displayName`? `kitType` seems redundant since that is the class name.

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
```diff
@@ -0,0 +1,35 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="PEPPER-833_rgp_kit_type_display_name" author="pegah">
+        <addColumn tableName="kit_type">
+            <column name="kit_display_name" type="VARCHAR(500)"></column>
+        </addColumn>
+    </changeSet>
+    <changeSet id="PEPPER-833_rgp_blood" author="pegah">
+        <insert tableName="kit_type">
+            <column name="kit_type_name" value="BLOOD"></column>
```
denniscunningham, line 14: Looks like this column is not unique? How does DSM deal with different records with the same `kit_type_name`?

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
```diff
@@ -0,0 +1,35 @@
...
+            <column name="bsp_material_type" value="Whole Blood Streck Cell-Free Preserved"></column>
+            <column name="bsp_receptacle_type" value="Vacutainer Cell-Free DNA Tube Camo-Top [10mL]"></column>
+            <column name="requires_insert_in_kit_tracking" value="1"></column>
+            <column name="kit_display_name" value="BLOOD &amp; RNA"></column>
+        </insert>
+    </changeSet>
+    <changeSet id="PEPPER-833_update_rgp_blood_subkit" author="pegah">
+        <update tableName="sub_kits_settings">
+            <column name="kit_type_id" valueComputed="(SELECT kit_type_id FROM kit_type WHERE kit_display_name = 'BLOOD &amp; RNA' )"></column>
+            <where>
```
denniscunningham, line 24: This feels odd. We are using the display name to figure out the kit type ID? Developers might feel they can change a field with the term `display` in it. 
Is there a reason we do not have a unique key in `kit_type`?
What's the overall goal here?
What is the `sub_kits_settings` table used for?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -122,7 +122,9 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                     + "LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id "
                     + "AND ex.ddp_participant_id = request.ddp_participant_id) "
                     + "LEFT JOIN ddp_kit_request_settings dkc ON (request.ddp_instance_id = dkc.ddp_instance_id "
-                    + "AND request.kit_type_id = dkc.kit_type_id) WHERE ex.ddp_participant_exit_id is null";
+                    + "AND request.kit_type_id = dkc.kit_type_id) "
+                    + "LEFT JOIN kit_type ktype ON (request.kit_type_id = ktype.kit_type_id) "
```
denniscunningham, line 130: Two things:

1. It is pretty much impossible to visually inspect a query this complex for a code review. It should really be refactored at some point.
2. I'm not sure what this new join provides since the test above is definitive.

We really need an automated test for this unless we have another way of evaluating that this is correct.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -517,11 +523,11 @@ public static Map<String, List<KitRequestShipping>> getKitRequests(@NonNull DDPI
     }
 
     private static Map<String, List<KitRequestShipping>> getKitRequests(@NonNull String realm, String target, String kitType,
-                                                                        boolean getAll) {
+                                                                        Integer kitTypeId, boolean getAll) {
```
denniscunningham: Perhaps the `kitType` parameter should be name `kitTypeName`, now that we have a `kitTypeId`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -684,9 +690,9 @@ private static boolean existsParticipant(KitRequestShipping kit, ElasticSearchPa
     }
 
     public static Map<String, List<KitRequestShipping>> getAllKitRequestsByRealm(@NonNull String realm, String target, String kitType,
-                                                                                 boolean getAll) {
+                                                                                 Integer kitTypeId, boolean getAll) {
```
denniscunningham, line 701: Perhaps the `kitType` parameter should be name `kitTypeName`, now that we have a `kitTypeId`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -751,14 +757,19 @@ public static Map<String, List<KitRequestShipping>> getAllKitRequestsByRealm(@No
     }
 
     private static PreparedStatement getPreparedStatement(@NonNull Connection conn, String target, @NonNull String realm, String type,
-                                                          boolean getAll) throws SQLException {
+                                                          Integer kitTypeId, boolean getAll) throws SQLException {
```
denniscunningham: Perhaps the `type` parameter should be name `kitTypeName`, now that we have a `kitTypeId`? `type` is not a good parameter name anyway, unless type is very obvious.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -751,14 +757,19 @@ public static Map<String, List<KitRequestShipping>> getAllKitRequestsByRealm(@No
...
         PreparedStatement stmt = null;
         if (getAll) {
             String query = SQL_SELECT_KIT_REQUEST.concat(QueryExtension.BY_REALM);
             stmt = conn.prepareStatement(query);
             stmt.setString(1, realm);
         } else {
-            if (StringUtils.isNotBlank(realm) && StringUtils.isNotBlank(type)) {
+            if (kitTypeId != null && kitTypeId != null) {
```
denniscunningham: Is this the same criteria in both clauses?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -751,14 +757,19 @@ public static Map<String, List<KitRequestShipping>> getAllKitRequestsByRealm(@No
...
+                    String query = addQueryExtension(target, SQL_SELECT_KIT_REQUEST.concat(QueryExtension.BY_REALM_AND_TYPE_ID));
+                    stmt = conn.prepareStatement(query);
+                    stmt.setString(1, realm);
+                    stmt.setInt(2, kitTypeId);
```
denniscunningham: Not sure I understand, looking at the next clause... `kitTypeId` and `type` are interchangeable?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -200,8 +198,8 @@ public Optional<ScanError> updateKitRequest(KitRequestShipping kitRequestShippin
             } catch (Exception ex) {
                 logger.error("Not able to update the kit for ddpLabel " + kitRequestShipping.getDdpLabel(), ex);
                 dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(),
-                        "Kit Label \"" + kitRequestShipping.getDdpLabel() + "\" was already scanned.\n"
-                                + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER);
+                        String.format("Kit Label \"%s\" w was already scanned.\n"
```
denniscunningham, line 190: Seems like this is a user input error, yet we are logging at error level?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -554,16 +554,16 @@ public Optional<ScanError> insertKitTracking(KitRequestShipping kitRequestShippi
                 int rowsAffected = stmt.executeUpdate();
                 if (rowsAffected != 1) {
                     dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(),
-                            "Kit Label \"" + kitRequestShipping.getKitLabel() + "\" does not exist.\n"
-                                    + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER);
-                    logger.error("The number of affected rows for kit label " + kitRequestShipping.getKitLabel() + "is not 1.");
+                            String.format("Kit Label \"%s\" does not exist.\n"
+                                    + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER, kitRequestShipping.getKitLabel()));
+                    logger.error(String.format("The number of affected rows for kit label %s is not 1", kitRequestShipping.getKitLabel()));
```
denniscunningham, line 548: Why doesn't this method rethrow these exceptions? Is it called during message processing? If so, perhaps the exceptions should be handled at the top of the message handling stack?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: A better fix is to use `{}` provided by the logger class as direct string replacement.

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="PEPPER-833_rgp_kit_type_display_name" author="pegah">
+        <addColumn tableName="kit_type">
+            <column name="kit_display_name" type="VARCHAR(500)"></column>
```
denniscunningham: 'display_name' is probably sufficient here, and `kit_display_name` may be confusing since it is actually the kit type display name, not the kit display name.

denniscunningham (changes requested):
I'm a bit confused about what's going on here. On one hand there is a new `display_name` field for `kit_type`, but there is also a new Integer parameter passed through the DAO code called (mostly) `kitTypeId`. How are these related?

Also, this is a scary PR to approve without some assurance that this code works properly. Are there existing automated tests that cover this, or some other way to demonstrate that this code works?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: The join with the kit_type now allows us to get the kit type's display names from the table.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: 😂 I don't know what happened there, thanks for catching it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Yes, it's based on what is chosen as the query extension. if it's `BY_REALM_AND_TYPE_ID` then the second one should be the type id, and if it's `BY_REALM_AND_TYPE` then the second one should be the type name. 

It would be far above this ticket's scope but we can make one ticket to refactor all of these kit methods.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: Yes, this scan error is sent to the frontend and is displayed for the user

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: It's because the inserting kit is called for an array of kit barcodes to get inserted, if one has an error we collect it as a scan error and log it but we want DSM to continue inserting the rest. The ones with an error are then returned to frontend and frontend displays them to the user so they can adjust the barcodes

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: Something interesting, I found out that we have an (  I am 99.9% sure ) unused column in `ddp_kit_request_settings` called `kit_type_display_name` . I think we need to investigate that more and get rid of it if it's not used. My first attempt was to get the code working with that but that would not work with sub kits

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: We have different blood kits for different studies in dev db, since every query has a matching with both realm and kit type name, so far DSM handles it ok because we didn't have a study with 2 blood kits. RGP would be a first and that's why I introduced searching by kit type ids instead

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: Yes, we kinda have to. The display name is the only thing making this field different than the others. Also kit types unfortunately are not unique, and my guess is that there could be many kits called blood but having different meanings for each study or different metadata. You can look at dev database to see it. 
I have added the general explanation of subkits to a new doc and will send it to you, but basically `sub_kits_settings` is the settings for when a package of kits contains multiple kit types and tubes in it. For example in this ticket RGP wants to send one package to each participant with 2 tubes, one for blood and one for RNA extraction

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Also confirmed with Kiara, we have general automated tests for kits :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: But do we need to log at error level, which will trigger a prod alert?
Should we first validate the data provided by the user? It feels like there might be a handful of reasons for this exception (which should probably be a SqlException), but this message concludes that it is only one.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: Thx for the explanation -- makes sense. 
As I mentioned above, logging at error level will cause a prod alert. Since it seems like the actual issue in some cases is bad user input, that does not feel right. As I mentioned above, perhaps it makes sense to first validate the user input, so that we can be confident that when `rowsAffected` is not 1, that there is a problem with table data, not with the input.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: String replacement for logging should use the `{}` pattern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: See comment above.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
denniscunningham: Is Sonar actually complaining about this pattern (simple append)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -639,7 +628,7 @@ public Optional<ScanError> updateKitLabel(KitRequestShipping kitRequestShipping)
                     dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(), "dsm_kit_id "
                             + kitRequestShipping.getDsmKitId() + " does not exist or already has a Kit Label");
                 } else {
-                    logger.info("Updated kitRequests for pt w/ shortId " + kitRequestShipping.getHruid());
+                    logger.info("Updated kitRequests for pt w/ shortId {}", kitRequestShipping.getHruid());
```
denniscunningham, line 628: Yes, this is the pattern for logging string replacement.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/KitRequestShippingMigrator.java
```diff
@@ -14,7 +14,7 @@ public KitRequestShippingMigrator(String index, String realm) {
 
     @Override
     protected Map<String, Object> getDataByRealm() {
-        Map<String, List<KitRequestShipping>> kitRequests = KitRequestShipping.getAllKitRequestsByRealm(realm, null, null, true);
+        Map<String, List<KitRequestShipping>> kitRequests = KitRequestShipping.getAllKitRequestsByRealm(realm, null, null,  true);
```
denniscunningham: Looks like this change is unnecessary?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -750,20 +762,18 @@ public static Map<String, List<KitRequestShipping>> getAllKitRequestsByRealm(@No
...
-                                                          boolean getAll) throws SQLException {
+    private static PreparedStatement getSelectKitReqPreparedStatement(@NonNull Connection conn, String target, @NonNull String realm,
+                                                                      Integer kitTypeId, boolean getAll) throws SQLException {
         PreparedStatement stmt = null;
         if (getAll) {
             String query = SQL_SELECT_KIT_REQUEST.concat(QueryExtension.BY_REALM);
             stmt = conn.prepareStatement(query);
             stmt.setString(1, realm);
         } else {
-            if (StringUtils.isNotBlank(realm) && StringUtils.isNotBlank(type)) {
```
denniscunningham, line 761: Does it make sense to check for these conditions and throw if they are not met, like an assertion?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -516,20 +526,19 @@ public static Map<String, List<KitRequestShipping>> getKitRequests(@NonNull DDPI
...
+            try (PreparedStatement stmt = getSelectKitReqPreparedStatement(conn, target, realm, kitTypeId, getAll)) {
                 if (stmt != null) {
                     try (ResultSet rs = stmt.executeQuery()) {
                         while (rs.next()) {
                             addKitRequest(rs, kitRequests);
                         }
                     }
                 } else {
-                    throw new RuntimeException("No prepareStatement was created " + target + " " + realm + " " + kitType);
+                    throw new DsmInternalError("No prepareStatement was created " + target + " " + realm + " " + kitTypeId);
```
denniscunningham: As things stand now `stmt` will never be null.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -588,17 +577,17 @@ private boolean booleanCheckFoundAsName(String kitLabel, String query) {
...
             if (dbVals.resultValue == null) {
-                throw new RuntimeException("Error checking if kit exists in tracking table ");
+                throw new DsmInternalError("Error checking if kit exists in tracking table ");
             }
-            logger.info("Found " + dbVals.resultValue + " kit in tracking table w/ kit_label " + kitLabel);
+            logger.info(String.format("Found %s kit in tracking table w/ kit_label %s ", dbVals.resultValue,  kitLabel));
             return dbVals;
         });
 
         if (results.resultException != null) {
```
denniscunningham, line 586: It looks like we are mixing two different error handling techniques here. It seems like `resultException` will always be null here.

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
```diff
@@ -0,0 +1,35 @@
...
+            <column name="display_name" type="VARCHAR(500)"></column>
+        </addColumn>
+    </changeSet>
+    <changeSet id="PEPPER-833_rgp_blood" author="pegah">
+        <insert tableName="kit_type">
+            <column name="kit_type_name" value="BLOOD"></column>
+            <column name="bsp_material_type" value="Whole Blood Streck Cell-Free Preserved"></column>
+            <column name="bsp_receptacle_type" value="Vacutainer Cell-Free DNA Tube Camo-Top [10mL]"></column>
+            <column name="requires_insert_in_kit_tracking" value="1"></column>
+            <column name="display_name" value="BLOOD &amp; RNA"></column>
```
zyme: Probably safer to replace `&` with `and`.  Might not make any difference, but might save us from some obscure character encoding/escaping headaches.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -76,9 +77,7 @@ public class KitDaoImpl implements KitDao {
             + "WHERE "
             + "dsm_kit_id = ?";
 
-    private static final String INSERT_KIT_TRACKING = "INSERT INTO "
-            + "ddp_kit_tracking "
-            + "SET "
+    private static final String INSERT_KIT_TRACKING = "INSERT INTO ddp_kit_tracking SET "
```
zyme, line 80: Neat!  I was not familiar with the `insert into ... set...` syntax, but I see the benefit.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitTypeImpl.java
```diff
@@ -31,14 +38,57 @@ public class KitTypeImpl implements KitTypeDao {
...
+                    }
+                }
+            } catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Error looking up kit type  ", results.resultException);
```
zyme: Please add more context to error message, like `realm` and `kitTypeName`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/KitSubKits.java
```diff
@@ -63,10 +64,10 @@ public static List<KitSubKits> getSubKits(@NonNull String realm, @NonNull String
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error looking up kit type  ", results.resultException);
+            throw new DsmInternalError("Error looking up sub kit with type  "+ kitType, results.resultException);
         }
         if (subKits.size() > 1) { //otherwise it found the normal kit
-            logger.info("Found " + subKits.size() + " subKits");
+            logger.info("Found {} subKits", subKits.size());
```
zyme: Maybe add "for `realm` and `kitTypeName`" here so that the logs are more informative--if we even need this as an info.  Might be better as a debug.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/KitRequestShippingMigrator.java
```diff
@@ -14,7 +14,7 @@ public KitRequestShippingMigrator(String index, String realm) {
 
     @Override
     protected Map<String, Object> getDataByRealm() {
```
zyme, line 19: Might be worth a method comment so that the caller knows the type of the values in the map.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
```diff
@@ -151,7 +151,7 @@ public Optional<ScanError> processRGPFinalScan(ScanPayload scanPayload) {
                  return result;
              }
             KitRequestShipping rgpRnaKit =
-                    subkits.stream().filter(subkit -> subkit.getKitTypeName().contains("RNA")).findFirst().orElseThrow();
+                    subkits.stream().filter(subkit -> subkit.getKitTypeName().equals("RNA")).findFirst().orElseThrow();
```
zyme: If `subkit` is null or `kit_type_name` is null, this will NPE.  A yoda statement like "RNA".equals(...) will save you from an NPE if kit_type_name is null (but not subkit).

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
zyme: What should be the unique key here?  Or should we add a UK to ddp_kit_request_settings to clarify that we do not support a ddp project that has multiple of the same named kit types?

zyme:
Overall this looks good, and after I probed in the code and database and thought about how clients are using kit type name, my opinion is that doing a db refactor to update the kit_type table so that kit_type_name is unique and creating backfill rows that reflect the study + type so that we have kit_type_names like `CMI_BOOD` or `DARWIN_SALIVA` is probably going to break lots of things in ways that will be a challenge to anticipate.  So my vote is to stay with the `display_name` column, and make this new column unique.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/DDPRequestRouteTest.java
```diff
@@ -136,7 +136,7 @@ public void checkForParticipant() throws Exception {
                                 "https://track.easypost.com/djE6dHJrXzY4NGJmYzU3ZjM5OTQ1Zjg5MjEzOGRmMWVmMjI1NWZl", null, 12L, false, "",
                                 12L, null, 12L, "so what", "mf_testLabel", false, "shp_f470591c3fb441a68dbb9b76ecf3bb3d", 12L, null,
                                 "44445", false, "NOT FOUND", null, null, null, null, 0L, false, null, null, null, null, null, null, null,
-                                null, null, null));
+                                null, null, null, null));
```
zyme, line 139: The only use of this new `displayName` in a test appears to only check cases where the value is null.  I can't tell if we have tests that hit the new display name with a non-null value.

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="PEPPER-833_rgp_kit_type_display_name" author="pegah">
+        <addColumn tableName="kit_type">
+            <column name="display_name" type="VARCHAR(500)"></column>
```
zyme, line 9: It looks like we are only setting this value for the new blood and rna type.  Is client code going to behave properly if this field is not null?  Should we put a unique key constraint on this new column and backfill it in some way that prefixes study guid or some such?

pegahtah: yes. I changed the client code to first look at the display name and show that, if there is no display name then it should just work as it currently does and show the kit_type_name

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: there is no unique key here (aside form the primary key of the table)
since in ddp_kit_request_settings, a realm is bundled together with a kit_type_id, I'm not sure how we can put a constraint on it to avoid pairing ddp_instance_id in that table with another value in kit_type table

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: Agreed, do you think GP is going to be ok with it? (I'm not sure where we are in terms of if we have shared screen shots with them for their sop )

zyme: Good question--can you ask them directly?

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
zyme: I think a unique key on (ddp_instance_id,kit_type_id) will work for ddp_kit_request.  These queries return the same results which to me says that the key won't break anything:

```
select concat(kit_type_id, '.', ddp_instance_id) from ddp_kit_request_settings

select distinct(concat(kit_type_id, '.', ddp_instance_id)) from ddp_kit_request_settings
```

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
zyme: Can we encapsulate the "if display name is null, show the kit type name" into whatever dto we're using to supply both values so that client code does not have to do this check each wherever it's looking for the value?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitTypeImpl.java
```diff
@@ -31,14 +38,57 @@ public class KitTypeImpl implements KitTypeDao {
...
+            + " WHERE realm.instance_name = ? AND type.kit_type_name = ?";
+
+    /**
+     * Getting kitTypes based on kit type name and realm
+     */
+    public static KitTypeDto getKitTypes(@NonNull String realm, @NonNull String kitTypeName) {
+        log.info("Finding kit typ for realm {} with kit type name {}", realm, kitTypeName);
+        List<KitTypeDto> kitTypes = new ArrayList<>();
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
```
denniscunningham: SimpleResult is probably not needed since you are filling `kitTypes` directly. You also handle exceptions directly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitTypeImpl.java
```diff
@@ -31,14 +38,57 @@ public class KitTypeImpl implements KitTypeDao {
...
+            }
+            return dbVals;
+        });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Error looking up kit type  ", results.resultException);
+        }
+        if (kitTypes.size() != 1) {
+            // JDBC does not fetch all the data at once, so we need to read all the kit types here and then later
+            // throw an error if there was more than one kit type for this realm with the same name
```
denniscunningham: I don't understand this comment. The test here is pretty clear that we are expecting only one result. If so, it probably makes sense to handle the results differently when processing the query,

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitDisplayNameTest.java
```diff
@@ -0,0 +1,102 @@
...
+                "https://easypost-files.s3-us-west-2.amazonaws"
+                        + ".com/files/postage_label/20200214/8240f1b66535494a82b1ec0d566c3f0f.png", "", "794685038506",
+                "9405536897846100551129", "https://track.easypost.com/djE6dHJrXzY4NGJmYzU3ZjM5OTQ1Zjg5MjEzOGRmMWVmMjI1NWZl",
+                null, 12L, false, "", 12L, null, 12L, "so what", "mf_testLabel", false, "shp_f470591c3fb441a68dbb9b76ecf3bb3d",
+                12L, null, "44445", false, "NOT FOUND", null, null, null, null, 0L, false, "STANDALONE", null, null, null,
+                null, null, null, null, null, null, BLOOD_RNA_KIT_TYPE_DISPLAY_NAME);
+        Assert.assertEquals(kitWithDisplayName.getDisplayName(), BLOOD_RNA_KIT_TYPE_DISPLAY_NAME);
+        Assert.assertEquals(kitWithDisplayName.getKitTypeName(), BLOOD_RNA_KIT_TYPE_NAME);
+
+        // test that not having a display name means the display name is set as the kit type name
```
zyme: Nice!  Now for the icing on the cake: since this test does not need any database setup, and is really testing `KitRequestShipping`, you can move this method to `KitRequestShippingTest`.  By being in the separate class, we'll get two wins:
1) The test will run much much faster since it doesn't require any of the db integration test setup
2) Authors of future changes to `KitRequestShipping` will know where to look for tests that target the class

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitDisplayNameTest.java
```diff
@@ -0,0 +1,102 @@
...
+        KitRequestShipping kitWithoutDisplayName = new KitRequestShipping(guid, "TestProject_2", null, "FAKE_DSM_LABEL_UID", "study", BLOOD_RNA_KIT_TYPE_NAME, 1L, 1L,
+                "https://easypost-files.s3-us-west-2.amazonaws"
+                        + ".com/files/postage_label/20200214/8240f1b66535494a82b1ec0d566c3f0f.png", "", "794685038506",
+                "9405536897846100551129", "https://track.easypost.com/djE6dHJrXzY4NGJmYzU3ZjM5OTQ1Zjg5MjEzOGRmMWVmMjI1NWZl",
+                null, 12L, false, "", 12L, null, 12L, "so what", "mf_testLabel", false, "shp_f470591c3fb441a68dbb9b76ecf3bb3d",
+                12L, null, "44445", false, "NOT FOUND", null, null, null, null, 0L, false, "STANDALONE", null, null, null,
+                null, null, null, null, null, null, null);
+        Assert.assertEquals(kitWithoutDisplayName.getDisplayName(), BLOOD_RNA_KIT_TYPE_NAME);
```
zyme: I think the args here are reversed.  `BLOD_RNA_KIT_TYPE_NAME` is the expected value and should be the first param, right?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitDisplayNameTest.java
```diff
@@ -21,32 +21,30 @@ public class KitDisplayNameTest extends DbTxnBaseTest {
...
-        KitUtil.deleteInstanceAndSettings();
+        testKitUtil.deleteInstanceAndSettings();
 
     }
 
     @Test
-    public void testSubKitWithDisplayNAme(){
-        subkitSettingsIds = KitUtil.createSubKitsForTheStudy("BLOOD",  "BLOOD and RNA", 0, "RNA", null, 1);
-        Assert.assertEquals(2, subkitSettingsIds.length);
+    public void testKitWithDisplayNAme(){
```
pegahtah: This test is working with the `createNonPepperTestKit` which is what is used for creating a juniper-likekit. Juniper Kit Creation is not set to use sub kits yet so instead I am just testing a single generic kit with a different display name and kit type name to make sure the kit returned has both values of kitTypeName and DisplayName set correctly

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: You're correct, this is a user error most of the times, so basically no we don't necessarily need to log at this level, but changing this is not in the scope of this ticket. There are a lot of these instances in DSM, I think if we want to change that I'd say this class alone needs it's own ticket based on the number of changes and checks we have to make.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: Correct, we can maybe discuss this more and make a plan for refactoring kit-related classes to do better checks?

pepper-apis/dsm-core/src/main/resources/liquibase/rgp/PEPPER-833_rgp_kit_display_name.xml
pegahtah: I think I'm confused,  since two kit entries in the `kit_type` table with different `kit_type_id`s could have the same kit_type_name which will pass the unique condition of `(ddp_instance_id,kit_type_id)` but would still be same kit type names

zyme: I am wondering whether there is an opportunity here to add a constraint that will avoid accidental inserts down the road.  Should we prevent having multiple rows in `ddp_kit_request_settings` that reference the same ddp_instance and the same kit_type?  In dev and prod, `ddp_kit_request_settings` does not have any duplication of rows that have the same kit_type_id and ddp_instance_id.  In this PR, I don't see any inserts that would create duplicates, so I'm wondering if we have a use case that requires duplicates.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Hey Dennis, I am trying to catch up with this comment but I can't understand what you mean, sorry but can you let me know if this is resolved?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitRequestShippingTest.java
```diff
@@ -0,0 +1,44 @@
+package org.broadinstitute.dsm.kits;
+
+import org.broadinstitute.dsm.db.KitRequestShipping;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class KitRequestShippingTest {
+    private static String BLOOD_RNA_KIT_TYPE_NAME = "BLOOD";
```
zyme: These could be public, and then then `KitDisplayNameTest` could reference them as well so we don't copy-pasta them across test cases.

zyme (approved):
Let's get this onto the RC!  I have some suggestions for how we can improve this in subsequent PRs and next sprint, but I think this can go as-is as long as we follow up in the coming days.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
denniscunningham: The test above `stmt != null` will always be true, and the else clause will never be hit.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -295,30 +296,33 @@ public void invalidKitTypeTest() {
...
-        Assert.assertEquals(KitResponse.ErrorMessage.UNKNOWN_KIT_TYPE, kitResponse.getErrorMessage());
-        Assert.assertEquals(kitResponse.getValue(), kitType);
-        List<KitRequestShipping> newKits = KitRequestShipping.getKitRequestsByRealm(instanceName, "overview", kitType);
-        Assert.assertEquals(newKits.size(), oldkits.size());
-
+        try {
+            KitRequestShipping.getKitRequestsByRealm(instanceName, "overview", kitType);
+        } catch (DsmInternalError e) {
+            Assert.assertTrue(e.getMessage().contains(String.format("No kit was found for study %s with kit type name %s", instanceName, kitType)));
+            String json = "{ \"firstName\":\"P\","
```
denniscunningham, line 304: I'm not sure what's going on here. In response to an error, the test creates a new Juniper Kit?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -295,30 +296,33 @@ public void invalidKitTypeTest() {
...
+
+            JuniperKitRequest juniperTestKit = new Gson().fromJson(json, JuniperKitRequest.class);
+            createdKitIds.add(juniperTestKit.getJuniperKitId());
+            KitResponse kitResponse =
+                    nonPepperKitCreationService.createNonPepperKit(juniperTestKit, kitType, mockEasyPostUtil, ddpInstance);
+            Assert.assertEquals(KitResponse.ErrorMessage.UNKNOWN_KIT_TYPE, kitResponse.getErrorMessage());
+            Assert.assertEquals(kitResponse.getValue(), kitType);
+            return;
+        }
+        Assert.fail();
```
denniscunningham: This fail should be in the try block above, and should add a message about the failure.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -0,0 +1,479 @@
+package org.broadinstitute.dsm.juniperkits;
```
denniscunningham, line 1: Is this test utility specific to Juniper kits? IOW, can it be used to support non-Juniper kits?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -0,0 +1,479 @@
...
+ * <p>
+ * The usage is by first creating an instance of this class by declaring the desired instance name and study guid,
+ * display name and collaborator prefix.
+ * Then call setupStudyWithKitsInstanceAndSettings() for initiating all the config in database.
+ * <p>
+ * When done, call deleteJuniperInstanceAndSettings() to delete everything
+ */
+
+@Slf4j
+public class TestKitUtil {
```
denniscunningham, line 58: Without getting into the specific methods here, in general tests should create entities using the core services, rather than creating them using separate means as is done here. This assures there is no drift from the service implementation, and has the benefit of not requiring separate maintenance of services and tests, as well as testing the service creation systems as a byproduct.

denniscunningham (commented):
I'm stuck on the new tests added since I believe they should be built with existing services, not with a separate system of direct DB creation and deletion.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
pegahtah: This is an old test, but yes, I first try to get BLOOD kits for a study that is not configured to have one, so DSM throws an error for it. So here I make sure that the assumption that each study gets to have only one kit type with a certain name is getting checked here. After that I try to create a kit with that type that is not set for this study and I check that the DSM response to that is an ERROR of  `UNKNOWN_KIT_TYPE`

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
pegahtah: This was my first goal, but basically changing DSM so we can do that means changing around 20 classes and refactoring lots of different kit classes just to create a kit through the upload, which was not worth this ticket. So I am just using the method of Juniper Kit Creation just as a way to create the kit.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
pegahtah: Agreed! But unfortunately most of the DSM kit ( or anything) creation starts at the route level, which requires lots of code changes to be able to get called in a junit test. Since there is no `KitCreationService` or `StudySetUp` or `ConfigureKitService` kind of classes in DSM right now, adding things to DB and removing them every time is the easiest way. I think you know that I would love to change that in the (hopefully) near future, but that change is very big. Here I first tried refactoring the kit upload route into service classes and response enums and the code grew exponentially out of the scope of this ticket.

denniscunningham (approved):

PR #2674, opened 2023-09-11
[DSM] PEPPER-1123 Mercury error fix

pegahtah:
## Context

Fixing DSM nacking faulty pubsub messages as well as large data not fitting in the database.
Also this PR contains lots of Sonar clean ups 

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -134,7 +134,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
     }
 
     public Map<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String ddpParticipantId) {
-        String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
+        String errorMessage = "Error getting eligible barcodes for a mercury order for participant "+ ddpParticipantId;
         HashMap<String, MercuryOrderDto> map = new HashMap<>();
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
@@ -158,8 +158,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
         });
 
         if (results.resultException != null) {
-            throw new DsmInternalError(String.format(errorMessage, ddpParticipantId),
-                    results.resultException);
+            throw new DsmInternalError(errorMessage, results.resultException);
         }
 
         results = inTransaction((conn) -> {
@@ -175,8 +174,6 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                         log.info("found related info about barcode " + mercuryOrderDto.getBarcode());
                         map.put(mercuryOrderDto.getBarcode(), mercuryOrderDto);
                     }
-                } catch (SQLException e) {
-                    throw new DsmInternalError(String.format(errorMessage, ddpParticipantId), e);
                 }
             } catch (SQLException ex) {
                 dbVals.resultException = ex;
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -67,10 +67,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             "UPDATE ddp_mercury_sequencing SET order_status = ?, status_date = ?, mercury_pdo_id = ?,  "
                     + " status_detail = ?, status_message = ? where order_id = ?";
 
-    String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
-
     private static final int MAX_CHAR_ALLOWED = 2000;
-
 
     public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) {
         long statusDate = System.currentTimeMillis();
@@ -137,6 +134,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
     }
 
     public Map<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String ddpParticipantId) {
+        String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
         HashMap<String, MercuryOrderDto> map = new HashMap<>();
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
@@ -219,8 +217,6 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                     } else {
                         dbVals.resultValue = false;
                     }
-                } catch (SQLException exception) {
-                    dbVals.resultException = exception;
                 }
             } catch (SQLException ex) {
                 dbVals.resultException = ex;
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -138,7 +138,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_KIT_FROM_BARCODE_KIT_LABEL)) {
                 stmt.setString(1, ddpParticipantId);
                 getKitsFromStatement(stmt, map);
-            } catch (Exception ex) {
+            } catch (SQLException ex) {
                 dbVals.resultException = ex;
             }
 
@@ -154,7 +154,18 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_KIT_FROM_BARCODE_SM_ID)) {
                 stmt.setString(1, ddpParticipantId);
-                getBarcodesFromResultSet(stmt, map, ddpParticipantId);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    while (rs.next()) {
+                        MercuryOrderDto mercuryOrderDto = new MercuryOrderDto(rs.getString(DBConstants.DDP_PARTICIPANT_ID),
+                                rs.getString(DBConstants.MERCURY_ORDER_CREATOR), rs.getString(DBConstants.SM_ID_VALUE),
+                                rs.getInt("ktype." + DBConstants.KIT_TYPE_ID), rs.getInt(DBConstants.DDP_INSTANCE_ID),
+                                rs.getLong(DBConstants.TISSUE_ID), null);
+                        log.info("found related info about barcode " + mercuryOrderDto.getBarcode());
+                        map.put(mercuryOrderDto.getBarcode(), mercuryOrderDto);
+                    }
+                } catch (SQLException e) {
+                    throw new DsmInternalError(String.format(errorMessage, ddpParticipantId), e);
+                }
             } catch (SQLException ex) {
                 dbVals.resultException = ex;
             }
@@ -168,22 +179,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
         return map;
     }
 
-    private void getBarcodesFromResultSet(PreparedStatement stmt, HashMap<String, MercuryOrderDto> map, String ddpParticipantId) throws DsmInternalError {
-        try (ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                MercuryOrderDto mercuryOrderDto = new MercuryOrderDto(rs.getString(DBConstants.DDP_PARTICIPANT_ID),
-                        rs.getString(DBConstants.MERCURY_ORDER_CREATOR), rs.getString(DBConstants.SM_ID_VALUE),
-                        rs.getInt("ktype." + DBConstants.KIT_TYPE_ID), rs.getInt(DBConstants.DDP_INSTANCE_ID),
-                        rs.getLong(DBConstants.TISSUE_ID), null);
-                log.info("found related info about barcode " + mercuryOrderDto.getBarcode());
-                map.put(mercuryOrderDto.getBarcode(), mercuryOrderDto);
-            }
-        } catch (Exception e) {
-            throw new DsmInternalError(String.format(errorMessage, ddpParticipantId), e);
-        }
-    }
-
-    private void getKitsFromStatement(PreparedStatement stmt, HashMap<String, MercuryOrderDto> map) throws Exception{
+    private void getKitsFromStatement(PreparedStatement stmt, HashMap<String, MercuryOrderDto> map) throws SQLException {
         try (ResultSet rs = stmt.executeQuery()) {
             while (rs.next()) {
                 MercuryOrderDto mercuryOrderDto = new MercuryOrderDto(rs.getString(DBConstants.DDP_PARTICIPANT_ID),
@@ -216,7 +212,15 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement selectKitRequest = conn.prepareStatement(SQL_SELECT_ORDER_NUMBER)) {
                 selectKitRequest.setString(1, orderNumber);
-                dbVals.resultValue = checkIfOrderNumberExists(selectKitRequest);
+                try (ResultSet rs = selectKitRequest.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = true;
+                    } else {
+                        dbVals.resultValue = false;
+                    }
+                } catch (SQLException exception){
+                    dbVals.resultException = exception;
+                }
             } catch (Exception ex) {
                 dbVals.resultException = ex;
             }
@@ -227,16 +231,6 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             throw new DsmInternalError("Error checking if values exist in db", results.resultException);
         }
         return (boolean) results.resultValue;
-    }
-
-    private Boolean checkIfOrderNumberExists(PreparedStatement selectKitRequest) throws SQLException {
-        try (ResultSet rs = selectKitRequest.executeQuery()) {
-            if (rs.next()) {
-                return true;
-            } else {
-                return false;
-            }
-        }
     }
 
     public void insertMercuryOrders(List<MercuryOrderDto> newOrders, String json) {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -70,10 +70,15 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
 
     String errorMessage = "Error getting possible mercury orders for participant %s";
 
+    private static final int MAX_CHAR_ALLOWED = 2000;
+
 
     public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) throws DSMPubSubException {
         long statusDate = System.currentTimeMillis();
-        String messageString = msgData.length() > 2000 ? msgData.substring(0, 1999) : msgData;
+        //check the length of fields to make sure they fit in the database
+        String messageString = msgData.substring(0, Math.min(msgData.length(), MAX_CHAR_ALLOWED));
+        String details = baseMercuryStatusMessage.getStatus().getDetails();
+        String statusDetail = details.substring(0, Math.min(details.length(), MAX_CHAR_ALLOWED));
         AtomicReference<ClinicalOrder> clinicalOrderAtomicReference = new AtomicReference<>();
         SimpleResult results = TransactionWrapper.inTransaction(conn -> {
             SimpleResult dbVals = new SimpleResult();
@@ -82,7 +87,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                 stmt.setString(1, mercuryStatusMessage.getOrderStatus());
                 stmt.setLong(2, statusDate);
                 stmt.setString(3, mercuryStatusMessage.getPdoKey());
-                stmt.setString(4, mercuryStatusMessage.getDetails());
+                stmt.setString(4, statusDetail);
                 stmt.setString(5, messageString);
                 stmt.setString(6, mercuryStatusMessage.getOrderID());
                 int result = stmt.executeUpdate();
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -97,13 +97,16 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                                 baseMercuryStatusMessage.getStatus().getJson(), baseMercuryStatusMessage.getStatus().getOrderID()));
                     }
                 } else {
-                    dbVals.resultException = new RuntimeException(
+                    dbVals.resultException = new DsmInternalError(
                             "Error updating Mercury status for order id " + mercuryStatusMessage.getOrderID()
                                     + " it was updating 0 rows");
                 }
             } catch (SQLException ex) {
                 dbVals.resultException =
-                        new RuntimeException("Error updating Mercury status for order id " + mercuryStatusMessage.getOrderID(), ex);
+                        new DsmInternalError(String.format(
+                                "Error encountered while updating Mercury status for order id %s, status message received was %s",
+                                        mercuryStatusMessage.getOrderID(), baseMercuryStatusMessage), ex);
+                return dbVals;
             }
             try (PreparedStatement stmt = conn.prepareStatement(SQL_SELECT_LAST_UPDATED_ORDER)) {
                 stmt.setString(1, mercuryStatusMessage.getOrderID());
@@ -121,12 +124,14 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                 }
             } catch (SQLException ex) {
                 dbVals.resultException =
-                        new DsmInternalError(String.format("Couldn't find an order based on the order id %s, ES update will fail", mercuryStatusMessage.getOrderID()), ex);
+                        new DsmInternalError(String.format(
+                                "Couldn't find an order based on the order id %s, ES update will fail. Status message received was %s",
+                                mercuryStatusMessage.getOrderID(), baseMercuryStatusMessage), ex);
             }
             return dbVals;
         });
         if (results.resultException != null) {
-            throw new DsmInternalError("Unable to process the status of the order " + baseMercuryStatusMessage, results.resultException);
+            throw (DsmInternalError) results.resultException;
         }
         MercuryOrderUseCase.exportStatusToES(baseMercuryStatusMessage, clinicalOrderAtomicReference.get(), statusDate);
     }
@@ -214,7 +219,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                     } else {
                         dbVals.resultValue = false;
                     }
-                } catch (SQLException exception){
+                } catch (SQLException exception) {
                     dbVals.resultException = exception;
                 }
             } catch (SQLException ex) {
@@ -224,7 +229,8 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
         });
 
         if (results.resultException != null) {
-            throw new DsmInternalError(String.format("Error checking if mercury order number {} exist in db", orderNumber), results.resultException);
+            throw new DsmInternalError(String.format("Error checking if mercury order number %s exist in db", orderNumber),
+                    results.resultException);
         }
         return (boolean) results.resultValue;
     }
@@ -277,7 +283,8 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
 
         });
         if (results.resultException != null) {
-            throw new DsmInternalError(String.format("Error inserting mercury order for order with json {}", messageJson), results.resultException);
+            throw new DsmInternalError(String.format("Error inserting mercury order for order with json {}", messageJson),
+                    results.resultException);
         }
         return (int) results.resultValue;
     }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -67,12 +67,12 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             "UPDATE ddp_mercury_sequencing SET order_status = ?, status_date = ?, mercury_pdo_id = ?,  "
                     + " status_detail = ?, status_message = ? where order_id = ?";
 
-    String errorMessage = "Error getting possible mercury orders for participant %s";
+    String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
 
     private static final int MAX_CHAR_ALLOWED = 2000;
 
 
-    public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) throws DsmInternalError {
+    public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) {
         long statusDate = System.currentTimeMillis();
         //check the length of fields to make sure they fit in the database
         String messageString = msgData.substring(0, Math.min(msgData.length(), MAX_CHAR_ALLOWED));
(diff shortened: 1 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -121,69 +132,74 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error getting the mercury orders", results.resultException);
+            throw new DsmInternalError("Error getting the mercury orders", results.resultException);
         }
         log.info(String.format("Found %d possible barcodes for participant %s", map.size(), ddpParticipantId));
         return map;
     }
 
+    private void getBarcodesFromResultSet(PreparedStatement stmt, HashMap<String, MercuryOrderDto> map, String ddpParticipantId) throws DsmInternalError {
```
denniscunningham: Why did you break these out of their prior methods? Was the linter complaining about something? With the change it's harder to associate the query with the statement result. Perhaps another way to do this is to break out just the part that creates the `MercuryOrderDto`? 

I'm also a bit concerned about changing anything unneeded unless there are automated tests covering at least the main code path. Do we have those for these?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -201,28 +217,29 @@ public boolean orderNumberExists(String orderNumber) {
...
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error checking if values exist in db", results.resultException);
+            throw new DsmInternalError("Error checking if values exist in db", results.resultException);
         }
         return (boolean) results.resultValue;
     }
 
+    private Boolean checkIfOrderNumberExists(PreparedStatement selectKitRequest) throws SQLException {
```
denniscunningham: This change seems to lessen the readability. Was there a problem reported by the linter?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,32 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
+@Slf4j
 public class MercuryOrderStatusListener {
-    private static final Logger logger = LoggerFactory.getLogger(MercuryOrderStatusListener.class);
-
+    private MercuryOrderStatusListener(){
+        throw new IllegalStateException("Utility class");
+    }
```
denniscunningham, line 25: The linter should say that you need a blank line here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,32 @@
...
         MessageReceiver receiver = (PubsubMessage message, AckReplyConsumer consumer) -> {
+            String messageId = message.getMessageId();
             // Handle incoming message, then ack the received message.
-            logger.info("Got STATUS message with Id: " + message.getMessageId());
-
+            log.info(String.format("Got STATUS message with Id: %s", messageId));
             try {
                 processOrderStatus(message);
-                logger.info("Processing the message finished");
+                log.info("Processing the message finished");
```
denniscunningham: My suggestion was to ack *before* the call to `processOrderStatus`. The only valid case for nack is to have the exact same message sent again. Will that help in some way here?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,12 +117,12 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
                             null, statusDate, null,
                             rs.getLong(DBConstants.TISSUE_ID), rs.getLong(DBConstants.DSM_KIT_REQUEST_ID), null));
                 } else {
-                    dbVals.resultException = new RuntimeException(
-                            "Couldn't get ddp instance id for order " + mercuryStatusMessage.getOrderID());
+                    dbVals.resultException = new DSMPubSubException(String.format("Couldn't get ddp instance id for order %s",
```
denniscunningham: I don't think DSMPubSubException is correct here. That implies something went wrong with PubSub per se. I think DsmInternalError is correct here. Likewise in the other places where you added it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -121,69 +132,74 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
-                                null, rs.getLong("kit." + DBConstants.DSM_KIT_REQUEST_ID));
-                        log.info("found related info about barcode " + mercuryOrderDto.getBarcode());
-                        map.put(mercuryOrderDto.getBarcode(), mercuryOrderDto);
-                    }
-                } catch (Exception e) {
-                    throw new RuntimeException("Error getting possible mercury orders for participant " + ddpParticipantId, e);
-                }
-            } catch (SQLException ex) {
+                getKitsFromStatement(stmt, map);
+            } catch (Exception ex) {
```
denniscunningham: I think we still want to catch SQLException here, not Exception. This may just lead to stacking DsmInternalErrors on top of each other.

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: Yeah I tried to reduce linter errors, sonar was really not happy with nested tries, however, I really hate these changes myself. Makes it less readable, and I wasn't sure what to put in for java docs. I think I'll just move them back and just try to fix it in the tries

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -65,9 +67,17 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             "UPDATE ddp_mercury_sequencing SET order_status = ?, status_date = ?, mercury_pdo_id = ?,  "
                     + " status_detail = ?, status_message = ? where order_id = ?";
 
+    String errorMessage = "Error getting possible mercury orders for participant %s";
```
denniscunningham: Nit: The term 'possible' seems unnecessary.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -65,9 +67,17 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
...
 
-    public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) throws Exception {
+    private static final int MAX_CHAR_ALLOWED = 2000;
+
+
+    public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusMessage, String msgData) throws DsmInternalError {
```
denniscunningham: This `throws` clause is unnecessary.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -65,9 +67,17 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
...
         long statusDate = System.currentTimeMillis();
+        //check the length of fields to make sure they fit in the database
+        String messageString = msgData.substring(0, Math.min(msgData.length(), MAX_CHAR_ALLOWED));
```
denniscunningham, line 75: 2000 seems generous for the message, esp since we provide a large details section.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,39 +116,28 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
                             null, statusDate, null,
                             rs.getLong(DBConstants.TISSUE_ID), rs.getLong(DBConstants.DSM_KIT_REQUEST_ID), null));
                 } else {
-                    dbVals.resultException = new RuntimeException(
-                            "Couldn't get ddp instance id for order " + mercuryStatusMessage.getOrderID());
+                    dbVals.resultException = new DsmInternalError(String.format("Couldn't get ddp instance id for order %s",
```
denniscunningham: Is this actually a user input error? Where does the order ID come from?
Also, is this error text correct? How does the DDP instance figure in here?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,39 +116,28 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
+                            mercuryStatusMessage.getOrderID()));
                 }
             } catch (SQLException ex) {
                 dbVals.resultException =
-                        new RuntimeException("Error getting ddp instance id for order " + mercuryStatusMessage.getOrderID(), ex);
+                        new DsmInternalError(String.format("Error getting ddp instance id for order %s", mercuryStatusMessage.getOrderID()), ex);
```
denniscunningham: Is this error text correct? How does the DDP instance figure in here?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,39 +116,28 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
             }
             return dbVals;
         });
         if (results.resultException != null) {
-            throw new DSMPubSubException("Unable to process the status of the order " + baseMercuryStatusMessage, results.resultException);
+            throw new DsmInternalError("Unable to process the status of the order " + baseMercuryStatusMessage, results.resultException);
```
denniscunningham: There is no need to stack the exceptions here. Simply throwing the existing exception is sufficient.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/DSMPubSubException.java
```diff
@@ -4,4 +4,8 @@ public class DSMPubSubException extends RuntimeException {
     public DSMPubSubException(String message, Throwable cause) {
         super(message, cause);
     }
+
+    public DSMPubSubException(String message) {
```
denniscunningham: Probably unused at this point?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
denniscunningham: Does not look like you added one?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,33 @@
...
+        throw new IllegalStateException("Utility class");
+    }
     public static void subscribeToOrderStatus(String projectId, String subscriptionId) {
         // Instantiate an asynchronous message receiver.
         MessageReceiver receiver = (PubsubMessage message, AckReplyConsumer consumer) -> {
+            String messageId = message.getMessageId();
             // Handle incoming message, then ack the received message.
-            logger.info("Got STATUS message with Id: " + message.getMessageId());
-
+            log.info(String.format("Got STATUS message with Id: %s", messageId));
```
denniscunningham: You should use the `{}` string replacement provided by the logger.
Some context in the message would be helpful. Which status does this refer to?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,33 @@
...
             try {
+                consumer.ack();
                 processOrderStatus(message);
-                logger.info("Processing the message finished");
+                log.info("Processing the message finished");
```
denniscunningham: Some context on which message processing would be helpful.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,33 @@
...
+            } catch (DSMPubSubException error) {
+                log.error("Error happened parsing Mercury Status Message, DSM will ack the message", error);
                 consumer.ack();
```
denniscunningham: This ack is now unnecessary.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,26 +11,33 @@
...
+                error.printStackTrace();
             } catch (Exception ex) {
-                logger.info("about to nack the message", ex);
+                log.error("Unexpected error: about to nack the message", ex);
                 consumer.nack();
```
denniscunningham: The error message should have some context about which message (not necessarily the content, but the message type, etc.).
This nack is now unnecessary.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -43,13 +50,13 @@ public static void subscribeToOrderStatus(String projectId, String subscriptionI
...
-            logger.info("Started pubsub subscription receiver for mercury order status subscription");
+            log.info("Started pubsub subscription receiver for mercury order status subscription");
         } catch (TimeoutException e) {
-            throw new RuntimeException("Timed out while starting pubsub subscription for mercury order status", e);
+            throw new DSMPubSubException("Timed out while starting pubsub subscription for mercury order status", e);
         }
     }
 
-    private static void processOrderStatus(PubsubMessage message) throws Exception {
+    private static void processOrderStatus(PubsubMessage message) throws DSMPubSubException {
```
denniscunningham: Does this actually throw `DSMPubSubException`?

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: So here messageString is the json we received from the Mercury. We keep that so if for any reason we miss the fields or something goes wrong (like the current error that was happening) we can always trace back to what was the actual message that we received. If it comes with a large details section we even need to keep more. 
I looked at the last huge message we received and that was too big for our columns that are set to TEXT type, so I thought 2000 would be a good balance. But I am more than willing to change. What number were you thinking?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: No it's not! great catch!! Not sure how that made it's way to here.... I will change it!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: I think someone did a wrong copy paste here long ago...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
pegahtah: But sonar tells me to not use `{}` :lolsob

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: If we don't stack it, it will just be treated as `Exception` in the functions that call it, so it needs to get added to the method signature. Is that really what we want?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
denniscunningham: I probably do not have enough information...
What is the expected content for each of the fields? Are the two fields related? 
From the field names I was guessing that the message would be brief human readable text (something that could be easily displayed) and the details were audit level details that would typically not be displayed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,39 +116,28 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
                 } else {
-                    dbVals.resultException = new RuntimeException(
-                            "Couldn't get ddp instance id for order " + mercuryStatusMessage.getOrderID());
+                    dbVals.resultException = new DsmInternalError(String.format("Couldn't find an order based on the order id %s",
+                            mercuryStatusMessage.getOrderID()));
                 }
             } catch (SQLException ex) {
                 dbVals.resultException =
-                        new RuntimeException("Error getting ddp instance id for order " + mercuryStatusMessage.getOrderID(), ex);
+                        new DsmInternalError(String.format("ECouldn't find an order based on the order id %s, ES update will fail", mercuryStatusMessage.getOrderID()), ex);
```
denniscunningham: Typo.
Also, I do not think the ES update will fail -- it just will not be performed. But in this case ES is still in sync with the DB, so is that information misleading? It seems to imply that something has to be done to get ES in sync with the DB.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
denniscunningham: Sorry, I was not clear...
On line 124 and 119 you already created exceptions, and here you are creating another and putting the first exception as its cause. You only need one exception here, and you most likely do not want to include the `baseMercuryStatusMessage` in the exception message. So one solution is to just throw `resultException` here without creating a redundant exception.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -164,26 +163,38 @@ public HashMap<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String
...
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error getting the mercury orders", results.resultException);
+            throw new DsmInternalError("Error getting the mercury orders", results.resultException);
         }
         log.info(String.format("Found %d possible barcodes for participant %s", map.size(), ddpParticipantId));
         return map;
     }
 
+    private void getKitsFromStatement(PreparedStatement stmt, HashMap<String, MercuryOrderDto> map) throws SQLException {
```
denniscunningham: I'm confused about this refactoring, since it seems to separate code at an odd boundary. It does not separate concerns well, but rather separates the statement execution from  statement creation, which creates an unnecessary break in the logic. The two solutions I can think of are:

1. Do not refactor. Leave the code as it was.
2. Pass in the result set rather than the statement.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -207,18 +218,17 @@ public boolean orderNumberExists(String orderNumber) {
...
                         dbVals.resultValue = false;
                     }
-
-                } catch (Exception e) {
-                    throw new RuntimeException("Error getting values from db", e);
+                } catch (SQLException exception){
+                    dbVals.resultException = exception;
                 }
-            } catch (SQLException ex) {
+            } catch (Exception ex) {
```
denniscunningham: This catch block is unnecessary, and does not follow the pattern we typically use for DSM DB queries.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -207,18 +218,17 @@ public boolean orderNumberExists(String orderNumber) {
...
                 dbVals.resultException = ex;
             }
             return dbVals;
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error checking if values exist in db", results.resultException);
+            throw new DsmInternalError("Error checking if values exist in db", results.resultException);
```
denniscunningham: This is a very generic message that does not help with debugging. A message saying that you could not find an Mercury order number in the DB, which includes the order number, would be better.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -271,7 +281,7 @@ public int create(MercuryOrderDto mercuryOrderDto, String messageJson) {
 
         });
         if (results.resultException != null) {
-            throw new RuntimeException("Error inserting mercury order ", results.resultException);
+            throw new DsmInternalError("Error inserting mercury order ", results.resultException);
```
denniscunningham: As above, including the Mercury order ID would help with debugging and resolution.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,27 +11,33 @@
...
                 consumer.ack();
+                processOrderStatus(message);
+                log.info("Processing the status message from Mercury finished");
+            } catch (DSMPubSubException error) {
+                log.error("Error happened parsing Mercury Status Message, DSM will ack the message", error);
+                error.printStackTrace();
             } catch (Exception ex) {
-                logger.info("about to nack the message", ex);
-                consumer.nack();
+                log.error("Unexpected error: about to nack the status message from Mercury", ex);
```
denniscunningham: This message is now incorrect.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
```diff
@@ -11,27 +11,33 @@
...
-
+            log.info(String.format("Got Mercury status message with Id: {}", messageId));
             try {
-                processOrderStatus(message);
-                logger.info("Processing the message finished");
                 consumer.ack();
+                processOrderStatus(message);
+                log.info("Processing the status message from Mercury finished");
+            } catch (DSMPubSubException error) {
+                log.error("Error happened parsing Mercury Status Message, DSM will ack the message", error);
```
denniscunningham: Are we still throwing DSMPubSubException in ProcessOrderStatus?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: The `msgData` is the whole json of the message that was received, and `baseMercuryStatusMessage` is the part that is related to this order and is parsed into a class.  `details` is just a field in there, which has exactly what you said, audit level details. 

We keep the message that was received in its original form for 2 reasons

1. in case DSM or Mercury break we don't want messages to get lost after acking pubsub topic
2. in case we need to go back and look at the structure of what was received (there were cases when we got concatenated messages from past added to the current message and everything was breaking and DSM was blamed for it, and having a record of what was passed to pubsub and what was received really helped )

In this current error, details was a huge string and message size grew because of that, but there were cases (like the example in the second bullet point ) where message was huge but details had a normal size.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: This was a sonar suggestion that I will remove

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: I understand now, I think given how it has been in the process of debugging errors on either side of Mercury and DSM it is going to be beneficial for us to log `baseMercuryStatusMessage` ( it does not contain PHI, just some barcode values and the error that happened ). We can log it to see what they sent back. Especially since sometimes they apparently send a huge string that we don't want to keep all of it, it's good to have a sense of what was happening.

denniscunningham: It does not look like you addressed the exception stacking issue?
Also, looking at line 104, do we want to continue if we encounter an error? Also, the `resultException` from this clause will just get overwritten by any errors in the following clause.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -271,7 +277,7 @@ public int create(MercuryOrderDto mercuryOrderDto, String messageJson) {
 
         });
         if (results.resultException != null) {
-            throw new RuntimeException("Error inserting mercury order ", results.resultException);
+            throw new DsmInternalError(String.format("Error inserting mercury order for order with json {}", messageJson), results.resultException);
```
denniscunningham: I trying to think about the best information to provide for troubleshooting and follow-up... Would the order ID be helpful? Perhaps more helpful than the message? Or should they both be provided?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/MercuryOrderStatusListener.java
denniscunningham: I think you can just get rid of this clause `catch (DsmInternalError...`, since the handling for any exception here should be the same. 
Also, I think you are getting a lint warning about the call to `printStackTrace`. You can remove it since the `log.error` call will provide a stack trace.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: order id and all of the other fields would be present in the json.

pegahtah: basically json is the message we want to publish to the topic. so if there was any errors with creating that order everything would be in the json

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: I fixed L104 and pushed again, I didn't change the stacking before because I thought the main reason for the stacking looking unnecessary to you was the `baseMercuryStatusMessage`  in there, which I think we should log. So now I moved the `baseMercuryStatusMessage` to when the original exception is created and removed the stacking

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -106,22 +119,24 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
...
-                        new RuntimeException("Error getting ddp instance id for order " + mercuryStatusMessage.getOrderID(), ex);
+                        new DsmInternalError(String.format(
+                                "Couldn't find an order based on the order id %s, ES update will fail. Status message received was %s",
+                                mercuryStatusMessage.getOrderID(), baseMercuryStatusMessage), ex);
             }
             return dbVals;
         });
         if (results.resultException != null) {
-            throw new DSMPubSubException("Unable to process the status of the order " + baseMercuryStatusMessage, results.resultException);
+            throw (DsmInternalError) results.resultException;
```
denniscunningham, line 131: I assume this cast is unnecessary? Did something suggest you add it?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -65,9 +67,17 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
             "UPDATE ddp_mercury_sequencing SET order_status = ?, status_date = ?, mercury_pdo_id = ?,  "
                     + " status_detail = ?, status_message = ? where order_id = ?";
 
+    String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
```
denniscunningham: Can you move this error message to the method where it is used? (Looks like it is only used in one place in this file.) When you do that, you can add the DdpParticipantId when it is declared, which makes the code easier to read and lessens the likelihood of mistakes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -207,9 +219,8 @@ public boolean orderNumberExists(String orderNumber) {
                     } else {
                         dbVals.resultValue = false;
                     }
-
-                } catch (Exception e) {
-                    throw new RuntimeException("Error getting values from db", e);
+                } catch (SQLException exception) {
```
denniscunningham: This is unnecessary since you are catching the same exception right below. Also, this method can be simplified by directly returning the results of the transaction, and not using SimpleResult, which does not add value. (The latter is a nit, that is not required for this PR.)

denniscunningham (commented):
Looks good!
A couple more comments.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: Yes otherwise we have to add `@sneakythrows` or throws Exception to method signature

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -137,6 +134,7 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
     }
 
     public Map<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String ddpParticipantId) {
+        String errorMessage = "Error getting eligible barcodes for a mercury order for participant %s";
```
denniscunningham: Sorry, I was not clear. You can add the ddpParticipantId right here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -164,26 +175,25 @@ public HashMap<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String
                         log.info("found related info about barcode " + mercuryOrderDto.getBarcode());
                         map.put(mercuryOrderDto.getBarcode(), mercuryOrderDto);
                     }
-                } catch (Exception e) {
-                    throw new RuntimeException("Error getting possible mercury orders for participant " + ddpParticipantId, e);
+                } catch (SQLException e) {
```
denniscunningham: Sorry, missed this one. This catch is also unnecessary.

denniscunningham (commented):
Looking good!
Just a couple last things.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: Oh sorry I misunderstood!

denniscunningham (approved):

PR #2677, opened 2023-09-18
[DSM] PEPPER-1058

pegahtah:
## Context
PEPPER-1058

I have added 3 extra fields in for the status response and tests to test them

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
@@ -1,0 +1,13 @@
+package org.broadinstitute.dsm.db.dao.kit;
+
+import lombok.Getter;
+
+@Getter
+public enum KitCurrentStatus {
+    QUEUE("Queue"), ERROR("Error"), RECEIVED("Received"), SENT("Sent"), KIT_WITHOUT_LABEL("Kit Without Label"), DEACTIVATED("Deactivated");
+    private final String value;
+
+    KitCurrentStatus(String name) {
+        this.value = name;
+    }
+}
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
@@ -4,10 +4,16 @@ import lombok.Getter;
 
 @Getter
 public enum KitCurrentStatus {
-    QUEUE("Queue"), ERROR("Error"), RECEIVED("Received"), SENT("Sent"), KIT_WITHOUT_LABEL("Kit Without Label"), DEACTIVATED("Deactivated");
+    QUEUE("Queue"),
+    ERROR("Error"),
+    RECEIVED("Received"),
+    SENT("Sent"),
+    KIT_WITHOUT_LABEL("Kit Without Label"),
+    DEACTIVATED("Deactivated");
     private final String value;
 
     KitCurrentStatus(String name) {
         this.value = name;
     }
+
 }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
@@ -129,7 +129,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
     }
 
     private static class BuildNonPepperKitStatusDto {
-        public static KitCurrentStatus calculateCurrentStatus(ResultSet foundKitResults) {
+        private static KitCurrentStatus calculateCurrentStatus(ResultSet foundKitResults) {
             try {
                 if (isDeactivatedKit(foundKitResults)) {
                     return KitCurrentStatus.DEACTIVATED;
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
@@ -10,6 +10,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.dao.Dao;
 import org.broadinstitute.dsm.db.dto.kit.nonPepperKit.NonPepperKitStatusDto;
@@ -19,11 +21,13 @@ import org.broadinstitute.dsm.model.nonpepperkit.NonPepperStatusKitService;
 import org.broadinstitute.dsm.statics.DBConstants;
 import org.broadinstitute.lddp.db.SimpleResult;
 
+@Slf4j
 public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
 
     private static final String SELECT_KIT_STATUS =
             " SELECT req.*, k.*, discard.*, tracking.tracking_id as return_tracking_number, tracking.scan_by as tracking_scan_by, "
-                    + " tracking.scan_date as tracking_scan_date, bsp_collaborator_sample_id, bsp_collaborator_participant_id FROM ddp_kit_request req "
+                    + " tracking.scan_date as tracking_scan_date, bsp_collaborator_sample_id, bsp_collaborator_participant_id "
+                    + " FROM ddp_kit_request req "
                     + " LEFT JOIN ddp_kit k on (k.dsm_kit_request_id = req.dsm_kit_request_id) "
                     + " LEFT JOIN ddp_kit_discard discard on  (discard.dsm_kit_request_id = req.dsm_kit_request_id) "
                     + " LEFT JOIN ddp_kit_tracking tracking on  (tracking.kit_label = k.kit_label) ";
@@ -82,12 +86,13 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                     list.add(builder.build(rs, users));
                 }
             } catch (Exception ex) {
-                dbVals.resultException = new Exception(String.format("Error getting kits with juniper kit id %s", juniperKitId), ex);
+                dbVals.resultException = ex;
             }
             return dbVals;
         });
         if (simpleResult.resultException != null) {
-            throw new DsmInternalError(simpleResult.resultException);
+            throw new DsmInternalError(String.format("Error getting kits with juniper kit id %s", juniperKitId),
+                    simpleResult.resultException);
         }
         return list;
     }
@@ -124,6 +129,85 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
     }
 
     private static class BuildNonPepperKitStatusDto {
+        public static KitCurrentStatus calculateCurrentStatus(ResultSet foundKitResults) {
+            try {
+                if (isDeactivatedKit(foundKitResults)) {
+                    return KitCurrentStatus.DEACTIVATED;
+                } else if (isQueuedKit(foundKitResults)) {
+                    return KitCurrentStatus.QUEUE;
+                } else if (isErrorKit(foundKitResults)) {
+                    return KitCurrentStatus.ERROR;
+                } else if (isReceivedKit(foundKitResults)) {
+                    return KitCurrentStatus.RECEIVED;
+                } else if (isSentKit(foundKitResults)) {
+                    return KitCurrentStatus.SENT;
+                } else if (isNewKit(foundKitResults) || isEasyPostLabelTriggeredKit(foundKitResults)) {
+                    return KitCurrentStatus.KIT_WITHOUT_LABEL;
+                } else {
+                    log.error(String.format("Unable to determine the current status of kit %s",
+                            foundKitResults.getString(DBConstants.DDP_KIT_REQUEST_ID)));
+                    return null;
+                }
+            } catch (SQLException e) {
+                throw new DsmInternalError(e);
+            }
+        }
+
+        private static boolean isEasyPostLabelTriggeredKit(ResultSet foundKitResults) throws SQLException {
+            return StringUtils.isBlank(foundKitResults.getString(DBConstants.EASYPOST_TO_ID))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.LABEL_URL_TO))
+                    && StringUtils.isNotBlank(foundKitResults.getString(DBConstants.LABEL_DATE))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.ERROR)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.ERROR)))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.KIT_COMPLETE)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.KIT_COMPLETE)));
+        }
+
+        private static boolean isDeactivatedKit(ResultSet foundKitResults) throws SQLException {
+            return StringUtils.isNotBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE))
+                    && !DBConstants.DEACTIVATION_REASON.equals(foundKitResults.getString(DBConstants.DEACTIVATION_REASON));
+        }
+
+        private static boolean isNewKit(ResultSet foundKitResults) throws SQLException {
+            return StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.ERROR)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.ERROR)))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.KIT_COMPLETE)) ||
+                    !"1".equals(foundKitResults.getString(DBConstants.KIT_COMPLETE)))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.LABEL_URL_TO)));
+        }
+
+        private static boolean isReceivedKit(ResultSet foundKitResults) throws SQLException {
+            return StringUtils.isNotBlank(foundKitResults.getString(DBConstants.DSM_RECEIVE_DATE))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE));
+        }
+
+        private static boolean isSentKit(ResultSet foundKitResults) throws SQLException {
+//        and kit.kit_complete = 1 and kit.deactivated_date is null
+            return ("1".equals(foundKitResults.getString(DBConstants.KIT_COMPLETE)))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_RECEIVE_DATE))
+                    && StringUtils.isNotBlank(foundKitResults.getString(DBConstants.DSM_SCAN_DATE));
+        }
+
+        private static boolean isErrorKit(ResultSet foundKitResults) throws SQLException {
+            return (StringUtils.isBlank(foundKitResults.getString(DBConstants.KIT_COMPLETE)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.KIT_COMPLETE)))
+                    && ("1".equals(foundKitResults.getString(DBConstants.ERROR)))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE));
+        }
+
+        private static boolean isQueuedKit(ResultSet foundKitResults) throws SQLException {
+            return (StringUtils.isBlank(foundKitResults.getString(DBConstants.KIT_COMPLETE)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.KIT_COMPLETE)))
+                    && (StringUtils.isBlank(foundKitResults.getString(DBConstants.ERROR)) ||
+                    "0".equals(foundKitResults.getString(DBConstants.ERROR)))
+                    && StringUtils.isNotBlank(foundKitResults.getString(DBConstants.LABEL_URL_TO))
+                    && StringUtils.isBlank(foundKitResults.getString(DBConstants.DSM_DEACTIVATED_DATE));
+
+        }
+
         public NonPepperKitStatusDto build(ResultSet foundKitResults, Map<Integer, UserDto> users) throws DsmInternalError {
             try {
                 return new NonPepperKitStatusDto.Builder()
@@ -161,7 +245,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                                 NonPepperStatusKitService.convertTimeStringIntoTimeStamp(foundKitResults.getLong(DBConstants.DISCARD_DATE)))
                         .withDiscardBy(
                                 NonPepperStatusKitService.getUserEmailForFields(foundKitResults.getString(DBConstants.DISCARD_BY), users))
-                        .withCurrentStatus(NonPepperStatusKitService.calculateCurrentStatus(foundKitResults))
+                        .withCurrentStatus(calculateCurrentStatus(foundKitResults).getValue())
                         .withCollaboratorParticipantId(foundKitResults.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID))
                         .withCollaboratorSampleId(foundKitResults.getString(DBConstants.BSP_COLLABORATOR_SAMPLE_ID))
                         .build();
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
@@ -23,7 +23,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
 
     private static final String SELECT_KIT_STATUS =
             " SELECT req.*, k.*, discard.*, tracking.tracking_id as return_tracking_number, tracking.scan_by as tracking_scan_by, "
-                    + " tracking.scan_date as tracking_scan_date FROM ddp_kit_request req "
+                    + " tracking.scan_date as tracking_scan_date, bsp_collaborator_sample_id, bsp_collaborator_participant_id FROM ddp_kit_request req "
                     + " LEFT JOIN ddp_kit k on (k.dsm_kit_request_id = req.dsm_kit_request_id) "
                     + " LEFT JOIN ddp_kit_discard discard on  (discard.dsm_kit_request_id = req.dsm_kit_request_id) "
                     + " LEFT JOIN ddp_kit_tracking tracking on  (tracking.kit_label = k.kit_label) ";
@@ -82,7 +82,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                     list.add(builder.build(rs, users));
                 }
             } catch (Exception ex) {
-                dbVals.resultException = new Exception(String.format("Error getting kits with juniper kit id %s", juniperKitId));
+                dbVals.resultException = new Exception(String.format("Error getting kits with juniper kit id %s", juniperKitId), ex);
             }
             return dbVals;
         });
@@ -140,6 +140,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                                 NonPepperStatusKitService.convertTimeStringIntoTimeStamp(foundKitResults.getLong(DBConstants.LABEL_DATE)))
                         .withLabelByEmail(
                                 NonPepperStatusKitService.getUserEmailForFields(foundKitResults.getString(DBConstants.LABEL_BY), users))
+                        .withMfBarcode(foundKitResults.getString(DBConstants.KIT_LABEL))
                         .withScanDate(
                                 NonPepperStatusKitService.convertTimeStringIntoTimeStamp(
                                         foundKitResults.getLong(DBConstants.DSM_SCAN_DATE)))
@@ -161,6 +162,8 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                         .withDiscardBy(
                                 NonPepperStatusKitService.getUserEmailForFields(foundKitResults.getString(DBConstants.DISCARD_BY), users))
                         .withCurrentStatus(NonPepperStatusKitService.calculateCurrentStatus(foundKitResults))
+                        .withCollaboratorParticipantId(foundKitResults.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID))
+                        .withCollaboratorSampleId(foundKitResults.getString(DBConstants.BSP_COLLABORATOR_SAMPLE_ID))
                         .build();
             } catch (SQLException e) {
                 throw new DsmInternalError("Error building the NonPepperKitStatusDto object from resultSet", e);
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
@@ -160,6 +160,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
                                 NonPepperStatusKitService.convertTimeStringIntoTimeStamp(foundKitResults.getLong(DBConstants.DISCARD_DATE)))
                         .withDiscardBy(
                                 NonPepperStatusKitService.getUserEmailForFields(foundKitResults.getString(DBConstants.DISCARD_BY), users))
+                        .withCurrentStatus(NonPepperStatusKitService.calculateCurrentStatus(foundKitResults))
                         .build();
             } catch (SQLException e) {
                 throw new DsmInternalError("Error building the NonPepperKitStatusDto object from resultSet", e);
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
@@ -15,6 +15,7 @@ public class NonPepperKitStatusDto {
     private String participantId;
     private String labelDate;
     private String labelByEmail;
+    private String mfBarcode;
     private String scanDate;
     private String scanByEmail;
     private String receiveDate;
@@ -29,9 +30,8 @@ public class NonPepperKitStatusDto {
     private String discardDate;
     private String discardBy;
     private String currentStatus;
-
-    public NonPepperKitStatusDto() {
-    }
+    private String collaboratorParticipantId;
+    private String collaboratorSampleId;
 
     /**
      * Creates a NonPepperKitStatusDto from a builder
@@ -57,6 +57,9 @@ public class NonPepperKitStatusDto {
         this.discardDate = builder.discardDate;
         this.discardBy = builder.discardBy;
         this.currentStatus = builder.currentStatus;
+        this.collaboratorParticipantId = builder.collaboratorParticipantId;
+        this.collaboratorSampleId = builder.collaboratorSampleId;
+        this.mfBarcode = builder.mfBarcode;
     }
 
     public static class Builder {
@@ -67,6 +70,7 @@ public class NonPepperKitStatusDto {
         private String participantId;
         private String labelDate;
         private String labelByEmail;
+        private String mfBarcode;
         private String scanDate;
         private String scanByEmail;
         private String receiveDate;
@@ -81,7 +85,8 @@ public class NonPepperKitStatusDto {
         private String discardDate;
         private String discardBy;
         private String currentStatus;
-
+        private String collaboratorParticipantId;
+        private String collaboratorSampleId;
 
         public Builder withJuniperKitId(String juniperKitId) {
             this.juniperKitId = juniperKitId;
@@ -105,6 +110,10 @@ public class NonPepperKitStatusDto {
 
         public Builder withLabelByEmail(String labelByEmail) {
             this.labelByEmail = labelByEmail;
+            return this;
+        }
+        public Builder withMfBarcode(String mfBarcode) {
+            this.mfBarcode = mfBarcode;
             return this;
         }
 
@@ -180,6 +189,16 @@ public class NonPepperKitStatusDto {
 
         public Builder withCurrentStatus(String currentStatus) {
             this.currentStatus = currentStatus;
+            return this;
+        }
+
+        public Builder withCollaboratorParticipantId(String collaboratorParticipantId) {
+            this.collaboratorParticipantId = collaboratorParticipantId;
+            return this;
+        }
+
+        public Builder withCollaboratorSampleId(String collaboratorSampleId) {
+            this.collaboratorSampleId = collaboratorSampleId;
             return this;
         }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
@@ -28,6 +28,7 @@ public class NonPepperKitStatusDto {
     private String errorMessage;
     private String discardDate;
     private String discardBy;
+    private String currentStatus;
 
     public NonPepperKitStatusDto() {
     }
@@ -55,6 +56,7 @@ public class NonPepperKitStatusDto {
         this.errorMessage = builder.errorMessage;
         this.discardDate = builder.discardDate;
         this.discardBy = builder.discardBy;
+        this.currentStatus = builder.currentStatus;
     }
 
     public static class Builder {
@@ -78,10 +80,8 @@ public class NonPepperKitStatusDto {
         private String errorMessage;
         private String discardDate;
         private String discardBy;
+        private String currentStatus;
 
-
-        public Builder() {
-        }
 
         public Builder withJuniperKitId(String juniperKitId) {
             this.juniperKitId = juniperKitId;
@@ -175,6 +175,11 @@ public class NonPepperKitStatusDto {
 
         public Builder withDiscardBy(String discardBy) {
             this.discardBy = discardBy;
+            return this;
+        }
+
+        public Builder withCurrentStatus(String currentStatus) {
+            this.currentStatus = currentStatus;
             return this;
         }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
@@ -29,7 +29,7 @@ public class NonPepperKitStatusDto {
     private String errorMessage;
     private String discardDate;
     private String discardBy;
-    private String currentStatus;
+    private final String currentStatus;
     private String collaboratorParticipantId;
     private String collaboratorSampleId;
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/jobs/LabelCreationJob.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/jobs/LabelCreationJob.java
@@ -22,7 +22,7 @@ public class LabelCreationJob implements Job {
             if (labelCreationStarted == 0) {
                 List<KitRequestCreateLabel> kitsLabelTriggered = KitUtil.getListOfKitsLabelTriggered();
                 if (!kitsLabelTriggered.isEmpty()) {
-                    KitUtil.createLabel(kitsLabelTriggered);
+                    KitUtil.createLabel(kitsLabelTriggered, null);
                 }
             } else if (labelCreationStarted < System.currentTimeMillis() - (SystemUtil.MILLIS_PER_HOUR * 2)) {
                 logger.error("Label creation job is running for over 2 hours now...");
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
@@ -57,11 +57,24 @@ public class NonPepperKitCreationService {
 
         // TODO for future if applicable, if the kit type has sub kits check that here
 
-        return validateAndInsertNewNonPepperKit(kitRequestSettings, juniperKitRequest, easyPostUtil, ddpInstance, kitType, kitTypeName);
+        return validateAndInsertNonPepperKit(kitRequestSettings, juniperKitRequest, easyPostUtil, ddpInstance, kitType, kitTypeName);
     }
+    /**
+     * Attempts to validate the juniperKitRequest received from Juniper, get an easy post id for its address and
+     * insert in the DB if all checks pass
+     *
+     * @param kitRequestSettings KitRequestSettings that shows the specific kit config for this study;
+     * @param juniperKitRequest the new JuniperKitRequest received
+     * @param easyPostUtil an instance of the EasyPostUtil
+     * @param ddpInstance DDPInstance representing the study for which we want to create kits
+     * @param kitType the KitType that the request wants to create
+     * @param kitTypeName the specific name of the request kit type
+     *
+     * @return an instance of KitResponse, either as an error or KitStatus
+     * **/
 
-    private KitResponse validateAndInsertNewNonPepperKit(KitRequestSettings kitRequestSettings, JuniperKitRequest juniperKitRequest,
-                                                         EasyPostUtil easyPostUtil, DDPInstance ddpInstance,
+    private KitResponse validateAndInsertNonPepperKit(@NonNull KitRequestSettings kitRequestSettings, @NonNull JuniperKitRequest juniperKitRequest,
+                                                         @NonNull EasyPostUtil easyPostUtil, DDPInstance ddpInstance,
                                                          KitType kitType, String kitTypeName) {
         //let's validate the participant's address
         //to validate the address, first we need create the Address instance
@@ -123,9 +136,9 @@ public class NonPepperKitCreationService {
             juniperKitRequestId = kit.getJuniperKitId();
             userId = JUNIPER;
         } else {
-            log.warn("Seems like {} is not configured as a JUNIPER study! ", ddpInstance.getName());
+            log.warn("Seems like {} is not configured as a Juniper study! ", ddpInstance.getName());
             transactionResults.resultException =
-                    new DsmInternalError(ddpInstance.getName() + " study is not configured to use this method.");
+                    new DsmInternalError(ddpInstance.getName() + " study is not configured as a Juniper study! ");
             return transactionResults;
         }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
@@ -72,7 +72,6 @@ public class NonPepperKitCreationService {
      *
      * @return an instance of KitResponse, either as an error or KitStatus
      * **/
-
     private KitResponse validateAndInsertNonPepperKit(@NonNull KitRequestSettings kitRequestSettings, @NonNull JuniperKitRequest juniperKitRequest,
                                                          @NonNull EasyPostUtil easyPostUtil, DDPInstance ddpInstance,
                                                          KitType kitType, String kitTypeName) {
@@ -94,9 +93,7 @@ public class NonPepperKitCreationService {
             }
             juniperKitRequest.setEasypostAddressId(easypostAddressId);
             SimpleResult result = TransactionWrapper.inTransaction(conn -> {
-                SimpleResult transactionResults = new SimpleResult();
-                return createKit(ddpInstance, kitType, juniperKitRequest, kitRequestSettings, easyPostUtil, kitTypeName, conn,
-                        transactionResults);
+                return createKit(ddpInstance, kitType, juniperKitRequest, kitRequestSettings, easyPostUtil, kitTypeName, conn);
 
                 //          order external kits here if external shipper name is set for that kit request, not needed for now
             });
@@ -128,7 +125,8 @@ public class NonPepperKitCreationService {
      */
     private SimpleResult createKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitType, JuniperKitRequest kit,
                                    @NonNull KitRequestSettings kitRequestSettings, @NonNull EasyPostUtil easyPostUtil,
-                                   @NonNull String kitTypeName, Connection conn, SimpleResult transactionResults) {
+                                   @NonNull String kitTypeName, Connection conn) {
+        SimpleResult transactionResults = new SimpleResult();
         String juniperKitRequestId;
         String userId;
         //checking ddpInstance.isHasRole() to know this is a Juniper Kit
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsm.model.nonpepperkit;
 
 import java.sql.Connection;
-import java.sql.SQLException;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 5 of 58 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
```diff
@@ -23,7 +23,7 @@ public class KitStatusDao implements Dao<NonPepperKitStatusDto> {
 
     private static final String SELECT_KIT_STATUS =
             " SELECT req.*, k.*, discard.*, tracking.tracking_id as return_tracking_number, tracking.scan_by as tracking_scan_by, "
-                    + " tracking.scan_date as tracking_scan_date FROM ddp_kit_request req "
+                    + " tracking.scan_date as tracking_scan_date, bsp_collaborator_sample_id, bsp_collaborator_participant_id FROM ddp_kit_request req "
```
pegahtah: 2 extra fields that Juniper wanted come straight from DB

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
```diff
@@ -160,6 +161,9 @@ public NonPepperKitStatusDto build(ResultSet foundKitResults, Map<Integer, UserD
                                 NonPepperStatusKitService.convertTimeStringIntoTimeStamp(foundKitResults.getLong(DBConstants.DISCARD_DATE)))
                         .withDiscardBy(
                                 NonPepperStatusKitService.getUserEmailForFields(foundKitResults.getString(DBConstants.DISCARD_BY), users))
+                        .withCurrentStatus(NonPepperStatusKitService.calculateCurrentStatus(foundKitResults))
```
pegahtah: This other field is calculated based on other values

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/jobs/LabelCreationJob.java
```diff
@@ -22,7 +22,7 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
             if (labelCreationStarted == 0) {
                 List<KitRequestCreateLabel> kitsLabelTriggered = KitUtil.getListOfKitsLabelTriggered();
                 if (!kitsLabelTriggered.isEmpty()) {
-                    KitUtil.createLabel(kitsLabelTriggered);
+                    KitUtil.createLabel(kitsLabelTriggered, null);
```
pegahtah, line 25: Added EasyPostUtil as the second arg so testing with mocks are easier

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -54,31 +55,56 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
...
-        }
-
-        SimpleResult result = TransactionWrapper.inTransaction(conn -> {
-            SimpleResult transactionResults = new SimpleResult();
-            return createKit(ddpInstance, kitType, juniperKitRequest, kitRequestSettings, easyPostUtil, kitTypeName, conn,
-                    transactionResults);
-
-            //          order external kits here if external shipper name is set for that kit request, not needed for now
-        });
+        return validateAndInsertNewNonPepperKit(kitRequestSettings, juniperKitRequest, easyPostUtil, ddpInstance, kitType, kitTypeName);
```
pegahtah: Moved the code to a new method for cleaner codes

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -127,7 +154,7 @@ private void addJuniperKitRequest(Connection conn, String kitTypeName, KitReques
                                       DDPInstance ddpInstance, int kitTypeId, String collaboratorParticipantId,
                                       String errorMessage, EasyPostUtil easyPostUtil, JuniperKitRequest kit,
                                       String externalOrderNumber, String juniperKitRequestId,
-                                      String ddpLabel, String userId, SimpleResult transactionResults) throws DsmInternalError {
```
pegahtah, line 130: ddpLabel was always null here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
```diff
@@ -49,6 +58,89 @@ public static String convertTimeStringIntoTimeStamp(Long dateMillis) {
         return instant.toString();
     }
 
+    public static String calculateCurrentStatus(ResultSet foundKitResults) {
```
pegahtah: This method calculates the current status of a kit based on the logic that kit pages use in DSM for choosing which kits to show as without label, queue, sent, etc

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
```diff
@@ -2,6 +2,10 @@
 
 public class DBConstants {
 
+    private DBConstants(){
+        throw new IllegalStateException("Utility class");
```
pegahtah, line 6: Sonar problem

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -307,45 +309,29 @@ public Shipment getShipment(String shipmentId) throws EasyPostException {
      * @return true if easypost was able to create the address, false otherwise
      */
 
-    public boolean checkAddress(@NonNull JuniperKitRequest juniperKitRequest, String phone) {
-        if ((StringUtils.isBlank(juniperKitRequest.getJuniperParticipantID()))
-                || StringUtils.isBlank(juniperKitRequest.getLastName())) {
-            return false;
+    public String getEasyPostAddressId(@NonNull JuniperKitRequest juniperKitRequest, String phone, DeliveryAddress deliveryAddress)
```
pegahtah: Changed the name of the method and it's return ro String so it is easier to test

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -307,45 +309,29 @@ public Shipment getShipment(String shipmentId) throws EasyPostException {
...
-                logger.warn("Easypost couldn't create an address for " + juniperKitRequest.getShortId(), e);
-                return false;
+                throw new DsmInternalError("Easypost couldn't create an address for " + juniperKitRequest.getShortId(), e);
             }
         }
-        //to validate the address, first we need create the Address instance
-        DeliveryAddress deliveryAddress =
-                new DeliveryAddress(juniperKitRequest.getStreet1(), juniperKitRequest.getStreet2(), juniperKitRequest.getCity(),
-                        juniperKitRequest.getState(),
-                        juniperKitRequest.getPostalCode(), juniperKitRequest.getCountry(), name, phone);
```
pegahtah, line 339: moved to NonPepperKitCreationService. Now DeliverAddress instance is calculated first and then passed to the easypost util

pepper-apis/dsm-core/src/main/resources/liquibase/order_transmitted_at_column_ddp_kit_request.xml
```diff
@@ -0,0 +1,16 @@
...
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+<changeSet id="add_order_transmitted_at_if_not_exists" author="pegah">
+    <preConditions onFail="MARK_RAN">
+        <not>
+            <columnExists tableName="ddp_kit_request" columnName="order_transmitted_at" />
+        </not>
+    </preConditions>
+    <addColumn tableName="ddp_kit_request">
+        <column name="order_transmitted_at" type="DATETIME(6)" />
+    </addColumn>
```
pegahtah, line 14: This column was missing from the liquibases and was not created in the test db

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -95,39 +143,75 @@ public void testStatusByJuniperKitId() {
...
+        when(mockEasyPostUtil.getEasyPostAddressId(any(), any(), any())).thenReturn("SOME_STRING");
+        try {
+            when(mockEasyPostUtil.createParcel(any(), any(), any(), any())).thenReturn(mockEasyPostParcel);
+            when(mockEasyPostUtil.getAddress(any())).thenReturn(mockEasyPostAddress);
+            when(mockEasyPostUtil.createAddressWithoutValidation(any(), any(), any(), any(), any(), any(), any(), any())).thenReturn(
+                    mockEasyPostAddress);
+            when(mockEasyPostUtil.buyShipment(any(), any(), any(), any(), any(), any(), any(), any())).thenReturn(mockEasyPostShipment);
+        } catch (EasyPostException e) {
+            throw new RuntimeException(e);
+        }
```
pegahtah, line 160: Lots of mocking added to mimic label creation

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -166,42 +254,11 @@ private JuniperKitRequest generateJuniperKit(int random) {
...
-                + "\"juniperParticipantID\":\"" + participantId + rand + "\","
-                + "\"forceUpload\":false,"
-                + "\"skipAddressValidation\":false,"
-                + "\"juniperStudyID\":\"Juniper-test-guid\"}";
 
-        JuniperKitRequest juniperTestKit = new Gson().fromJson(json, JuniperKitRequest.class);
-        log.info("Juniper test kit id is {} ", juniperTestKit.getJuniperKitId());
-        when(mockEasyPostUtil.checkAddress(any(), anyString())).thenReturn(true);
+        int rand = new Random().nextInt() & Integer.MAX_VALUE;
+        JuniperKitRequest juniperTestKit = generateJuniperKit(rand);
```
pegahtah: We have a method that does the same thing so no need to double code it here

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -292,7 +347,6 @@ public void invalidKitTypeTest() {
                 + "\"phoneNumber\":\" 111 - 222 - 3344\","
                 + "\"juniperKitId\":\"JuniperTestKitId_" + rand + "\","
                 + "\"juniperParticipantID\":\"" + participantId + rand + "\","
-                + "\"forceUpload\":false,"
```
pegahtah, line 295: old name for variable

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -407,4 +408,13 @@ private static void deleteInstance(int instanceId) {
         });
     }
 
+    public void setUpByConfig(String ddpInstanceId, String ddpGroupId) {
+        if (StringUtils.isBlank(ddpInstanceId) || StringUtils.isBlank(ddpGroupId)) {
```
pegahtah: I need the study roles to be set up by the instance id and group id from my previous set up of juniper

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/nonPepperKit/NonPepperKitStatusDto.java
```diff
@@ -28,9 +29,9 @@ public class NonPepperKitStatusDto {
     private String errorMessage;
     private String discardDate;
     private String discardBy;
-
-    public NonPepperKitStatusDto() {
-    }
+    private String currentStatus;
```
denniscunningham: These can probably be `final`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -54,31 +55,56 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
...
-                    transactionResults);
-
-            //          order external kits here if external shipper name is set for that kit request, not needed for now
-        });
+        return validateAndInsertNewNonPepperKit(kitRequestSettings, juniperKitRequest, easyPostUtil, ddpInstance, kitType, kitTypeName);
+    }
 
-        if (result.resultException != null) {
-            log.error(String.format("Unable to create Juniper kit for %s", juniperKitRequest), result.resultException);
+    private KitResponse validateAndInsertNewNonPepperKit(KitRequestSettings kitRequestSettings, JuniperKitRequest juniperKitRequest,
```
denniscunningham: Nit: The term `New` here seems to add no value to the method name.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
```diff
@@ -13,13 +15,20 @@
...
 import org.broadinstitute.dsm.db.dto.user.UserDto;
+import org.broadinstitute.dsm.exception.DsmInternalError;
 import org.broadinstitute.dsm.statics.DBConstants;
 
 /**
  * This service returns a Juniper kit's status
  */
 @Slf4j
 public class NonPepperStatusKitService {
+    public static final String QUEUE = "Queue";
```
denniscunningham: This would be clearer as an enum.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -307,45 +309,29 @@ public Shipment getShipment(String shipmentId) throws EasyPostException {
      * @return true if easypost was able to create the address, false otherwise
      */
 
-    public boolean checkAddress(@NonNull JuniperKitRequest juniperKitRequest, String phone) {
-        if ((StringUtils.isBlank(juniperKitRequest.getJuniperParticipantID()))
-                || StringUtils.isBlank(juniperKitRequest.getLastName())) {
-            return false;
+    public String getEasyPostAddressId(@NonNull JuniperKitRequest juniperKitRequest, String phone, DeliveryAddress deliveryAddress)
+            throws DsmInternalError, DSMBadRequestException {
```
denniscunningham: These exceptions do not need to be listed in the method signature.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -407,4 +408,13 @@ private static void deleteInstance(int instanceId) {
         });
     }
 
+    public void setUpByConfig(String ddpInstanceId, String ddpGroupId) {
```
denniscunningham: A few suggestions:

- Move this near createRealmAndStudyGroup, since it is an alternate way to establish the ddpInstanceId and the studyGroupId.
- Rename the `ddpGroupId` parameter to `studyGroupId` to make this consistent with the terminology in this class.
- Add JavaDoc for this method describing its function. (If you think the term `ddpGroupId` is important for callers, you could state that the `studyGroupId` equates to the ID for the ddp_group table.)
- Since this class manages resources, it will remove the realm and study group when `deleteGeneratedData` is called (required), so it is odd to have a service delete something created elsewhere. So I suggest pointing that out in the JavaDoc.
- Rename this method to something like setExistingRealmAndStudyGroup. This makes it consistent with the existing methods, and better describes the function provided. In general, method names should hint at the function provided, not guess about the context the caller might be calling them from. In this case the term `Config` is unlikely to help other callers understand what this method does.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
```diff
@@ -118,7 +118,7 @@ public UserAdminService(String operatorId, String studyGroup) {
         initialized = false;
     }
 
-    protected void initialize() {
+    public void initialize() {
```
denniscunningham: This change should not be needed. By making this public the UserAdminService contract is changed, which is not backed up with additional automated tests.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
```diff
@@ -82,7 +82,7 @@ public List<NonPepperKitStatusDto> getKitsByJuniperKitId(String juniperKitId, Ma
                     list.add(builder.build(rs, users));
                 }
             } catch (Exception ex) {
-                dbVals.resultException = new Exception(String.format("Error getting kits with juniper kit id %s", juniperKitId));
+                dbVals.resultException = new Exception(String.format("Error getting kits with juniper kit id %s", juniperKitId), ex);
```
denniscunningham: Nit, but creating a new Exception here seems unnecessary. Why not construct this error message directly in the DsmInternalError below? The `juniperKitId` is available there.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
denniscunningham: It seems odd to do this here. Effectively reaching above on the stack to the service. Why isn't `calculateCurrentStatus` a method in this class?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
denniscunningham: As I mentioned elsewhere, it feels like this should be in the Dao, particularly since it is handling SQLException. All the static methods in this class that deal with result sets should also be in the Dao, unless I am missing something. Each is transforming DB information into a DTO, which is the job of the DAOs in Pepper.
Sorry if I missed this in an earlier review.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
```diff
@@ -49,6 +58,89 @@ public static String convertTimeStringIntoTimeStamp(Long dateMillis) {
...
+            } else if (isReceivedKit(foundKitResults)) {
+                return RECEIVED;
+            } else if (isSentKit(foundKitResults)) {
+                return SENT;
+            } else if (isNewKit(foundKitResults) || isEasyPostLabelTriggeredKit(foundKitResults)) {
+                return KIT_WITHOUT_LABEL;
+            } else if (isDeactivatedKit(foundKitResults)) {
+                return DEACTIVATED;
+            } else {
+                log.error(String.format("Unable to decide the current status of kit %s",
```
denniscunningham: Nit: `determine` might be a better term than `decide`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
denniscunningham: I think you can ignore these warnings. This is not really a problem.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -307,45 +309,29 @@ public Shipment getShipment(String shipmentId) throws EasyPostException {
      * @return true if easypost was able to create the address, false otherwise
```
denniscunningham: The JavaDoc is no longer accurate.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -120,14 +121,15 @@ public class KitUtil {
     private static final String EASYPOST_FAILURE_STATUS = "failure";
     private static final String EASYPOST_RETURN_SENDER_STATUS = "return_to_sender";
     private static final String EASYPOST_ERROR_STATUS = "error";
-    public static final String PECGS_RESEARCH = "PECGS_RESEARCH";
 
-    public static void createLabel(List<KitRequestCreateLabel> kitsLabelTriggered) {
+    public static void createLabel(List<KitRequestCreateLabel> kitsLabelTriggered, EasyPostUtil easyPostUtil) {
```
denniscunningham, line 132: How are we able to pass this in if a new EasyPostUtil gets created for each instance when it is null?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -24,7 +45,7 @@
 @Slf4j
```
denniscunningham, line 53: This doc might be better if it actually named the methods used to set things up and tear them down.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -167,14 +192,63 @@ private static String getPrimaryKey(ResultSet rs, String table) throws SQLExcept
...
+        //check if kit is from a pt which is withdrawn
+        Optional<BSPKitStatus> result = bspKit.getKitStatus(optionalBSPKitDto.get(), notificationUtil);
+        if (!result.isEmpty()) {
+            return;
+        }
+        //kit found in ddp_kit table
+        bspKit.receiveKit(kitLabel, optionalBSPKitDto.get(), notificationUtil, "BSP").orElseThrow();
+    }
+
+    public void setupJuniperInstanceAndSettings() {
```
denniscunningham: It feels like this method should be towards the top of this file, particularly if it is required when using this class. (The JavaDoc for this class might want to be clear on whether this method *must* be called.)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -97,6 +121,7 @@ public static void deleteJuniperInstanceAndSettings() {
                 delete(conn, "ddp_instance_group", "instance_group_id", ddpInstanceGroupId);
                 delete(conn, "ddp_instance", "ddp_instance_id", ddpInstanceId);
                 delete(conn, "ddp_group", "group_id", ddpGroupId);
+                cmiAdminUtil.deleteGeneratedData();
```
denniscunningham, line 149: I am surprised this is not failing since `deleteGeneratedData` will delete the instance and group, which are already deleted.
Stepping back, perhaps a better way to handle this is to call `UserAdminTestUtil.createRealmAndStudyGroup` instead of `createDdpGroupForJuniper` and `createDdpInstanceForJuniper` (and get rid of `UserAdminTestUtil. setUpByConfig). If this util needs more non-constant info in the instance table, add that as parameters to `createRealmAndStudyGroup`. That takes care of the problems with sharing responsibility for managing realm and study group, and leaves just one place where we create those for tests.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -167,14 +192,63 @@ private static String getPrimaryKey(ResultSet rs, String table) throws SQLExcept
...
+        }
+
+    }
+
+    public static void changeKitToSent(JuniperKitRequest juniperTestKit) {
+        List<SentAndFinalScanPayload> scanPayloads = new ArrayList<>();
+        DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(instanceName).orElseThrow();
+        SentAndFinalScanPayload sentAndFinalScanPayload = new SentAndFinalScanPayload(juniperTestKit.getDdpLabel(), "SOME_RANDOM_KIT_LABEL");
+        scanPayloads.add(sentAndFinalScanPayload);
+        KitFinalScanRoute kitFinalScanRoute = new KitFinalScanRoute();
```
denniscunningham: This feels awkward to be doing at the route level. 
I'm not sure what we want to accomplish here, but it feels like if there is service level code at the route level, it should be moved into an appropriate service.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
```diff
@@ -18,8 +18,8 @@
 public abstract class KitStatusChangeRoute extends RequestHandler {
 
     protected NotificationUtil notificationUtil;
-    protected KitPayload kitPayload;
-    protected List<ScanError> scanErrorList;
+    public KitPayload kitPayload;
```
denniscunningham: See my comments below, but it is unclear why we are using code at this level in a test util.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
```diff
@@ -49,6 +58,89 @@ public static String convertTimeStringIntoTimeStamp(Long dateMillis) {
         return instant.toString();
     }
 
+    public static String calculateCurrentStatus(ResultSet foundKitResults) {
+        try {
+            if (isQueuedKit(foundKitResults)) {
```
denniscunningham: How did you pick this order of conditionals? It is unclear from looking at the `isXxxKit` methods if the conditions are mutually exclusive.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -98,7 +124,8 @@ private SimpleResult createKit(@NonNull DDPInstance ddpInstance, @NonNull KitTyp
             userId = JUNIPER;
         } else {
             log.warn("Seems like {} is not configured as a JUNIPER study! ", ddpInstance.getName());
-            transactionResults.resultException = new DsmInternalError(ddpInstance.getName() + " study is not configured to use this method.");
+            transactionResults.resultException =
+                    new DsmInternalError(ddpInstance.getName() + " study is not configured to use this method.");
```
denniscunningham: "this method" seems vague. Can you say something like "study is not configured as a Juniper study", since that seems to be the specific case here? (And matches the warning message.)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -86,6 +111,29 @@ public void testStatusKitEndpointByJuniperStudyGuid() {
         }
     }
 
+    @Test
+    public void testCurrentStatusFieldAfterChanges() {
```
denniscunningham: Nit: `AfterChanges` may be more confusing than helpful. This appears to be the main test for the current status field, so `testCurrentStatusField` seems like an appropriate summary.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -86,6 +111,29 @@ public void testStatusKitEndpointByJuniperStudyGuid() {
...
+            assertStatusKitResponse(kitResponse, juniperTestKit, rand, NonPepperStatusKitService.QUEUE);
+            juniperTestKit.setDdpLabel(kitResponse.getKits().get(0).getDsmShippingLabel());
+            JuniperSetupUtil.changeKitToSent(juniperTestKit);
+            kitResponse = nonPepperStatusKitService.getKitsBasedOnJuniperKitId(juniperTestKit.getJuniperKitId());
+            assertStatusKitResponse(kitResponse, juniperTestKit, rand, NonPepperStatusKitService.SENT);
+            JuniperSetupUtil.changeKitToReceived();
+            kitResponse = nonPepperStatusKitService.getKitsBasedOnJuniperKitId(juniperTestKit.getJuniperKitId());
+            assertStatusKitResponse(kitResponse, juniperTestKit, rand, NonPepperStatusKitService.RECEIVED);
+        } finally {
+            createdKitIds.add(juniperTestKit.getJuniperKitId());
```
denniscunningham: Is there a reason to have this in a finally block? It seems like it could go immediately after the call to `generateJuniperKit`.

denniscunningham (changes requested):
Added some comments.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
pegahtah: We need to create a new EasyPostUtil because some studies use different easypost accounts. The change I made was for when we want to test this and it enables us to pass a mock easypost. Otherwise in real worlds for each if the members of kitsLabelTriggered a new easypostUtil is created. I see this is not ideal but changing the createLabel method would be out of scope imho

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
pegahtah: I want to make sure if anything goes wrong while the new kit is getting added the new kit still gets added to the createdKitIds so it get deleted in the end

denniscunningham: Right. But isn't that accomplished by just adding the kit ID right after the call to `generateJuniperKit`? Just from the structure of the code (and not knowing the content) that seems true.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
denniscunningham: So for that passed-in EasyPostUtil, the instance/realm does not matter? If so, you might want to add some JavaDoc, since otherwise other readers may have the same question.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
pegahtah: I wanted to make sure we call the real code as it is written and not a portion of it

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
pegahtah: I got it removed and applied your other comment from above

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/admin/UserAdminService.java
pegahtah: Yes my bad, forgot to revert

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
pegahtah: makes sense now that you point it, I will move it 
I was fooled by this method not running any direct query

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
pegahtah: Moved! Thanks for noticing this!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
pegahtah: No specific order,  but I went through them again and I didn't find conditions that overlap. I added some more checks too

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.dsm.db.dao.kit;
+
+import lombok.Getter;
+
+@Getter
+public enum KitCurrentStatus {
+    QUEUE("Queue"), ERROR("Error"), RECEIVED("Received"), SENT("Sent"), KIT_WITHOUT_LABEL("Kit Without Label"), DEACTIVATED("Deactivated");
```
denniscunningham: Nit: Typically each enum value is on a new line.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitCurrentStatus.java
```diff
@@ -0,0 +1,13 @@
...
+    private final String value;
+
+    KitCurrentStatus(String name) {
+        this.value = name;
+    }
+}
```
denniscunningham, line 19: You may be getting a checkstyle warning about a missing newline.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitStatusDao.java
```diff
@@ -124,6 +129,85 @@ public List<NonPepperKitStatusDto> getKitsByKitIdArray(String[] kitIdsArray, Map
     }
 
     private static class BuildNonPepperKitStatusDto {
+        public static KitCurrentStatus calculateCurrentStatus(ResultSet foundKitResults) {
```
denniscunningham: Nit: This can probably be private or protected?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -54,31 +55,69 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
...
+     * @param kitRequestSettings KitRequestSettings that shows the specific kit config for this study;
+     * @param juniperKitRequest the new JuniperKitRequest received
+     * @param easyPostUtil an instance of the EasyPostUtil
+     * @param ddpInstance DDPInstance representing the study for which we want to create kits
+     * @param kitType the KitType that the request wants to create
+     * @param kitTypeName the specific name of the request kit type
+     *
+     * @return an instance of KitResponse, either as an error or KitStatus
+     * **/
+
```
denniscunningham: Nit: Are you getting a lint warning about this blank line?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -54,31 +55,69 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
...
-                    transactionResults);
-
-            //          order external kits here if external shipper name is set for that kit request, not needed for now
-        });
-
-        if (result.resultException != null) {
-            log.error(String.format("Unable to create Juniper kit for %s", juniperKitRequest), result.resultException);
+        name += juniperKitRequest.getLastName();
+        DeliveryAddress deliveryAddress =
+                new DeliveryAddress(juniperKitRequest.getStreet1(), juniperKitRequest.getStreet2(), juniperKitRequest.getCity(),
```
denniscunningham, line 87: Not for now, but DeliveryAddress is a great candidate for a builder pattern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -54,31 +55,69 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
...
+                        juniperKitRequest.getState(), juniperKitRequest.getPostalCode(), juniperKitRequest.getCountry(), name, phone);
+        try {
+            String easypostAddressId = easyPostUtil.getEasyPostAddressId(juniperKitRequest, phone, deliveryAddress);
+            if (StringUtils.isBlank(easypostAddressId)) {
+                throw new DsmInternalError("EasyPost address Id should not be null!");
+            }
+            juniperKitRequest.setEasypostAddressId(easypostAddressId);
+            SimpleResult result = TransactionWrapper.inTransaction(conn -> {
+                SimpleResult transactionResults = new SimpleResult();
```
denniscunningham: Is there are reason 'transactionResults' is passed into `createKit`? It seems like `createKit` could create this.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
pegahtah: Right, I removed ` createdKitIds.add(juniperTestKit.getJuniperKitId());` from finally and moved it into `createAndAssertNonPepperCreation`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -116,7 +156,7 @@ private SimpleResult createKit(@NonNull DDPInstance ddpInstance, @NonNull KitTyp
         //In case it's needed, subkits handling should be added here
         try {
             addJuniperKitRequest(conn, kitTypeName, kitRequestSettings, ddpInstance, kitType.getKitTypeId(), collaboratorParticipantId,
-                    errorMessage, easyPostUtil, kit, externalOrderNumber, juniperKitRequestId, null, userId, transactionResults);
+                    errorMessage, easyPostUtil, kit, externalOrderNumber, juniperKitRequestId, userId, transactionResults);
```
denniscunningham, line 157: Not for now, but `addJuniperKitRequest` has *way* too many positional parameters. This is an anti-pattern which can be solved, for instance, by refactoring kit request into its own class, likely with a builder. And I'm sure there are also other ways to fix this.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -174,9 +214,7 @@ private void addJuniperKitRequest(Connection conn, String kitTypeName, KitReques
                 String dsmKitRequestId =
                         KitRequestShipping.writeRequest(ddpInstance.getDdpInstanceId(), juniperKitRequestId, kitTypeId, participantID,
                                 collaboratorParticipantId, collaboratorSampleId, userId, addressId, errorMessage,
-                                kit.getExternalOrderNumber(),
-                                false,
-                                null, ddpInstance, bspCollaboratorSampleType, ddpLabel);
+                                kit.getExternalOrderNumber(), false, null, ddpInstance, bspCollaboratorSampleType, null);
```
denniscunningham, line 215: As I mentioned above, this many parameter anti-pattern should be replaced, although perhaps not for this PR.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperStatusKitService.java
```diff
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsm.model.nonpepperkit;
 
+import java.sql.ResultSet;
```
denniscunningham: I'm guessing these imports are no longer needed?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
denniscunningham: It looks like the code is still in the finally block?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -66,68 +87,109 @@ public static void deleteJuniperInstance() {
...
+        when(mockEasyPostAddress.getId()).thenReturn("SOME_STRING");
+        when(mockEasyPostUtil.getEasyPostAddressId(any(), any(), any())).thenReturn("SOME_STRING");
+        try {
+            when(mockEasyPostUtil.createParcel(any(), any(), any(), any())).thenReturn(mockEasyPostParcel);
+            when(mockEasyPostUtil.getAddress(any())).thenReturn(mockEasyPostAddress);
+            when(mockEasyPostUtil.createAddressWithoutValidation(any(), any(), any(), any(), any(), any(), any(), any())).thenReturn(
+                    mockEasyPostAddress);
+            when(mockEasyPostUtil.buyShipment(any(), any(), any(), any(), any(), any(), any(), any())).thenReturn(mockEasyPostShipment);
+        } catch (EasyPostException e) {
+            throw new RuntimeException(e);
```
denniscunningham: In general RuntimeException is a bad choice, but since this is in a test context it probably makes more sense to Assert.fail here, perhaps with a stack trace for the exception.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -66,68 +87,109 @@ public static void deleteJuniperInstance() {
...
 
-        } finally {
-            createdKitIds.add(juniperTestKit.getJuniperKitId());
-        }
     }
 
     private void createAndAssertNonPepperCreation(JuniperKitRequest juniperTestKit) {
-        List<KitRequestShipping> oldkits = KitRequestShipping.getKitRequestsByRealm(instanceName, "overview", salivaKitType);
-        KitResponse kitResponse = nonPepperKitCreationService.createNonPepperKit(juniperTestKit, salivaKitType, mockEasyPostUtil, ddpInstance);
+        createdKitIds.add(juniperTestKit.getJuniperKitId());
```
denniscunningham: I'm not sure I understand what's going on here. It seems like this method is called at the beginning of each test method. Perhaps it needs JavaDoc? Or maybe the method name is confusing? 
Is it basically creating a non-Pepper kit for testing? If so, perhaps a better method name would be `createNonPepperKit` or `createNonPepperTestKit`?
Perhaps I am also confused by the parameter name, since it is actually a JuniperKitRequest (not a kit), and `generateJuniperKit`, which does not actually generate a kit, but a kit request. Since `generateJuniperKit` is (necessarily) called before this method, perhaps it should be the first line in this method?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -66,68 +87,109 @@ public static void deleteJuniperInstance() {
...
+                newKits.stream().filter(kitRequestShipping -> kitRequestShipping.getDdpParticipantId()
+                                .equals(juniperTestKit.getJuniperParticipantID()))
+                        .findAny().get();
+        Assert.assertEquals(newKit.getBspCollaboratorParticipantId(), bspPrefix + "_" + juniperTestKit.getJuniperParticipantID());
+        Assert.assertEquals(newKit.getBspCollaboratorSampleId(),
+                bspPrefix + "_" + juniperTestKit.getJuniperParticipantID() + "_" + salivaKitType);
     }
 
-    private void assertStatusKitResponse(KitResponse kitResponse, JuniperKitRequest juniperTestKit, int testRandomNumber) {
+    private NonPepperKitStatusDto assertStatusKitResponse(KitResponse kitResponse, JuniperKitRequest juniperTestKit, int testRandomNumber,
```
denniscunningham: It does not look like this return value is used? Also, since `assert` is a very specific term in Junit, perhaps `verifyKitResponse` is a better method name?
Also, perhaps no need to fix now, but it would probably be better to pass in a fully formed kit ID. The parameter `testRandonNumber` is not actually random.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
denniscunningham: Given the way the test file is constructed now, this test can probably be deleted.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -66,7 +69,70 @@ public class UserAdminTestUtil {
             "DELETE FROM ddp_instance WHERE ddp_instance_id = ?";
 
 
-    public UserAdminTestUtil() {}
+
+    public UserAdminTestUtil() {
+    }
+
+    private static int createInstanceGroup(String instanceName, String studyGuid, String collaboratorPrefix, int studyGroupId) {
```
denniscunningham: Why did you move these in the file? The history tracing gets messed up when methods are moved. Can you move them back?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
pegahtah: So the reason we make different ones for each realm is to pass the right apikey and secret to real easypost. For the mock one we just have to mock everything and not use the real easypost.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -135,11 +203,12 @@ public int setStudyAdminAndRoles(String adminEmail, String adminRole, List<Strin
 
     /**
      * Add roles for study
+     *
      * @param roles actual DSM roles
      */
     public void addStudyRoles(List<String> roles) {
         assertInitialized();
-        for (String role: roles) {
+        for (String role : roles) {
```
denniscunningham: Why this change and other similar changes?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -177,21 +247,22 @@ public void deleteStudyAdminAndRoles() {
         new UserDao().delete(userAdminId);
         userAdminId = -1;
 
-        for (int groupRoleId: createdGroupRoles) {
+        for (int groupRoleId : createdGroupRoles) {
```
denniscunningham: Why this change?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -42,6 +44,7 @@
  *      Call close()
  */
 @Slf4j
+@Getter
```
denniscunningham: Why is a getter needed? That changes the contract for this class.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -95,23 +161,25 @@ private void assertInitialized() {
...
-        ddpInstanceId = createInstanceGroup(realmName, studyGroupId);
+        ddpInstanceId = createInstanceGroup(realmName, studyGuid, collaboratorPrefix, studyGroupId);
     }
 
     /**
      * Setup study admin and study roles
      *
-     * @param adminEmail for new admin user account
-     * @param adminRole PEPPER_ADMIN or STUDY_USER_ADMIN
+     * @param adminEmail    for new admin user account
```
denniscunningham: This change seems unnecessary.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -56,25 +86,61 @@ public class JuniperSetupUtil {
...
+        this.groupName = groupName;
+    }
 
+    public void setupJuniperInstanceAndSettings() {
+        //everything should get inserted in one transaction
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult simpleResult = new SimpleResult();
+            try {
+                cmiAdminUtil.createRealmAndStudyGroup(instanceName, studyGuid, collaboratorPrefix, groupName);
+                ddpInstanceId = String.valueOf(cmiAdminUtil.getDdpInstanceId());
```
denniscunningham, line 110: There should be no reason to do this testing. It is the responsibility of UserAdminTestUtil to assure that it created these things or return an error.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -56,25 +86,61 @@ public class JuniperSetupUtil {
...
+            return simpleResult;
+        });
+        if (results.resultException != null) {
+            log.error("Error creating juniper data ", results.resultException);
+            deleteJuniperInstanceAndSettings();
+        }
     }
 
-    private static String createDdpGroupForJuniper(Connection conn) throws SQLException {
+    private static String createDdpGroupForJuniper(Connection conn, String groupName) throws SQLException {
```
denniscunningham: This should be obsolete now, right?

denniscunningham (changes requested):
A few more comments.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
pegahtah: I don't undertsand... I need the ddpInstanceId to insert a bunch more stuff in other tables, does this makes sense as of why I am getting the ddpInstanceId from the User if that's the concern?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: I think I should revisit my lint set up again

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
pegahtah: Correct! I will make a ticket for this

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
denniscunningham: That makes sense. It is unclear why you are testing the value for null.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
denniscunningham: I can see why you need the ddpInstanceId and studyGroupId, but I suggest making explicit getters for them, since that limits the contract for this class. IOW, this class is not a bean, so should not have a blanket getters or setters.

denniscunningham (commented):
Responded to your question.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
pegahtah: testRandomNumber is generated everytime with `int rand = new Random().nextInt() & Integer.MAX_VALUE;` and then passed into this. it makes finding newly created kits easier by their id

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
pegahtah: this method creates the juniperTestKitRequest in the database  by calling `NonPepperKitCreationService.createNonPepperKit` and verifies the response is as expected
The `generateJuniperKit` generates a kit request  with random ids and a test passes it to this for making sure it gets created

pegahtah: I need to call `generateJuniperKit` (to be renamed to `generateJuniperKitRequest` separately because the returned value of that is useful outside of the scope of this method to my tests

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
denniscunningham: Right, I get that. I was simply suggesting constructing the kit ID and passing that around instead, since this 'random' number is really just part of an ID at this point, which you reconstruct in this method. Just makes things easier to follow if you pass around the actual ID. As I said, just a suggestion, not necessary for this review.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
pegahtah: I think my checkstyle fixes actually broke this file. I'll revert it and fix my checkstyle later

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
pegahtah: Understood!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
denniscunningham: I no longer see the comments I mentioned here (perhaps I did not save them), but do these need to be public?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
denniscunningham: It still feels like the parameter name here does not match the parameter type, and the method name still misses the fundamental purpose of this method, which is to create a kit for the caller. Perhaps a nit, but I do not think the purpose is clear at all, so if you like the method name, perhaps just some JavaDoc explaining that this method creates a kit? All the other stuff in the method seems secondary.

denniscunningham (commented):
Thanks for all your changes!
Just a couple more comments (in this review block and I think one other outside this review block).

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperKitCreationStatusTest.java
```diff
@@ -66,68 +87,101 @@ public static void deleteJuniperInstance() {
...
-    private void createAndAssertNonPepperCreation(JuniperKitRequest juniperTestKit) {
-        List<KitRequestShipping> oldkits = KitRequestShipping.getKitRequestsByRealm(instanceName, "overview", salivaKitType);
-        KitResponse kitResponse = nonPepperKitCreationService.createNonPepperKit(juniperTestKit, salivaKitType, mockEasyPostUtil, ddpInstance);
+    /**
+     * this method creates the juniperTestKitRequest in the database  by calling
+     * `NonPepperKitCreationService.createNonPepperKit` and verifies the response is as expected
+     * @param juniperTestKitRequest a JuniperKitRequest that can be passed to the kti creation service
+     * ***/
+
+    private void createNonPepperTestKit(JuniperKitRequest juniperTestKitRequest) {
```
pegahtah, line 143: @denniscunningham I think github was hiding some changes last time, I am not sure if I made a mistake with pushing or not, but the javadoc comment and the new name is showing now

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitFinalScanRoute.java
```diff
@@ -13,7 +13,7 @@ public KitFinalScanRoute() {
     }
 
     @Override
-    protected void processRequest() {
+    public void processRequest() {
```
denniscunningham: Can you remind me why we need this change?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
denniscunningham: Still wondering about this.

denniscunningham (commented):
Looks good!
You may have missed a question from I prior review, so I added that to this review.
Once we are straight on that this will be good to go.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
pegahtah: I missed this, sorry!

denniscunningham (approved):
Looks great!
Thx for your patience.

PR #2679, opened 2023-09-21
[DSM] PEPPER-798 smooth deletions of onc history and tissue fields

pegahtah:
## Context
PEPPER-798

I changed how the delete works by creating a new DeletePatchFactory class and different classes for Tissue and OncHistory. For each delete request submitted for any of the onc history or tissue, the factory collects the other objects that belong to it that need to get deleted and creates a patch for request for them.

For handling Unique constraint, I added a new annotation so we can check the uniqueness in java instead of in database, however we need to drop our unique constraints by hand for this to work

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -458,7 +466,7 @@ public static String createNewOncHistoryDetail(@NonNull String medicalRecordId,
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error adding new oncHistoryDetail for medicalRecord w/ id " + medicalRecordId,
+            throw new DsmInternalError("Error adding new oncHistoryDetail for medicalRecord w/ id " + medicalRecordId,
                     results.resultException);
```
pegahtah: checkstyle fixes here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
```diff
@@ -97,9 +99,11 @@ public class Tissue {
     @DbDateConversion (SqlDateConverter.STRING_DAY)
     private String scrollsReceived;
 
+    @UniqueField(true)
     @ColumnName (DBConstants.SK_ID)
     private String skId;
 
+    @UniqueField(true)
     @ColumnName (DBConstants.SM_ID)
     private String smId;
```
pegahtah, line 109: These 3 fields are the only fields that need unique value checks. I added a new annotation so we know them and we can ignore the values from deleted tissues when we are checking the uniqueness of the input. For example, if a tissue gets deleted and wants to get reentered, crc should be able to enter the sm id again

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueDao.java
```diff
@@ -75,8 +78,30 @@ public Map<String, List<Tissue>> getTissuesByStudy(String study) {
...
+            } catch (SQLException ex) {
+                execResult.resultException = ex;
+            }
+            return execResult;
+        });
+        if (results.resultException != null) {
+            throw new DsmInternalError("Error getting tissue id by onc history detail id " + oncHistoryDetailId, results.resultException);
+        }
+        return tissueIds;
+    }
```
pegahtah, line 106: This method is made to find the tissues that belong to a specific onc history. The idea behind this is if an onc history gets deleted, the corresponding tissues for it should also get deleted

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -13,30 +13,26 @@
...
-    private static final Logger logger = LoggerFactory.getLogger(TissueSMIDDao.class);
-
     public static final String SQL_GET_SM_ID_BASED_ON_TISSUE_ID =
-            " SELECT * from sm_id sm where sm.tissue_id= ?   and NOT sm.deleted <=> 1";
+            " SELECT * from sm_id sm where sm.tissue_id= ?";
     public static final String SQL_GET_SEQUENCING_SM_ID_BASED_ON_TISSUE_ID = " SELECT * from sm_id sm "
             + "left join sm_id_type smtype on (sm.sm_id_type_id = smtype.sm_id_type_id) "
             + "where sm.tissue_id = ? and (smtype.sm_id_type = \"uss\" or smtype.sm_id_type = \"scrolls\") "
-            + "and sm.received_date is not null ";
+            + "and sm.received_date is not null and NOT sm.deleted <=> 1 ";
```
pegahtah, line 29: Changes added so the deleted sm ids are not included in a clinical order to Mercury

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -48,15 +44,17 @@ public class TissueSMIDDao {
                     + "LEFT JOIN ddp_participant as p ON (p.participant_id = ins.participant_id) "
                     + "LEFT JOIN ddp_instance as realm ON (realm.ddp_instance_id = p.ddp_instance_id) "
                     + "WHERE realm.instance_name = ?";
+    private static final Logger logger = LoggerFactory.getLogger(TissueSMIDDao.class);
 
-    public String getTypeForName(String type) {
```
pegahtah, line 52: This method was moved down

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -48,15 +44,17 @@ public class TissueSMIDDao {
...
             SimpleResult dbVals = new SimpleResult();
-            try (PreparedStatement stmt = conn.prepareStatement(SQL_TYPE_ID_FOR_TYPE)) {
-                stmt.setString(1, type);
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_SM_ID_BASED_ON_TISSUE_ID)) {
+                stmt.setString(1, tissueId);
                 try (ResultSet rs = stmt.executeQuery()) {
-                    if (rs.next()) {
-                        dbVals.resultValue = rs.getString(DBConstants.SM_ID_TYPE_ID);
+                    while (rs.next()) {
+                        smIds.add(rs.getString(DBConstants.SM_ID_PK));
```
pegahtah: This method finds the sm ids that all belong to the same tissue by the tissue id. The idea is if that tissue gets deleted, the sm ids belonging to it should also get deleted

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -211,7 +211,7 @@ public Object handle(Request request, Response response) {
                 String shortId = maybeParticipantByParticipantId.get().getProfile().map(Profile::getHruid).get();
                 addCollaboratorSampleId(tissueId, ddpInstance, ddpParticipantId, shortId);
             }
-            new TissueSMIDDao().createNewSMIDForTissueWithValue(tissueId, ffpeUser, smIdType, kitLabel);
```
pegahtah, line 212: These two methods were doing the same thing so I removed `reateNewSMIDForTissueWithValue` and used `createNewSMIDForTissue`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/structure/DBElement.java
```diff
@@ -14,16 +14,19 @@ public class DBElement {
...
         this.tableAlias = tableAlias;
         this.primaryKey = primaryKey;
         this.columnName = columnName;
+        this.isUniqueField = false;
     }
 
-    public DBElement(String tableName, String tableAlias, String primaryKey, String columnName, DbDateConversion dateConverter) {
+    public DBElement(String tableName, String tableAlias, String primaryKey, String columnName, DbDateConversion dateConverter, Boolean isUniqueField) {
         this(tableName, tableAlias, primaryKey, columnName);
+        this.isUniqueField = isUniqueField;
```
pegahtah, line 29: This new annotation is added so that we know if we have to check for the uniqueness of the value before patching

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -4,14 +4,22 @@
...
+import org.broadinstitute.dsm.exception.DSMBadRequestException;
 import org.broadinstitute.dsm.model.elastic.export.generate.PropertyInfo;
+import org.broadinstitute.dsm.statics.DBConstants;
 import org.broadinstitute.dsm.util.NotificationUtil;
 
 public class PatchFactory {
 
+    private PatchFactory(){
+        throw new IllegalStateException("Utility class");
+    }
```
pegahtah, line 16: sonar

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/PatchUtil.java
```diff
@@ -133,9 +136,9 @@ private static void getColumnNames(Class<?> obj) {
                     if (StringUtils.isNotBlank(columnPrefix)) {
                         fieldKey = columnPrefix.concat("_").concat(field.getName());
                     }
-
+                    boolean isUnique = Arrays.stream(uniqueFields).anyMatch(uniqueField -> uniqueField.value());
```
pegahtah: check if the field had the `@UniqueField` annotation

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.dsm.model.patch;
+
+import lombok.extern.slf4j.Slf4j;
+import org.broadinstitute.dsm.util.NotificationUtil;
+
+@Slf4j
+public class DeleteOncHistoryPatch extends ExistingOncHistoryPatch {
+    private final NotificationUtil notificationUtil;
+    public DeleteOncHistoryPatch(Patch patch, NotificationUtil notificationUtil) {
```
pegahtah, line 10: So originally the delete patch requests were considered an existing patch, so I made these two classes `DeleteOncHistoryPatch` and `DeleteTissuePatch` that are extended from the original class and the only thing I changed is that I added one extra step in their `doPatch` method to call `DeletePatchFactory.setDeletedForChildrenFields(this.patch, notificationUtil);` I have explained that more in its class

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+import lombok.NonNull;
+import org.apache.commons.lang3.NotImplementedException;
+import org.broadinstitute.dsm.db.dao.ddp.tissue.TissueDao;
+import org.broadinstitute.dsm.db.dao.ddp.tissue.TissueSMIDDao;
+import org.broadinstitute.dsm.model.NameValue;
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.util.NotificationUtil;
+
+public class DeletePatchFactory {
```
pegahtah, line 15: This DeletePatchFactory decides what kind of patch class should be assigned to the request. The PatchRoute checks the patch and if it is a Delete Patch instance, this class will create the appropriate type of `Delete` patcher

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,70 @@
...
+        if (originalPatch.getNameValue().getName().equals("t.deleted") && originalPatch.getTableAlias().equals(DBConstants.DDP_TISSUE_ALIAS)) {
+            deletePatches = getPatchForSmIds(originalPatch);
+        } else if (originalPatch.getNameValue().getName().equals("oD.deleted") &&
+                originalPatch.getTableAlias().equals(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS)) {
+            deletePatches = getPatchForTissues(originalPatch);
+        }
+        for (Patch childPatch : deletePatches) {
+            BasePatch patcher = PatchFactory.makePatch(childPatch, notificationUtil);
+            patcher.doPatch();
+        }
```
pegahtah, line 46: This method finds the tissues that belong to this onc history that just got deleted, or the sm ids that belong to the tissue,  and creates `Patch` requests for each of those tissues. The `doPatch()` method of those patches will then be called one by one

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -123,6 +137,11 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
         }
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
+            if (dbElement.isUniqueField) {
+                if(!isValueUnique(dbElement, String.valueOf(nameValue.getValue()))){
```
pegahtah: For this to work as expected, we need to manually drop the unique constraints on the table

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -195,16 +195,24 @@ public class OncHistoryDetail implements HasDdpInstanceId {
...
     private String changedBy;
+
     @ColumnName(DBConstants.DELETED)
     private boolean deleted;
+
     @ColumnName(DBConstants.UNABLE_OBTAIN_TISSUE)
     private boolean unableObtainTissue;
+
     private String participantId;
+
```
denniscunningham: These blank lines are probably unnecessary and do not align with the general style used in DSM.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -444,10 +452,10 @@ public static String createNewOncHistoryDetail(@NonNull String medicalRecordId,
                             dbVals.resultValue = oncHistoryDetailId;
                         }
                     } catch (Exception e) {
-                        throw new RuntimeException("Error getting id of new institution ", e);
+                        throw new DsmInternalError("Error getting id of new institution ", e);
```
denniscunningham: It looks like we are mixing two styles here, so if you are fixing the exceptions you should probably fix that too.
Typically we just capture exception here for processing outside of this block. See line 469.

Also, it look like there are no changes to this file other than linter fixes? If so, perhaps leave this for another day, since it may just cause conflicts with actual content changes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueDao.java
```diff
@@ -75,8 +78,30 @@ public Map<String, List<Tissue>> getTissuesByStudy(String study) {
...
+                try (ResultSet rs = stmt.executeQuery()) {
+                    while (rs.next()) {
+                        tissueIds.add(rs.getString(DBConstants.TISSUE_ID));
+                    }
+                }
+            } catch (SQLException ex) {
+                execResult.resultException = ex;
+            }
+            return execResult;
+        });
```
denniscunningham, line 101: Nit, but this can be simplified by moving the throw into the transaction clause and just returning the results from that clause directly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
denniscunningham: Why? It makes for confusing change and history makes reviewing the changes more difficult

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -124,7 +107,7 @@ public String createNewSMIDForTissue(String tissueId, String userId, @NonNull St
                 if (result == 1) {
                     try (ResultSet rs = stmt.getGeneratedKeys()) {
                         if (rs.next()) {
-                            logger.info("Created new sm id for tissue w/ id " + tissueId);
+                            logger.info(String.format("Created new sm id for tissue w/ id %s ", tissueId));
```
denniscunningham: As mentioned elsewhere, use the logger `{}` string substitution instead.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -123,6 +137,11 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
+                    throw new DuplicateException(dbElement.getColumnName());
```
denniscunningham: Where is this exception handled?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
denniscunningham: Nit: Unnecessary, despite what Sonar says.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -29,17 +37,26 @@ public static BasePatch makePatch(Patch patch, NotificationUtil notificationUtil
...
                 patcher = new ParticipantRecordPatch(patch);
             }
-        } else if (isParentParticipantId(patch) && !isMedicalRecordAbstractionFieldId(patch)
-                    && StringUtils.isNotBlank(patch.getDdpParticipantId())) {
+        } else if (isOncHistoryDetailPatch(patch)) {
             patcher = new OncHistoryDetailPatch(patch);
         }
         if (patcher instanceof NullPatch) {
-            throw new RuntimeException("Id and parentId was null");
+            throw new DSMBadRequestException("Id and parentId was null");
```
denniscunningham: Perhaps mention "patch" in this message?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -160,12 +179,13 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
         return true;
     }
 
-    public static Boolean isValueUnique(@NonNull DBElement dbElement) {
+    public static Boolean isValueUnique(@NonNull DBElement dbElement, String value) {
```
denniscunningham, line 187: This method needs some work since it appears (at least) exceptions are swallowed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -25,7 +25,7 @@
 @Data
 public class Patch {
 
-    public static final String SQL_CHECK_UNIQUE = "SELECT * FROM $table WHERE ($colName = ? ) and deleted <=> 0 ";
+    public static final String SQL_CHECK_UNIQUE = "SELECT * FROM $table WHERE ($colName = ? ) and deleted <> 1 ";
```
denniscunningham: Why this change? Is this for consistency with how we do this elsewhere?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -160,12 +179,13 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
     }
 
-    public static Boolean isValueUnique(@NonNull DBElement dbElement) {
+    public static Boolean isValueUnique(@NonNull DBElement dbElement, String value) {
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement stmt = conn.prepareStatement(
                     SQL_CHECK_UNIQUE.replace(TABLE, dbElement.getTableName()).replace(COL_NAME, dbElement.getColumnName()))) {
                 try {
+                    stmt.setString(1, value);
```
denniscunningham, line 193: So this never worked before? Can we build a test for this method?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/structure/UniqueField.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface UniqueField {
+    boolean value();
```
denniscunningham: The value does not seem to provide anything useful. Can you create an annotation without it?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
denniscunningham: You mean via Liquibase, rght?

denniscunningham (changes requested):
I have not completed my review, but there are some actionable comments. I also left some questions on the ticket.
More broadly, it is going to be very hard to approve this PR without automated tests or some other way of demonstrating that this code works properly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
pegahtah: It gets sent to frontend with a message and the ui shows it to the user to change the value. Look at PatchRoute:72

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
pegahtah: Changed it to `NOT deleted <=> 1` to be consistent

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
pegahtah: Sure!
No it was not even called before but it was there... waiting for someone to use it
I think if I go back into old repo and look at history I can find out what happened to this method

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
pegahtah: I agree about not being necessary but the style was not consistent, if you look into the variables defined above this line they have a blank line in between, that's why I added them. I am more than happy to change them all to without blank lines too. I think it added some ease of readability for all the annotations

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
pegahtah: I'll revert it so we can fix it better in future

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
pegahtah: Not intentionally, it was moved because I wrote a new method on top of it and removed another method  (`createNewSMIDForTissueWithValue`) entirely from this file so git kinda lost it and _thinks_ it was moved and renamed. I think its because they share some lines of code (like creating the transaction) and I'm not sure how to avoid that.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueDao.java
pegahtah: Correct, but I think as you mentioned (maybe in another PR) that is not the standard of how we do it. Maybe there is no standard and both patterns are in DSM, so we should decide on one and change all the other ones we see?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -402,7 +404,7 @@ public static Map<String, List<OncHistoryDetail>> getOncHistoryDetails(@NonNull
                     }
                     //add tissues to their onc history
                     for (Tissue tissue : tissues.values()) {
-                        long tissueOncHistoryDetailId = tissue.getOncHistoryDetailId();
+                        Integer tissueOncHistoryDetailId = tissue.getOncHistoryDetailId();
```
denniscunningham: This can be just an `int`, right?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
```diff
@@ -207,7 +212,8 @@ public static Tissue getTissue(@NonNull ResultSet rs) throws SQLException {
                 rs.getString(DBConstants.TISSUE_SEQUENCE), (Integer) rs.getObject(DBConstants.SCROLLS_COUNT),
                 (Integer) rs.getObject(DBConstants.USS_COUNT), (Integer) rs.getObject(DBConstants.BLOCKS_COUNT),
                 (Integer) rs.getObject(DBConstants.H_E_COUNT), rs.getString(DBConstants.BLOCK_ID_SHL),
-                new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
+                new ArrayList<>(), new ArrayList<>(), new ArrayList<>(),
+                rs.getInt(DBConstants.DDP_TISSUE_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.DELETED));
```
denniscunningham: This should probably be one constant since the name of this field should not vary, and in this case the term `DELIMITER` does not apply.
It is also impacts readability since the reader has to look up each of these constants to construct the name of the field.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/onchistory/OncHistoryDetailDaoImpl.java
```diff
@@ -143,7 +143,8 @@ public OncHistoryDetail getRandomOncHistoryDetail(@NonNull String oncHistoryDeta
                                         + DBConstants.ADDITIONAL_VALUES_JSON), null, rs.getString(DBConstants.TISSUE_PROBLEM_OPTION),
                                 rs.getString(DBConstants.DESTRUCTION_POLICY), rs.getBoolean(DBConstants.UNABLE_OBTAIN_TISSUE),
                                 rs.getString(DBConstants.PARTICIPANT_ID), rs.getString(DBConstants.DDP_PARTICIPANT_ID),
-                                rs.getLong(DBConstants.DDP_INSTANCE_ID));
+                                rs.getLong(DBConstants.DDP_INSTANCE_ID),
+                                rs.getInt(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.DELETED));
```
denniscunningham: See my comment above.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -225,9 +208,9 @@ public Map<String, List<SmId>> getSmIdsByParticipantByStudy(String instanceName)
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Couldn't get list of smIds for instance " + instanceName, results.resultException);
+            throw new DsmInternalError("Couldn't get list of smIds for instance " + instanceName, results.resultException);
         }
-        logger.info("Got " + smIds.size() + " participants smIds in DSM DB for " + instanceName);
+        logger.info(String.format("Got %d participants smIds in DSM DB for %s ", smIds.size(), instanceName));
```
denniscunningham: As I mentioned in the prior round of comments, for the logging methods `{}` is the correct way to do string substitution.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public static BasePatch produce(Patch patch, NotificationUtil notificationUtil) {
+        BasePatch patcher;
+        if (DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS.equals(patch.getTableAlias())) {
+            patcher = new DeleteOncHistoryPatch(patch, notificationUtil);
+        } else if (PatchFactory.isTissueRelatedOncHistoryId(patch)) {
+            patcher = new DeleteTissuePatch(patch, notificationUtil);
+        } else {
+            throw new NotImplementedException("This method should not reach here");
```
denniscunningham: This should probably be an DsmInternalError, to assure the proper HTTP response handling. Or you can add `NotImplementedException` to the Spark exception handling in `DsmServer`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
denniscunningham: You can just keep getTypeForName as the first method in this class, unless you feel strongly that getSmIdPksForTissue needs to be the first method in this class.
Accurate diffs for reviews are important.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/DuplicateException.java
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.dsm.exception;
+
+public class DuplicateException extends RuntimeException {
```
denniscunningham: We just got rid of this in a recent PR since error handling for this was ambiguous.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public Object doPatch() {
+        Object o = null;
+        try {
+            o = super.doPatch();
+        } finally {
+            DeletePatchFactory.setDeletedForChildrenFields(this.patch, notificationUtil);
+            return o;
```
denniscunningham, line 24: This return needs to be outside of the `finally` block.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.dsm.model.patch;
+
+import lombok.extern.slf4j.Slf4j;
+import org.broadinstitute.dsm.util.NotificationUtil;
+
+@Slf4j
+public class DeleteOncHistoryPatch extends ExistingOncHistoryPatch {
+    private final NotificationUtil notificationUtil;
```
denniscunningham: Why is there a separate instance of `NotificationUtil` for this class?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
```diff
@@ -0,0 +1,25 @@
...
+        this.notificationUtil = notificationUtil;
+        this.dbElementBuilder = new DefaultDBElementBuilder();
+    }
+
+    @Override
+    public Object doPatch() {
+        Object o = null;
+        try {
+            o = super.doPatch();
+        } finally {
```
denniscunningham: Is this the behavior you want?
It appears that you want to continue to delete things even if the upstream patch fails? Will that leave things dangling?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.commons.lang3.NotImplementedException;
+import org.broadinstitute.dsm.db.OncHistoryDetail;
+import org.broadinstitute.dsm.db.dao.ddp.tissue.TissueDao;
+import org.broadinstitute.dsm.db.dao.ddp.tissue.TissueSMIDDao;
+import org.broadinstitute.dsm.model.NameValue;
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.util.NotificationUtil;
+
+public class DeletePatchFactory {
+    public final static String ONC_HISTORY_DELETED_FIELD = "oD.deleted";
```
denniscunningham: Why are these constants public? (Also, the convention is `static final`.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,84 @@
...
+            patcher = new DeleteOncHistoryPatch(patch, notificationUtil);
+        } else if (PatchFactory.isTissueRelatedOncHistoryId(patch)) {
+            patcher = new DeleteTissuePatch(patch, notificationUtil);
+        } else {
+            throw new NotImplementedException("This method should not reach here");
+        }
+        return patcher;
+    }
+
+    protected static void setDeletedForChildrenFields(@NonNull Patch originalPatch, NotificationUtil notificationUtil) {
```
denniscunningham: This is not a setter -- it appears to actually delete things -- so `set...` is not a good name.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,84 @@
...
+        List<Patch> deletePatches = null;
+        if (isTissuePatch(originalPatch)) {
+            deletePatches = getPatchForSmIds(originalPatch);
+        } else if (isOncHistoryPatch(originalPatch)) {
+            deletePatches = getPatchForTissues(originalPatch);
+        }
+        for (Patch childPatch : deletePatches) {
```
denniscunningham, line 43: What happens when `deletePatches` is null?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,84 @@
...
+            BasePatch patcher = PatchFactory.makePatch(childPatch, notificationUtil);
+            patcher.doPatch();
+        }
+    }
+
+    private static boolean isOncHistoryPatch(Patch originalPatch) {
+        return originalPatch.getNameValue().getName().equals(ONC_HISTORY_DELETED_FIELD) &&
```
denniscunningham, line 50: Just curious, why do both of these fields need to be set? It seems like each would convey enough info on their own.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteTissuePatch.java
```diff
@@ -0,0 +1,26 @@
+package org.broadinstitute.dsm.model.patch;
+
+import lombok.extern.slf4j.Slf4j;
+import org.broadinstitute.dsm.util.NotificationUtil;
+
+@Slf4j
+public class DeleteTissuePatch extends ExistingRecordPatch {
```
denniscunningham, line 7: See my comments on `DeleteOncHistoryPatch`... same applies here since these are largely cut and paste.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
denniscunningham: That line does not seem to exist in this branch.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -185,6 +206,10 @@ public static Boolean isValueUnique(@NonNull DBElement dbElement) {
             }
             return dbVals;
         });
+        if (results.resultException != null) {
+            throw new RuntimeException(String.format("Error while checking the uniqueness of the value for {} ", dbElement.getColumnName()),
```
denniscunningham: This should be a `DsmInternalError`.
The `String.format` substitution character appears to be incorrect. (You can also achieve this with a simple in-line string append.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -29,17 +37,26 @@ public static BasePatch makePatch(Patch patch, NotificationUtil notificationUtil
...
         return patcher;
     }
 
+    private static boolean isOncHistoryDetailPatch(Patch patch) {
+        return isParentParticipantId(patch) && !isMedicalRecordAbstractionFieldId(patch)
+                && StringUtils.isNotBlank(patch.getDdpParticipantId());
+    }
+
+    public static boolean isDeletePatch(Patch patch) {
+        return patch.getNameValue().getName().contains(".deleted") &&
```
denniscunningham, line 57: Probably worth a comment about why a NameValue name needs to contain `.deleted`, since that is not apparent.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/TestParticipantUtil.java
```diff
@@ -0,0 +1,361 @@
...
+import org.broadinstitute.dsm.db.OncHistoryDetail;
+import org.broadinstitute.dsm.db.SmId;
+import org.broadinstitute.dsm.db.Tissue;
+import org.broadinstitute.dsm.db.dao.ddp.tissue.TissueSMIDDao;
+import org.broadinstitute.dsm.exception.DsmInternalError;
+import org.broadinstitute.dsm.service.admin.UserAdminTestUtil;
+import org.broadinstitute.lddp.db.SimpleResult;
+
+@Slf4j
+public class TestParticipantUtil {
```
denniscunningham: There is an existing `TestParticipantUtil` class that does similar operations. I suggest renaming this to avoid confusion.
To the extent there are similar operations/methods it may make sense for this class to inherit from the existing `TestParticipantUtil` class.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/TestParticipantUtil.java
```diff
@@ -0,0 +1,361 @@
...
+    private static final UserAdminTestUtil cmiAdminUtil = new UserAdminTestUtil();
+    private static final String INSERT_PARTICIPANT =
+            "INSERT INTO ddp_participant (ddp_participant_id, ddp_instance_id, last_version_date, last_changed) VALUES (?, ?, ?,?) ";
+    private static final String INSERT_INSTITUTION =
```
denniscunningham: There are existing non-test methods that create and delete medical records and institutions. By using those instead of creating new methods, those methods get the benefit of this testing, and there will be no drift if the underlying tables or relationships change.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+        String tissueSkIdUniquePatch = "{\"id\":" + tissueId + ",\"user\":\"" + testParticipantUtil.userEmail +
+                "\",\"nameValue\":{\"name\":\"t.skId\",\"value\":\"123\"},\"ddpParticipantId\":\""
+                + participantGuid + "\",\"parent\":\"oncHistoryDetailId\",\"parentId\":" + oncHistoryId +
+                ",\"tableAlias\":\"t\",\"isUnique\":true,\"realm\":\"" + instanceName + "\"}";
+        Patch patch = new Gson().fromJson(tissueSkIdUniquePatch, Patch.class);
+        BasePatch patcher = PatchFactory.makePatch(patch, null);
+        try {
+            patcher.doPatch();
+        } catch (Exception e){
+            log.error("", e);
```
denniscunningham, line 124: What's going on here? Seems like this could be an `Assert.fail`?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+        List<String> participantIds = new ArrayList<>();
+        participantIds.add(participantGuid);
+        OncHistoryDetail oncHistoryDetail = testParticipantUtil.getOncHistoryDetailById(oncHistoryId);
+        Assert.assertFalse(oncHistoryDetail.isDeleted());
+        Tissue tissue = oncHistoryDetail.getTissues().stream().findFirst().get();
+        Assert.assertEquals(String.valueOf(tissue.getTissueId()), tissueId);
+        SmId tissueSmId = tissue.getUssSMID().stream().filter(smId -> String.valueOf(smId.getSmIdPk()).equals(smIdPk)).findAny().get();
+        Assert.assertEquals(String.valueOf(tissueSmId.getTissueId()), tissueId);
+
+        String tissueSkIdUniquePatch = "{\"id\":" + tissueId + ",\"user\":\"" + testParticipantUtil.userEmail +
```
denniscunningham: You might want to explain what you are doing as you go along, at least in the testing parts. This looks like at this point you are testing patching as upserts rather than deletes?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        tissueSkIdUniquePatch = "{\"id\":" + tissueId + ",\"user\":\"" + testParticipantUtil.userEmail +
+                "\",\"nameValue\":{\"name\":\"t.skId\",\"value\":\"123\"},\"ddpParticipantId\":\""
+                + participantGuid + "\",\"parent\":\"oncHistoryDetailId\",\"parentId\":" + oncHistoryId +
+                ",\"tableAlias\":\"t\",\"isUnique\":true,\"realm\":\"" + instanceName + "\"}";
+        patch = new Gson().fromJson(tissueSkIdUniquePatch, Patch.class);
+        patcher = PatchFactory.makePatch(patch, null);
+        try {
+            patcher.doPatch();
```
denniscunningham, line 134: If you are expecting an exception here, then you should put an `Assert.fail` after the call to `doPatch`. Otherwise you will not know if this is actually working.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+            Assert.assertTrue(e instanceof DuplicateException);
+        }
+
+        String oncHistoryDeleteJson = "{\"id\":\"" + oncHistoryId + "\",\"user\":\"" + testParticipantUtil.userEmail
+                + "\",\"nameValue\":{\"name\":\"oD.deleted\",\"value\":true},\"ddpParticipantId\":" + participantGuid
+                + ",\"parent\":\"participantId\",\"parentId\":" + participantId
+                + ",\"tableAlias\":\"oD\",\"isUnique\":true,\"realm\":\"" + instanceName + "\"}";
+        Patch oncHistoryDeletePatch = new Gson().fromJson(oncHistoryDeleteJson, Patch.class);
+        patcher = DeletePatchFactory.produce(oncHistoryDeletePatch, null);
+        patcher.doPatch();
```
denniscunningham, line 145: Perhaps catch exceptions here since this is the primary code under test.

denniscunningham (changes requested):
Thanks for adding the test!

My main concern, as I detail in the comments, is with TestParticipantUtil, since it should be using existing methods instead of creating new methods to create and delete entities. As things are now we are missing opportunities to test existing methods, and creating maintenance issues to keep the proposed new code in sync. Also, since the new test methods are separate from the existing code, we cannot be sure they is creating things in the same way as existing non-test methods. 
If you are unclear on how to to that, you can look at existing test methods or I can walk you through some examples.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -62,7 +62,7 @@ public class OncHistoryDetail implements HasDdpInstanceId {
                     + "collaborator_sample_id, block_sent, scrolls_received, sk_id, sm_id, "
                     + "sent_gp, first_sm_id, additional_tissue_value_json, expected_return, return_date, return_fedex_id, "
                     + "shl_work_number, block_id_shl, tumor_percentage, tissue_sequence, "
-                    + "scrolls_count, uss_count, h_e_count, blocks_count, sm.sm_id_value, sm.sm_id_type_id, sm.sm_id_pk, sm.deleted, "
+                    + "scrolls_count, uss_count, h_e_count, blocks_count, t.deleted, sm.sm_id_value, sm.sm_id_type_id, sm.sm_id_pk, sm.deleted, "
```
zyme, line 67: Having multiple columns in the select with the same name (`deleted`) might make for runtime errors if code is addressing columns by column name vs. index.  Maybe change these to two different aliases and update `FieldSettingsDao.DELETED`?

zyme: I do see us doing things like `DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.DELETED`, which makes for some error prone result set fetching.  Maybe make some new constants like `DBConstants.DDP_ONC_HISTORY_DETAIL_DELETED` so that we don't need to remember to concat various things when looking for a column?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -297,6 +297,7 @@ public OncHistoryDetail(Integer oncHistoryDetailId, Integer medicalRecordId, Str
         this.participantId = participantId;
         this.ddpParticipantId = ddpParticipantId;
         this.ddpInstanceId = ddpInstanceId;
+        this.deleted =  (deleted != null && deleted == 1);
```
zyme: It would be great to backfill `deleted` and make sure it's always non-null so that we can use `int` (or better yet, `boolean`) and avoid null checks like this.

We could also pull out a `Integer.valueOf(1)` into a constant like `INTEGER_TRUE` and then simplify this with `INTEGER_TRUE.equals(deleted)`

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-798_drop_unique_index.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet author="pegah" id="PEPPER-798_drop_indices">
```
zyme, line 8: It's not clear to me why we have to drop the indexes here in order to have proper unique checks at the code level.  Dropping the indexes will have performance ramifications, and if we need to do any direct db work, the lack of the unique indexes here will increase the risk of dirty data.  Can we take another look at keeping the indexes (or changing to unique constraints) and enabling code-level unique checks?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/elastic/migration/BaseCollectionMigratorTest.java
```diff
@@ -30,18 +30,18 @@ public void transformObject() {
     private List mockOncHistoryDetail() {
         OncHistoryDetail oncHistoryDetail =
                 new OncHistoryDetail(23, 0, null, null, null, null, null, null, null, null, null, null, null, null, null,
-                        null, null, null, null, null, null, null, null, null, mockTissues(), null, null, false, null, null, 0);
+                        null, null, null, null, null, null, null, null, null, mockTissues(), null, null, false, null, null, 0, null);
         return Collections.singletonList(oncHistoryDetail);
     }
 
     private List<Tissue> mockTissues() {
         List<Tissue> fieldValue = new ArrayList<>(List.of(new Tissue(11, 22,
```
zyme, line 38: These constructors are really hard to use.  Can you create new constructors that only use the non-null constructors?  That will save many lines of code and make it easier for us to understand which fields are actually required in combination with (or absence of) other fields.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+        testParticipantUtil.deleteSmId(createdSmIdPks);
+        testParticipantUtil.deleteTissue(createdTissueIds);
+        List<String> participantIds = new ArrayList<>();
+        participantIds.add(participantGuid);
+        testParticipantUtil.deleteOncHistoryDetailAndMedicalRecord(createdOncHistoryIds, participantIds);
+    }
+
+    @Test
+    public void testIsDeletePatchMethod() {
+        String oncHistoryDeleteJson =
```
zyme, line 50: Another way to do this that might be more readable is to new up an instance of `Patch` code-wise, and skip all the escaped json to deserialization to get `oncHistoryDeletePatch` in the desired state.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SmId.java
```diff
@@ -37,7 +37,7 @@ public class SmId {
     @ColumnName(DBConstants.SM_ID_PK)
     private Integer smIdPk;
     @ColumnName(DBConstants.DELETED)
-    private Boolean deleted;
+    private boolean deleted;
```
zyme: I'm not sure it's safe to do this.  I see a bunch of `null` values in the `sm_id` field, and there's an existing constructor that attempts to unbox `Boolean -> boolean` on L54.  I think this change will result in an NPE somewhere.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/PatchUtil.java
```diff
@@ -121,6 +122,7 @@ private static void getColumnNames(Class<?> obj) {
             }
 
             ColumnName[] columnName = field.getAnnotationsByType(ColumnName.class);
+            UniqueField isUniqueField = field.getAnnotation(UniqueField.class);
```
zyme, line 125: I don't understand the intent and scope of `UniqueField`.  Can you say more about what is being checked for uniqueness?  Is this trying to check for uniqueness in an unbounded static map that grows with every unique field DSM tries to insert for however long the VM is running?  This doesn't appear to query the database, so the uniqueness is confined to the VM, and clients have to know to call `getDataBaseMap()` and check it in some way, so it looks like it's easy for client code to ignore this, except for `Patch`, which checks and throws an exception if something isn't unique in the static map.

There is also no synchronization around the static `PatchUtil` fields, which might lead to improper state in these maps under load.

I know this general approach predates the PR, but the lack of tests that show proper functioning of `PatchUtil`'s concepts of uniqueness makes me wonder if this is going to solve the issue, or if it's better to leave the indexes in place and look at `equals()` and `hashCode()` to help client code remove duplicates before persisting to the db.

Should we be thinking about an upsert approach in `Patch` that takes into account the `deleted` column?

At the end of the day, if something is marked as `unique` by all of this, the external behavior change is the throwing of `DuplicateException`, which we had removed from the codebase recently because nothing responds to it in production--it's an exception without any different handling (although I do see that a new test does check for it).

pegahtah: Of course, let me explain better: 
 > Can you say more about what is being checked for uniqueness?

The primary purpose of the `UniqueField` within our codebase is to ensure the uniqueness of fields while exempting records marked as 'deleted'. The motivation behind this feature is to enable reuse of data from deleted records without triggering SQL duplicate exceptions. To illustrate, consider a scenario where a CRC erroneously deletes a tissue entry. With this implementation, they should be able to recreate the tissue, reusing the same data, without any hitches. Consequently, to accommodate this functionality, our current unique indexes in the tables would need to be removed. The UniqueField annotation informs the Patch Service about fields that should maintain uniqueness across all undeleted data. Before any insertion, the system verifies whether the data already exists, acting as a "soft check" on the field's uniqueness.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/TestParticipantUtil.java
```diff
@@ -0,0 +1,361 @@
...
+            }
+        }
+    }
+
+    public void deleteTissue(List<String> tissueIds) {
+        for (String tissueId : tissueIds) {
+            try {
+                delete("ddp_tissue", "tissue_id", tissueId);
+            } catch (Exception e) {
+                log.error("unable to delete tissue {}", tissueId, e);
```
zyme: We should probably be throwing exceptions here, rather than silently logging.  Client code will not know that something has gone wrong and they probably should.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/PatchUtil.java
pegahtah: > Is this trying to check for uniqueness in an unbounded static map that grows with every unique field DSM tries to insert for however long the VM is running?

UniqueField employs a static map initialized during server startup. This map plays a crucial role in all patch requests, serving primarily to associate frontend field names with their corresponding backend columns and tables. The incorporation of the UniqueField annotation merely adds an extra characteristic to this existing mapping process.

The uniqueness check operates through this static map. If an entry corresponds to a field marked with this annotation, the system prompts Patch to verify the value's presence among all "not deleted" records—effectively excluding any records marked with deleted=true—before proceeding with an insertion.

It's important to note that this approach shouldn't impose additional load on the server. The static map is a foundational component that’s already employed for other operations, and the UniqueField simply integrates an extra layer of validation within this existing framework. Consequently, while we're enhancing data integrity by ensuring uniqueness where required, the overhead is negligible because we're capitalizing on pre-existing infrastructure (the static map) that's fundamental to the system’s operation.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
zyme: Might be worth breaking some of these longer tests into separate test methods so it's easier to see what the blocks of tests are doing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
pegahtah: My thoughts here was that if the ES upsert fails, we should still continue deleting other "child" records in db so  the patch and delete is completely done, but now I see your point, I'll make changes so we differentiate between SQL and ES exceptions

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeleteOncHistoryPatch.java
pegahtah: Can you explain the difference a bit, from my understanding everything in the finally block is always executed ( in this case it means there is no difference between where the return is)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -248,9 +248,18 @@ private void addWrapperToList(ElasticSearchParticipantDto elasticSearchParticipa
...
+            }
             List<SomaticResultUpload> somaticResultUpload = esDsm.getSomaticResultUpload();
             List<Tissue> tissues = esDsm.getTissue();
+            if (tissues != null) {
+                tissues.removeIf(tissue -> tissue.isDeleted() == true);
+            }
             List<SmId> smIds = esDsm.getSmId();
+            if (smIds != null) {
+                smIds.removeIf(smId -> smId.isDeleted() == true);
+            }
```
pegahtah, line 263: @zyme @denniscunningham thank you for all of your comments and review! While I'm working on them, do you have any concerns regarding these changes? I added them based on my conversation with Zim, where we talked about having a guard in the backend  to make sure we are not sending deleted records to the frontend (instead of frontend having all the control of what to show and what to hide from user). 

Based on the java docs, the `removeIf` is O(n) and at one time DSM will only have one participant's data to process in the `elasticSearchParticipantDto`. However I would love your feedback about if this is a good change in prod based on the current environment settings and the usage.

Alternatives to this are : 
1. not adding deleted data to the ES from the beginning ( changing the upsert and the export code and rebuilding the indices) 
2. changing all of the filtering and query classes to add the condition check to all of the queries we send to ES. This is very scary and would be a big change and bigger to test.

I appreciate your feedback on this

denniscunningham: @pegahtah Are you talking about a specific commit, or just the code in ParticipantWrapper?

pegahtah: @denniscunningham specifically this code in the `ParticipantWrapper` class. streaming the array of tissues and onc histories and removing them on the spot if they are deleted

denniscunningham: Makes sense to me, and I agree that it is better to do this here since we have no cases where the frontend wants to deal with deleted ptp data.
I'm not worried about the performance. The lists should be small, and I'd bet any perf issues we have would swamp whatever this adds.

What else consumes ES ptp records and will have to detected deleted entities?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
pegahtah: Fixed!! Sorry I must have had older versions of code

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/DeletePatchTest.java
pegahtah: Thanks that's a great idea!

pegahtah:
Change of plans, will not merge this

PR #2683, opened 2023-09-25
Moving DSM emails to the production sendgrid account

zyme:
PEPPER-784 Centralized some sendgrid-related code so that DSM can use the egress proxy as part of switching to the production sendgrid account.

DSM uses sendgrid utilities that are left over from the LDDP days, and were not aware of the egress proxy that we use as part of our IP allow list in production.  The main change here is the creation of `SendGridFactory`, a simple factory that creates `SendGrid` instances that are proxy-aware.  Most of the code in this class was factored out of DSS's `SendGridMailUtil`.

DSM is still relying on the `SendGridClient` class, which is the main holdover from LDDP.  This class had some state management complexity that required callers to set state and then queue emails.  I've removed this rules-of-use type code and replaced it with parameters to existing methods.

## FUD Score

- [ x] :sweat_smile: There might be some issues here

@pegahtah I'll need some help with exercising these changes, maybe adjusting some quartz jobs and getting data into the right state in DSM for it to send us emails in non-prod.

## Testing

@ssettipalli we'll want to turn on email sending briefly to verify that DSS is sending emails properly with these changes. 

## Release

In addition to the code here, we will need to change the sendgrid API key in DSM secret manager, and update the template ids as per [this sheet](https://docs.google.com/spreadsheets/d/1rx0W6KNmNoS2ab8P2KLbKAji7B6gAanhJ9D0Mn9CHmU/edit#gid=0) that shows the current template ids in dev and the new template ids in prod.

### DIFF
--- a/pepper-apis/ddp-common/pom.xml
+++ b/pepper-apis/ddp-common/pom.xml
@@ -101,5 +101,12 @@
             <artifactId>mysql</artifactId>
             <scope>compile</scope>
         </dependency>
+        <!-- sendgrid for email sending -->
+        <dependency>
+            <!-- https://mvnrepository.com/artifact/com.sendgrid/sendgrid-java -->
+            <groupId>com.sendgrid</groupId>
+            <artifactId>sendgrid-java</artifactId>
+        </dependency>
+
     </dependencies>
 </project>
--- a/pepper-apis/ddp-common/pom.xml
+++ b/pepper-apis/ddp-common/pom.xml
@@ -106,6 +106,7 @@
             <!-- https://mvnrepository.com/artifact/com.sendgrid/sendgrid-java -->
             <groupId>com.sendgrid</groupId>
             <artifactId>sendgrid-java</artifactId>
+            <version>4.8.1</version>
         </dependency>
 
     </dependencies>
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/SendGridFactory.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/SendGridFactory.java
@@ -1,0 +1,41 @@
+package org.broadinstitute.ddp.util;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import com.sendgrid.Client;
+import com.sendgrid.SendGrid;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.http.HttpHost;
+import org.apache.http.impl.NoConnectionReuseStrategy;
+import org.apache.http.impl.client.HttpClients;
+
+/**
+ * Central factory for creating {@link SendGrid} instances in
+ * DSM and DSS.
+ */
+@Slf4j
+public class SendGridFactory {
+
+    /**
+     * Creates a new SendGrid instance using the proxy url for
+     * outbound egress if not null.
+     */
+    public static SendGrid createSendGridInstance(String sendGridApiKey, String proxy) {
+        var httpClientBuilder = HttpClients.custom();
+        if (proxy != null && !proxy.isBlank()) {
+            URL proxyUrl;
+            try {
+                proxyUrl = new URL(proxy);
+            } catch (MalformedURLException e) {
+                throw new IllegalArgumentException("proxy needs to be a valid url");
+            }
+            httpClientBuilder.setProxy(new HttpHost(proxyUrl.getHost(), proxyUrl.getPort(), proxyUrl.getProtocol()));
+            httpClientBuilder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
+            log.info("Using SendGrid proxy: {}", proxy);
+        }
+        var client = new Client(httpClientBuilder.build());
+        return new SendGrid(sendGridApiKey, client);
+    }
+
+}
--- a/pepper-apis/dsm-core/pom.xml
+++ b/pepper-apis/dsm-core/pom.xml
@@ -255,16 +255,6 @@
             <artifactId>google-analytics-java</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.sendgrid</groupId>
-            <artifactId>sendgrid-java</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.httpcomponents</groupId>
-                    <artifactId>httpcore</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
             <groupId>net.minidev</groupId>
             <artifactId>json-smart</artifactId>
             <version>2.4.11</version>
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
@@ -11,10 +11,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
+import com.google.gson.*;
 import com.typesafe.config.Config;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
@@ -83,8 +80,8 @@ public class NotificationUtil {
     public synchronized void startup(@NonNull Config config) {
         emailKey = config.getString(ApplicationConfigConstants.EMAIL_KEY);
         emailProxyUrl = config.getString(ApplicationConfigConstants.EMAIL_PROXY_URL);
-        JsonObject senderInfo = (JsonObject) (new JsonParser().parse(config.getString(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS)));
-        emailSender = new EmailSender(senderInfo.get("sendGridFrom").getAsString(), senderInfo.get("sendGridFromName").getAsString());
+
+        new Gson().fromJson(config.getString(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS), EmailSender.class);
         logger.info("Will send email from " + emailSender.getFromEmailAddress() + " " + emailSender.getFromName());
 
         JsonArray array = (JsonArray) (new JsonParser().parse(config.getString(ApplicationConfigConstants.EMAIL_NOTIFICATIONS)));
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
@@ -11,7 +11,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
@@ -29,6 +28,7 @@ import org.broadinstitute.dsm.statics.ApplicationConfigConstants;
 import org.broadinstitute.dsm.statics.DBConstants;
 import org.broadinstitute.lddp.db.SimpleResult;
 import org.broadinstitute.lddp.email.EmailRecord;
+import org.broadinstitute.lddp.email.EmailSender;
 import org.broadinstitute.lddp.email.Recipient;
 import org.broadinstitute.lddp.email.SendGridClient;
 import org.broadinstitute.lddp.exception.EmailQueueException;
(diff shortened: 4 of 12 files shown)


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/SendGridFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * DSM and DSS.
+ */
+@Slf4j
+public class SendGridFactory {
+
+    /**
+     * Creates a new SendGrid instance using the proxy url for
+     * outbound egress if not null.
+     */
+    public static SendGrid createSendGridInstance(String sendGridApiKey, String proxy) {
```
zyme, line 24: Moved from [SendGridMailUtil](https://github.com/broadinstitute/ddp-study-server/pull/2683/files#r1335942595).

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/util/SendGridMailUtil.java
```diff
@@ -96,21 +90,4 @@ private static void submitMailMessage(Mail mailMessage, String sendGridApiKey, S
             log.info("Successfully submitted message to SendGrid with subject {}", mailMessage.subject);
         }
     }
-
-    private static SendGrid createSendGridInstance(String sendGridApiKey, String proxy) {
```
zyme, line 100: Moved to [SendGridFactory](https://github.com/broadinstitute/ddp-study-server/pull/2683/files#r1335942101).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -20,29 +20,27 @@
...
 
     private static final Logger logger = LoggerFactory.getLogger(SendGridClient.class);
 
     private static final String LOG_PREFIX = "SendGrid";
+    private final SendGrid sendGrid;
 
-    private String sendGridKey;
-    private JsonObject settings = null;
-
-    public void configure(String sendGridKey, JsonObject settings) {
```
zyme, line 34: Calls to `configure()` have been replaced by adding `settings` params to various methods.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -20,29 +20,27 @@
...
-        this.sendGridKey = sendGridKey;
-        this.settings = settings;
+    public SendGridClient(SendGrid sendGrid) {
```
zyme, line 33: Instead of passing in key and settings values and mutating state via `configure()`, we just pass in a fully formed `SendGrid` object and then clients pass in `settings` as needed when sending emails.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/NotificationJobTest.java
```diff
@@ -133,12 +133,8 @@ public void getUnsentKits() {
     }
 
     @Test
-    @Ignore("Throws that error when run at ALL Tests, not if run alone! Caused by: java.security.cert.CertificateException: "
```
zyme, line 136: I think this test really doesn't test anything.  I've uncommented it to see what happens.  It might better be removed at this point, but let's see what happens in circleCI.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -20,29 +20,27 @@
...
+        this.sendGrid = sendGrid;
     }
 
     /**
      * Used to send emails using SendGrid templates.
      */
-    public void sendSingleEmail(@NonNull String sendGridTemplate, @NonNull Recipient recipient) {
+    public void sendSingleEmail(@NonNull String sendGridTemplate, @NonNull Recipient recipient, JsonObject settings) {
```
denniscunningham: Since JsonObject does not tell the reader anything about the expected content, a JavaDoc explanation would be helpful. Also, right below this, ArrrayList on the left hand side should be a List.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -67,10 +65,11 @@ private void sendEmail(@NonNull String sendGridTemplate, @NonNull Collection<Rec
     /**
      * Sends a request per recipient to SendGrid.
      */
-    private void sendRequestUsingTemplate(@NonNull Recipient recipient, @NonNull String sendGridTemplate, @NonNull SendGrid sendGrid)
+    private void sendRequestUsingTemplate(@NonNull Recipient recipient, @NonNull String sendGridTemplate,
+                                          @NonNull SendGrid sendGrid, JsonObject settings)
```
denniscunningham: Are we passing an instance variable to a private method (SendGrid)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
```diff
@@ -202,12 +205,12 @@ public void sentAbstractionExpertQuestion(@NonNull String from, @NonNull String
...
         try {
-            SendGridClient sendGridClient = new SendGridClient();
+            SendGridClient sendGridClient = new SendGridClient(SendGridFactory.createSendGridInstance(emailKey, emailProxyUrl));
             JsonObject emailClientSettings = new JsonObject();
             emailClientSettings.addProperty("sendGridFrom", from);
             emailClientSettings.addProperty("sendGridFromName", name);
-            sendGridClient.configure(emailKey, new Gson().fromJson(emailClientSettings.toString(), JsonObject.class));
-            sendGridClient.sendSingleEmail(sendGridTemplate, emailRecipient);
+            sendGridClient.sendSingleEmail(sendGridTemplate, emailRecipient,
+                    new Gson().fromJson(emailClientSettings.toString(), JsonObject.class));
```
denniscunningham: Ideally `emailClientSettings` would be a class, so the expected content would be obvious and type checked.

pepper-apis/ddp-common/pom.xml
```diff
@@ -101,5 +101,12 @@
             <artifactId>mysql</artifactId>
             <scope>compile</scope>
         </dependency>
+        <!-- sendgrid for email sending -->
+        <dependency>
+            <!-- https://mvnrepository.com/artifact/com.sendgrid/sendgrid-java -->
+            <groupId>com.sendgrid</groupId>
+            <artifactId>sendgrid-java</artifactId>
+        </dependency>
```
denniscunningham, line 110: Can we remove this dependency from other POM files?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
```diff
@@ -202,12 +205,12 @@ public void sentAbstractionExpertQuestion(@NonNull String from, @NonNull String
...
         } catch (Exception ex) {
             logger.error("An error occurred trying to send abstraction expert question.", ex);
```
denniscunningham, line 210: Is this error message always correct? Is there a reason we are logging an error rather than throwing? (I don't know the calling context.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
```diff
@@ -51,13 +49,13 @@ public void sendSingleEmail(@NonNull String sendGridTemplate, @NonNull Recipient
...
+                           JsonObject settings) {
         try {
             if (recipientList.size() == 0) {
                 throw new IllegalArgumentException("RecipientList cannot be empty.");
             }
-            SendGrid sendGrid = new SendGrid(sendGridKey);
-            sendRequestUsingTemplate((Recipient) (recipientList.toArray())[0], sendGridTemplate, sendGrid);
+            sendRequestUsingTemplate((Recipient) (recipientList.toArray())[0], sendGridTemplate, sendGrid, settings);
         } catch (Exception ex) {
             throw new NotificationSentException("An error occurred trying to send emails. " + ex.getMessage());
```
denniscunningham: Seems like `NotificationSentException` is kind of useless, since nothing appears to catch it. Is this likely an internal error (500)? If so, DsmInternalError is a better choice.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
zyme: Not for long! :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
zyme: Agreed, I don't see anything responding to NotificationSentException.  Will change.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/email/SendGridClient.java
zyme: I think the only thing being pulled out of `settings` is the from email and the from name.  I can create a separate class so this is more clear.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
pegahtah: I think this error message is only correct for one of the usecases of the emails, not all. Most probably we forgot to update this while new usecases were being added to Notification class

zyme: I'm slightly terrified of changing this from logging to a throw since this gets called from a lot of patch-related endpoints.

pepper-apis/ddp-common/pom.xml
zyme: Done!

pepper-apis/dsm-core/pom.xml
```diff
@@ -254,16 +254,6 @@
             <groupId>com.brsanthu</groupId>
             <artifactId>google-analytics-java</artifactId>
         </dependency>
-        <dependency>
```
zyme, line 257: No longer needed since sendgrid is imported from ddp-common.

pepper-apis/dss-core/pom.xml
```diff
@@ -91,13 +91,6 @@
 
         </dependency>
 
-        <!-- sendgrid for email sending -->
```
zyme, line 94: No longer needed since sendgrid is imported from ddp-common.

pepper-apis/parent-pom.xml
```diff
@@ -111,14 +111,6 @@
 
         </dependency>
 
-        <!-- sendgrid for email sending -->
```
zyme, line 114: No longer needed since sendgrid is imported from ddp-common.

pepper-apis/pom.xml
```diff
@@ -191,14 +191,6 @@
                 <version>2.3.0</version>
             </dependency>
 
-            <!-- sendgrid for email sending -->
```
zyme, line 194: No longer needed since sendgrid is imported from ddp-common.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
```diff
@@ -80,7 +82,10 @@ public static Long getLastChanged(@NonNull String participantId) {
 
     public synchronized void startup(@NonNull Config config) {
         emailKey = config.getString(ApplicationConfigConstants.EMAIL_KEY);
-        emailClientSettings = (JsonObject) (new JsonParser().parse(config.getString(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS)));
+        emailProxyUrl = config.getString(ApplicationConfigConstants.EMAIL_PROXY_URL);
+        JsonObject senderInfo = (JsonObject) (new JsonParser().parse(config.getString(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS)));
+        emailSender = new EmailSender(senderInfo.get("sendGridFrom").getAsString(), senderInfo.get("sendGridFromName").getAsString());
```
denniscunningham: Nit: You can skip the JsonObject stuff by deserializing `EMAIL_CLIENT_SETTINGS` directly into `EmailSender`, like this: `Gson().fromJson(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS, EmailSender.class)`.

denniscunningham (approved):
Looks good. 
Left one suggestion.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
zyme: I think it's a little different:
`new Gson().fromJson(config.getString(ApplicationConfigConstants.EMAIL_CLIENT_SETTINGS), EmailSender.class);`

And for good measure, I'll add `@SerializedName` annotations to `EmailSender`

PR #2685, opened 2023-09-25
[DSM] PEPPER-1057 added extra group role

pegahtah:
I forgot to include this in the original PR.


denniscunningham (approved):
I don't think I've updated a change set file before, but it should work?

pegahtah:
I _think_ as long as it has it's own changeset it should work

PR #2686, opened 2023-09-26
[DSM] Fix for Mercury's long status message breaking while ES update

pegahtah:
Kiara caught this bug that when the code chooses to only update the database with the first 2000 characters in the details it is not reflected in the ES upsert. That's because the string we use in update is not changed in the original baseMercuryStatusMessage object and we still use the original object to upsert.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -89,6 +89,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                 int result = stmt.executeUpdate();
                 if (result != 0) {
                     log.info("Updated Mercury status for order id " + mercuryStatusMessage.getOrderID());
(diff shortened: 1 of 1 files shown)


Rinchan20 (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -89,6 +89,7 @@ public static void updateOrderStatus(BaseMercuryStatusMessage baseMercuryStatusM
                 int result = stmt.executeUpdate();
                 if (result != 0) {
                     log.info("Updated Mercury status for order id " + mercuryStatusMessage.getOrderID());
+                    baseMercuryStatusMessage.getStatus().setDetails(statusDetail);
```
denniscunningham, line 92: Would it make more sense to update this right after line 77? This feels like a non-sequitur.

Also, for the for the error cases below the error handling seems wrong. It looks like in one case we proceed with the second query after the first one fails, and what is the intention with the code just below? We log the error but proceed with the second query and the ES update? That case probably needs a comment if it is indeed the correct behavior.

PR #2691, opened 2023-09-28
PEPPER-1152 completely removing slack appender from dss and dsm.

zyme:
PEPPER-1152 completely removes all `SlackAppender`-related code and config from DSM and DSS.  The class is gone, as is all uses of it in production code and tests.  Logback config files have been purged of slackappender as well.

I've been using this script locally to reproduce the slackappender-induced lockup and validate that these changes eliminate the problem.  When you run this locally outside of this branch to reproduce the issue, be sure to use a config file that has all the slack stuff enabled.  Once things lock up, you can see where the thread locks show up.  Running the script on this branch results in proper logging and no blocking threads.

```
package org.broadinstitute.ddp.log;

import lombok.extern.slf4j.Slf4j;

@Slf4j
public class SlackAppenderError {

        new Thread(new Runnable() {
            @Override
            public void run() {
                for (;;) {
                    log.error("Here's an error");
                }
            }
        }).start();

        new Thread(new Runnable() {
            @Override
            public void run() {
                for (;;) {
                    log.info("this is info");
                }
            }
        }).start();

    }
}

```

## FUD Score

## How do we test and demo these changes?

Run housekeeping and watch it *NOT* completely lock up due to synchronization locks when API calls to slack fail.

## Release

Although these code changes will give us the desired stability, we also should remove slack stanzas in vault and DSM's secret manager, and then remove the API keys from slack itself.


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -1112,33 +1110,6 @@ private void setupRouteGenericErrorHandlers() {
         });
     }
 
-    protected void setupErrorNotifications(Config config, String schedulerName) {
```
zyme, line 1115: No longer needed since all this method does is set up slackappender alerting.

ssettipalli:
looks good to me

pegahtah (approved):
LGTM, my only concern is if we have to change something on GCP so we can still see the errors in the slack

zyme:
@pegahtah I've turned on GCP based slack alerting on prod, so we should start seeing more traffic there as of now.

PR #2696, opened 2023-09-29
PEPPER-784 proxyUrl config value is optional

zyme:
Follow-up for PEPPER-784.  The proxyUrl config value is not required.  DSM is failing to boot because the config value is currently required and it's not present.  We only use this value in production.  In order to test the behavior when the field is present, we will need to enable it in non-prod and ensure that the proxy works properly.  But first, let's ket DSM to a bootable state.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
@@ -11,10 +11,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
+import com.google.gson.*;
 import com.typesafe.config.Config;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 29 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/NotificationUtil.java
```diff
@@ -79,8 +79,9 @@ public static Long getLastChanged(@NonNull String participantId) {
 
     public synchronized void startup(@NonNull Config config) {
         emailKey = config.getString(ApplicationConfigConstants.EMAIL_KEY);
-        emailProxyUrl = config.getString(ApplicationConfigConstants.EMAIL_PROXY_URL);
-
+        if (config.hasPath(ApplicationConfigConstants.EMAIL_PROXY_URL)) {
+            emailProxyUrl = config.getString(ApplicationConfigConstants.EMAIL_PROXY_URL);
```
denniscunningham, line 83: What happens elsewhere in the code if `emailProxyUrl` is not set?

zyme: Nothing--the part of the code that responds to its null-ness is unchanged.  See [SendGridFactory](https://github.com/broadinstitute/ddp-study-server/blob/eedc468244e942c6e8dd0b29ccdc11007478bb01/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/SendGridFactory.java) for details.

ssettipalli (approved):

denniscunningham (approved):

PR #2699, opened 2023-09-29
Elasticsearch container for automated DSM testing

denniscunningham:
## Context
PEPPER-855

The fundamental feature of this PR is to enable JUnit testing with Elasticsearch via test containers, but it also contains two tests that use Elastic containers, and several collateral changes to support those tests. It also includes code cleanup in areas directly used by the example tests. 

The PR includes:
- The Elastic container infrastructure, along with abstract test classes that allow the container to be used with or without the DB test container. This includes:
    - Code to setup an Elastic container, ElasticTestContainer
    - An Elastic container to use with CircleCI, since test containers don't really work there.
    - Restructuring client initialization in ElasticSearchUtil
    - Adding ElasticBaseTest, DbAndElasticBaseTest
    - A smoke test for Elastic test containers, ElasticContainerTest
    - Modifying DBTestUtil to accept an ES index
- Two tests that use an Elastic container: `OncHistoryDetailTest` and `OncHistoryDetailPatchTest`. The latter is a starting point for building automated tests to debug and verify fixes for `OncHistoryDetailPatch`.
- Created `ElasticTestUtil` to support ES testing. 
- Added a handful of test resources to support ES tests.
- Cleaned up the interfaces for `ElasticSearchUtil.getClientForElasticsearchCloudCF` `ElasticSearchUtil. fetchESDataByParticipantId `, `ElasticSearchUtil.getParticipantESDataByParticipantId` and `ElasticSearch. parseSourceMap`, which removed complexity from several callers. (Covid19OrderRegistrar, BSPDummyKitDao, ClinicalKitDto, BasicDefaultDataMaker, ReferralSourceService, ElasticSearch, GBFOrderFinder, CreateClinicalDummyKitRoute, ElasticSearchTest, ParticipantUtil, TestHelper)
- Added ElasticSearchUtil.updateParticipant to do updates via Json string.
- Cleaned up client initialization and retrieval in `ElasticSearchUtil`.
- Created a single base ES search method in `ElasticSearchUtil` to unify behavior and centralize error handling and logging.
- Updated the base ES update method `ElasticSearchUtil.doUpdate` to use IMMEDIATE refresh policy, since we've seen 
- Updated `MedicalRecord` and `OncHistoryDetail` to support tests. Also removed some unneeded complexity from the latter. 
- Added logging in several files to help with test development. (Including DDPInstitutionDao.java, MedicalRecordDao.java, OncHistoryDetailDaoImpl.java, ElasticDataExportAdapter.java, OncHistoryDetailPatch.java, Patch.java)
- Added delete implementation to ParticipantRecordDao.
- Provided an alternative constructor for UpsertPainless.
- Simplified the syntax in DDPMedicalRecordDataRequest.
- version conflict errors when the refresh lags. (This becomes more obvious when writing ES tests.)
- Improved error handling in `MedicalRecordUtil`.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/db/OncHistoryDetailTest.java
```diff
@@ -51,46 +58,89 @@ public void deleteParticipants() {
...
             OncHistoryDetailDto updateRec1 = oncHistoryDetailDao.get(recId1).orElseThrow();
             Assert.assertEquals("5", updateRec1.getColumnValues().get("destruction_policy"));
             OncHistoryDetailDto updateRec2 = oncHistoryDetailDao.get(recId2).orElseThrow();
             Assert.assertEquals("5", updateRec2.getColumnValues().get("destruction_policy"));
             OncHistoryDetailDto updateRec3 = oncHistoryDetailDao.get(recId3).orElseThrow();
             Assert.assertEquals("3", updateRec3.getColumnValues().get("destruction_policy"));
+
+            ElasticSearchParticipantDto esParticipant =
+                    ElasticSearchUtil.getParticipantESDataByParticipantId(esIndex, ddpParticipantId);
+            Dsm dsm = esParticipant.getDsm().orElseThrow();
```
zyme, line 125: Huzzah!  Great to see ES utils under test!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -0,0 +1,208 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
+            Assert.fail("Unexpected exception creating index: " + e);
+        }
+        return indexName;
+    }
+
+    public static void deleteIndex(String esIndex) {
+        try {
+            RestHighLevelClient client = ElasticSearchUtil.getClientInstance();
```
zyme, line 69: Do we want some additional safeguards here so that destructive operations like wiping out an index can only be done on disposable ES containers?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
```diff
@@ -108,12 +108,14 @@ public long getTotalCount() {
         return totalCount;
     }
 
-    public Optional<ElasticSearchParticipantDto> parseSourceMap(Map<String, Object> sourceMap) {
-        if (sourceMap == null) {
-            return Optional.of(new ElasticSearchParticipantDto.Builder().build());
+    public ElasticSearchParticipantDto parseSourceMap(Map<String, Object> sourceMap) {
```
zyme, line 111: Thank you for simplifying the optionals!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -220,13 +218,10 @@ public Object handle(Request request, Response response) {
...
 
-    private boolean participantIsEnrolled(
-            Optional<ElasticSearchParticipantDto> maybeParticipantByParticipantId) {
-        if (maybeParticipantByParticipantId.orElseThrow().getStatus().isEmpty()) {
+    private boolean participantIsEnrolled(ElasticSearchParticipantDto esParticipantDto) {
+        if (esParticipantDto.getStatus().isEmpty()) {
             return false;
         }
-        return maybeParticipantByParticipantId.orElseThrow().getStatus().get().equals("ENROLLED");
+        return esParticipantDto.getStatus().get().equals("ENROLLED");
```
zyme, line 225: Nit: I prefer yoda conditionals, so this would be:

```
"ENROLLED".equals(esParticipantDto.getStatus().get())
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -143,15 +147,17 @@ private static void initialize() {
         }
     }
 
-    public static synchronized void initClient() {
+    public static void initClient() {
```
zyme, line 150: Safe to remove synchronization because this is only called from already-synchronized 'initialize()'

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -143,15 +147,17 @@ private static void initialize() {
...
-            } catch (MalformedURLException e) {
-                throw new DsmInternalError("Error while initializing ES client", e);
-            }
+            initClient(DSMConfig.getSqlFromConfig(ApplicationConfigConstants.ES_URL),
+                    DSMConfig.getSqlFromConfig(ApplicationConfigConstants.ES_USERNAME),
+                    DSMConfig.getSqlFromConfig(ApplicationConfigConstants.ES_PASSWORD), getProxy());
+        }
+    }
+
+    public static synchronized void initClient(String url, String username, String password, String proxy) {
```
zyme, line 158: Maybe some javadoc  and or a log entry that tells the caller about blank password implying local/test instance?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -173,51 +179,67 @@ public static RestHighLevelClient getClientInstance() {
...
-        URL url = new URL(baseUrl);
+    public static RestHighLevelClient getClientForElasticsearchCloudCF(@NonNull String baseUrl, @NonNull String userName,
+                                                                       @NonNull String password, String proxy) {
         return getClientForElasticsearchCloud(baseUrl, userName, password, proxy);
     }
 
+    protected static SearchResponse search(SearchRequest searchRequest) {
+        try {
+            SearchResponse res = client.search(searchRequest, RequestOptions.DEFAULT);
+            logger.info("ES search returned {}", res.getHits().getTotalHits());
```
zyme, line 250: Maybe add "hits" to the log entry so readers know it's units and not a return code.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -706,17 +711,22 @@ private static void updateOrCreateMap(@NonNull Object id, @NonNull String object
...
+                                 RestHighLevelClient client) {
         String participantId = ParticipantUtil.isGuid(ddpParticipantId) ? ddpParticipantId :
                 getParticipantESDataByAltpid(client, index, ddpParticipantId).getProfile().map(Profile::getGuid).orElse(ddpParticipantId);
         UpdateRequest updateRequest =
-                new UpdateRequest().index(index).type("_doc").id(participantId).doc(objectsMapES).docAsUpsert(true).retryOnConflict(5);
+                new UpdateRequest().index(index).type("_doc")
+                        .id(participantId)
+                        .doc(objectsMapES)
+                        .docAsUpsert(true)
+                        .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
```
zyme, line 743: Could you say more about the choice of refresh policy here?  In the PR description, I think something got cut off:

> Updated the base ES update method ElasticSearchUtil.doUpdate to use IMMEDIATE refresh policy, since we've seen

pepper-apis/pom.xml
```diff
@@ -521,7 +521,13 @@
             <dependency>
                 <groupId>org.testcontainers</groupId>
                 <artifactId>mysql</artifactId>
-                <version>1.17.6</version>
+                <version>1.18.1</version>
```
zyme, line 524: Looks like we set the mysql version via `DatabaseTestUtil.MYSQL_VERSION`, separate from the version of testcontainers, so no concerns here.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -0,0 +1,208 @@
...
+import org.elasticsearch.xcontent.XContentType;
+import org.junit.Assert;
+
+@Slf4j
+public class ElasticTestUtil {
+
+    public static String createIndexWithMappings(String realm, String mappingsFile) {
+        String indexName = createIndex(realm);
+        try {
+            String mappingsJson = TestUtil.readFile(mappingsFile);
```
zyme, line 42: @ssettipalli what can we do here to ensure that when we redefine indexes during deployments, we make the appropriate changes here to our tests so the index structure doesn't drift from reality?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestContainer.java
```diff
@@ -0,0 +1,88 @@
...
+                    cfg.getString(ConfigFile.ELASTICSEARCH_USERNAME),
+                    cfg.getString(ConfigFile.ELASTICSEARCH_PASSWORD), null);
+            initialized = true;
+            return;
+        }
+        log.info("Using ElasticSearch test container");
+
+        container = new ElasticsearchContainer(ELASTIC_IMAGE)
+                .withExposedPorts(9200)
+                .withEnv("xpack.security.enabled", "false")
```
zyme, line 45: Can you drop a comment here and in the circle yaml so everyone knows that this is a deliberate choice *just for testing*?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/patch/OncHistoryDetailPatchTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @BeforeClass
+    public static void setup() throws Exception {
+        esIndex = ElasticTestUtil.createIndexWithMappings(instanceName, "elastic/lmsMappings.json");
+        ddpInstanceDto = DBTestUtil.createTestDdpInstance(ddpInstanceDao, instanceName, esIndex);
+    }
+
+    @AfterClass
+    public static void tearDown() {
+        try {
+            Assert.assertNull(testParticipant);
```
zyme, line 58: Are there things that can only be asserted during teardown and shouldn't be asserted in the test methods?  Or should this be a delete-only-if-exists without asserts?

zyme (commented):
Really great stuff here!  I think this will really take our testing to the next level.  I was able to run the tests locally, set breakpoints, etc. without any issues.  I have a few small suggestions and requests for clarification.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
pegahtah: Yes EXACTLY!!!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantDto.java
```diff
@@ -35,6 +36,14 @@ public void setParticipantId(int participantId) {
         this.participantId = Optional.of(participantId);
     }
 
+    public int getParticipantIdOrThrow() {
+        return getParticipantId().orElseThrow(() -> new DsmInternalError("Participant ID cannot be null"));
```
pegahtah: This should not happen, but in case we ever reached here can we add more to this log ? maybe logging the ddp_participant_id

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantDto.java
```diff
@@ -35,6 +36,14 @@ public void setParticipantId(int participantId) {
...
+    }
+
+    public String getDdpParticipantIdOrThrow() {
+        return getDdpParticipantId().orElseThrow(() -> new DsmInternalError("DDP participant ID cannot be null"));
```
pegahtah: Same here :)

pepper-apis/dsm-core/src/test/resources/elastic/lmsMappings.json
```diff
@@ -0,0 +1,3812 @@
+{
```
pegahtah, line 1: How is this mapping generated? Is there a way to get it automatically generated instead of adding it to the code and modify it with every change ?

pegahtah (commented):
LGTM just left some comments/questions.
Thank you for making these changes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantDto.java
denniscunningham: Good idea!
Will do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
denniscunningham: Will do.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
denniscunningham: The response from `getTotalHits` is not a number but a string that includes the term "hits".

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
denniscunningham: Yup, great question.
I don't know that I've seen the issue here for sure, but I have definitely observed the issue with upserts and changed the code there a couple months ago (see UpsertPainless.java)...

If we do not set refresh to IMMEDIATE then ES updates the index every second (if I recall the setting properly). That leads to version conflicts if we try to update the same document (typically participant document) before the prior changes are written to the index.

That error is stealthy, at least in the upsert case, since you have to examine the results to notice that an update did not happen. I added code to UpsertPainless to detect that situation.

I'd have to examine UpdateResponse to recall how version conflicts are handled for updates.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/patch/OncHistoryDetailPatchTest.java
denniscunningham: I suppose it could be implemented either way, but in this case I prefer the discipline of asserting. This is affirming that the @After method worked properly, so it feels sloppy to say "but just in case" here, rather than remaining firm that the after method did its job. 
Does that make sense?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestContainer.java
denniscunningham: Will do.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
denniscunningham: Hmmm... I'll have to think about ways we can automatically confirm that this test utility is only being used for tests.

pepper-apis/dsm-core/src/test/resources/elastic/lmsMappings.json
denniscunningham: I ran this ES query: ` curl -u pepper_engineering:'xxxxxx' -XGET --header 'Content-Type: application/json' https://f7df559584f84f2cab27de822618fb89.us-central1.gcp.cloud.es.io:9243/participants_structured.cmi.cmi-lms/_mapping`.

Refreshing the mapping from an ES instance at test time generates a dependency on an external resource, which could be a problem. Also, this is just the mapping I chose for these tests, so other tests may prefer other mappings.

You could imagine a scheduled job that updates the test resources from an ES instance, to catch any mapping changes, but I wonder it the mapping update process is better done deliberately by hand?

Pulling back and thinking through scenarios... First, how often do ES mappings change? (DSM code updates mappings on the fly, but these are limited to dynamic fields.) If we make a deliberate mapping change, would if make sense to first verify the change with a test using the ES test containers? 

Is it better to deal with broken tests from automatically updating mappings, or is it better to have tests break because their mapping is out of date with DSM ES code under test? If we do not change mappings very often, does it make more sense to spend our development time increasing test coverage of DSM ES code, or automating test mapping update?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
zyme: Aha, in that case, never mind.

pepper-apis/dsm-core/src/test/resources/elastic/lmsMappings.json
zyme: ES mappings will not change much from here on out.  I think it's better to have tests break because their mapping is out of date with ES code under test.  I think we need to add the curl snippet and a few sentences about the new ES testing infrastructure to our internal dev documentation so that when people do go to change the mappings, they are leverage this new infrastructure for testing.  @denniscunningham @ssettipalli  can you drop some updates in those docs?

denniscunningham: I can update those docs, but I am not sure where they are. @ssettipalli, what do you suggest?

zyme (approved):
LGTM!  Excited to add this new testing capability.

pegahtah (approved):
Thank you!

PR #2700, opened 2023-09-29
PEPPER-784 added very basic test to cover NotificationUtil startup

zyme:
PEPPER-784 did not include a unit test for the portion of NotificationUtil that was being changed.  This PR includes a unit test, albeit a perfunctory one that only checks that startup runs properly given an appropriate config file.  I could not find an existing example in DSM of a test that uses a config file but does not do all the integration test setup, so I added a resource file with the minimal hocon needed to exercise NotificationUtil.startup.

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
+++ b/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
@@ -9,11 +9,11 @@ import org.junit.Test;
 public class NotificationUtilTest {
 
     @Test
-    public void testStartup() {
+    public void testInitialization() {
         Config cfg = ConfigFactory.parseResources("NotificationUtilTest.conf");
         try {
             var notificationUtil = new NotificationUtil(cfg);
-            notificationUtil.startup(cfg);
+            notificationUtil.initialize(cfg);
         } catch (Exception e) {
             fail("Could not start notification util due to " + e.getMessage());
         }
--- a/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
+++ b/pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
@@ -1,0 +1,21 @@
+package org.broadinstitute.dsm.util;
+
+import static org.junit.Assert.fail;
+
+import com.typesafe.config.Config;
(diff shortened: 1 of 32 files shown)


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
```diff
@@ -0,0 +1,21 @@
...
+import com.typesafe.config.ConfigFactory;
+import org.junit.Test;
+
+public class NotificationUtilTest {
+
+    @Test
+    public void testStartup() {
+        Config cfg = ConfigFactory.parseResources("NotificationUtilTest.conf");
+        try {
+            var notificationUtil = new NotificationUtil(cfg);
```
denniscunningham, line 15: FWIW, the DSM backend code uses `var` sparingly. I use it occasionally, and only in `for` loops.
You may want to bring up usage of `var` in Process and Standards, since in the case of DSM backend, it is currently non-standard. (And Java programmers may have strong opinions about its use.)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
```diff
@@ -0,0 +1,21 @@
...
+            notificationUtil.startup(cfg);
```
denniscunningham: Is this actually "starting" anything? The code looks like mostly initialization? 
I ask because this can be confusing when evaluating test code, since presumably if a test starts something -- like a thread -- it needs to get stopped.

denniscunningham (commented):
A couple tangental comments.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
zyme: Agreed, it's pretty confusingly named.  Switching to `initialize` since there's no threading going on here.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/NotificationUtilTest.java
zyme: Added to the agenda, will leave this for now and see where the discussion goes next week.

denniscunningham (approved):

PR #2701, opened 2023-10-02
[DSM] more logging in the RC

pegahtah:


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -42,15 +43,19 @@ protected void fillBulkRequestWithTransformedMapAndExport(Map<String, Object> pa
...
+                    transformObject(participantDetails);
+                    Map<String, Object> finalMapToUpsert = generate();
+                    bulkExportFacade.addDataToRequest(finalMapToUpsert, participantId);
+                }
+                if (isReadyToExport(participantsIterator)) {
+                    totalExported += bulkExportFacade.executeBulkUpsert();
+                    bulkExportFacade.clear();
+                }
+            }catch (Exception e){
+                throw new DsmInternalError(String.format("Error occurred while trying to export data for participant %s", participantId), e);
```
zyme: Could we log more than just the participant id?  Is `participantDetails` too much?

pegahtah: yes of course

pegahtah: just a note - that can have phi so we probably should remove that before it makes it to production

denniscunningham (approved):

PR #2703, opened 2023-10-11
PEPPER-786 More details in PatchRoute exceptions

zyme:
A review of error logs from PEPPER-786 hasn't turned up a specific root cause, but it did identify some areas where the code is ignoring some potentially useful exception details, and some areas where we could log a bit more detail to help us identify what's going on.

I also removed `DuplicateException` since it was buying us nothing but added complexity, and replaced it with `DSMInternalError`.

## FUD Score

## Testing
Because these changes are in exception handling code that is difficult to reproduce, there aren't any tests that test that the logging is indeed adding more details.  This means that we need to focus our reviews on verifying that 1) the changes will add useful information and 2) the changes won't lose any more information than is currently being lost.

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/DuplicateException.java
```diff
@@ -5,4 +5,8 @@ public class DuplicateException extends RuntimeException {
     public DuplicateException(String message) {
         super(message);
     }
+
+    public DuplicateException(String message, Exception cause) {
```
zyme: Pretty much any exception class we create should have the `(message,cause)` constructor in addition to just the `(message)` constructor, since our exceptions tend to wrap other exceptions.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -65,11 +65,11 @@ public Object processRequest(Request request, Response response, String userId)
                 throw new AuthorizationException("User is not authorized to patch participant data");
             }
         } catch (JsonSyntaxException e) {
-            throw new DSMBadRequestException("Invalid request payload format for patch");
+            throw new DSMBadRequestException("Invalid request payload format for patch", e);
```
zyme: Here we are losing the underlying`JsonSyntaxException`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -65,11 +65,11 @@ public Object processRequest(Request request, Response response, String userId)
...
         } catch (DuplicateException e) {
             // TODO: fix this by inspecting the DuplicateException throwers to see if cases are due to
             // bad request data (400)
-            throw new DsmInternalError("Duplicate value", e);
+            throw new DsmInternalError("Duplicate value:" + e.getMessage(), e);
```
zyme: Some code that instantiates `DuplicateException` contains really useful information in the `message` field.  So let's pass it along to our logging.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -147,7 +147,8 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
                                     + " rows");
                 }
             } catch (SQLIntegrityConstraintViolationException ex) {
-                throw new DuplicateException(dbElement.getColumnName());
+                throw new DuplicateException("Error updating " + dbElement.getTableName() + "."
```
zyme: We can do better than just "an attempt to duplicate a value in column X was rejected".  Added table, column, value, and pk id to the exception text, which should be passed up the stack to `PatchRoute` and then out to the logs.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
denniscunningham: As my comment above mentions, ideally we convert the DuplicateException in Patch.java to a DsmInternalError, thus we don't have to catch it here (at least that particular instance). That allows us to avoid this particular fancy footwork.
DuplicateException is not buying us much here (or elsewhere), since we can (and do) describe the particular error in the exception message. What we want to be sure of is whether this is an internal error (and thus mapped to 500) or an issue with a user input. In the latter case we want to use DSMBadRequestException, which gets mapped to 400.
Since you are already working around the problems with DuplicateException, perhaps now is the right time to make the change in Patch.java? (Which would also match our general error handling pattern in SQL query code.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/DuplicateException.java
denniscunningham: See my comment below, but I'd just as soon get rid of DuplicateException rather than extend it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -142,12 +142,13 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
-                    throw new RuntimeException(
-                            "Error updating " + dbElement.getTableName() + " record of w/ id " + id + " it was updating " + result
-                                    + " rows");
+                    throw new RuntimeException("Error updating " + dbElement.getTableName() + " with "
+                            + dbElement.getPrimaryKey() + "=" + id + ".  " + result + " rows updated.");
                 }
             } catch (SQLIntegrityConstraintViolationException ex) {
-                throw new DuplicateException(dbElement.getColumnName());
+                throw new DuplicateException("Error updating " + dbElement.getTableName() + "."
+                        + dbElement.getColumnName() + " with value " + nameValue.getValue() + " for "
```
denniscunningham: This would be a lot easier to read with String.format.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
zyme: I've replaced `DuplicateException`, but I don't think now is the right time to make decisions about what is a 400.  I'm in favor of getting some road time with this and seeing if the additional logging can help us determine what should be a 400.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -65,11 +63,7 @@ public Object processRequest(Request request, Response response, String userId)
                 throw new AuthorizationException("User is not authorized to patch participant data");
             }
         } catch (JsonSyntaxException e) {
-            throw new DSMBadRequestException("Invalid request payload format for patch");
-        } catch (DuplicateException e) {
-            // TODO: fix this by inspecting the DuplicateException throwers to see if cases are due to
-            // bad request data (400)
-            throw new DsmInternalError("Duplicate value", e);
+            throw new DSMBadRequestException("Invalid request payload format for patch: " + requestBody, e);
```
zyme, line 66: Let me know if anyone is worried about logging a potentially very large requestBody.  I think we want to log this for a while to help us determine what kind of posts should result in a 400 vs. 500.

denniscunningham: The request bodies I've seen are manageable, so I do not think logging it is an issue.
If the body cannot be mapped to `Patch.class`, that is a 400, so this code is correct as DSMBadRequestException (before and after this change). But I may be missing the intent of your comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -142,12 +142,13 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
                 if (result == 1) {
                     logger.info("Updated " + dbElement.getTableName() + " record w/ id " + id);
                 } else {
-                    throw new RuntimeException(
-                            "Error updating " + dbElement.getTableName() + " record of w/ id " + id + " it was updating " + result
-                                    + " rows");
+                    throw new RuntimeException("Error updating " + dbElement.getTableName() + " with "
```
denniscunningham: Sorry, I missed this the first time around...
We want to replace `RuntimeException` with something more meaningful that indicates how to generate the response. We introduced DSMBadRequestException (400) and DsmInternalError (500) for that purpose. In this case DsmInternalError makes sense.
Also, `String.format` would make this code more readable.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -142,12 +142,12 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
+                    throw new DsmInternalError(toErrorMessage(dbElement.getTableName(), dbElement.getColumnName(),
+                            nameValue.getValue(), dbElement.getPrimaryKey(), id) + ": " + result + " rows updated");
                 }
             } catch (SQLIntegrityConstraintViolationException ex) {
-                throw new DuplicateException(dbElement.getColumnName());
+                throw new DsmInternalError(toErrorMessage(dbElement.getTableName(), dbElement.getColumnName(),
+                        nameValue.getValue(), dbElement.getPrimaryKey(), id), ex);
             } catch (SQLException ex) {
                 dbVals.resultException = ex;
             }
```
denniscunningham, line 153: Nit: Does it make sense to convert the RuntimeException below to DsmInternalError?

denniscunningham (approved):
Looks good!
One small comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
zyme: Done!

PR #2704, opened 2023-10-13
[DSM| Hotfix] DON'T MERGE - Hotfix for Sprint 14

pegahtah:
## Context

Addressing [this ticket PEPPER-1176](https://broadworkbench.atlassian.net/browse/PEPPER-1176)
DSM is losing the stack trace while logging an exception during kit accessioning. This branch is made off of the last DSM RC branch for Sprint 14, I have made some changes so that the stack traces are not lost and we have better logs.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
@@ -7,6 +7,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -433,6 +434,7 @@ public class OncHistoryDetail implements HasDdpInstanceId {
 
     public static Map<String, List<OncHistoryDetail>> getOncHistoryDetailsByParticipantIds(@NonNull String realm,
                                                                                            List<String> participantIds) {
+        logger.info("Getting onc histories for participants {}", Arrays.toString(participantIds.toArray()));
         String queryAddition = " AND p.ddp_participant_id IN (?)".replace("?", DBUtil.participantIdsInClause(participantIds));
         return getOncHistoryDetails(realm, queryAddition);
     }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
@@ -425,7 +425,7 @@ public class OncHistoryDetail implements HasDdpInstanceId {
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Couldn't get list of oncHistories ", results.resultException);
+            throw new DsmInternalError("Couldn't get list of oncHistories ", results.resultException);
         }
 
         logger.info("Got " + oncHistory.size() + " participants oncHistories in DSM DB for " + realm);
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
@@ -425,7 +425,7 @@ public class OncHistoryDetail implements HasDdpInstanceId {
         });
 
         if (results.resultException != null) {
-            throw new DsmInternalError("Couldn't get list of oncHistories ", results.resultException);
+            throw new DsmInternalError(String.format("Couldn't get list of oncHistories with queryAddition '%s' and for realm %s", queryAddition, realm), results.resultException);
         }
 
         logger.info("Got " + oncHistory.size() + " participants oncHistories in DSM DB for " + realm);
@@ -434,7 +434,7 @@ public class OncHistoryDetail implements HasDdpInstanceId {
 
     public static Map<String, List<OncHistoryDetail>> getOncHistoryDetailsByParticipantIds(@NonNull String realm,
                                                                                            List<String> participantIds) {
-        logger.info("Getting onc histories for participants {}", Arrays.toString(participantIds.toArray()));
+        logger.info("Getting onc histories for participants {}", String.join(", ", participantIds));
         String queryAddition = " AND p.ddp_participant_id IN (?)".replace("?", DBUtil.participantIdsInClause(participantIds));
         return getOncHistoryDetails(realm, queryAddition);
     }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
@@ -24,6 +24,7 @@ import org.broadinstitute.dsm.db.structure.ColumnName;
 import org.broadinstitute.dsm.db.structure.DbDateConversion;
 import org.broadinstitute.dsm.db.structure.SqlDateConverter;
 import org.broadinstitute.dsm.db.structure.TableName;
+import org.broadinstitute.dsm.exception.DsmInternalError;
 import org.broadinstitute.dsm.statics.DBConstants;
 import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
 import org.broadinstitute.lddp.db.SimpleResult;
@@ -316,8 +317,8 @@ public class Tissue {
                 default:
                     logger.error("Unknown type");
             }
-        } catch (SQLException throwables) {
-            throwables.printStackTrace();
+        } catch (SQLException e) {
+            throw new DsmInternalError(e);
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
@@ -317,7 +317,7 @@ public class Tissue {
                     logger.error("Unknown type");
             }
         } catch (SQLException e) {
-            throw new DsmInternalError(e);
+            throw new DsmInternalError(String.format("Error setting type for smId with id: ", tissueSmId.getSmIdPk()), e);
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
@@ -317,7 +317,7 @@ public class Tissue {
                     logger.error("Unknown type");
             }
         } catch (SQLException e) {
-            throw new DsmInternalError(String.format("Error setting type for smId with id: ", tissueSmId.getSmIdPk()), e);
+            throw new DsmInternalError("Error setting type for smId with id: " + tissueSmId.getSmIdPk(), e);
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
@@ -304,7 +304,6 @@ public class Tissue {
                 case "he": {
                     this.heSMID.add(tissueSmId);
                     break;
-
                 }
                 case "uss": {
                     this.ussSMID.add(tissueSmId);
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
@@ -119,7 +119,7 @@ public class ClinicalKitDto {
                             ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, null);
             this.setCollaboratorParticipantId(collaboratorParticipantId);
         } catch (Exception e) {
-            throw new RuntimeException("Participant doesn't exist / is not valid for kit " + e.getMessage());
+            throw new RuntimeException("Participant doesn't exist / is not valid for kit", e);
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
@@ -113,7 +113,7 @@ public class ClinicalKitDto {
             this.setDateOfBirth(elasticSearchParticipantDto.getDsm().map(Dsm::getDateOfBirth).orElse(""));
             this.setFirstName(elasticSearchParticipantDto.getProfile().map(Profile::getFirstName).orElse(""));
             this.setLastName(elasticSearchParticipantDto.getProfile().map(Profile::getLastName).orElse(""));
-            this.setGender(getParticipantGender(elasticSearchParticipantDto, ddpInstance.getName()));
+            this.setGender(getParticipantGender(elasticSearchParticipantDto, ddpInstance.getName(), ddpParticipantId));
             String shortId = elasticSearchParticipantDto.getProfile().map(Profile::getHruid).orElse("");
             String collaboratorParticipantId =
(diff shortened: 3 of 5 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,25 +101,25 @@ public void setGender(String genderString) {
...
         try {
-            this.setDateOfBirth(maybeParticipantESDataByParticipantId.get().getDsm().map(Dsm::getDateOfBirth).orElse(""));
-            this.setFirstName(maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getFirstName).orElse(""));
-            this.setLastName(maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getLastName).orElse(""));
-            this.setGender(getParticipantGender(maybeParticipantESDataByParticipantId.get(), ddpInstance.getName()));
-            String shortId = maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getHruid).orElse("");
+            this.setDateOfBirth(elasticSearchParticipantDto.getDsm().map(Dsm::getDateOfBirth).orElse(""));
+            this.setFirstName(elasticSearchParticipantDto.getProfile().map(Profile::getFirstName).orElse(""));
+            this.setLastName(elasticSearchParticipantDto.getProfile().map(Profile::getLastName).orElse(""));
+            this.setGender(getParticipantGender(elasticSearchParticipantDto, ddpInstance.getName()));
```
ssettipalli: I am not sure what value is in the name but we are setting "name" into "Gender" ?

ssettipalli:
Do we have a jira ticket to track these changes ?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -433,6 +434,7 @@ public static Map<String, List<OncHistoryDetail>> getOncHistoryDetails(@NonNull
 
     public static Map<String, List<OncHistoryDetail>> getOncHistoryDetailsByParticipantIds(@NonNull String realm,
                                                                                            List<String> participantIds) {
+        logger.info("Getting onc histories for participants {}", Arrays.toString(participantIds.toArray()));
```
ssettipalli: I believe this can be simplified as 
        logger.info("Getting onc histories for participants {}", participantIds);

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,25 +101,25 @@ public void setGender(String genderString) {
...
-
+        ElasticSearchParticipantDto elasticSearchParticipantDto = maybeParticipantESDataByParticipantId.get();
         try {
-            this.setDateOfBirth(maybeParticipantESDataByParticipantId.get().getDsm().map(Dsm::getDateOfBirth).orElse(""));
-            this.setFirstName(maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getFirstName).orElse(""));
-            this.setLastName(maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getLastName).orElse(""));
-            this.setGender(getParticipantGender(maybeParticipantESDataByParticipantId.get(), ddpInstance.getName()));
-            String shortId = maybeParticipantESDataByParticipantId.get().getProfile().map(Profile::getHruid).orElse("");
+            this.setDateOfBirth(elasticSearchParticipantDto.getDsm().map(Dsm::getDateOfBirth).orElse(""));
+            this.setFirstName(elasticSearchParticipantDto.getProfile().map(Profile::getFirstName).orElse(""));
```
denniscunningham, line 114: This code should first establish whether `profile` and `dsm` sections exist in the participant document, and handle those situations separately. (Do we want to proceed is those cases or throw an error?)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,25 +101,25 @@ public void setGender(String genderString) {
     }
 
     public void setNecessaryParticipantDataToClinicalKit(String ddpParticipantId, DDPInstance ddpInstance) {
+        logger.info("Setting PHI to clinical kit's accessioning data for {}", ddpParticipantId);
         Optional<ElasticSearchParticipantDto> maybeParticipantESDataByParticipantId =
                 ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
         if (maybeParticipantESDataByParticipantId.isEmpty()) {
             throw new RuntimeException("Participant ES Data is not found for " + ddpParticipantId);
```
denniscunningham: This can be a DsmInternalError.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -323,7 +323,9 @@ public static String getKitCollaboratorId(@NonNull String ddpParticipantId, @Non
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error getting list of assignees ", results.resultException);
+            throw new RuntimeException(
+                    String.format("Error getting kitCollaboratorId for participant %s in realm with id %s", ddpParticipantId, realmId),
+                    results.resultException);
```
denniscunningham, line 329: This can be a DsmInternalError.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -346,7 +348,9 @@ public static String getTissueCollaboratorId(@NonNull String ddpParticipantId, @
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Error getting list of assignees ", results.resultException);
+            throw new RuntimeException(
+                    String.format("Error getting TissueCollaboratorId for participant %s in realm with id %s", ddpParticipantId, realmId),
+                    results.resultException);
```
denniscunningham, line 354: This can be a DsmInternalError.

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
```diff
@@ -424,7 +425,7 @@ public static Map<String, List<OncHistoryDetail>> getOncHistoryDetails(@NonNull
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Couldn't get list of oncHistories ", results.resultException);
+            throw new DsmInternalError("Couldn't get list of oncHistories ", results.resultException);
```
zyme: Worth adding the `realm` and `queryAddition` params here?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,25 +101,25 @@ public void setGender(String genderString) {
...
+            String shortId = elasticSearchParticipantDto.getProfile().map(Profile::getHruid).orElse("");
             String collaboratorParticipantId =
                     KitRequestShipping.getCollaboratorParticipantId(ddpInstance.getBaseUrl(), ddpInstance.getDdpInstanceId(),
-                            ddpInstance.isMigratedDDP(),
-                            ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, null);
+                            ddpInstance.isMigratedDDP(), ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, shortId, null);
             this.setCollaboratorParticipantId(collaboratorParticipantId);
         } catch (Exception e) {
-            throw new RuntimeException("Participant doesn't exist / is not valid for kit " + e.getMessage());
+            throw new RuntimeException(String.format("Participant {} doesn't exist / is not valid for kit", ddpParticipantId), e);
```
denniscunningham: This exception handling should be pushed down to `setCollaboratorParticipantId` or lower.
At this point we cannot deduce the reason for the exception and I suspect "Participant doesn't exist" is typically not accurate.
If that fix is not in scope for this PR, then we should create a ticket to fix this in the near furture.

denniscunningham (commented):
Sorry, forgot to save one of my comments.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/OncHistoryDetail.java
zyme: Or `String.join(", ", participantIds);`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
```diff
@@ -316,8 +316,8 @@ public void setSmIdBasedOnType(SmId tissueSmId, ResultSet rs) {
                 default:
                     logger.error("Unknown type");
             }
-        } catch (SQLException throwables) {
-            throwables.printStackTrace();
+        } catch (SQLException e) {
+            throw new DsmInternalError(e);
```
zyme: Would be helpful to include `tissueSmId` in the error message.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,25 +101,25 @@ public void setGender(String genderString) {
     }
 
     public void setNecessaryParticipantDataToClinicalKit(String ddpParticipantId, DDPInstance ddpInstance) {
+        logger.info("Setting PHI to clinical kit's accessioning data for {}", ddpParticipantId);
```
zyme: Let's add the `ddpInstance` name here too.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
zyme: This error message is pretty ambiguous and I'm not sure it's accurate any more.  Since this is failing during `setNecessaryParticipantDataToClinicalKit`, perhaps a more accurate error is `Could not set participant data for clinical kit`.  This should also be a DSMInternalError, right @denniscunningham ?  I'm still trying to figure out the proper use of that exception.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -135,6 +135,7 @@ private String getParticipantGender(ElasticSearchParticipantDto participantBySho
                 return oncHistoryWithGender.get().getGender();
             }
         }
+        logger.info("Participant {} did not have gender on tissue pages, will look into activities", participantByShortId.getParticipantId());
```
denniscunningham, line 143: We need to check that the `ElasticSearchParticipantDto` has a participant ID before using it (starting at line 129 above). I suspect a missing participant ID is the cause of this particular bug.
Likewise for getting activities, unless `getGenderFromActivities` is robust for that already.

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
pegahtah: the name is the study name, like `cmi-lms` in this case

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
pegahtah: Unless I am mistaken that should be handled by `orElse` . Am I wrong?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
pegahtah: I made the changes to the error's text, will continue to communicate with @denniscunningham about the Exception type

pegahtah: I made the changes to the error's text but I didn't understand where you think the error handling should be moved to. The `setCollaboratorParticipantId` is just a setter

pegahtah: Also I think it is a DSMInternalError ( at least as much as the user is concerned), you agree?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
```diff
@@ -316,8 +316,8 @@ public void setSmIdBasedOnType(SmId tissueSmId, ResultSet rs) {
                 default:
                     logger.error("Unknown type");
             }
-        } catch (SQLException throwables) {
-            throwables.printStackTrace();
+        } catch (SQLException e) {
+            throw new DsmInternalError(String.format("Error setting type for smId with id: ", tissueSmId.getSmIdPk()), e);
```
denniscunningham: Looks like you are missing `%` in `StringFormat`? (The linter should spot this for you.)
(You can also just skip `StringFormat` here and append in-line.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
denniscunningham: Yup, but I was wondering if we want to proceed if either DSM or profile are empty, but get that's probably out of scope for this PR.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,30 +102,33 @@ public void setGender(String genderString) {
...
         } catch (Exception e) {
-            throw new RuntimeException("Participant doesn't exist / is not valid for kit " + e.getMessage());
+            throw new DsmInternalError(String.format("Error getting participant data for %s", ddpParticipantId), e);
         }
     }
 
     private String getParticipantGender(ElasticSearchParticipantDto participantByShortId, String realm) {
         // if gender is set on tissue page use that
+        if(StringUtils.isBlank(participantByShortId.getParticipantId())){
+            throw new DsmInternalError(String.format("The participant %s is missing participant id", participantByShortId.getProfile().get().getGuid()));
```
pegahtah: @denniscunningham I added the check for participant id here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -101,30 +102,33 @@ public void setGender(String genderString) {
...
+        }
```
denniscunningham, line 132: We have not established that the ptp has a profile at this point, so this may fail. But we do have a ddpParticipantId that we can pass into this method.
(This is also a good place to assign the return value of `participantByShortId.getParticipantId()` to improve readability.)

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
denniscunningham: Oops, typo: `getCollaboratorParticipantId` (and `getParticipantGender` which you did with your latest commit).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
pegahtah: Yeah not sure if that's an error or an empty string should be returned in that case

denniscunningham (approved):

pegahtah:
Closing this PR after it was approved because it is not supposed to get merged

PR #2708, opened 2023-10-18
PEPPER-1182 Cancer Translations

zyme:
## Context

PEPPER-1182 is about adding Spanish versions of cancer names to our cancer list.  There were a lot of existing tests for cancer list, which made these changes pretty quick.

The overall approach here is:
* Change the list of `String`s that we passed around to a list of `CancerItem`, which is a cancer name and a language.  You'll see a lot of changes that are all about `List<String>` -> `List<CancerItem>`.
* Update `CancerStore` to work off of lists of `CancerItems`, maintaining multiple lists internally (one for each language)
* Change the `cancer_list` table to add an `iso_language_code` column, backfill values, and add new Spanish values as per the ticket

## FUD Score

Load the OS project once it's got translated content and has Spanish enabled as a supported language, and verify that the cancer list has Spanish values when you change your language to Spanish.  Then flip back and forth between English and Spanish.

Bonus: in theory this should work for any other study that uses the cancer list and has Spanish configured as a supported language for the study.

## Testing

Plenty of existing tests for cancer list management in DSM and DSS, so I adjusted those to pass.  I also added some tests to verify that the language filtering works properly, and that search using non-ASCII characters works properly.

## Release


pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1182_cancer_list_translations.xml
```diff
@@ -0,0 +1,283 @@
...
+            <column name="iso_language_code" type="varchar(5)"/>
+        </addColumn>
+
+        <sql>update cancer_list
+             set iso_language_code = 'en'</sql>
+
+        <addUniqueConstraint tableName="cancer_list" columnNames="display_name,iso_language_code"
+                             constraintName="cancer_list_uk"/>
+
+        <insert tableName="cancer_list">
```
zyme, line 19: Autogenerated this from the spreadsheet Whitney attached to the original ticket by way of google sheet concatenation magic [here](https://docs.google.com/spreadsheets/d/1faPBOzx4dsOsd9xUCsqPVphnqao1ejGrhAzVN6nGIfo/edit#gid=1107868038).

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/dto/CancerItem.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * The name of a cancer and its language
+ */
+public class CancerItem {
```
zyme, line 16: The cancer list was just a bunch of strings, so we needed a new class that could represent both the cancer name and the language.  I put this class into ddp-common so that both DSM and DSS can use it.

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1182_cancer_list_translations.xml
```diff
@@ -0,0 +1,283 @@
...
+
+    <changeSet author="andrew" id="PEPPER-1182_cancer_list_translations">
+        <addColumn tableName="cancer_list">
+            <column name="iso_language_code" type="varchar(5)"/>
+        </addColumn>
+
+        <sql>update cancer_list
+             set iso_language_code = 'en'</sql>
+
+        <addUniqueConstraint tableName="cancer_list" columnNames="display_name,iso_language_code"
```
zyme, line 16: Java code does a bunch of `Set`-type flattening to avoid duplicates, but we can also avoid them with this unique key.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/model/dsm/Cancer.java
```diff
@@ -10,8 +10,12 @@ public class Cancer {
     @SerializedName("description")
     private String description;
 
-    public Cancer(String name, String description) {
+    public Cancer(String name) {
```
zyme, line 13: `description` was never used, so I removed it.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRouteStandaloneTest.java
```diff
@@ -184,7 +254,6 @@ public void givenLookupWithCyrillicCharacterThatDoesNotMatchAnything() {
     }
 
     private static class TestData {
-        public static final String STUDY_GUID = "test-study1";
```
zyme, line 187: This test was using a phony study guid that did not exist in the database, which was causing me grief when trying to adjust the study's list of configured languages.  I've removed it and changed the test so that it uses a real study guid.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/util/TestDataSetupUtil.java
```diff
@@ -621,7 +622,7 @@ public static UserProfile createTestingProfile(Handle handle, long userId, boole
                 .preferredLangId(LanguageStore
                         .get(TestConstants.TEST_USER_PROFILE_PREFERRED_LANGUAGE)
                         .getId())
-                .preferredLangCode(null)
+                .preferredLangCode(TEST_USER_PROFILE_PREFERRED_LANGUAGE)
```
zyme, line 625: Fixing a bug that has been with us for a while: the generated profile data had a language id, but not a corresponding language code, which was leading to breakages in the test when asserting that the test user's preferred language had been reset properly (in addition to it not reflecting the reality of the dao, that uses the language id to lookup and populate the language code).

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/dto/CancerItem.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getCancerName() {
+        return cancerName;
+    }
+
+    public String getIsoLanguageCode() {
+        return isoLanguageCode;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
zyme, line 25: `equals()` and `hashCode()` are here because `CancerStore` deals in `Sets`, and  although the new unique key constraint will help avoid duplicates, some tests were looking to verify that duplicates were filtered out at the code level.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/cache/LanguageStore.java
```diff
@@ -13,7 +13,11 @@
  */
 public class LanguageStore {
 
-    public static final String DEFAULT_LANG_CODE = "en";
+    public static final String SPANISH_LANG_CODE = "es";
```
zyme, line 16: Just trying to keep it D-R-Y

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/CancerStore.java
```diff
@@ -31,21 +32,28 @@ public static CancerStore getInstance() {
...
-    public Set<Cancer> getCancerList() {
-        return cancers;
+    /**
+     * Returns all the cancers for the given language
+     */
+    public Set<CancerItem> getCancerList(String languageIsoCode) {
+        if (cancersByLanguage.get(languageIsoCode) == null) {
+            return Collections.emptySet();
+        } else {
+            return Collections.unmodifiableSet(cancersByLanguage.get(languageIsoCode));
```
zyme, line 55: My own two cents: it seems unnecessary to make all this stuff `unmodifiable`, since other lists that are backed by database tables don't get such treatment, but since there are tests that enforce this, I went along with the approach here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/CancerRoute.java
```diff
@@ -1,30 +1,17 @@
...
     public Object handle(Request request, Response response) throws Exception {
-        List<String> cancerList = new ArrayList<>();
-        try {
-            cancerList = Cancer.getCancers();
-            return cancerList;
-        } catch (Exception e) {
-            logger.error("Cancer list attempt gave an error: ", e);
-            return new Result(500, UserErrorMessages.CONTACT_DEVELOPER);
-        }
+        return Cancer.getCancers();
```
zyme, line 15: I thought this could be simplified quite a bit--if something blows up, let it blow up naturally instead of doing all the wrapping.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/dto/CancerItem.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public class CancerItem {
+
+    @SerializedName("name")
+    private final String cancerName;
+
+    @SerializedName("language")
+    private final String isoLanguageCode;
+
+    public CancerItem(String cancerName, String isoLanguageCode) {
```
ssettipalli: May be use Lombok annotations and get rid of constructor , getter and setters
@Data
@AllArgsConstructor

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRoute.java
```diff
@@ -33,53 +35,53 @@ public class GetCancerSuggestionsRoute implements Route {
     private static final String CANCER_QUERY_REGEX = "\\w+";
     private static final int DEFAULT_LIMIT = 100;
     private final CancerStore cancerStore;
-    
 
     @Override
     public CancerSuggestionResponse handle(Request request, Response response) {
+        String cancerListLanguage = RouteUtil.getUserLanguage(request).getIsoCode();
```
ssettipalli, line 41: wondering if the request will always contain the language code !
As a fallback can we query users preferred language from user_profile.preferred_language_id ? If both fail go with default "en" !

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/ListCancersRoute.java
```diff
@@ -3,19 +3,28 @@
...
 @AllArgsConstructor
 public class ListCancersRoute implements Route {
     private final CancerService cancerService;
 
     @Override
-    public List<Cancer> handle(Request request, Response response) {
-        return cancerService.fetchCancers();
+    public List<CancerItem> handle(Request request, Response response) {
+        String cancerLanguage = LanguageStore.DEFAULT_LANG_CODE;
+        LanguageDto userLanguage = getUserLanguage(request);
```
ssettipalli, line 24: Same as previous comment.
If not in request, may be check in user_profile before deciding to default "en"

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRouteStandaloneTest.java
ssettipalli: wonder how was the test passing until now !

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRoute.java
zyme: It should.  `StudyLanguageResolutionFilter` is invoked as a `before()` ahead of this, and it will default to the default language for the study, or barring that, just "en".  So I think we're safe without any additional safeguards here.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRouteStandaloneTest.java
zyme: `StudyLanguageResolutionFilter` doesn't throw errors if the study isn't real.  It was defaulting to English, and since `GetCancerSuggestionsRoute` doesn't do anything with the study guid, it worked.  Once I tried to update the study to support Spanish, things broke since the study didn't exist in the database.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/ListCancersRoute.java
zyme: As per previous comment, will change this to rely on the request attribute alone, since adding this default appears redundant.

zyme: Never mind, I'm reverting this--this route does not pick up language from upstream filters, but the autocomplete route does.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRoute.java
ssettipalli: If no language in the request, we are using study default  (which is always en ) rather than querying from DB table user_profile which has users preferred language ?

ssettipalli (approved):
looks good to me

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1182_cancer_list_translations.xml
pegahtah: Wow!! That's real magic

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/CancerStore.java
```diff
@@ -1,23 +1,24 @@
...
-import java.util.stream.Collectors;
 
-import org.apache.commons.collections4.CollectionUtils;
-
-import org.broadinstitute.ddp.model.dsm.Cancer;
+import org.broadinstitute.ddp.db.dto.CancerItem;
 
 public class CancerStore {
 
-    private static Set<Cancer> cancers;
```
pegahtah, line 15: I wonder why it was a set before?

pegahtah (commented):
Thank you so much Zim! Sorry for the late review, it looks great!!

PR #2709, opened 2023-10-18
PEPPER-1170 Generic Task to search and update PI name changes 

ssettipalli:
Generic  study-builder task to search and update PI name changes 

PR has some unused code and quite and bit of //todo
Code was left in there because it might be needed to implement the todo's which will be implemented as part of any new future tickets.
Current version of the implementation is good enough to take care of Pancan PI name changes
Code will: Search for matched strings ,  replace with new string and will version (configurable) if needed
Source of truth is current DB
String ADD and DELETE are not implemented in current version
UpdateActivityContentSourceDB.java is the main class that does all the work


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/TemplateActionType.java
```diff
@@ -0,0 +1,7 @@
+package org.broadinstitute.ddp.studybuilder.task;
+
+public enum TemplateActionType {
+    ADD,
```
zyme: Maybe remove `ADD` and `DELETE` but leave a note here saying that these might be things we want to build out later.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateActivityContentSourceDB.java
```diff
@@ -0,0 +1,690 @@
...
+ * General task to iterate through activity content in a study and update needed content from passed patch file.
+ * For each activity of the study, iterate through each section, block, each template and for each template, search and
+ * compare the template text, the variables, and the variable translations, updating these things in-place with revisioning if needed.
+ * Source of truth is DB.
+ * Iterate through each translation variable in DB and find matching text and replace
+ * Iterate through each template text in DB and find matching text and replace
+ */
+@Slf4j
+@NoArgsConstructor
+public class UpdateActivityContentSourceDB extends SimpleRevisionTask {
```
zyme, line 91: How about renaming this so that it more accurately conveys that this is a special case of finding and replacing content, constrained to pancan since there are hardcoded file paths that reference pancan files?

ssettipalli: Moved file paths to the other Pancan class
In current functionality they are of not much needed, may be in future.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateActivityContentSourceDB.java
```diff
@@ -0,0 +1,696 @@
...
+        //todo .. query DB and act accordingly
+
+        //steps: revision the block, revision the template, insert new template, create new block_content
+
+        //iterate through and apply ALL changes
+        String newTemplateText = template.getTemplateText();
+        log.info("revisioning and updating template : {}", template.getTemplateText());
+        for (TemplateUpdateInfo updateInfo : updateInfos) {
+            newTemplateText = newTemplateText.replace(updateInfo.getSearchString(), updateInfo.getReplaceString());
+            newTemplateText = newTemplateText.replace("<li></li>", ""); //remove empty tags if any
```
zyme: Is there a way to generalize this with jsoup so we can remove any empty element?

zyme (approved):
LGTM--we'll iterate on this as we go.

pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateActivityContentSourceDB.java
ssettipalli: That was result of removing one PI name . 
Actually I could just add these tags to the search strings and that eliminates this need. 
updated the patch file and removed this line

PR #2710, opened 2023-10-19
PEPPER-1182 improving cancer search experience for Spanish

zyme:
## Context

After a quick demo of PEPPER-1182, we realized that we could improve the search experience across different keyboard settings by considering unaccented characters as their accented counterparts and returning the accented results.  So the search for Spanish now uses a regex so that "ü/u", "ñ/n", "é/e", "á/a", "í/a", "ó/o", "ú/u" are considered equivalent, in addition to looking for exact matches.

## FUD Score

Load up the "other cancers" question in OS and toggle the language setting back and forth between English and Spanish.  Try some matches with accented and unaccented chars.

## Testing

## Release


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/GetCancerSuggestionsRoute.java
```diff
@@ -44,48 +35,15 @@ public CancerSuggestionResponse handle(Request request, Response response) {
...
                 log.warn("Cancer query contains non-alphanumeric characters!");
                 throw ResponseUtil.haltError(response, HttpStatus.SC_BAD_REQUEST,
                         new ApiError(ErrorCodes.MALFORMED_CANCER_QUERY, "Invalid cancer query"));
             }
-            return getCancerSuggestions(cancerQuery, cancerListLanguage, limit);
+            return CancerStore.getInstance().getCancerSuggestions(cancerQuery, cancerListLanguage, limit);
         }
     }
 
-    private CancerSuggestionResponse getUnfilteredCancerSuggestions(String cancerQuery, String language, int limit) {
```
zyme, line 58: Moved these methods to `CancerStore` because 1) I needed to test them explicitly and 2) they seem like general purpose, and shouldn't be confined to a route class.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/CancerStore.java
```diff
@@ -56,4 +65,67 @@ public Set<CancerItem> getCancerList(String languageIsoCode) {
         }
     }
 
+    /**
```
zyme, line 68: Brought these methods over from `GetCancerSuggestionsRoute` and refactored them a bit.

ssettipalli (approved):
looks good to me

pegahtah (approved):

PR #2711, opened 2023-10-19
PEPPER-1182 hit start/end for Spanish

zyme:
## Context

PEPPER-1182.  This change sets the proper hit index and length Spanish cancer name typeaheads where we substitute in accented characters.  It should fix the highlighting below:

Screenshot 2023-10-19 at 3 27 34 PM

## FUD Score

## Testing

## Release


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/db/CancerStore.java
```diff
@@ -115,14 +117,24 @@ public CancerSuggestionResponse getCancerSuggestions(String cancerQuery, String
...
         var suggestionComparator = new StringSuggestionTypeaheadComparator(upperCancerQuery);
         cancerMatches.stream()
                 .sorted((lhs, rhs) -> suggestionComparator.compare(lhs.getCancerName(), rhs.getCancerName()))
                 .limit(limit)
                 .forEach(cancer -> {
-                    int offset = cancer.getCancerName().toUpperCase().indexOf(upperCancerQuery);
-                    sortedSuggestions.add(new CancerSuggestion(new Cancer(cancer.getCancerName()),
-                            Collections.singletonList(new PatternMatch(offset, cancerQuery.length()))));
+                    Matcher matcher = pattern.matcher(cancer.getCancerName().toUpperCase());
+                    if (matcher.find()) {
```
zyme, line 119: The start position and length should be calculated using the regex-based match, not the indexOf() and length of the query string.

ssettipalli (approved):

PR #2712, opened 2023-10-21
PEPPER-1171_KitTrackingPKClashes

zyme:
DDP-1171 has a good amount of background, so take a look there before diving into this one.

The main change here was updating `KitDaoImpl` so that instead of preventing a duplicate row by waiting for the unique key to break on insert, it looks for a row that would break the unique key prior to insert.  In its previous form, because a unique key violation triggers a `SQLException` which in turn becomes an alert, which is a distraction for us because this a user error that does not require any intervention on our part.  What's important is that the user gets feedback about a duplicate scan, which still happens after this PR, but without any alerting and without any unique key issues.

Note that [PR to remove the `KitDao` interface](https://github.com/broadinstitute/ddp-study-server/pull/2739) will be merged into this branch *after* this PR is approved.  Since much is changing in `KitDaoImpl`, it's easier to keep these changes separate.

Likewise, a separate PR for renaming `ScanError` -> `ScanResult` is [here](https://github.com/broadinstitute/ddp-study-server/pull/2743) and will be merged after this PR is approved.

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -530,30 +574,58 @@ public KitRequestShipping getKitRequestShipping(@NonNull ResultSet rs) throws SQ
...
                 stmt.setString(3, kitRequestShipping.getTrackingId());
                 stmt.setString(4, kitRequestShipping.getKitLabel());
+                stmt.setString(5, kitRequestShipping.getTrackingId());
+                stmt.setString(6, kitRequestShipping.getKitLabel());
                 int rowsAffected = stmt.executeUpdate();
                 if (rowsAffected != 1) {
-                    dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(),
-                            String.format("Kit Label \"%s\" does not exist.\n"
-                                    + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER, kitRequestShipping.getKitLabel()));
-                    logger.error(String.format("The number of affected rows for kit label %s is not 1", kitRequestShipping.getKitLabel()));
```
zyme, line 548: The change here is that instead of erroring out when > 1 row is updated, we check for the row first and return a scan error without breaking the constraint and sending an alert.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitTrackingScanUseCase.java
```diff
@@ -17,22 +19,8 @@ public KitTrackingScanUseCase(KitPayload kitPayload, KitDao kitDao) {
 
     @Override
     protected Optional<ScanError> process(ScanPayload scanPayload) {
-        String kitLabel = scanPayload.getKitLabel();
```
zyme, line 20: This logic has moved into the dao, so that we can test more code without having to hit the route.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -191,7 +190,7 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
     private static final String GET_COUNT_KITS_WITH_SAME_COLLABORATOR_SAMPLE_ID_AND_KIT_TYPE =
             " SELECT count(*) kitRequestCount from ddp_kit_request where bsp_collaborator_sample_id REGEXP \"^%1\" and kit_type_id = ?";
 
-    private static final String GET_FOUND_IF_KIT_WITH_DDP_LABEL_ALREADY_EXISTS =
+    public static final String SELECT_1_FROM_KIT_REQUEST_WITH_DDP_LABEL =
```
zyme, line 193: Needed to run this query from another dao, so I made it public.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1243,7 +1243,7 @@ public static int getKitCounter(@NonNull Connection conn, String collaboratorSam
      * and checking it against db if unique
      */
     public static String generateDdpLabelID() {
-        return generateDdpLabelID(15, GET_FOUND_IF_KIT_WITH_DDP_LABEL_ALREADY_EXISTS);
+        return generateDdpLabelID(15, SELECT_1_FROM_KIT_REQUEST_WITH_DDP_LABEL);
```
zyme, line 1246: Renamed this to make it a bit more clear what the query does.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1748,4 +1748,9 @@ public boolean isKitRequiringTrackingScan() {
     public boolean hasTrackingScan() {
         return StringUtils.isNotBlank(trackingId);
     }
+
+    // todo arz test me
+    public boolean hasBSPCollaboratorParticipantId() {
+        return StringUtils.isNotBlank(bspCollaboratorParticipantId);
```
zyme, line 1753: I saw the check for blank bsp collaborator participant id in a few places so I extracted a method for it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -530,30 +646,59 @@ public KitRequestShipping getKitRequestShipping(@NonNull ResultSet rs) throws SQ
...
+                stmt.setInt(2, userId);
                 stmt.setString(3, kitRequestShipping.getTrackingId());
                 stmt.setString(4, kitRequestShipping.getKitLabel());
+                stmt.setString(5, kitRequestShipping.getTrackingId());
+                stmt.setString(6, kitRequestShipping.getKitLabel());
                 int rowsAffected = stmt.executeUpdate();
                 if (rowsAffected != 1) {
-                    dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(),
-                            String.format("Kit Label \"%s\" does not exist.\n"
-                                    + UserErrorMessages.IF_QUESTIONS_CONTACT_DEVELOPER, kitRequestShipping.getKitLabel()));
```
zyme, line 547: Instead of having "is not 1", expanded the queries to help me understand whether 1 row or 0 rows is expected in different situations.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -530,30 +646,59 @@ public KitRequestShipping getKitRequestShipping(@NonNull ResultSet rs) throws SQ
...
-                    logger.error(String.format("The number of affected rows for kit label %s is not 1", kitRequestShipping.getKitLabel()));
+                    dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(), errorMessage);
                 } else {
-                    logger.info("Added tracking for kit w/ kit_label " + kitRequestShipping.getKitLabel());
+                    logger.info("Added tracking id {} for kit {}", kitRequestShipping.getTrackingId(),
+                            kitRequestShipping.getKitLabel());
+                    return dbVals;
                 }
-            } catch (Exception ex) {
-                dbVals.resultValue = new ScanError(kitRequestShipping.getKitLabel(),
```
zyme, line 553: Rather than catching `Exception` and issuing an alert, this now catches only `SQLException` so that we can return a ScanError for some situations without sending an alert.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
```diff
@@ -37,23 +33,18 @@ public List<ScanError> get() {
...
     protected abstract Optional<ScanError> process(ScanPayload scanPayload);
+
     protected abstract Optional<ScanError> processRGPFinalScan(ScanPayload scanPayload);
 
     protected boolean isKitUpdateSuccessful(Optional<ScanError> maybeScanError, String bspCollaboratorParticipantId) {
-        return maybeScanError.isEmpty() || maybeScanError.get().isScanErrorOnlyBspParticipantId( bspCollaboratorParticipantId);
-    }
-
-
-    protected BaseKitUseCase getDecoratedScanUseCase() {
```
zyme, line 53: Unused method, so I removed it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/BaseKitUseCase.java
```diff
@@ -37,23 +33,18 @@ public List<ScanError> get() {
...
-        if (Objects.isNull(decoratedScanUseCase)) {
-            decoratedScanUseCase = new NullObjectDecorator();
-        }
-        return decoratedScanUseCase;
+        return maybeScanError.isEmpty()
+                || maybeScanError.get().isScanErrorOnlyBspParticipantId(bspCollaboratorParticipantId);
```
zyme, line 48: Little bit of checkstyle clean up.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/ScanError.java
```diff
@@ -1,27 +1,39 @@
 package org.broadinstitute.dsm.model.kit;
 
 import lombok.Data;
-import spark.utils.StringUtils;
+import org.apache.commons.lang3.StringUtils;
```
zyme: Wrong import--this should be directly from apache.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -163,32 +196,98 @@ public Boolean hasTrackingScan(String kitLabel) {
...
+                    return dbVals;
+                }
+                int numKitsWithLabel = rs.getInt(1);
+
+                if (numKitsWithLabel == 0) {
+                    dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(),
+                            "kit " + kitRequestShipping.getDdpLabel() + " not found.");
+                    return dbVals;
+                } else if (numKitsWithLabel > 1) {
+                    // todo arz is this an error?  Expected to have multiple kits with the same label?
```
zyme: @pegahtah can you confirm that this should indeed be an error?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -163,32 +196,98 @@ public Boolean hasTrackingScan(String kitLabel) {
...
-                    dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(), "ddp_label "
-                            + kitRequestShipping.getDdpLabel() + " does not exist or already has a Kit Label");
-                } else {
-                    logger.info("Updated kitRequests w/ ddp_label " + kitRequestShipping.getDdpLabel());
-                    if (StringUtils.isNotBlank(kitRequestShipping.getBspCollaboratorParticipantId())) {
+                if (rowsAffected == 1) {
+                    logger.info("Updated ddp_kit.kit_label to {} for kit request  {}.",
+                            kitRequestShipping.getKitLabel(), kitRequestShipping.getDdpLabel());
+                    if (kitRequestShipping.hasBSPCollaboratorParticipantId()) {
+                        // todo arz how does this work?  A scan error with no error?
```
zyme: @pegahtah can you help me understand this special block of code that shows up when there's a bsp collab participant id?

pegahtah: Of course, for understanding this I need to first tell you  that the ScanError is used to communicate anything (not just errors) with the frontend. If the kit has a bsp participant id, we will send it to the frontend. So the frontend will show the successful scan using the returned response. (GP wants to see that as a final step to make sure they are sending the kit to the right participant)
 As an example Angio Dev returned this `[{"kit":"W9O6M23K73T36N8","shortId":"ASCProject_P9846A"}]` for a successful scan, in which `kit` is the `DSM shipping label` or `ddpLabel` and frontend uses it to display the bspParticipantId under the correct corresponding scan line
Screenshot 2023-11-22 at 11 09 45 AM

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -612,6 +757,39 @@ public List<KitRequestShipping> getKitsByHruid(String hruid) {
         return kitRequestList;
     }
 
+    public void updateKitRequestDDPLabel(int dsmKitRequestId, String ddpLabel) {
```
pegahtah: I am not able to find the usage of this method, but in general I can't think of the use case for changing the ddpLabel of a kit. It is one of the unique ids of the kit

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -612,6 +757,39 @@ public List<KitRequestShipping> getKitsByHruid(String hruid) {
...
+                            + "dsm kit request id %s", numRows, ddpLabel, dsmKitRequestId));
+                }
+            } catch (SQLException e) {
+                throw new DsmInternalError(String.format("Could not update ddp_label to %s for dsm kit request id%s",
+                        ddpLabel, dsmKitRequestId), e);
+            }
+        });
+    }
+
+    public void updateKitComplete(long dsmKitId, boolean isKitComplete) {
```
pegahtah, line 777: Can this be private and `VisibleForTesting` maybe?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: Gaah, I forgot to remove this--was a leftover from an earlier approach.  Will delete.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: What would you say to changing `ScanError` -> `ScanResponse` to reflect the fact that it's not just for errors?

zyme: PR for *just* changing ScanError -> ScanResult is [here](https://github.com/broadinstitute/ddp-study-server/pull/2743).  Keeping it separate so we can evaluate whether we want to do it independently.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
pegahtah: Yes, I double checked, ddp_label should be unique per kit entry

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -66,19 +68,31 @@ public class KitDaoImpl implements KitDao {
             + "(SELECT 1 FROM "
             + "ddp_kit_tracking tracking "
             + "WHERE tracking.kit_label = ?) AS existing_rows";
-    private static final String UPDATE_KIT_REQUEST = "UPDATE ddp_kit SET "
+
+    private static final String UPDATE_KIT_REQUEST_DDP_LABEL = "update ddp_kit_request set ddp_label = ? where "
```
pegahtah: Now that the method is removed do we need this one?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -163,32 +196,98 @@ public Boolean hasTrackingScan(String kitLabel) {
...
+                                    numRows, kitRequestShipping.getDdpLabel()));
+                    return dbVals;
+                } // else one row found, as expected
+            } catch (SQLException ex) {
+                logger.error("Not able to query the kit request for ddpLabel " + kitRequestShipping.getDdpLabel(), ex);
+                dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(),
+                        "An unexpected error has occurred");
+            }
+
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_NUM_KITS_BY_LABEL)) {
```
pegahtah: Do we need this after checking the ddpLabel exists in Line 211 with `SELECT_1_FROM_KIT_REQUEST_WITH_DDP_LABEL`?  Maybe we can do only `SQL_NUM_KITS_BY_LABEL` which is checking both ddp_kit and ddp_kit_request table?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -163,32 +196,98 @@ public Boolean hasTrackingScan(String kitLabel) {
...
+                } else if (numRows > 1) {
+                    dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(),
+                            String.format("Found %s matching kit request rows for ddp_label %s.",
+                                    numRows, kitRequestShipping.getDdpLabel()));
+                    return dbVals;
+                } // else one row found, as expected
+            } catch (SQLException ex) {
+                logger.error("Not able to query the kit request for ddpLabel " + kitRequestShipping.getDdpLabel(), ex);
+                dbVals.resultValue = new ScanError(kitRequestShipping.getDdpLabel(),
+                        "An unexpected error has occurred");
```
pegahtah: Should we return the dbVals here? Or do we want to continue after a SQL exception?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -369,14 +468,31 @@ public Integer deleteKitRequest(Long kitRequestId) {
         });
 
         if (simpleResult.resultException != null) {
-            throw new DsmInternalError(String.format("Error deleting kit request with id: %d", kitRequestId),
-                    simpleResult.resultException);
+            throw new RuntimeException("Error deleting kit request with id: " + kitRequestId, simpleResult.resultException);
```
pegahtah: I think DSMInternalError is more accurate for here since we are moving on from RuntimeExcpetions

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: Indeed we don't

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: I think we need both?  One checks for existence of the kit request, the other for the ddp kit.

zyme: On the other hand, all the tests pass after I remove this query, so I think we can ditch it!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
zyme: Nice catch!  Yes, this should return early.

pegahtah (approved):

PR #2714, opened 2023-10-23
PEPPER-784 moved email.notifications to secret manager

zyme:
Follow-up to PEPPER-784!  Moved `email.notifications` from `application.conf` to secret manager so that different environments can use different notification template ids, instead of having one set of template ids apply to all environments.  Along the way, I expanded `NotificationUtilTest` so that it asserts the proper structure of email templates, although this does not actually verify these changes directly.

## FUD Score

After deploying these changes, notifications should be sent as per usual, even though the `email.notifications` stanza in `application.conf` is gone.  It will be in secret manager, just like with `reminderNotifications`.

## Testing

The code that does hocon configuration is doing a "merge" of sorts with config files, using a `withFallback` approach that will read `email.notifications` from either SM or `application.conf`.  If `email.notifications` is found in the static config file when `DSMServer` starts, a warning will be logged.

## Release

- [x ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

### DIFF
--- a/pepper-apis/dsm-core/src/main/resources/application.conf
+++ b/pepper-apis/dsm-core/src/main/resources/application.conf
@@ -1,14 +1,3 @@
-email {
-  notifications: """
-        [{"reason": "PARTICIPANT_ASSIGNED", "sendGridTemplate": "c163e30a-8e5d-4fd0-b239-798554015575", "url": ""},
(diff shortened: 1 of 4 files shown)


pepper-apis/dsm-core/src/main/resources/application.conf
```diff
@@ -1,14 +1,3 @@
-email {
```
zyme, line 1: Moved to secret manager.

pegahtah (approved):

PR #2719, opened 2023-10-25
[DSM] PEPPER-1195 LabelCreationJob breaking for subkits

pegahtah:
## Context

The label creation was throwing an error because the sub kit id is not presen tdirectly in the ddp_kit_request_settings table, so the join between the tables on kit type was not returning any results and therefore we were seeing exceptions when trying to buy shipments. 

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestCreateLabel.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestCreateLabel.java
@@ -38,7 +38,7 @@ public class KitRequestCreateLabel {
     private final String participantCollaboratorId;
     private final String collaboratorIdPrefix;
     private final boolean hasAuth0Token;
-    private final KitRequestSettings kitRequestSettings;
+    private KitRequestSettings kitRequestSettings;
     private final KitType kitTyp;
     private final String billingReference;
     private final String participantIndexES;
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestCreateLabel.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestCreateLabel.java
@@ -38,7 +38,7 @@ public class KitRequestCreateLabel {
     private final String participantCollaboratorId;
     private final String collaboratorIdPrefix;
     private final boolean hasAuth0Token;
-    private KitRequestSettings kitRequestSettings;
+    private final KitRequestSettings kitRequestSettings;
     private final KitType kitTyp;
     private final String billingReference;
     private final String participantIndexES;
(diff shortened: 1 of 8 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestCreateLabel.java
```diff
@@ -38,7 +38,7 @@ public class KitRequestCreateLabel {
     private final String participantCollaboratorId;
     private final String collaboratorIdPrefix;
     private final boolean hasAuth0Token;
-    private final KitRequestSettings kitRequestSettings;
+    private KitRequestSettings kitRequestSettings;
```
pegahtah: Had to change it from final so that if it was null in the first try, we can find it directly by the subKit's kit type id and replace it with right values

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/KitRequestSettings.java
```diff
@@ -46,6 +46,31 @@ public class KitRequestSettings {
...
+                    + "WHERE subK.ddp_kit_request_settings_id = dkc2.ddp_kit_request_settings_id "
+                    + " AND dkc2.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id) AS has_sub_kits "
+                    + "FROM ddp_kit_request_settings dkc "
+                    + "LEFT JOIN kit_dimension dim ON (dkc.kit_dimension_id = dim.kit_dimension_id) "
+                    + "LEFT JOIN carrier_service cs_to ON (dkc.carrier_service_to_id=cs_to.carrier_service_id) "
+                    + "LEFT JOIN carrier_service cs_return ON (dkc.carrier_service_return_id=cs_return.carrier_service_id) "
+                    + "LEFT JOIN kit_return_information ret ON (dkc.kit_return_id=ret.kit_return_id) "
+                    + "LEFT JOIN sub_kits_settings subK ON (subK.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id) "
+                    + "LEFT JOIN kit_type kt ON (subK.kit_type_id = kt.kit_type_id) ";
+
```
pegahtah, line 74: This is very close to the query `SQL_SELECT_CARRIER` , and looks very complicated but is basically joining the ddp_kit_request_settings table with all of the other shipmen tables (like return information, kit dimensions, carriers, etc ) and then the difference is that the kit type here is found by joining it with the sub_kits_settings table instead of kit_type table

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -224,6 +235,17 @@ public static void createLabel(List<KitRequestCreateLabel> kitsLabelTriggered, E
...
+                && kitRequestSettings.getCarrierReturn() == null && kitRequestSettings.getServiceReturn() == null && kitRequestSettings.getCarrierRetrunAccountNumber() == null
+                && kitRequestSettings.getWeight() == null && kitRequestSettings.getLength() == null && kitRequestSettings.getHeight() == null && kitRequestSettings.getWeight() == null
+                && kitRequestSettings.getCollaboratorSampleTypeOverwrite() == null && kitRequestSettings.getCollaboratorParticipantLengthOverwrite() == null && kitRequestSettings.getReturnName() == null
+                && kitRequestSettings.getReturnStreet1() == null && kitRequestSettings.getReturnStreet2() == null && kitRequestSettings.getReturnCity() == null && kitRequestSettings.getReturnZip() == null
+                && kitRequestSettings.getReturnState() == null && kitRequestSettings.getReturnCountry() == null && kitRequestSettings.getPhone() == null && kitRequestSettings.getDisplayName() == null
+                && kitRequestSettings.getExternalShipper() == null && kitRequestSettings.getExternalShipperKitName() == null));
```
pegahtah: Not all values will be null, so I found the ones that would mean that the query went wrong. if one of these values has a value it could mean that it's not a sub kits problem because the joins between tables are working. Also we should not look at empty strings because they would also mean the value was found and only the config was set in a way that a value is null (hence the comparison to null instead of StringUtils.isBlank())

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -84,7 +85,39 @@ public class KitUtil {
                     + "LEFT JOIN carrier_service cs_to on (dkc.carrier_service_to_id=cs_to.carrier_service_id) "
                     + "LEFT JOIN carrier_service cs_return on (dkc.carrier_service_return_id=cs_return.carrier_service_id) "
                     + "LEFT JOIN kit_return_information ret on (dkc.kit_return_id=ret.kit_return_id) "
-                    + "LEFT JOIN kit_type t on (request.kit_type_id = t.kit_type_id) where ex.ddp_participant_exit_id is null";
+                    + "LEFT JOIN kit_type t on (request.kit_type_id = t.kit_type_id) where ex.ddp_participant_exit_id is null "
+                    + "and ddp_kit_request_settings_id is not null";
```
pegahtah, line 93: only return kits that we can get their kit settings directly. This will make sure there is no overlap

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -232,38 +265,7 @@ public static List<KitRequestCreateLabel> getListOfKitsLabelTriggered() {
                     SQL_SELECT_KIT_REQUESTS_ALL_INFO.concat(QueryExtension.KIT_LABEL_TRIGGERED))) {
                 try (ResultSet rs = bspStatement.executeQuery()) {
                     while (rs.next()) {
-                        kitsLabelTriggered.add(new KitRequestCreateLabel(rs.getString(DBConstants.DSM_KIT_ID),
```
pegahtah, line 235: some refactoring here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
zyme: You mean no overlap with subkits?  If so, this is a generally useful remark so please enshrine it in the code so that those who come after us will always know why it's here :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
```diff
@@ -162,7 +161,6 @@ public Optional<ScanError> processRGPFinalScan(ScanPayload scanPayload) {
                 //successfully scanned and going to update db and ES
                 if (StringUtils.isEmpty(rgpRnaKit.getKitLabel())) {
                     rgpRnaKit.setKitLabel(RNA);
-                    rgpRnaKit.setDdpLabel(rgpRnaKit.getDdpLabel());
```
pegahtah, line 165: redundant and not needed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/kit/KitFinalScanUseCase.java
```diff
@@ -135,7 +135,6 @@ public Optional<ScanError> processRGPFinalScan(ScanPayload scanPayload) {
                 if (StringUtils.isNotEmpty(rgpBloodKit.getKitLabel()) && kitLabel.equals(rgpBloodKit.getKitLabel())
                         || StringUtils.isEmpty(rgpBloodKit.getKitLabel())) {
                     rgpBloodKit.setKitLabel(kitLabel);
-                    rgpBloodKit.setDdpLabel(ddpLabel);
```
pegahtah, line 138: I found this bug while testing again. This `rgpBloodKit.setDdpLabel(ddpLabel)` is assuming something that is not always true, even though it may not break all the time there is no need to setDdpLabel at this place. `rgpBloodKit` always has a ddpLabel

zyme (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -288,7 +288,7 @@ public static List<KitRequestCreateLabel> getListOfKitsLabelTriggered() {
         Map<String, KitRequestCreateLabel> subKitsLabelTriggered = KitUtil.getListOfSubKitsThatNeedLabels();
         for (Map.Entry<String, KitRequestCreateLabel> subKitEntry : subKitsLabelTriggered.entrySet()) {
             if (!kitsLabelTriggered.containsKey(subKitEntry.getKey()) ||
-                    kitsLabelTriggered.get(subKitEntry.getKey()).getKitRequestSettings().getReturnCountry() == null) {
+                    kitsLabelTriggered.get(subKitEntry.getKey()).getKitRequestSettings().getReturnName() == null) {
```
pegahtah, line 291: This is to fix the breaking tests, and is also a better check because not only the name should also not be null ( it has a `@NonNull ` check when creating return address) but all the test setup is focused more on names.

PR #2721, opened 2023-10-27
PEPPER-1188 Bumping redisson library

zyme:
Appsec flagged our redisson library as a vulnerable, so we're upgrading to 3.24.2

## FUD Score

Let our existing tests do their thing!

## Release


pepper-apis/pom.xml
```diff
@@ -25,6 +25,7 @@
         <elasticsearch.version>7.17.10</elasticsearch.version>
         <velocity.version>2.3</velocity.version>
         <jackson.version>2.12.1</jackson.version>
+        <redisson.version>3.24.2</redisson.version>
```
zyme, line 28: Extracting this so we only have to update the version in one place in the future.

ssettipalli (approved):

PR #2722, opened 2023-10-27
PEPPER-1200 log participant id when default values are missing

zyme:
PEPPER-1200.  We sometimes see `Could not add AT default values ` errors coming from DSM.  These changes will improve things by including the participant id in the error message when default values are missing.

Just calling `getParticipantId()` when creating the error message did not do the trick because the id comes from the profile, which itself is empty when this exception is thrown.  I have introduced a new field called `queriedParticipantId` to track which participant id was used when hitting ES.  **Not all code paths query by participant id, so there are situations in which there will not be a participant id to include in the error message**, but many code paths do query by a single participant id.  This `queriedParticipantId` does **not** get set in ES.  It is only set in the DTO on-the-fly.

Thanks to Dennis, I was able to put together some tests that hit the new ES container to verify these changes.  I made some updates to the ES container classes to help clarify use for others down the road.  Because A-T has many fields in its index, the test ran into an error because `index.mapping.total_fields.limit` (default: 1000) was not sufficient to handle all of A-T's fields ([ES docs here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-settings-limit.html)).  Similar to how we keep a mappings json in the resources directory for each index, we can now also create a per-index settings.json, which is where this value is set.

I have updated some classes to change visibility of various fields from `protected` to `public` for testability.

I suggest starting the review at [ATDefaultValues](https://github.com/broadinstitute/ddp-study-server/pull/2722/files#diff-338ce4c6381c01e0f47a8548a7fe31d06716df94e74a77f503db1392a6d90764) to see how this new field shows up in the stack trace, then moving to [ElasticSearchParticipantDto](https://github.com/broadinstitute/ddp-study-server/pull/2722/files#diff-0c6339475a6bfa8ae8e6fb5ece84a01e10b3113a63f6252bf97434dc8d75356f), and [ElasticSearch](https://github.com/broadinstitute/ddp-study-server/pull/2722/files#diff-fc214be67f7abaa63f0fda9ef29b8b8565b999959d24493d3e2aaf209ca31bf0) to see how the `queriedParticipantId` is set, and then checking out the updated tests and [ElasticTestUtil](https://github.com/broadinstitute/ddp-study-server/pull/2722/files#diff-5aa4d843226f8ab90cce84858bc4348debb418a7871a7b4b1512d2a37ac838ec).

One question I have for reviewers: is it possible to get non-prod into a state where we can see this error?  Can we delete a profile for an A-T participant and verify that we get the participant id in the stack trace?

## FUD Score

## Testing

- [ x] I have added regression test labels to the jira ticket

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -114,6 +123,11 @@ public List<String> getGovernedUsers() {
         return governedUsers == null ? Collections.emptyList() : governedUsers;
     }
 
+
+    /**
+     * Returns the participant id by checking the profile
+     * for guid and altpid, preferentially returning the guid.
+     */
```
zyme, line 129: Initially I had changed `getParticipantId()` so that it would return the `queriedParticipantId` when the profile was null or the guids were empty, but this felt pretty clunky and overly complex to me, so I opted for a separate field whose purpose is more clear.

pepper-apis/dsm-core/src/test/resources/elastic/atcpSettings.json
```diff
@@ -0,0 +1,9 @@
+{
```
zyme, line 1: A-T has > 1000 fields. 
 https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-settings-limit.html

pepper-apis/dsm-core/src/test/resources/elastic/atcpMappings.json
```diff
@@ -0,0 +1,5172 @@
+{
```
zyme, line 1: Extracted from prod

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -34,11 +34,11 @@ public class ElasticSearchParticipantDto {
     private List<Map<String, Object>> workflows;
     private String status;
     private Dsm dsm;
+    private String queriedParticipantId; // optional, used for troubleshooting to report
```
denniscunningham: Nit: perhaps put this comment above `queriedParticipantId`. As is it appears to be two comments, the second applying to the getter.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -76,14 +92,37 @@ public static void deleteIndex(String esIndex) {
         }
     }
 
+    /**
+     * Updates the mappings for the index.  When creating an index for the first time,
+     * query it from an existing elastic instance by running a command like this
+     */
```
denniscunningham, line 98: Like what?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/BasicDefaultDataMaker.java
```diff
@@ -28,7 +28,8 @@ public abstract class BasicDefaultDataMaker implements Defaultable {
     protected ElasticSearchParticipantDto elasticSearchParticipantDto;
 
 
-    protected abstract boolean setDefaultData();
+    @VisibleForTesting
+    abstract boolean setDefaultData();
```
denniscunningham: Is this actually used in a test somewhere?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/ATDefaultValues.java
```diff
@@ -42,9 +42,11 @@ public class ATDefaultValues extends BasicDefaultDataMaker {
     private Dao dataAccess;
 
     @Override
-    protected boolean setDefaultData() {
+    boolean setDefaultData() {
         if (isParticipantDataNotInES()) {
-            throw new ESMissingParticipantDataException("Participant does not yet have profile and activities in ES");
+            throw new ESMissingParticipantDataException(
+                    String.format("Participant %s does not yet have profile and activities in ES",
+                    elasticSearchParticipantDto.getQueriedParticipantId()));
```
denniscunningham, line 49: I'm not sure I understand when the queried ptp ID would ever be set in this case.

denniscunningham (commented):
Before I go any further with this review, we know why we are seeing the behavior this code is hoping to flush out, and there is a straightforward fix. See https://broadworkbench.atlassian.net/browse/PEPPER-1077 for an explanation.
In those cases there is no upstream code that is looking for a particular participant, so this code is not relevant. 
Have we seen other instances in prod that do not fit the pattern in https://broadworkbench.atlassian.net/browse/PEPPER-1077?

Regarding this PR, is the queried ptp ID ever written to the index, or just added to the DTO after it is read from ES?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
zyme: Whoops, I moved the guts of this to the createIndex method.  Updating!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/BasicDefaultDataMaker.java
zyme: No--I forgot to undo this as I experimented with different approaches.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/ATDefaultValues.java
zyme: It'll be set in `ElasticSearch.parseSourceMap`: https://github.com/broadinstitute/ddp-study-server/pull/2722/files#diff-9b912998c53490bce5548304c2dbaa2977c084100ace3d6b8d38bceafcd1e8ed

zyme: But only when we are querying a participant record by a participant id.

denniscunningham (approved):

PR #2725, opened 2023-11-01
PEPPER-1192. Add 2 new gastro cancers for pedHCC

ssettipalli:
Add 2 new gastro cancers for pedHCC to pancan FAMILY_HISTORY cancer list


Rinchan20 (approved):
took a look in dev - seems to show up in English and Spanish

PR #2727, opened 2023-11-03
[DSM] PEPPER-1204

pegahtah:
This PR implements the modifications specified in PEPPER-1204, transitioning the DSM's deletion feature from a soft delete—where the 'deleted' flag is toggled to '1'—to a hard delete, which involves the complete removal of the record from the database.

The process starts at `PatchFactory.makePatch()`, where DSM evaluates if the patch request from the frontend seeks to delete an entity. If so, `DeletePatchFactory.produce()` generates the appropriate `DeletePatch`. This new `DeletePatch` significantly alters the behavior of the `doPatch` and `handleSingleNameValue` methods previously used by an `ExistingRecordPatch`. Essentially, `DeletePatch.doPatch()` first ascertains the safety of deletion by verifying the record's absence in other tables. Following this, it proceeds to gather any associated 'children' records—such as tissues linked to an OncHistory, or SmIds associated with a Tissue. These associated records must be removed before the main object to maintain SQL foreign key integrity. The record's JSON representation is preserved in a `deleted_object` table before its deletion. Additionally, `doPatch` ensures the removal of the object from the ES index document post-deletion (look at `ElasticSearchDeleteUtil`).


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -24,13 +28,13 @@ public static BasePatch makePatch(Patch patch, NotificationUtil notificationUtil
                 patcher = new TissuePatch(patch);
             } else if (isSmIdCreation(patch)) {
                 patcher = new SMIDPatch(patch);
-            } else if (isParentParticipandDataId(patch)) {
```
pegahtah, line 27: typo fixed

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -1086,6 +1087,13 @@ private void setupRouteGenericErrorHandlers() {
             response.status(500);
             response.body(exception.getMessage());
         });
+        // todo not sure if this will remain in this ticket or not
+        exception(UnsafeDeleteError.class, (exception, request, response) -> {
+            logger.error("DSM is unable to delete the object {}",  exception.toString());
+            exception.printStackTrace();
+            response.status(500);
+            response.body(exception.getMessage());
+        });
```
pegahtah, line 1095: A way to respond to frontend for in case the object they want to delete is in the a clinical order

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -161,30 +185,67 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
         return true;
     }
 
-    private static String toErrorMessage(String table, String column, Object value, String primaryKey, String id) {
-        return String.format("Error updating %s.%s with value %s for %s=%s", table, column, value, primaryKey, id);
-    }
+    /****/
 
-    public static Boolean isValueUnique(@NonNull DBElement dbElement) {
```
pegahtah, line 168: This was not used

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/UnsafeDeleteError.java
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.dsm.exception;
+
+public class UnsafeDeleteError extends RuntimeException {
+    public UnsafeDeleteError(String message) {
+        super(message);
+    }
+
+}
```
pegahtah, line 8: I created a special exception for this just in case, we can delete later if we decide we want to do something different

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
```diff
@@ -0,0 +1,48 @@
...
+        if (Patch.deletePatch(patch.getId(), patch.getUser(), patch.getNameValue(), dbElement, patch)) {
+// todo pegah add remove and export to ES here
+            return nameValues;
+        }
+        return nameValues;
+    }
+
+    //TODO this is supposed to get handled by PEPPER-1209
+    private boolean isSafeToDelete() {
+        return true;
```
zyme: This will basically be ClinicalOrderDao.getClinicalOrdersForTissueIds, then check that returned map has size 0 for all contained lists.

Will need to pass tissueIds to this method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -24,13 +28,13 @@ public static BasePatch makePatch(Patch patch, NotificationUtil notificationUtil
...
+            } else if (isParentParticipantDataId(patch)) {
                 patcher = new ParticipantDataPatch(patch);
             } else if (isParticipantIdForRecord(patch)) {
                 patcher = new ParticipantRecordPatch(patch);
             }
         } else if (isParentParticipantId(patch) && !isMedicalRecordAbstractionFieldId(patch)
-                    && StringUtils.isNotBlank(patch.getDdpParticipantId())) {
+                && StringUtils.isNotBlank(patch.getDdpParticipantId())) {
```
zyme: Consider making a `hasDDPParticipantId() method on patch instead of having clients do things like `StringUtils.isNotBlank(patch.getDdpParticipantId())`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,86 @@
...
+            deletePatches = getPatchForTissues(originalPatch);
+        }
+        for (Patch childPatch : deletePatches) {
+            BasePatch patcher = PatchFactory.makePatch(childPatch, notificationUtil);
+            patcher.doPatch();
+        }
+    }
+
+    private static boolean isOncHistoryPatch(Patch originalPatch) {
+        return originalPatch.getNameValue().getName().equals(ONC_HISTORY_DELETED_FIELD) &&
```
zyme: Consider moving these methods into `Patch` class itself so that clients can just ask `Patch`: are you an onc history patch?  And then Patch itself can have control over how it answers the question.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
```diff
@@ -0,0 +1,201 @@
...
+            + " from information_schema.columns where table_name = 'ddp_tissue'";
+
+    private static final String SQL_GET_JSON_OBJECT_FOR_SM_ID = "select concat('select json_object(', group_concat( "
+            + " concat(quote(column_name), ', ', column_name)), ') from sm_id where sm_id_pk = ? ;') into @sql "
+            + " from information_schema.columns where table_name = 'sm_id'";
+
+    private static final String SQL_GET_JSON_OBJECT_FOR_ONC_HISTORY = "select concat('select json_object(', group_concat( "
+            + " concat(quote(column_name), ', ', column_name)),') from ddp_onc_history_detail where onc_history_detail_id = ? ;')"
+            + " into @sql from information_schema.columns where table_name = 'ddp_onc_history_detail'";
+
```
pegahtah, line 33: I made 3 different strings that look the same to avoid the risk of SQL injection, let me know if you find this unnecessary

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -52,13 +52,6 @@
  */
 @Slf4j
 public class JuniperSetupUtil {
-    private static final String INSERT_JUNIPER_GROUP =
-            "INSERT INTO ddp_group (name) VALUES (?) ON DUPLICATE KEY UPDATE name = 'juniper test';";
-    private static final String INSERT_JUNIPER_INSTANCE =
-            "INSERT INTO ddp_instance (instance_name, study_guid, display_name, is_active, bsp_organism, "
-                    + " collaborator_id_prefix, auth0_token) VALUES (?, ?, ?, 1, 1, ?, 0) ON DUPLICATE KEY UPDATE auth0_token = 0;";
-    private static final String INSERT_DDP_INSTANCE_GROUP = "INSERT INTO ddp_instance_group (ddp_instance_id, ddp_group_id) "
-            + " VALUES (?, ?) ON DUPLICATE KEY UPDATE ddp_group_id = ?;";
```
pegahtah, line 61: All moved to the new DdpInstanceGroupTestUtil

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/JuniperSetupUtil.java
```diff
@@ -340,31 +334,4 @@ private String createDdpInstanceRole(Connection conn) throws SQLException {
...
-        PreparedStatement stmt = conn.prepareStatement(INSERT_JUNIPER_INSTANCE, Statement.RETURN_GENERATED_KEYS);
-        stmt.setString(1, instanceName);
-        stmt.setString(2, studyGuid);
-        stmt.setString(3, displayName);
-        stmt.setString(4, collaboratorPrefix);
-        stmt.executeUpdate();
-        ResultSet rs = stmt.getGeneratedKeys();
-        return getPrimaryKey(rs, "ddp_instance");
-    }
-
```
pegahtah, line 369: Not needed anymore

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/admin/UserAdminTestUtil.java
```diff
@@ -309,109 +320,6 @@ public int getRoleId(String roleName) {
...
-            try (PreparedStatement stmt = conn.prepareStatement(SQL_DELETE_DDP_INSTANCE)) {
-                stmt.setInt(1, instanceId);
-                return stmt.executeUpdate();
-            } catch (SQLException ex) {
-                String msg = String.format("Error deleting DDP instance: ddpInstanceId=%d", instanceId);
-                throw new DsmInternalError(msg, ex);
-            }
-        });
-    }
-
```
pegahtah, line 415: I created the DdpInstanceGroupTestUtil to avoid having to do this everytime

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final String PK_PLACE_HOLDER = "$pk";
+
+    private static final String SQL_DELETE_VALUES = "DELETE FROM $table WHERE $pk = ?";
+    private static final String SQL_INSERT_DELETED_VALUES = "INSERT INTO deleted_object SET original_table = ?, original_primary_key = ? ,"
+            + " data =?, deleted_by = ?,  deleted_date = ? ";
+
+    private static final String SQL_GET_DELETED_OBJECT_BY_ORIGINAL_ID_TABLE = "SELECT data FROM deleted_object WHERE original_table = ? "
+            + " AND original_primary_key = ? ";
+
+    public static boolean deletePatch(@NonNull String id, @NonNull String user,
```
zyme: `id` is numeric, right?  If so, change this to an int/long (whichever is appropriate given the size of the column), use setInt/Long instead of setString, and have client callers do any string -> int conversion necessary.

pegahtah: This id that comes from frontend is a String. I will change it both in the frontend code and here and make sure they match.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
```diff
@@ -0,0 +1,201 @@
...
+     * table later.
+     *
+     * @param conn       DB connection from the transaction that will do the whole delete procedure
+     * @param query      query to select from the right table
+     * @param id         primary key /id that is supposed to get deleted
+     * @param deleteType DeleteType of the patch.
+     * @return JSON representation of the object that will be deleted later
+     */
+
+    private static String getJsonObject(Connection conn, String query, String id, DeleteType deleteType) {
```
zyme: Instead of using `@sql` in other methods and then querying the string value here via `SQL_GET_JSON_OBJECT_FOR...`, it might be simpler and more testable to do 

```
select
    column_name from information_schema.columns where table_name = 'sm_id'
 ```
and then concatenate everything on the java side so the java code makes a big select  json_object statement like:

```
select
    json_object('sm_id_value', sm_id_value, 'changed_by', changed_by)

from sm_id where sm_id_pk = 1
```

and then execute that query.

File: pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java [hunk 14]
```diff
@@ -0,0 +1,201 @@
...
+    private static final String SQL_GET_DELETED_OBJECT_BY_ORIGINAL_ID_TABLE = "SELECT data FROM deleted_object WHERE original_table = ? "
+            + " AND original_primary_key = ? ";
+
+    public static boolean deletePatch(@NonNull String id, @NonNull String user,
+                                      @NonNull DBElement dbElement, DeleteType deleteType) {
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            //Select everything in the db as the current state of the object into a json string , this is better than creating a Dto because
+            // this can be done independent of class
+            String currentData = getJsonRepOfDeletedObject(conn, id, deleteType);
```
zyme, line 52: I think the transaction management here looks right--since we are using sql vars in this process, it's critical that we're not opening up separate transactions and setting `@sql`, since doing so could wipe out (or mis-set) `@sql` values across web requests.  I _think_ we're safe, but I'm flagging this as a concern if future editors split transactions in the wrong place.  Might be worth a comment or two and some warnings in the queries that use `@sql`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
zyme: SQL injection issues usually come up when code is blindly concatenating strings or running raw strings as SQL.  I think this isn't doing that--the are parameterized values and user input only gets into statements via stmt.setFoo().  If you are concerned that this is using SQL to generate SQL, that's not an injection hazard per se because the only part of the generated SQL that comes from user values is already sanitized upstream.

Having said that, this is one reason why setting ids as ints/longs is important--if you can only set an id as an int value, there's no string value that could possibly cause problems.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
zyme: By changing `id` to an int/long (and calling setInt/Long) this code will be much harder to exploit with sql injection.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DdpInstanceGroupTestUtil.java
```diff
@@ -0,0 +1,162 @@
...
+        return ddpInstanceDto;
+    }
+
+    public static void deleteInstanceGroup(int instanceId) {
+        inTransaction(conn -> {
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_DELETE_DDP_INSTANCE_GROUP)) {
+                stmt.setInt(1, instanceId);
+                return stmt.executeUpdate();
+            } catch (SQLException ex) {
+                String msg = String.format("Error deleting DDP instance group: ddpInstanceId=%d", instanceId);
```
zyme, line 97: Good good, exceptions with lots of details :)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
```diff
@@ -0,0 +1,105 @@
...
+                            participantId, instanceName, userEmail);
+            int oncHistoryDetailId = Integer.parseInt((String) response.get("oncHistoryDetailId"));
+            OncHistoryDetail oncHistoryDetail =
+                    OncHistoryDetail.getOncHistoryDetail(String.valueOf(oncHistoryDetailId), ddpInstanceDto.getInstanceName());
+            Assert.assertNotNull(oncHistoryDetail);
+            dsmOncHistoryCreatorUtil.deleteOncHistory(ParticipantDto.getDdpParticipantIdOrThrow(), participantId, instanceName, userEmail,
+                    oncHistoryDetailId);
+            dsmOncHistoryCreatorUtil.assertOncHistoryIsDeleted(guid, oncHistoryDetail, oncHistoryDetailId, ddpInstanceDto, true, true);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
zyme: Please add some context on this one--method params, etc.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DBTestUtil.java
```diff
@@ -1128,23 +1128,7 @@ public static void deleteMessage(int userId) {
         }
     }
 
-    public static DDPInstanceDto createTestDdpInstance(DDPInstanceDao ddpInstanceDao, String ddpInstanceName) {
-        return createTestDdpInstance(ddpInstanceDao, ddpInstanceName, null);
-    }
 
-    public static DDPInstanceDto createTestDdpInstance(DDPInstanceDao ddpInstanceDao, String ddpInstanceName, String esIndex) {
```
zyme, line 1135: Where did these go?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DSMOncHistoryCreatorUtil.java
```diff
@@ -0,0 +1,258 @@
...
+import org.mockito.Mock;
+import org.testcontainers.shaded.org.apache.commons.lang3.StringUtils;
+
+/**
+ * This class is to create participants with Medical Record and Onc Histories in DSM
+ * It utilises the UserAdminTestUtil to create a ddpInstance and a user with privileges
+ * and to use that to be
+ **/
+@Slf4j
+public class DSMOncHistoryCreatorUtil {
```
zyme: I think the naming patter we are using is *TestUtil, so this should be `DSMOncHIstoryCreatorUtil` or `DSMOncHistoryTestUtil`.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DSMOncHistoryCreatorUtil.java
```diff
@@ -0,0 +1,258 @@
...
+    public void initialize() {
+        ddpInstanceGroupTestUtil.createTestDdpInstance(instanceName, esIndex);
+        userAdminUtil.createRealmAndStudyGroup(instanceName, studyGuid, collabPrefix, groupName, esIndex);
+        userAdminUtil.setStudyAdminAndRoles("adminUserPatchTest@unittest.dev", USER_ADMIN_ROLE, Arrays.asList(MR_VIEW));
+        userId = Integer.toString(userAdminUtil.createTestUser(userEmail, Collections.singletonList("mr_view")));
+        //todo create instance with role
+    }
+
+    public void deleteEverything() {
+        //this bit is failing now for foreign keys
```
zyme: Let's clean this up so we get clean deletes.

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1204_dsm_delete.xml
```diff
@@ -0,0 +1,23 @@
...
+            <column name="deleted_object_id" autoIncrement="true" type="INT">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="original_table" type="VARCHAR(200)">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="original_primary_key" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="data" type="text"></column>
```
zyme, line 18: Is `text` large enough to capture the biggest json-ified versions of the tables we are deleting?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatchFactory.java
```diff
@@ -0,0 +1,39 @@
...
+    public static BasePatch produce(Patch patch, NotificationUtil notificationUtil) {
+        BasePatch patcher;
+        if (PatchFactory.isOncHistoryDetailPatch(patch)) {
+            patcher = new DeletePatch(patch, notificationUtil, DeleteType.ONC_HISTORY_DETAIL);
+        } else if (PatchFactory.isTissueRelatedOncHistoryId(patch)) {
+            patcher = new DeletePatch(patch, notificationUtil, DeleteType.TISSUE);
+        }  else if (PatchFactory.isSmIdPatch(patch)) {
+            patcher = new DeletePatch(patch, notificationUtil, DeleteType.SM_ID);
+        } else {
+            throw new DsmInternalError("This method should not reach here");
```
zyme: How about `No support for deleting [some kind of type information]`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
zyme: Great to see pruning of irrelevant code :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -161,30 +179,50 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
+        for (String smIdPk : smIds) {
+            NameValue nameValue = new NameValue(SM_ID_DELETED_FIELD, TRUE_FLAG);
+            deletePatches.add(
+                    new Patch(smIdPk, TissuePatch.TISSUE_ID, tissuePatch.getId(), tissuePatch.getUser(), nameValue, null, true,
+                            tissuePatch.getDdpParticipantId(), tissuePatch.getRealm()));
+        }
+        return deletePatches;
+    }
+
+    public static boolean hasDDPParticipantId(Patch patch) {
```
zyme: How about adding this as an instance method on `Patch`?  Having a static method that takes in an instance to inspect the instance's field is a bit of an antipattern.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    Object handleSingleNameValue() {
+        List<NameValue> nameValues = new ArrayList<>();
+        if (DeletedObjectDao.deletePatch(patch.getId(), patch.getUser(), dbElement, deleteType)) {
+            DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(patch.getRealm()).orElseThrow();
+            try {
+                ElasticSearchDeleteUtil.deleteFromIndexById(patch.getDdpParticipantId(), Integer.parseInt(patch.getId()), ddpInstanceDto, deleteType);
+            } catch (Exception e) {
+                throw new DsmInternalError("Unable to delete from ES for patch with id "+patch.getId() +" for NameValue: "+patch.getNameValue().toString(),e);
```
zyme: Is there something we can log here to help us recover from a successful db delete but a failed ES delete?  If we see this exception thrown, how would we go about synchronizing the state of ES?  A full export?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
```diff
@@ -0,0 +1,105 @@
...
+    static String userEmail = "patchTestUser1@unittest.dev";
+    static String guid = "PATCH_TEST_PARTICIPANT";
+    static String guid2 = "PATCH_PARTICIPANT";
+    private static String esIndex;
+    private static DDPInstanceDto ddpInstanceDto;
+    private static ParticipantDto testParticipant = null;
+
+    @BeforeClass
+    public static void doFirst() {
+        esIndex = ElasticTestUtil.createIndexWithMappings(instanceName, "elastic/lmsMappings.json");
```
zyme: It's worth it to take the time to try dump another index here if the main driver is another project like OS.  It's pretty straightforward--ElasticTestUtil has some pointers on how to grab index/settings files from an existing instance.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.broadinstitute.dsm.db.dto.ddp.instance.DDPInstanceDto;
+import org.broadinstitute.dsm.db.dto.ddp.participant.ParticipantDto;
+import org.broadinstitute.dsm.util.DSMOncHistoryCreatorUtil;
+import org.broadinstitute.dsm.util.ElasticTestUtil;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ObjectDeleteTest extends DbAndElasticBaseTest {
```
zyme, line 22: YAY!  More ES tests!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
zyme: Or if this should be a Assert.fail() on exception...

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
```diff
@@ -0,0 +1,105 @@
...
+            Assert.assertNotNull(oncHistoryDetail);
+            dsmOncHistoryCreatorUtil.deleteOncHistory(ParticipantDto.getDdpParticipantIdOrThrow(), participantId, instanceName, userEmail,
+                    oncHistoryDetailId);
+            dsmOncHistoryCreatorUtil.assertOncHistoryIsDeleted(guid, oncHistoryDetail, oncHistoryDetailId, ddpInstanceDto, true, true);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+    }
+
```
zyme, line 77: Should there be additional tests here for deleting tissues and samples, independent of onc histories?  Those would help shine a light on any issues with underlying tissue/sm deletes, especially if users can delete them individually.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
zyme: I'm not requesting any changes here--just highlighting that by moving more work out from the database to the middleware, it might be easier to test/maintain.  Something to bear in mind down the road.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/UnsafeDeleteError.java
zyme: A couple questions to decide whether there is value in this exception:
1. What do we do programmatically in response to this class?
2. What would we do if we saw this pop up in an alert?

If the answers are "nothing", then this probably isn't worth keeping.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
zyme: I'm not sure we want to do anything in response to this-if someone tries to do something that the system should prevent, and the system prevents it, there's no need for us to be alerted to it, although the user should receive some sort of indication of what happened.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/GroupDao.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final String SQL_DELETE_GROUP = "DELETE FROM ddp_group WHERE group_id = ? ";
+
+
+    @Override
+    public int create(String groupName) {
+        SimpleResult results = inTransaction(conn -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_INSERT_GROUP, Statement.RETURN_GENERATED_KEYS)) {
+                stmt.setString(1, groupName);
+                stmt.executeUpdate();
```
zyme: Should the return value of `executeUpdate()` be checked?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
```diff
@@ -0,0 +1,201 @@
...
+     * @param deleteType DeleteType of the patch.
+     * @return JSON representation of the object that will be deleted later
+     */
+
+    private static String getJsonObject(Connection conn, String query, String id, DeleteType deleteType) {
+        //Select everything in the db as the current state of the object into a json string , this is better than creating a Dto because
+        // this can be done independent of class and changes to the schema
+        String data = null;
+        try (PreparedStatement stmt = conn.prepareStatement(query)) {
+            stmt.executeQuery();
```
zyme, line 160: Should the return value of `executeQuery()` be checked?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
zyme: It might not be worth changing it on the frontend--the important thing is that by the time it gets into this method, it's treated as an int (which could be a change to the signature, or could just be internal to the method to convert to an int/long.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ElasticSearchDeleteUtil.java
```diff
@@ -54,10 +55,10 @@ public static void deleteOncHistoryDetailById(String ddpParticipantId, int oncHi
                 (List<Map<String, Object>>) ((Map<String, Object>) esDsmMap.get(ESObjectConstants.DSM)).get(
                         ESObjectConstants.ONC_HISTORY_DETAIL);
         oncHistoryDetails.removeIf(stringObjectMap -> (int) stringObjectMap.get("oncHistoryDetailId") == oncHistoryDetailId);
-        esDsmMap.put(ESObjectConstants.ONC_HISTORY_DETAIL, oncHistoryDetails);
-
+        ((HashMap<String,List<Map<String, Object>>>) esDsmMap.get(ESObjectConstants.DSM)).put(ESObjectConstants.ONC_HISTORY_DETAIL,
```
zyme: Is there any chance that `esDsmMap.get(ESObjectConstants.DSM)` will return a null value?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SmId.java
```diff
@@ -49,6 +44,12 @@ public SmId(Integer smIdPk, String smIdType, String smIdValue, Integer tissueId)
         this.tissueId = tissueId;
     }
 
+    public SmId(Integer smIdPk, String smIdValue, Integer tissueId) {
```
zyme: `smIdPk` should be an int--looking at these constructor usages, they are all called with `getInt()`, which returns an `int`, so it should be safe to changed `smIdPk` in the constructors to `int` and then change calling classes to use `int foo = rs.getInt(...)`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SmId.java
```diff
@@ -77,10 +78,6 @@ public static SmId getSMIdsForTissueId(ResultSet rs) {
         return tissueSmId;
     }
 
-    public static boolean isUniqueSmId(String smIdValue, String id) {
```
zyme, line 80: Nice cleanup!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -297,4 +300,28 @@ public static List<String> getSmIdPksForTissue(String tissueId) {
         logger.info("Got %d smId pks in DSM DB for tissue with id %s", smIds.size() , tissueId);
         return smIds;
     }
+
+    public SmId get(int id) {
```
zyme: How about a method name like `findById()`?  `get()` implies a simple return of state, whereas this method is doing a db lookup.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
```diff
@@ -17,6 +20,7 @@
 @Slf4j
 public class DeletePatch extends ExistingRecordPatch {
     DeleteType deleteType;
+    static ClinicalOrderDao clinicalOrderDao = new ClinicalOrderDao();
```
zyme: Although `ClinicalOrderDao` appears stateless, it would be safer to make `clinicalOrderDao` an instance variable rather than a static, and then create a new ClinicalOrderDao on-the-fly in `isSafeToDelete()`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
```diff
@@ -72,9 +75,25 @@ Optional<Object> processSingleNameValue() {
...
-    private boolean isSafeToDelete() {
-        return true;
+    public static boolean isSafeToDelete(List<Integer> tissueIds) {
+        if (tissueIds == null || tissueIds.isEmpty())
+            return true;
+        return clinicalOrderDao.getClinicalOrdersForTissueIds(tissueIds).values().stream().filter(list -> !list.isEmpty()).findAny()
+                .isEmpty();
+    }
+
+    public static boolean isSafeToDelete(Patch patch) {
```
zyme: Consider this non-static implementation instead:

```
public boolean isSafeToDelete() {
        List<Integer> tissueIds = new ArrayList<>();
        if (Patch.isTissueDeletePatch(patch)) {
            tissueIds.add(Integer.parseInt(patch.getId()));
        } else if (Patch.isSmIdDeletePatch(patch)) {
            int tissueId = new TissueSMIDDao().get(Integer.parseInt(patch.getId())).getTissueId();
            tissueIds.add(tissueId);
        } else if (Patch.isOncHistoryDeletePatch(patch)) {
            tissueIds = TissueDao.getTissuesByOncHistoryDetailId(Integer.parseInt(patch.getId()));
        }
        return isSafeToDelete(tissueIds);
    }
```

zyme: Similarly:

```
public boolean isSafeToDelete(List<Integer> tissueIds) {
        if (tissueIds == null || tissueIds.isEmpty()) {
            return true;
        }
        return new ClinicalOrderDao().getClinicalOrdersForTissueIds(tissueIds).values().stream().filter(list -> !list.isEmpty()).findAny()
                .isEmpty();
    }
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -184,25 +185,26 @@ private static String toErrorMessage(String table, String column, Object value,
...
-    static boolean isTissuePatch(Patch originalPatch) {
+    static boolean isTissueDeletePatch(Patch originalPatch) {
         return originalPatch.getNameValue().getName().equals(TISSUE_DELETED_FIELD) &&
                 originalPatch.getTableAlias().equals(DBConstants.DDP_TISSUE_ALIAS);
     }
 
     //creates a list of patches for deleting tissues that belong to an onc history
     protected static List<Patch> getPatchForTissues(Patch oncHistoryPatch) {
-        List<String> tissueIds = TissueDao.getTissuesByOncHistoryDetailId(oncHistoryPatch.getId());
+        List<Integer> tissueIds = TissueDao.getTissuesByOncHistoryDetailId(Integer.parseInt(oncHistoryPatch.getId()));
```
zyme: Consider adding a `getIdAsInt()` method to `Patch`, and then calling that when you need the int instead of the string.  This will facilitate code re-use and relieve callers of the burden of casting/parsing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -226,6 +228,12 @@ public static boolean hasDDPParticipantId(Patch patch) {
         return StringUtils.isNotBlank(patch.getDdpParticipantId());
     }
 
+    public static boolean isSmIdDeletePatch(Patch originalPatch) {
```
zyme: Consider these non-static versions:

```
public boolean isSmIdDeletePatch() {
        return (nameValue.getName().contains(".deleted")) &&
                DBConstants.SM_ID_TABLE_ALIAS.equals(tableAlias) &&
                TissuePatch.TISSUE_ID.equals(parent);
    }

    boolean isOncHistoryDeletePatch() {
        return nameValue.getName().equals(ONC_HISTORY_DELETED_FIELD) && 
                tableAlias.equals(DBConstants.DDP_ONC_HISTORY_DETAIL_ALIAS);
    }

    boolean isTissueDeletePatch() {
        return nameValue.getName().equals(TISSUE_DELETED_FIELD) &&
                tableAlias.equals(DBConstants.DDP_TISSUE_ALIAS);
    }
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SmId.java
pegahtah: Correct! Method is auto-generated by intellij and I overlooked the types, will fix

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
pegahtah: Yes the inserting the json of the we want to delete and deleting it from its original table should all be done in one transaction, I will leave a comment explaining it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
pegahtah: Actually that is a great concern of mine too, I made a ticket to be able to export only one participant data by a  `ddp_participant_id`, so maybe we can think of prioritizing that ticket? 

One other solution may be to roll back a db delete if ES delete does not work. I don't think it would be bad to consider that. What do you think about that?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
pegahtah: Just as you also pointed out below, less and less statics is my goal for today. I will remove static from this and `isSafeToDelete` implementations

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
pegahtah: Is there any concerns to make a ClinicalOrder globally for the class and call it here? I want to make sure we avoid any unnecessary memory consumtions

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1204_dsm_delete.xml
pegahtah: If I'm not mistaken `text` can. hold up to 2GB of text data. I do hope that's enough but that's a great point, I will check the biggest objects we have on prod and change this if needed

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DBTestUtil.java
pegahtah: all in `DdpInstanceGroupTestUtil`

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DSMOncHistoryCreatorUtil.java
pegahtah: Fixed

pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1204_dsm_delete.xml
zyme: Seems like plenty of space!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
zyme: Happy to up-prioritize the export of a single participant!

For rollback, we want database transactions to be as short as possible to avoid db contention, and avoid external network calls as much as possible while the transaction is open.  Otherwise, when those external calls slow down, database contention can become a big issue quickly.  So let's leave this as-is and work on the single participant export in a separate ticket as you suggest.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
zyme: I'm not sure what you mean by "globally for the class".  Do you mean an instance variable?  If so, that should be fine IMO.  The memory use for a field is vanishingly small compared to all the other operations.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
pegahtah: No this returns nothing, just a `@sql` variable is created in the DB

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
pegahtah: No this returns nothing, just a @sql variable is created in the DB, trying to check what it returned results in a SQLException that says this does not return a ResultSet

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
pegahtah: I see!! I am not really scared of creating the query and storing it in `@sql` but was more afraid of the other, more refactored approach (that I did not use) of having only one query and adding the table names and primary keys with string concatenations to it based on what we are deleting from. Unfortunately the `stmt.setString()` can not be used to add the table name and column name, just the values so it's either creating separating queries, doing some replacing in the query or just concatenations and I think this was the best choice.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/DeletedObjectDao.java
pegahtah: That's a great approach! I did not see it that way

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
pegahtah: I am just watching out for ES or RB failure exceptions. I will add a log to say it is unexpected if it happens

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/exception/UnsafeDeleteError.java
pegahtah: To your point, I changed it so that nothing pops up in the error channel for us and this is only logged as a warn, but frontend is programmed to look at the message and not the type

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/ClinicalOrderDao.java
```diff
@@ -29,15 +29,15 @@ public class ClinicalOrderDao implements Dao<ClinicalOrderDto> {
...
+            "SELECT t.collaborator_sample_id, m.order_id, m.order_status, m.order_date, m.status_date, m.status_detail,"
+            + "st.sm_id_type, t.tissue_id, st.sm_id_type as sample_type, m.mercury_sequencing_id "
+            + "FROM ddp_mercury_sequencing m, ddp_tissue t , sm_id s, sm_id_type st "
+            + "where t.tissue_id = m.tissue_id "
+            + "and s.tissue_id = t.tissue_id "
+            + "and s.sm_id_type_id = st.sm_id_type_id "
             + "and t.tissue_id in (:list:)"; // will be replaced with actual params
 
-    public static String SQL_GET_ALL_ORDERS_FOR_CLINICAL_KIT_PAGE = "select ms.order_id, ms.ddp_participant_id, "
+    public static String SQL_GET_ALL_ORDERS_FOR_CLINICAL_KIT_PAGE = "select ms.mercury_sequencing_id, ms.order_id, ms.ddp_participant_id, "
```
pegahtah, line 40: The lack of `mercury_sequencing_id` was breaking the PW tests so I just added them here

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/patch/ObjectDeleteTest.java
```diff
@@ -0,0 +1,183 @@
...
+            oncHistoryTestUtil.assertOncHistoryIsDeleted(guid3, oncHistoryDetail, oncHistoryDetailId, ddpInstanceDto, false, false);
+
+        } catch (Exception e) {
+            throw new RuntimeException("Unexpected exception ", e);
+        }
+
+    }
+
+    @Test
+    public void deleteJustTheSmId() {
```
pegahtah, line 140: This test creates 2 sm ids for one single onc history and only deletes one of them, then make sure nothing else was deleted

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/DeletePatch.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public boolean isSafeToDelete(List<Integer> tissueIds) {
+        if (tissueIds == null || tissueIds.isEmpty())
+            return true;
+        return clinicalOrderDao.getClinicalOrdersForTissueIds(tissueIds).values().stream().filter(list -> !list.isEmpty()).findAny()
+                .isEmpty();
+    }
+
+    public boolean isSafeToDelete(Patch patch) {
```
zyme: Now that this isn't static, can you remove the `patch` parameter and query what you need from the state if `this`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -161,34 +180,66 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
+        }
+        return deletePatches;
+    }
+
+    public boolean hasDDPParticipantId() {
+        return StringUtils.isNotBlank(this.ddpParticipantId);
+    }
+
+    public boolean isSmIdDeletePatch() {
+        return (this.getNameValue().getName().contains(".deleted")) &&
```
zyme: Since this is now an instance method, you can skip the getter and just use `nameValue` instead of `getNameValue()`.

zyme (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
```diff
@@ -266,9 +266,6 @@ private DBConstants(){
     public static final String SCROLLS_COUNT = "scrolls_count";
     public static final String H_E_COUNT = "h_e_count";
     public static final String BLOCKS_COUNT = "blocks_count";
-    public static final String USS_SMIDS = "uss_sm_ids";
-    public static final String SCROLL_SMIDS = "scrolls_sm_ids";
-    public static final String HE_SMIDS = "he_sm_ids";
```
pegahtah, line 271: unused

PR #2728, opened 2023-11-03
PEPPER-1209 Service to check whether tissues are in an order

zyme:
PEPPER-1209 is a part of our larger tissue/sample delete epic, focused just on a dao to check whether a list of tissue ids are referenced in a mercury order.  The main changes are in [ClinicalOrderDao.getClinicalOrdersForTissueIds()](https://github.com/broadinstitute/ddp-study-server/pull/2728/files#r1382141538), but there was a lot of related DAO code that needed some work in order to make useful tests.

I also changed `TestKitUtil` to improve state management.  CircleCI tests were failing in interesting ways because of how `TestKitUtil` was storing almost everything as static vars.  `TestKitUtil` also was not deleting tables in the right order, and was silently failing in ways that were difficult to troubleshoot.  By passing up the exceptions, I was able to identify a number of bugs in the cleanup code and get the deletes to happen appropriately.

Along the way, I noticed many places where we are passing around `String` values for primary keys, which is just wrong--our primary keys are numbers, and the java types should be numeric.  I changed the types in many places, but didn't go totally bonkers fixing everything.

I also added some utility classes so that others can have an easier go of creating and deleting Tissues, Samples, and Mercury Orders.

The DAO will be used in PEPPER-1204.

## Testing

- [ x] I have written automated negative tests

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/Tissue.java
```diff
@@ -253,18 +253,19 @@ public static String createNewTissue(@NonNull String oncHistoryId, @NonNull Stri
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
             try (PreparedStatement stmt = conn.prepareStatement(SQL_INSERT_TISSUE, Statement.RETURN_GENERATED_KEYS)) {
-                stmt.setString(1, oncHistoryId);
```
zyme, line 256: We seem to have an antipattern of implicit conversions.  As I see these, I am trying to fix them case-by-case.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/db/dao/mercury/ClinicalOrderDaoTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+    private static final DDPInstitutionDao institutionDao = new DDPInstitutionDao();
+
+    private static final Tissue tissueDao = new Tissue();
+
+    private static int medicalRecordId = -1;
+
+    private static final MercuryOrderDao mercuryOrderDao = new MercuryOrderDao();
+
+    private static final ClinicalOrderDao clinicalOrderDao = new ClinicalOrderDao();
```
zyme, line 51: I'm not proud of the messiness of the test setup/teardown.  I would like to clean this up so that others can make use of utility classes for samples/tissues/onchistory/mr.  I would like to break that work out to a separate PR so that we can get this DAO onto the mainline for integration testing.

edit: I've pulled out a few *TestUtil classes to make this better.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -82,14 +78,14 @@ public class TestKitUtil {
             "INSERT INTO  sub_kits_settings (ddp_kit_request_settings_id, kit_type_id, kit_count, hide_on_sample_pages, external_name) "
                     + " VALUES (?, ?, 1, ?, '') ;";
     private static final String SELECT_DSM_KIT_REQUEST_ID = "SELECT dsm_kit_request_id from ddp_kit_request where ddp_kit_request_id like ? ";
-    private static final UserAdminTestUtil cmiAdminUtil = new UserAdminTestUtil();
+    public static final UserAdminTestUtil adminUtil = new UserAdminTestUtil();
```
zyme: This isn't a cmi thing--I renamed this so others don't end up confused.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/mercury/ComparisonTest.java
```diff
@@ -24,9 +24,9 @@ public void testLongComparison() {
 
         ArrayList<ClinicalOrderDto> clinicalOrderDtoArrayList = new ArrayList<>();
         clinicalOrderDtoArrayList.add(new ClinicalOrderDto("FAKE_SHORT_ID", "FAKE_SAMPLE", "FAKE_ORDER", "FAKE_STATUS",
-                1L, 1L, null, "FAKE_TYPE"));
+                1L, 1L, null, "FAKE_TYPE", 1));
```
zyme, line 27: Since we added the mercury_sequencing_id to the ClinicalOrderDto, we needed to add values to these c'tors.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -190,12 +194,13 @@ public Map<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String ddp
 
     @Override
     public int create(MercuryOrderDto mercuryOrderDto) {
-        throw new IllegalStateException("This method should not be used");
+        throw new IllegalStateException("Call the version of this method that includes the raw json payload");
     }
 
     @Override
     public int delete(int id) {
-        return 0;
```
zyme, line 198: It made me sad to see this--having a method do absolutely nothing, with no exceptions or anything...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -190,12 +194,13 @@ public Map<String, MercuryOrderDto> getPossibleBarcodesForParticipant(String ddp
 
     @Override
     public int create(MercuryOrderDto mercuryOrderDto) {
-        throw new IllegalStateException("This method should not be used");
+        throw new IllegalStateException("Call the version of this method that includes the raw json payload");
```
zyme, line 197: I do wonder why we have these interfaces--a number of implementations throw exceptions or do nothing (see the delete() method below).  I'm trying to make this a little less cryptic without refactoring the universe.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -82,36 +78,36 @@ public class TestKitUtil {
             "INSERT INTO  sub_kits_settings (ddp_kit_request_settings_id, kit_type_id, kit_count, hide_on_sample_pages, external_name) "
                     + " VALUES (?, ?, 1, ?, '') ;";
     private static final String SELECT_DSM_KIT_REQUEST_ID = "SELECT dsm_kit_request_id from ddp_kit_request where ddp_kit_request_id like ? ";
-    private static final UserAdminTestUtil cmiAdminUtil = new UserAdminTestUtil();
```
zyme, line 85: Keeping so many things in static state while also having a constructor that sets some of the static state makes it hard to keep the state correct, which resulted in some cryptic errors in circleCI with `Realm and study group already initialized`, which came from `UserAdminTestUtil.createRealmAndStudyGroup()` because it was a singleton here.  Running a few tests locally never uncovered this issue, but because circleCI runs a pile of tests in a single VM, the issue was exposed only in circleCI.  The changes here move most things into instance state, so if one test fails to clean up after itself, it should only impact that one test.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/ClinicalOrderDto.java
```diff
@@ -1,20 +1,55 @@
...
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
 
 @AllArgsConstructor
 @Getter
 public class ClinicalOrderDto {
 
-
+    @SerializedName("shortId")
```
zyme, line 16: Down the road, we do plan on making a query for orders accessible to the frontend, so these annotations are here to help us do that in a subsequent sprint.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/GetClinicalOrdersForTissuesRoute.java
```diff
@@ -0,0 +1,35 @@
...
+import spark.Response;
+
+import java.util.List;
+
+/**
+ * Consumes a list of tissue id ints from a post and returns
+ * a map where the keys are the tissue ids and the values
+ * are the clinical orders for the tissue.
+ */
+public class GetClinicalOrdersForTissuesRoute extends RequestHandler {
```
zyme, line 18: Down the road, we do plan on making a query for orders accessible to the frontend.  This route is not yet used, so it does not have any tests, but I've kept it here so we can use it in a subsequent sprint.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/ClinicalOrderDao.java
```diff
@@ -158,4 +171,60 @@ private String createShortIdFromCollaboratorSampleId(String collabSampleId) {
         }
         return collabSampleId.substring(index1 + 1, index2);
     }
+
+    /**
+     * Looks up all clinical orders for a list of tissue ids
+     * @param tissueIds a list of tissue primary key ids
+     * @return a map where the tissue id is the key and the value is a list of orders for that tissue
+     */
+    public Map<Integer, Collection<ClinicalOrderDto>> getClinicalOrdersForTissueIds(List<Integer> tissueIds) {
```
zyme, line 180: This is where the actual functionality is in this PR.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -110,6 +116,22 @@ public String createNewSMIDForTissueWithValue(String tissueId, String userId, St
         }
     }
 
+    /**
+     * Deletes the tissue and smid
+     */
+    public void deleteTissueAndSmId(int tissueId) {
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            DaoUtil.deleteSingleRowById(tissueId, DELETE_SM_ID);
```
pegahtah: I could be mistaken about the usage here, but because of 1:n relationship between tissue and sm_id, this can delete more than one row which will throw an exception in the `DaoUtil.deleteSingleRowById`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: Thanks for cleaning up the error! We talked about abstract Dao implementations but  I would also vote for stopping the use of Dao interfaces in DSM. I don't think they are adding any values.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -187,4 +187,8 @@ public static Boolean isValueUnique(@NonNull DBElement dbElement) {
         });
         return (Boolean) results.resultValue;
     }
+
+    public int getParentIdAsInt() {
```
zyme, line 191: Instead of going full bore and changing various fields from String -> Int/Long, I added this utility method to make some incremental progress.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPDummyKitDao.java
```diff
@@ -115,7 +115,7 @@ public Optional<String> getRandomParticipantIdForStudy(String ddpInstanceId) {
         return Optional.ofNullable((String) results.resultValue);
     }
 
-    public String getRandomOncHistoryForStudy(String ddpInstanceName) {
```
zyme, line 118: Renamed for clarity while fixing return type

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -82,36 +78,36 @@ public class TestKitUtil {
...
-    private static String groupName;
-    private static String kitTypeName;
-    private static String studyGuid;
-    private static String collaboratorPrefix;
-    private static String userWithKitShippingAccess;
+    public final UserAdminTestUtil adminUtil = new UserAdminTestUtil();
+    public Integer ddpGroupId;
+    public Integer ddpInstanceId;
+    public Integer ddpInstanceGroupId;
+    public Integer instanceRoleId;
```
zyme, line 86: We really need to stop declaring primary key values as strings...

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SmId.java
```diff
@@ -85,21 +85,4 @@ public static boolean isUniqueSmId(String smIdValue) {
         return new TissueSMIDDao().isUnique(smIdValue);
     }
 
-    public String createNewSmId(@NonNull String tissueId, String userId, @NonNull List<NameValue> smIdDetails) {
```
zyme, line 88: Nothing was using this method, so I removed it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
zyme: Good catch!  I've changed this to delete the smid by its pk instead of tissue id.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitDisplayNameTest.java
```diff
@@ -15,9 +15,9 @@
 import org.junit.Test;
 
 public class KitDisplayNameTest extends DbTxnBaseTest {
-    private static TestKitUtil
-            testKitUtil = new TestKitUtil("kit_test_instance", "kit_test_instance_guid", "some_prefix", "kit_test_group",
-            KitRequestShippingTest.BLOOD_RNA_KIT_TYPE_NAME);
+    private static TestKitUtil testKitUtil = new TestKitUtil("kit_test_instance", "kit_test_instance_guid",
+            "some_prefix", "kit_test_group", KitRequestShippingTest.BLOOD_RNA_KIT_TYPE_NAME,
+            KitRequestShippingTest.BLOOD_RNA_KIT_TYPE_DISPLAY_NAME);
```
zyme, line 20: @pegahtah I changed this to include the display name because the test was currently passing in `null` for the value and the test was failing.  Can you confirm that this change is appropriate?  I can't tell how the test was passing before since the display name was null and `BLOOD_RNA_KIT_TYPE_DISPLAY_NAME` is different from `BLOOD_RNA_KIT_TYPE_NAME`.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/SampleIdTestUtil.java
```diff
@@ -0,0 +1,25 @@
...
+
+    private final Set<Integer> createdSampleIds = new HashSet<>();
+
+    public int createSampleForTissue(int tissueId, String user, String sampleType, String sampleName) {
+        int sampleId = smIdDao.createNewSMIDForTissue(tissueId, user, sampleType, sampleName);
+        createdSampleIds.add(sampleId);
+        return sampleId;
+    }
+
+    public void deleteCreatedSamples() {
```
zyme, line 20: There's probably some generalized pattern lurking here--our test utils create rows and then have various methods to delete the things that were created, usually called within a teardown method.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -82,71 +79,72 @@ public class TestKitUtil {
...
-                delete(conn, "ddp_group", "group_id", ddpGroupId);
-                cmiAdminUtil.deleteGeneratedData();
+                adminUtil.deleteGeneratedData();
             } catch (Exception e) {
                 dbVals.resultException = e;
             }
             return dbVals;
         });
+        if (results.resultException != null) {
+            throw new DsmInternalError("Error deleting instance and settings", results.resultException);
```
zyme, line 139: These errors were being ignored, leading to downstream failures.  Once the exceptions became visible, it was clear that the rows needed to be deleted in a different order.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -82,71 +79,72 @@ public class TestKitUtil {
...
         this.instanceName = instanceName;
         this.studyGuid = studyGuid;
         this.collaboratorPrefix = collaboratorPrefix;
         this.groupName = groupName;
         this.kitTypeName = kitTypeName;
+        this.kitTypeDisplayName = kitTypeDisplayName;
     }
 
-    public static void deleteInstanceAndSettings() {
+    public void deleteGeneratedData() {
```
zyme, line 123: Renamed because this method is doing more deletes than just instance and settings.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -170,86 +168,83 @@ private static void delete(Connection conn, String tableName, String primaryColu
...
-
     }
 
-    public static void deleteKit(String ddpKitRequestId) {
+    public void deleteKit(String ddpKitRequestId) {
         SimpleResult results = inTransaction((conn) -> {
-            List<String> dsmKitRequestId = new ArrayList<>();
+            List<Integer> dsmKitRequestId = new ArrayList<>();
             try (PreparedStatement stmt = conn.prepareStatement(SELECT_DSM_KIT_REQUEST_ID)) {
-                stmt.setString(1, ddpKitRequestId.concat("\\_%"));
```
zyme, line 184: @pegahtah when I set a breakpoint here, it became clear that this pattern match resulted in zero rows being deleted, instead of the given `ddpRequestId`, which then lead to downstream errors during cleanup methods because ddp_instance rows could not be deleted because of the fk from ddp_kit_request.  Can you confirm that this pattern match is not needed?  Once I changed the delete to use the request id exactly, tests started passing.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/juniperkits/TestKitUtil.java
```diff
@@ -258,27 +253,27 @@ public void setupInstanceAndSettings() {
...
+        if (ddpKitRequestSettingsId != null) {
             return ddpKitRequestSettingsId;
         }
         PreparedStatement stmt = conn.prepareStatement(INSERT_DDP_KIT_REQUEST_SETTINGS, Statement.RETURN_GENERATED_KEYS);
-        stmt.setString(1, ddpInstanceId);
-        stmt.setString(2, kitTypeId);
-        stmt.setString(3, kitReturnId);
-        stmt.setString(4, carrierId);
-        stmt.setString(5, kitDimensionId);
+        stmt.setInt(1, ddpInstanceId);
```
zyme, line 265: Ids are numeric...

ssettipalli (commented):
overall looks good to me.
noticed that tissueId is "int" in some places and  Integer in some which is probably ok depending on where the value is coming from and assured that value exists.

ssettipalli (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
```diff
@@ -110,13 +115,21 @@ public String createNewSMIDForTissueWithValue(String tissueId, String userId, St
...
     }
 
-    public String createNewSMIDForTissue(String tissueId, String userId, @NonNull String smIdType, @NonNull String smIdValue) {
-        String smIdtypeId = getTypeForName(smIdType);
+    public void deleteTissueById(int tissueId) {
+        DaoUtil.deleteSingleRowById(tissueId, DELETE_TISSUE);
+    }
+
+    public void deleteSampleById(int sampleId) {
+        DaoUtil.deleteSingleRowById(sampleId, DELETE_SM_ID);
```
pegahtah, line 123: Do we want to use sm id instead of sample all over the code so grepping would show all the usage maybe? sample to me could be a saliva/blood sample too

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/SampleIdTestUtil.java
pegahtah: I can try to change this use the util I created after it's merged

pegahtah (approved):
LGTM! Thank you!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/tissue/TissueSMIDDao.java
zyme: Maybe?  I think the confusing thing to me is that SM ids in BSP are ids for samples, and "SM" doesn't really capture that.  Plus the "SMID id" stuff like "smid_id" makes my head hurt sometimes.  I find myself looking at `smId` wondering "is this an actual numeric id or a sample name"?  Which is made more complicated by the fact that sometimes we pass around numeric sample id pks as strings.

PR #2736, opened 2023-11-16
Pepper-1190 kit rule update

ssettipalli:
update kit creation rules for pe-cgs studies lms and osteo2
In addition to patch, updated original study config file to keep the rule in sync


aweng98 (approved):
LGTM

PR #2737, opened 2023-11-20
[DSM] PEPPER-1059

pegahtah:
## Context

The aim for this ticket is to fix export so that when a value is deleted directly from DB running export will also remove it from ES. 

The bug was that the export would only take what was in the DB and upsert it in ES, so if a participant had all of their onc histories removed, no data was found for them in the DB and export was not aware about replacing their onc history data in ES with an empty list. 

The fix is to first get all of the participant ids in the ES and if they don't have anything in the DB, change export to have an empty list for those ids instead.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/OncHistoryTestUtil.java
```diff
@@ -445,4 +451,24 @@ public void assertSmIdIsNOTDeleted(String participantGuid,  int tissueId, int sm
         }
 
     }
+
+    public void deleteOncHistoryDirectlyFromDB(int oncHistoryDetailId){
```
zyme: Maybe move this to a dao?  Seems like it could be a useful thing outside of testing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/DSMtasksSubscription.java
```diff
@@ -120,7 +120,7 @@ public static void doESExport(Map<String, String> attributesMap, String data) {
         }
     }
 
-    public static void migrateToES(ExportToES.ExportPayload exportPayload) {
+    public static boolean migrateToES(ExportToES.ExportPayload exportPayload) {
```
zyme: When would this be expected to return false?

pegahtah: No I changed it so the functionality can be used synchronously in test, and tests wait for a return value from this method to make sure export is completed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -32,11 +34,18 @@ public BaseMigrator(String index, String realm, String object) {
         this.object = object;
         elasticSearch = new ElasticSearch();
         objectTransformer = new ObjectTransformer(realm);
+        if (index != null) {
+            participantsInTheStudy = elasticSearch.getAllParticipantsInIndex(index);
+        }
     }
 
     protected void fillBulkRequestWithTransformedMapAndExport(Map<String, Object> participantRecords) {
+        if (participantRecords == null || participantRecords.isEmpty()) {
```
zyme: I think we could get the same amount of information but with fewer lines of code by putting a `@NonNull` on the `participantRecords`, and then removing L43-46 and adding the size of `participantRecords` to the log entry on L48.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/elasticExport/DeletedDataExportTest.java
```diff
@@ -0,0 +1,98 @@
...
+    static OncHistoryTestUtil oncHistoryTestUtil;
+    static String userEmail = "deleteTestUser1@unittest.dev";
+    String guid = "DELETE_PARTICIPANT";
+    String guid1 = "DELETE1_PARTICIPANT";
+    String guid2 = "DELETE2_PARTICIPANT";
+    String guid3 = "DELETE3_PARTICIPANT";
+
+    @BeforeClass
+    public static void doFirst() {
+        esIndex = ElasticTestUtil.createIndex(instanceName, "elastic/lmsMappings.json", null);
```
zyme: I am wondering if we are relying too much on `lmsMappings`...I see it in a lot of our tests.  It could be fine, but if there's something about LMS that isn't representative of other studies, we might be missing something.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/OncHistoryTestUtil.java
```diff
@@ -445,4 +448,8 @@ public void assertSmIdIsNOTDeleted(String participantGuid,  int tissueId, int sm
         }
 
     }
+
+    public void deleteOncHistoryDirectlyFromDB(int oncHistoryDetailId) {
```
zyme: I don't think we need this method, since it's just adding indirection on top of the dao delete call, and the method name looks like it's trying to say "Really, just delete it from the database", which is what daos are all about.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/elastic/migration/BaseCollectionMigratorTest.java
```diff
@@ -15,7 +15,7 @@ public class BaseCollectionMigratorTest {
 
     @Test
     public void transformObject() {
-        BaseCollectionMigrator baseCollectionMigrator = new MockBaseCollectionMigrator("index", "realm", "object");
+        BaseCollectionMigrator baseCollectionMigrator = new MockBaseCollectionMigrator(null, "realm", "object");
```
zyme: Consider creating a two-arg constructor for `MockBaseCollectionMigrator` and adding some comments there about when to use one vs. the other.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
```diff
@@ -156,6 +160,53 @@ String getDdpFromIndex(String searchHitIndex) {
...
+        return participantIds;
+    }
+
+    private void clearScroll(String scrollId, String index) {
+        try {
+            ClearScrollRequest clearScrollRequest = new ClearScrollRequest();
+            clearScrollRequest.addScrollId(scrollId);
+            ElasticSearchUtil.getClientInstance().clearScroll(clearScrollRequest, RequestOptions.DEFAULT);
+        } catch (IOException e){
+            logger.error("Unable to clear the scroll in the connection to index: "+index, e);
```
zyme: Why should this not throw the exception?  If the code continuously fails to clear scrolls, what happens?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
```diff
@@ -156,6 +160,53 @@ String getDdpFromIndex(String searchHitIndex) {
         return searchHitIndex.substring(dotIndex + 1);
     }
 
+    public List<String> getAllParticipantsInIndex(String esParticipantsIndex) {
```
zyme, line 161: Could this instead by `getParticipantsWithinRange()` from 0 (or 1) to whatever the max is (`getParticipantsSize()`?  Or `getAllParticipantsDataByInstanceIndex()`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/NullAdditionalParticipantMigrator.java
```diff
@@ -13,4 +13,13 @@ public NullAdditionalParticipantMigrator() {
     protected Map<String, Object> getDataByRealm() {
         return new HashMap<>();
     }
+
```
zyme, line 23: Class comments, please: what does this class do?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -32,11 +34,18 @@ public BaseMigrator(String index, String realm, String object) {
         this.object = object;
         elasticSearch = new ElasticSearch();
         objectTransformer = new ObjectTransformer(realm);
+        if (index != null) {
```
zyme: Should we be creating what looks like a potentially very big object in the constructor, when it's only ever used when `export()` is called?  This will have an impact on memory use and performance of all derived classes since it looks like they tend to call `super()` in their constructors.  I think `participantsInTheStudy` should move from an instance variable to a variable whose scope is limited to the `export()` method.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/elasticExport/DeletedDataExportTest.java
```diff
@@ -0,0 +1,98 @@
...
+        ExportToES.ExportPayload exportPayload = new ExportToES.ExportPayload();
+        exportPayload.setStudy(instanceName);
+        exportPayload.setIsMigration(true);
+        DSMtasksSubscription.migrateToES(exportPayload);
+        Map<String, Object> esDsmMap = ElasticSearchUtil.getObjectsMap(ddpInstanceDto.getEsParticipantIndex(), guid, ESObjectConstants.DSM);
+        List<Map<String, Object>> oncHistoryDetails =
+                (List<Map<String, Object>>) ((Map<String, Object>) esDsmMap.get(ESObjectConstants.DSM)).get(ESObjectConstants.ONC_HISTORY_DETAIL);
+        long countNumberOfOncHistoriesInEs = oncHistoryDetails.stream().filter(stringObjectMap ->
+                (int) stringObjectMap.get("oncHistoryDetailId") == oncHistoryDetailId).count();
+        Assert.assertEquals(0L, countNumberOfOncHistoriesInEs);// Currently the test is failing here because the value is still in ES
```
zyme: Nice!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/elasticExport/DeletedDataExportTest.java
pegahtah: I understand your concern but LMS and OS2 are the best options for these features since their biggest usecase would be in the PECGS studies. I can get a mapping from OS index and try a test with that as well

zyme: Ah that makes perfect sense--we have just the right index for the testing, in that case :)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
pegahtah: I understand the concern, but checking further into the code shows that this constructor is only used for calling the `export`. The accessors of this constructor are only classes that are inheriting from this migrator to migrate different objects to ES (Onc Histories, Medical Records, etc) and they are called only when DSM preparing to do a full export . We can add the code to create the list here so all of the classes that are inheriting from this class will create the list when initialized, or we can add it separately in each of their export methods. But we would still need it for each of those classes to have a way of creating and accessing this list before the export begins. The memory usage is the same in both of the cases.

zyme: I think the hazard to mitigate here is future use, where all of a sudden we might have more instances floating around for whatever reason, without them calling `export()`.  Let's future-proof this by lazily creating participantsInTheStudy in `BaseMigrator.export`.  At least that way, the cost of creating the object will only be incurred during export.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
pegahtah: From what I understand, if the scroll is not cleared in Elasticsearch, it can lead to unnecessary resource usage on the Elasticsearch cluster. Each scroll request maintains a search context in the cluster, which consumes memory and other resources. If these contexts are not cleared, they will eventually time out based on the scroll timeout setting, but until they do, they can consume resources and potentially affect the performance of the Elasticsearch cluster, especially if there are many unclosed scrolls or if the cluster is under heavy load.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearch.java
pegahtah: Oh I missed this!! I don't know what is the usage of `getAllParticipantsDataByInstanceIndex()` , it's only used in the tests. Let me see if we can replace this.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -90,7 +94,16 @@ private Map<String, Object> replaceLegacyAltPidKeysWithGuids(Map<String, Object>
     @Override
     public void export() {
         Map<String, Object> dataByRealm = getDataByRealm();
+        if (index != null) {
+            participantsInTheStudy = elasticSearch.getAllParticipantsInIndex(index);
```
zyme, line 98: I think `participantsInTheStudy` should be have method scope--otherwise it'll stick around and take up space when it's not needed after the export.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -90,7 +93,17 @@ private Map<String, Object> replaceLegacyAltPidKeysWithGuids(Map<String, Object>
     @Override
     public void export() {
         Map<String, Object> dataByRealm = getDataByRealm();
+        List<String> participantsInTheStudy = null;
+        if (index != null) {
+            participantsInTheStudy = elasticSearch.getAllParticipantsInIndex(index);
+        }
+        for (String ddpParticipantId : participantsInTheStudy) {
```
zyme: This will NPE if `participantsInTheStudy` is null.  Since there's a check for index being non-null just above this, it seems like this for loop (from L100 - L104) should be moved to within the if block in L97?

pegahtah: Nice catch!! Fixed, thank you

zyme (approved):

PR #2739, opened 2023-11-20
PEPPER-1171 Removed KitDao interface, replacing it with KitDaoImpl

zyme:
Since we decided it's fair game to remove our dao interfaces and replace them with their single implementation, I have done so in the context of PEPPER-1171, changing KitDao from an interface to an implementation.

This looks like a big scary PR but it was just a simple set of refactors in intellij:

1. Remove `KitDao.java`, with no regard for dependencies, "safe delete" deselected.
2. Use the rename refactor to rename `KitDaoImpl` -> `KitDao`
3. Clean up the broken `@Override` annotations and a few other glitches


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDaoImpl.java
```diff
@@ -1,842 +0,0 @@
-package org.broadinstitute.dsm.db.dao.kit;
```
zyme, line 1: This was moved to `KitDao` after the original `KitDao` was deleted.

pegahtah (approved):
THANK YOU!!

PR #2741, opened 2023-11-27
PEPPER-1205 added script to hard delete previously soft-deleted onc histories, etc.

zyme:
PEPPER-1205 is about cleaning up onc histories, tissues, and sm ids marked deleted.  The cleanup involves identifying all soft-deleted rows and hard deleting them after writing them to the `deleted_object` table.

Overall this looked straightforward, since just finding rows where `deleted = 1` seemed like it should be sufficient.  But then I found some relational integrity issues because some `sm_id` rows were *not* marked deleted but they referenced `ddp_tissue` rows marked for deletion.  Similar story for `ddp_tissue` rows not marked deleted but reference `ddp_onc_history_detail` rows that were marked for deletion.  The script I wrote breaks out the straightforward cases and the "inferred as should be deleted" cases.

The delete script does not pull in `sm_id.accession_date` because although this column exists in dev, it does not exist in test, staging, or prod.

I ran a bunch of read-only queries in test and prod to try to anticipate any data wrinkles that might not be present in dev, so I'm feeling pretty good about there being no surprises when this is applied to prod.

The overall volume of deletes is pretty small here, so transactions should be pretty quick on prod.  I don't anticipate any lock issues, so I think we can just run this script on prod at any point without any impact on production operations.


pepper-apis/dsm-core/src/main/resources/liquibase/DDP-1205_CleanupDeletedOncHistoryEtc.sql
```diff
@@ -0,0 +1,116 @@
...
+-- deleted, but a bug prevented proper deletion
+insert into deleted_object (original_table, original_primary_key, data)
+    (select 'ddp_tissue',
+            tissue_id,
+            json_object('tissue_id',tissue_id,'onc_history_detail_id',onc_history_detail_id,'notes',notes,'count_received',count_received,'h_e_count',h_e_count,'scrolls_count',scrolls_count,'blocks_count',blocks_count,'uss_count',uss_count,'tissue_type',tissue_type,'tissue_site',tissue_site,'tumor_type',tumor_type,'h_e',h_e,'pathology_report',pathology_report,'collaborator_sample_id',collaborator_sample_id,'block_sent',block_sent,'block_id_shl',block_id_shl,'shl_work_number',shl_work_number,'scrolls_received',scrolls_received,'sk_id',sk_id,'sm_id',sm_id,'first_sm_id',first_sm_id,'sent_gp',sent_gp,'tissue_sequence',tissue_sequence,'tumor_percentage',tumor_percentage,'additional_tissue_value',additional_tissue_value,'additional_tissue_value_json',additional_tissue_value_json,'last_changed',last_changed,'changed_by',changed_by,'deleted',deleted,'expected_return',expected_return,'return_fedex_id',return_fedex_id,'return_date',return_date)
```
zyme: I generated these params to `json_object()` by running one of the commented out queries at the bottom of the file.  So many columns...

pepper-apis/dsm-core/src/main/resources/liquibase/DDP-1205_CleanupDeletedOncHistoryEtc.sql
```diff
@@ -0,0 +1,116 @@
+-- this file will be run manually, not by liquibase, but it is here to track
+-- our approach to DDP-1205.
+
+-- insert deleted onc history details
+insert into deleted_object (original_table, original_primary_key, data)
```
ssettipalli: Maybe use systemdate to store into deleted_date if we think it helps in anyway.
Since all deletions are same date, maybe we can add to the ticket when we deleted, once we run this delete in prod

ssettipalli (approved):
looks good to me

pepper-apis/dsm-core/src/main/resources/liquibase/DDP-1205_CleanupDeletedOncHistoryEtc.sql
zyme: Good catch, and this makes me think we should make `data`, `deleted_by`, and `deleted_date` non-nullable.  @pegahtah what do you think of making these fields required?

pepper-apis/dsm-core/src/main/resources/liquibase/DDP-1205_CleanupDeletedOncHistoryEtc.sql
```diff
@@ -0,0 +1,128 @@
...
+    (select 'ddp_onc_history_detail',
+            onc_history_detail_id,
+            json_object('onc_history_detail_id',onc_history_detail_id,'medical_record_id',medical_record_id,'date_px',date_px,'type_px',type_px,'location_px',location_px,'histology',histology,'accession_number',accession_number,'facility',facility,'phone',phone,'fax',fax,'notes',notes,'additional_values',additional_values,'additional_values_json',additional_values_json,'request',request,'fax_sent',fax_sent,'fax_sent_by',fax_sent_by,'fax_confirmed',fax_confirmed,'fax_sent_2',fax_sent_2,'fax_sent_2_by',fax_sent_2_by,'fax_confirmed_2',fax_confirmed_2,'fax_sent_3',fax_sent_3,'fax_sent_3_by',fax_sent_3_by,'fax_confirmed_3',fax_confirmed_3,'tissue_received',tissue_received,'tissue_problem',tissue_problem,'tissue_problem_text',tissue_problem_text,'tissue_problem_option',tissue_problem_option,'destruction_policy',destruction_policy,'unable_obtain',unable_obtain,'gender',gender,'unable_obtain_tissue',unable_obtain_tissue,'last_changed',last_changed,'changed_by',changed_by,'deleted',deleted),
+            'andrew@broadinstitute.org',
+            unix_timestamp() * 1000
```
zyme, line 10: @ssettipalli I added unix time * 1000 (since DSM stores millis since epoch)

pegahtah (approved):
Thank you for making this Zim!

PR #2743, opened 2023-11-27
PEPPER-1171 Rename ScanError to ScanResult

zyme:
`ScanError` isn't an error--it's used to pass information from the backend to the frontend during normal operation, and is not used just for errors.


pegahtah (approved):
Thanks

PR #2756, opened 2023-12-05
PEPPER-1240 Concurrency Bug in DSM Kit Scan Pages

zyme:
In PEPPER-1240, we discovered an interesting case where the final scan page for kit "X" returns an error about kit "Y".

Our routes are generally singletons, and they are generally stateless with regard to http request/response values, but `KitStatusChangeRoute` and derived classes used instance variables `kitPayload` and `scanResultList`.  As a result, it was possible that concurrent calls into derived instances of `KitStatusChangeRoute` (which includes 6 kit-related routes) could result in payload and response being overwritten such that the http inputs into one instance would be mixed up with the http inputs to a different instance (and similar with the outputs).

The fix is straightforward: move the stateful instance variables that keep track of request/response information values KitStatusChangeRoute to method params and return values so that concurrent calls don't result in inputs/outputs being overwritten. [ Start your review here to see the primary change to the superclass](https://github.com/broadinstitute/ddp-study-server/pull/2756/files#r1415859897), and then look at changes to each derived class.

image

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
@@ -18,13 +18,16 @@ import spark.Response;
 public abstract class KitStatusChangeRoute extends RequestHandler {
 
     protected NotificationUtil notificationUtil;
-    protected KitPayload kitPayload;
-    protected List<ScanResult> scanResultList;
 
     public KitStatusChangeRoute(NotificationUtil notificationUtil) {
         this.notificationUtil = notificationUtil;
     }
 
+    /**
+     * Checks the auth, converts the payload, and then calls
+     * {@link #processRequest(KitPayload)} and returns the resulting
+     * list of scan results.
+     */
     @Override
     public Object processRequest(Request request, Response response, String userId) throws Exception {
         String requestBody = request.body();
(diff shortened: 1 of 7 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
```diff
@@ -18,13 +18,16 @@
 public abstract class KitStatusChangeRoute extends RequestHandler {
 
     protected NotificationUtil notificationUtil;
-    protected KitPayload kitPayload;
```
zyme, line 21: We should never track request/response values in instance variables in a route this way because our routes are singletons, and different threads can overwrite these values when multiple requests are running simultaneously.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/kit/KitStatusChangeRoute.java
```diff
@@ -34,12 +37,12 @@ public Object processRequest(Request request, Response response, String userId)
         DDPInstanceDto ddpInstanceDto = new DDPInstanceDao().getDDPInstanceByInstanceName(realm).orElseThrow();
         if (UserUtil.checkUserAccess(null, userId, "kit_shipping", userIdRequest) || UserUtil.checkUserAccess(null, userId, "kit_receiving",
                 userIdRequest)) {
-            scanResultList = new ArrayList<>();
+            List<ScanResult> scanResultList = new ArrayList<>();
```
zyme, line 40: Instead of relying on instance variables, we create the payload and results objects inside of the method, thereby eliminating shared state across threads.

pegahtah (approved):
OMG thank you!

ssettipalli (approved):

PR #2760, opened 2023-12-07
[DSM] PEPPER-526 Phi Manifest

pegahtah:
## Context

I have made the PhiManifestService to do all the job around creating the report.  This PR currently only has the eligibility check and the tests around that.
We check that the sequencing order number is valid, belongs to the participant and to the realm, and has a PDO number which means it was accepted by Mercury
We also need to check that the participant is consented to receive shared learnings and has consented to tumor collection. 
There are new utility methods and classes in test built around creating a Mercury order and a participant who has consented to the above criteria. 

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+import org.broadinstitute.dsm.model.elastic.search.ElasticSearchParticipantDto;
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.statics.RoutePath;
+import org.broadinstitute.dsm.util.ElasticSearchUtil;
+import spark.QueryParamsMap;
+
+@Slf4j
+public class PhiManifestService {
+
+    public Object generateReport(QueryParamsMap queryParams, String realm) {
```
zyme: For ease of testing, break out seq order and participant id into method args, and then let the route layer do the conversion from params map.  Then the test methods will be simpler and the class can be re-used outside of a route.

zyme: Think about what would be the most useful thing to return here to avoid casting, make it more testable, and help callers know exactly what they're getting back explicitly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+        String sequencingOrderNumber = queryParams.get(RoutePath.SEQ_ORDER_NUMBER).value();
+        String ddpParticipantId = queryParams.get(RoutePath.DDP_PARTICIPANT_ID).value();
+        DDPInstanceDto ddpInstanceDto = DDPInstanceDao.of().getDDPInstanceByInstanceName(realm).orElseThrow();
+        String maybeErrorMessage = null;
+        if (StringUtils.isBlank(ddpParticipantId)) {
+            maybeErrorMessage = "No DdpParticipantId was provided";
+        }
+        if (StringUtils.isBlank(sequencingOrderNumber)) {
+            maybeErrorMessage = "No sequencingOrderNumber was provided";
```
zyme: I think `maybeErrorMessage` is really `errorMessage`, right?  It's being set when there are errors, not when there might be errors.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        if (!isSequencingOrderValid(sequencingOrderNumber, ddpParticipantId, ddpInstanceDto)) {
+            maybeErrorMessage = String.format("Sequencing order number %s is not a valid order for participant %s ", sequencingOrderNumber,
+                    ddpParticipantId);
+        }
+        if (!isParticipantConsented(ddpParticipantId, ddpInstanceDto)) {
+            maybeErrorMessage = String.format("Participant %s has not consented to receive shared learning ", ddpParticipantId);
+        }
+        if (StringUtils.isNotBlank(maybeErrorMessage)) {
+            log.warn(maybeErrorMessage);
```
zyme: Probably don't need a warn here--the exception thrown below will be handled/logged somewhere up the stack.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+        return false;
+    }
+
+
+    @VisibleForTesting
+    public boolean isSequencingOrderValid(@NonNull String sequencingOrderId, @NonNull String ddpParticipantId,
+                                          @NonNull DDPInstanceDto ddpInstanceDto) {
+        List<MercuryOrderDto> orders = new MercuryOrderDao().getByOrderId(sequencingOrderId);
```
zyme: Consider moving this into `MercuryOrderDto`, and making the method name more explicit.  In this case it looks like it could be `order.containsParticipantInRealm(participantId, instanceName)` or something similar.  The "valid" section of the method name leaves me wondering "what makes an order valid".

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+        if (orders.isEmpty() || orders.stream().anyMatch(order -> StringUtils.isBlank(order.getMercuryPdoId())
+                || ddpInstanceDto.getDdpInstanceId() != order.getDdpInstanceId()
+                || !ddpParticipantId.equals(order.getDdpParticipantId()))) {
+            return false;
+        }
+        return true;
+    }
+
+    public int calculateAge(@NonNull String dateOfBirth) {
```
zyme: Consider moving to `DateTimeUtils` and changing the param to a stronger type, like date/localdate, and then calling the method something like ageInYears().

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /** checks if the answer provided by a participant to a survey question matches the criteria
+     * @param participant participant data from ES.
+     * */
+    private boolean checkAnswerToActivity(ElasticSearchParticipantDto participant, String activityName, String question, Object answer) {
+        Optional<Activities> maybeActivity = participant.getActivities().stream().filter(activities -> activities.getActivityCode()
+                .equals(activityName)).findAny();
+        if (maybeActivity.isPresent()) {
+            Activities activities = maybeActivity.get();
```
zyme: Consider moving the guts of this method into `activities.hasAnswerToQuestion(String questionId, Object answer)`.  Then we can add tests to `Activities` and get re-use of the activity object.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -234,13 +236,56 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static Dsm addParticipantDsmFromFile(String esIndex, String fileName, String ddpParticipantId, String dob) {
```
zyme: How does `participantDsm` differ from just `participant`?  If this is adding a participant to the dsm document, it might be better conveyed as `addParticipantToDsmDoc/Index`.

zyme: Since this is a test utility, it's okay to just have the method declare `throws Exception`.  If we want to avoid throwing the exception, consider narrowing the scope of the `try` so that it only captures the checked exception, like so:

```
Gson gson = new Gson();
        String json = null;
        try {
            json = TestUtil.readFile(fileName);
        } catch (Exception e) {
            e.printStackTrace();
            Assert.fail("Could not read test json for participant " + ddpParticipantId);
            return null;
        }
 ```

Putting too much stuff into the `try` block tends to result in exception handling that is hard to debug.  If we narrow it here, we can make a more specific error message, instead of a generic "oops, something went wrong".  We know exactly what went wrong--we could read the json file.

File: pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/MercuryOrderTestUtil.java [hunk 8]
```diff
@@ -1,21 +1,22 @@
...
     private final MercuryOrderDao mercuryOrderDao = new MercuryOrderDao();
 
-    private final Set<Long> createdOrderIds = new HashSet<>();
+    private final Set<Integer> createdOrderIds = new HashSet<>();
 
     public MercuryOrderDto createMercuryOrder(String ddpParticipantId, String orderBarcode,
                                               int kitTypeId, int ddpInstanceId, int tissueId) {
         MercuryOrderDto orderDto = new MercuryOrderDto(ddpParticipantId, ddpParticipantId, orderBarcode,
-                kitTypeId, ddpInstanceId, Integer.toUnsignedLong(tissueId), null);
+                kitTypeId, ddpInstanceId, tissueId, null);
```
zyme, line 19: Nice catch!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/OncHistoryTestUtil.java
```diff
@@ -104,6 +104,19 @@ public ParticipantDto createParticipant(String guid, DDPInstanceDto ddpInstanceD
         return testParticipant;
     }
 
+    public ParticipantDto createSharedLearningParticipant(String guid, DDPInstanceDto ddpInstanceDto, String dob) {
```
zyme: Could we move this to `TestParticipantUtil`?  Since it's creating a participant, it should probably go in a class that deals with participants.

zyme (commented):
Good stuff!  It was easy for me to understand the intent and get an overall sense of where this is going.  Thanks for the draft share.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: It will add the participant info that is kept in the `DSM` object in the ES doc.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -70,8 +71,7 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
     public static String SQL_UPDATE_ORDER_STATUS =
             "UPDATE ddp_mercury_sequencing SET order_status = ?, status_date = ?, mercury_pdo_id = ?,  "
                     + " status_detail = ?, status_message = ? where order_id = ?";
-
-    private static final int MAX_CHAR_ALLOWED = 2000;
+    private static String SQL_GET_MERCURY_ORDER_ID = " SELECT * FROM ddp_mercury_sequencing WHERE order_id = ? ";
```
zyme: How about `SQL_GET_MERCURY_ORDER_BY_ID`?  GET_MERCURY_ORDER_ID makes it sound like this is just going to return the order id.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,267 @@
...
+        ElasticSearchParticipantDto participant = ElasticSearchUtil.getParticipantESDataByParticipantId(
+                ddpInstanceDto.getEsParticipantIndex(), ddpParticipantId);
+        if (participant.getProfile().isEmpty()) {
+            throw new DsmInternalError("There is no profile in ES for participant " + ddpParticipantId);
+        }
+        PhiManifest finalPhiManifest = generateDataForReport(participant, orders, ddpInstanceDto);
+        return createResponse(finalPhiManifest, ddpParticipantId, sequencingOrderId);
+    }
+
+    private PhiManifestResponse createResponse(PhiManifest finalPhiManifest, String ddpParticipantId, String sequencingOrderId) {
```
zyme: This can move to a method in `PhiManifest`, perhaps in a `toDataArray()` method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,267 @@
...
+
+    /**
+     * checks if the answer provided by a participant to a survey question matches the criteria
+     *
+     * @param participant  participant data from ES.
+     * @param activityName stable id of the activity
+     * @param question     stable id of the question
+     * @param answer       desired answer
+     */
+    private boolean checkAnswerToActivity(ElasticSearchParticipantDto participant, String activityName, String question, Object answer) {
```
zyme: Can you move this to `ElasticSearchParticipantDto` and rename to something like `hasAnswerInActivity()`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * checks if the answer provided by a participant to a survey question matches the criteria
+     *
+     * @param participant  participant data from ES.
+     * @param activityName stable id of the activity
+     * @param question     stable id of the question
+     */
+    private Optional<Object> getParticipantAnswerInSurvey(ElasticSearchParticipantDto participant, String activityName, String question) {
```
zyme: This should also move to `ElasticSearchParticipantDto`

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/OncHistoryTestUtil.java
```diff
@@ -148,6 +150,17 @@ public Object createSmId(String guid, int tissueId, String realm, String userEma
         return patcher.doPatch();
     }
 
+    public Object createSmId(ParticipantDto participantDto, String eligibleSmId, DDPInstanceDto ddpInstanceDto) throws Exception {
```
zyme: This should return a more specific type.  If we know it's going to be a `SMIDPatch`, have the method declare that as the return type, even if internally things need to be cast.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/OncHistoryTestUtil.java
```diff
@@ -461,4 +474,26 @@ public OncHistoryDetail createOncHistoryAndParseResponse(String guid, int partic
         Assert.assertNotNull(oncHistoryDetail);
         return oncHistoryDetail;
     }
+
+    public Object createPatchRequest(String guid, int participantId, String realm, String userEmail, String name, String value,
```
zyme, line 481: Same comment here as L153.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -197,15 +197,66 @@ public int create(MercuryOrderDto mercuryOrderDto) {
...
+                stmt.setLong(2, mercuryOrderDto.getOrderDate());
+                stmt.setString(3, mercuryOrderDto.getDdpParticipantId());
+                stmt.setInt(4, mercuryOrderDto.getKitTypeId());
+                stmt.setString(5, mercuryOrderDto.getBarcode());
+                stmt.setInt(6, mercuryOrderDto.getDdpInstanceId());
+                stmt.setString(7, mercuryOrderDto.getCreatedBy().orElse(""));
+                stmt.setString(11, messageJson);
+                stmt.setNull(12, Types.VARCHAR);
+                if (mercuryOrderDto.getTissueId() != null) {
+                    stmt.setString(8, String.valueOf(mercuryOrderDto.getTissueId()));
```
zyme: We're setting this as a string if it's not null, but then if it is null, we set it to null of type integer?  Seems like this should be setInt.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -244,47 +295,56 @@ public void insertMercuryOrders(List<MercuryOrderDto> newOrders, String json) {
...
-                    results.resultException);
+        if (result.resultException != null) {
+            throw new DsmInternalError(result.resultException);
         }
-        return (int) results.resultValue;
+        log.info("Found {} barcodes for order number {} ", ordersWithOrderId.size(), orderId);
+        return ordersWithOrderId;
+    }
+
+    public boolean isSequencingOrderValidForPhiReport(@NonNull List<MercuryOrderDto> orders, @NonNull String ddpParticipantId,
```
zyme: I think I had an earlier comment about this--what "valid" means is unclear, so moving this to MercuryOrderDto.containsParticipantForInstance() and then iterating through the list would help with clarity and code re-use.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/ParticipantUtil.java
```diff
@@ -110,4 +118,40 @@ public static ParticipantData findApplicantData(String ddpParticipantId, List<Pa
     public static boolean isLegacyAltPid(String participantId) {
         return !isGuid(participantId);
     }
+
+    public static String getParticipantGender(ElasticSearchParticipantDto participantByShortId, String realm, String ddpParticipantId) {
```
zyme: On second thought, why not move this and dependent methods into ElasticSearchParticipantDto?  These methods are basically just checking internal state of that class.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -234,13 +236,57 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static Dsm addDsmObjectToParticipantFromFile(String esIndex, String fileName, String ddpParticipantId, String dob) {
+        Gson gson = new Gson();
+        String json = null;
+        try {
+            json = TestUtil.readFile(fileName);
+            json = json.replace("<dateOfBirth>", dob);
```
zyme, line 245: Do you want to catch an exception in json.replace?  Or is it really just readFile() that needs exception handling?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/phimanifest/PhiManifestResponse.java
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsm.model.phimanifest;
+
+import lombok.Setter;
+
+@Setter
+public class PhiManifestResponse {
```
zyme: If this is specific to route handling, it should probably be closer to the related route, not in the model.  Or it could move to a static class inside of PhiManifest, and then be returned in a method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: Old code, fixed now, thanks for catching it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
pegahtah: I misunderstood that comment and instead added what it is valid for (hence the "forPhiReport" part), I changed it now and moved it. Thx!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: just the readFile, it's only in tests and most probably will just happen if the tests are running with wrong env variables

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/phimanifest/PhiManifestResponse.java
pegahtah: Good idea, I moved it

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -805,19 +805,19 @@ public static void deactivateKitRequest(long dsmKitRequestId, @NonNull String de
         }
     }
 
-    public static KitRequestShipping getKitRequest(@NonNull String kitRequestId) {
+    public static KitRequestShipping getKitRequest(String dsmKitRequestId) {
```
zyme: While we're changing this, can we change `dmsKitRequestId` to an int (along with setInt on L812?  It's type in the database is int.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/phimanifest/PhiManifestResponse.java
```diff
@@ -0,0 +1,18 @@
+//package org.broadinstitute.dsm.model.phimanifest;
```
zyme: If we don't need this file, let's remove it.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/DdpInstanceGroupTestUtil.java
```diff
@@ -116,8 +116,10 @@ private static void deleteInstance(int instanceId) {
 
     public static int getDdpInstanceId(String instanceName) {
```
zyme: Although this method name isn't being changed directly in this PR, I'll mention that a getter is not expected to create something--it should just return a thing.  I think it's more clear to have the "create if not exists" embedded in the method name.  Something like "createDDPInstanceIfNotFound".

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
```diff
@@ -19,14 +22,14 @@ public class MercuryOrderDto {
     String mercuryPdoId;
     String createdBy;
     int ddpInstanceId;
-    Long tissueId;
-    Long dsmKitRequestId;
+    Integer tissueId;
```
zyme, line 25: If tissueId and dsmKitRequestId are required, consider making them int instead of Integer.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/KitRequest.java
```diff
@@ -42,7 +42,7 @@ public KitRequest(String participantId, String shortId, DDPParticipant participa
         this(null, participantId, shortId, null, externalOrderNumber, participant, null, null, null);
     }
 
-    public KitRequest(Long dsmKitRequestId, String participantId, String shortId, String ddpLabel, String externalOrderNumber,
+    public KitRequest(Integer dsmKitRequestId, String participantId, String shortId, String ddpLabel, String externalOrderNumber,
```
zyme, line 45: Is it normal for there to be a kit request without a `dsmKitRequestId`?  If not, this should be an `int` to avoid creating an object with invalid/unexpected state.

zyme (approved):

PR #2764, opened 2023-12-19
Reducing chattiness of housekeeping logs.

zyme:
## Context
Housekeeping logs are full of noisy entries that aren't useful and make it harder to find items of interest.  These log statements have been moved from `warn` or `info` to `debug`:

```
o.b.ddp.service.KitCheckService Participant ... was ineligible for a kit
 o.b.ddp.service.AgeUpService Age-up candidate with guid ... in study ... does not have birth date, skipping
o.b.ddp.service.KitCheckService Participant ... has an invalid mailing address
 o.b.ddp.service.AgeUpService No applicable age-of-majority rules found for participant .... in study ...
```

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
@@ -124,7 +124,7 @@ public class KitCheckService {
 
         if (candidate.getAddressValidationStatus() == null
                 || candidate.getAddressValidationStatus() == DSM_INVALID_ADDRESS_STATUS) {
-            log.warn("Participant {} has an invalid mailing address", userGuid);
+            log.debug("Participant {} has an invalid mailing address", userGuid);
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/KitCheckService.java
```diff
@@ -124,7 +124,7 @@ public KitCheckResult processPotentialKitRecipient(String studyGuid, long kitTyp
 
         if (candidate.getAddressValidationStatus() == null
                 || candidate.getAddressValidationStatus() == DSM_INVALID_ADDRESS_STATUS) {
-            log.warn("Participant {} has an invalid mailing address", userGuid);
+            log.debug("Participant {} has an invalid mailing address", userGuid);
```
ssettipalli, line 127: Rare cases , investigating in prod why kit wasnt created, these statements used to help that ptp was skipped. 
ok to change though to reduce noise because Housekeeping keeps checking and logging same

ssettipalli (approved):

PR #2765, opened 2023-12-19
PEPPER-1261 removing redis ping logic from logback to separate thread.

zyme:
## Context

In the now-abandoned isolated instance (which also used an upgraded version of redis), I noticed some connectivity issues between dss backend and redis.  I traced this how we've been using logback to run redis pings, and when I removed the logback-based pings and replaced them with a simple executor, things got better.  So I am bringing those changes over to the mainline in light of some errors we have seen since the redis upgrade (and the VPC upgrade).

Existing PW tests should pick up any issues.

## FUD Score

## Release


pepper-apis/dss-core/src/main/resources/logback.xml
```diff
@@ -8,17 +8,8 @@
         </encoder>
     </appender>
 
-    <appender name="REDISCONNVALIDATOR" class="org.broadinstitute.ddp.log.RedisConnectionValidatorAppender">
```
zyme, line 11: Removing redis pings from logback

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/log/RedisConnectionValidatorAppender.java
```diff
@@ -1,12 +0,0 @@
-package org.broadinstitute.ddp.log;
```
zyme, line 1: Removing this class since it's only used in logback-based redis pings, and we are moving the pings to a separate thread outside of logback.

ssettipalli (approved):

PR #2766, opened 2023-12-20
PEPPER-1235. upgrade logback version

ssettipalli:
upgrade logback version

### DIFF
--- a/pepper-apis/ddp-common/pom.xml
+++ b/pepper-apis/ddp-common/pom.xml
@@ -12,7 +12,7 @@
 
     <!-- -->
     <properties>
-        <logback.version>1.4.5</logback.version>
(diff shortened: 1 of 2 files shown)


pepper-apis/ddp-common/pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <!-- -->
     <properties>
-        <logback.version>1.4.5</logback.version>
+        <logback.version>1.4.14</logback.version>
```
ssettipalli, line 15: As of now 1.4.14 seem to be the version with no vulnerabilities, so going with it
https://mvnrepository.com/artifact/ch.qos.logback/logback-core

zyme (approved):
LGTM

PR #2768, opened 2024-01-04
PEPPER-1181 OS2 Spanish pdfs

ssettipalli:
Osteo2 spanish pdf templates and pdf configuration changes .
Only latest version of activities/pdfs are impacted with spanish pdfs.

Except 2 files, all other 27 files are spanish pdf templates.
1. OsteoPdfUpdatesES.java : StudyBuilder task to populate new pdf templates and update the pdf configuration to include spanish pdf's
2. study-pdfs-en-es.conf: new / updated pdf config with both en and es languages. Task uses this file to cleanup old pdf config and populate new pdf configs.


aweng98:
Is existing `/study-pdfs.conf` still useful to keep? It could be confusing to someone if we have both files.

aweng98 (approved):

ssettipalli:
> Is existing `/study-pdfs.conf` still useful to keep? It could be confusing to someone if we have both files.

should be ok for this study, Osteo. It went through lot of changes.

PR #2769, opened 2024-01-05
PEPPER-1191 added script to aid in purging data for a study.

zyme:
For PEPPER-1191, refining the original doc on deleting study data [from here](https://docs.google.com/document/d/1xC2-m6wIHwbzqqKuZfGt_amSV__NhvYxJHbrMrUnoCo/edit).  This script does not include erasure of medical records, onc history, clinical orders, and other data that are only relevant to current CMI cancer studies.


ssettipalli (approved):
DSS side of deletion sqls looks good to me.
On DSM side , I dont know of all different DSM tables that need deletions but from what I know it looks good on DSM side deletions too.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
```diff
@@ -0,0 +1,183 @@
...
+
+-- delete kits
+select
+    concat('delete from ddp_kit where dsm_kit_id = ',k.dsm_kit_id, ';')
+from
+    ddp_kit_request req,
+    ddp_instance i,
+    ddp_kit k
+where
+        i.study_guid = @study_guid
```
denniscunningham, line 73: Probably need to confirm the study name (not really a GUI) on each environment, since these are not always consistent across environments.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
```diff
@@ -0,0 +1,183 @@
+-- scripts that will:
+-- generate a bunch of sql statements and shell script calls to be run to erase data for a study from dsm and dss.
+-- it also sets queued events to happen in the infinite future, effectively disabling them
+
+-- set the study to be deleted
+set @study_guid = 'singular' collate 'utf8mb4_unicode_ci';
+
+-- set the gcp environment in which to run the user deletes for dss
+set @gcp_project = 'broad-ddp-test';
```
denniscunningham, line 10: Is there a way to parameterize the env?
Seem like a new PR is required for each env?
Or do you expect this code is throw away after its use on each env?

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
```diff
@@ -0,0 +1,183 @@
...
+    not exists (select 1 from user_governance g where g.operator_user_id = u.user_id)
+  and
+        s.guid = @study_guid;
+
+
+use test_dsm_db;
+
+-- delete kits
+select
+    concat('delete from ddp_kit where dsm_kit_id = ',k.dsm_kit_id, ';')
```
denniscunningham, line 67: Not sure I understand why this pattern is necessary? (i.e. "concate('delete...")

denniscunningham (commented):
Just trying to understand the context...
This is intended for limited use to remove a set of studies now? IOW, it is not intended for deleting studies in the future? 
It does not include medical records, etc., so would not be appropriate for all studies, as is.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
zyme: The script outputs a bunch of delete statements to be run, in addition to doing some deletes directly.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
zyme: I am expecting that this script can be used to wipe out any study.  Just set these parameters and go, no need for a new PR.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
zyme: Added a comment to remind us to do this.

zyme:
The goal here is a script that can be used to semi-automate the erasure of studies that do not include medical records or onc history.  In other words, anything but CMI studies.

pepper-apis/scripts/admin/delete-participant-data-for-study.sql
denniscunningham: Sorry, still trying to wrap my head around this, but I am apparently not familiar with this technique.
Is the resulting select statement as follows:
`select delete from ddp_kit where dsm_kit_id=12; from ddp_kit_request req...`?

denniscunningham (approved):

PR #2771, opened 2024-01-18
PEPPER-1289 fix to handle user preferred language for AOM emails during Ageup process

ssettipalli:
PEPPER-1289 
handle the scenario where Agedup users email is passed in the message during Ageup prep
If the message contains participantGuid, try to get user preferred language

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/PubSubMessageBuilder.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/PubSubMessageBuilder.java
@@ -106,6 +106,14 @@ public class PubSubMessageBuilder {
                         if (jmlDtoOpt.isPresent()) {
                             userPreferredLangCode = jmlDtoOpt.get().getLanguageCode();
                         }
+                    } else if (queuedNotificationDto.getParticipantGuid() != null) {
+                        //handle the scenario where Agedup users email is passed in the message during Ageup prep
+                        //try to get user preferred language
+                        UserProfile profile = apisHandle.attach(UserProfileDao.class)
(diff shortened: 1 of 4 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/PubSubMessageBuilder.java
```diff
@@ -106,6 +106,14 @@ public PubsubMessage createMessage(String ddpMessageId,
                         if (jmlDtoOpt.isPresent()) {
                             userPreferredLangCode = jmlDtoOpt.get().getLanguageCode();
                         }
+                    } else if (queuedNotificationDto.getParticipantGuid() != null) {
```
ssettipalli, line 109: handle the scenario where Agedup users email is passed in the message during Ageup prep
During AGEUP prep , we collect aged-up users emailID and it is used to send Ageup related emails requesting to consent. User is not yet signed up on own but is a governed user and message will contain the user guid.
If the message contains participantGuid, try to get user preferred language to that emails can be sent in preferred language rather than study default.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/housekeeping/PubSubMessageBuilder.java
```diff
@@ -226,6 +234,9 @@ public PubsubMessage createMessage(String ddpMessageId,
                     }
                 }
 
+                log.info("Looking up notification template(s) for event configId: {} study: {} and user preferred language={} "
```
ssettipalli, line 237: just some extra logging

Rinchan20 (approved):

PR #2776, opened 2024-01-25
Osteo spanish updates

ssettipalli:
Osteo pe-cgs spanish updates


Rinchan20 (approved):
checked changes in test env - looks good

PR #2777, opened 2024-01-25
PEPPER-526 PHI Manifest Yes/No/Blank fixes

zyme:
## Context

As Whitney [pointed out in slack](https://broadinstitute.slack.com/archives/C043W8G8SS3/p1706121846474169?thread_ts=1706118450.961619&cid=C043W8G8SS3), there are a few columns in the PHI manifest where default values of "No" are misleading.  This addresses the problem by writing blank `""` values in a few columns instead of defaulting to "No" when the value is blank.

## FUD Score

_Overall, how are you feeling about these changes?_

- [x ] :relaxed: All good, business as usual!
- [ ] :sweat_smile: There might be some issues here
- [ ] :scream: I'm sweaty and nervous

## How do we demo these changes?

_How does one observe these changes in a deployed system? Note that **user visible** encompasses many personas--not just patients and study staff, but also ops duty, your fellow devs, compliance, etc._

- [ x] They are user-visible in dev as a regular user journey and require no additional instructions.  Load the [PHI manifest in DSM](https://dsm-dev.datadonationplatform.org/osteo2/phiManifest) for participant `HNCX820VERQH07GYIGAP` and order  `I46YYWYEYXJSJSC7XEHK` and verify that the Somatic Assent Addendum Response column and the Somatic Consent Tumor Response column are blank.

## Testing

- [ ] I have added regression test labels to the jira ticket
- [ ] I have added verification steps to the jira ticket
- [ x] I have written automated positive tests
- [ x] I have written automated negative tests
- [ ] I have written zero automated tests but have poked around locally to verify proper functionality
- [ ] The jira ticket has acceptance criteria and QA has the needed information to test changes

## Release

- [x ] These changes require no special release procedures--just code!
- [ ] Releasing these changes requires special handling and I have documented the special procedures in the release plan document

PR #2781, opened 2024-01-26
PEPPER-526 Adult Consent

zyme:
Follow-on to https://github.com/broadinstitute/ddp-study-server/pull/2777.  The adult consent field was also being defaulted to "No" even when the participant did not answer the question.

The fix is similar to the other PR.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
@@ -68,22 +68,23 @@ public class ElasticSearchParticipantDto {
     /**
      * Changes the value for the given question's answer.
      * Does not make any modification to underlying elastic data.
+     * @return true if the value was changed, false if the activity
+     * and question do not exist.
      */
     @VisibleForTesting
-    public void changeQuestionAnswer(String activityCode, String questionStableId, String value) {
+    public boolean changeQuestionAnswer(String activityCode, String questionStableId, String value) {
         for (Activities activity : getActivities()) {
             if (activity.getActivityCode().equals(activityCode)) {
                 for (Map<String, Object> questionAnswer : activity.getQuestionsAnswers()) {
                     if (questionAnswer.containsKey(questionStableId)) {
                         questionAnswer.replace(ESObjectConstants.ANSWER, value);
                         questionAnswer.replace(questionStableId, value);
-                        return;
+                        return true;
                     }
                 }
             }
         }
-        throw new DsmInternalError(String.format("Could not change answer to %s.%s because question %s was not found "
-                + "in any activities.", activityCode, questionStableId, questionStableId));
+        return false;
     }
 
     public Optional<Address> getAddress() {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
@@ -158,9 +158,10 @@ public class PhiManifestService {
         String assentAddendumResponse = convertBooleanActivityAnswerToString(participant.getParticipantAnswerInSurvey(
                 CONSENT_ADDENDUM_PEDIATRICS_ACTIVITY_CODE, SOMATIC_ASSENT_ADDENDUM_QUESTION_STABLE_ID));
         phiManifest.setSomaticAssentAddendumResponse(assentAddendumResponse);
-        Optional<Boolean> consentAnswer = getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid());
+        Optional<Object> consentAnswer = getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid());
+
         if (consentAnswer.isPresent()) {
-            phiManifest.setSomaticConsentTumorResponse(convertBooleanToYesNo(consentAnswer.get()));
+            phiManifest.setSomaticConsentTumorResponse(convertBooleanToYesNo((Boolean)consentAnswer.get()));
         } else {
             phiManifest.setSomaticConsentTumorResponse("");
         }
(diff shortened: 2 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -68,22 +68,23 @@ protected ElasticSearchParticipantDto() {  }
     /**
      * Changes the value for the given question's answer.
      * Does not make any modification to underlying elastic data.
+     * @return true if the value was changed, false if the activity
+     * and question do not exist.
      */
     @VisibleForTesting
-    public void changeQuestionAnswer(String activityCode, String questionStableId, String value) {
+    public boolean changeQuestionAnswer(String activityCode, String questionStableId, String value) {
```
zyme, line 75: The way adult consent for tumor is determined requires for PE-CGS requires looking for either an OS-specific answer or an LMS-specific answer.  Rather than change the queries in the manifest test to be aware of LMS/OS study guids, I changed this method to return whether or not the field is found.  Existing clients are tests that will fail for various reasons if the expected fields are not set, so I've left those calls alone.  Effectively, nothing is checking this value as of now, and I think that's okay.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -203,13 +204,11 @@ private boolean hasParticipantConsentedToTumor(ElasticSearchParticipantDto parti
      * If the participant has answered the question, it will be returned
      * as a boolean.  If they have not, an empty optional will be returned.
      */
-    private static Optional<Boolean> getAdultParticipantConsentedToTumorAnswer(ElasticSearchParticipantDto participant, String studyGuid) {
+    private static Optional<Object> getAdultParticipantConsentedToTumorAnswer(ElasticSearchParticipantDto participant, String studyGuid) {
```
zyme, line 207: This was the source of the issue Whitney identified.  Instead of interpreting "no answer" to mean "false", this method now returns an empty optional, leaving it to clients to figure out what to do with an empty response.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -222,7 +221,8 @@ private boolean hasParticipantConsentedToSharedLearning(ElasticSearchParticipant
         String dateOfBirth = participant.getDsm().get().getDateOfBirth();
         String dateOfMajority = (String) participant.getDsm().get().getDateOfMajority();
         if (StringUtils.isBlank(dateOfMajority) || DateTimeUtil.isAdult(dateOfMajority)) {
-            return getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid()).get();
+            return (Boolean)getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid())
```
zyme, line 224: In this usage, we *do* want the absence of a value to be considered "false", since they have not given consent.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/phimanifest/PhiManifestTest.java
```diff
@@ -235,5 +249,19 @@ private void assertPhiManifest(PhiManifest phiManifest, List<MercuryOrderDto> or
                         SOMATIC_ASSENT_ADDENDUM_QUESTION_STABLE_ID));
         Assert.assertTrue(
                 StringUtils.equals(somaticAssentAddendumReportValue, phiManifest.getSomaticAssentAddendumResponse()));
+
+        String os2SomaticConsent = PhiManifestService.convertBooleanActivityAnswerToString(
```
zyme, line 253: Checks for the OS-specific or LMS-specific consent fields.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -158,9 +158,10 @@ public static PhiManifest generateDataForReport(@NonNull ElasticSearchParticipan
         String assentAddendumResponse = convertBooleanActivityAnswerToString(participant.getParticipantAnswerInSurvey(
                 CONSENT_ADDENDUM_PEDIATRICS_ACTIVITY_CODE, SOMATIC_ASSENT_ADDENDUM_QUESTION_STABLE_ID));
         phiManifest.setSomaticAssentAddendumResponse(assentAddendumResponse);
-        Optional<Boolean> consentAnswer = getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid());
+        Optional<Object> consentAnswer = getAdultParticipantConsentedToTumorAnswer(participant, ddpInstanceDto.getStudyGuid());
```
zyme, line 161: In this usage, we do want to respond differently to an empty value since it is going to be written to the phi manifest.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -158,9 +158,10 @@ public static PhiManifest generateDataForReport(@NonNull ElasticSearchParticipan
...
+
         if (consentAnswer.isPresent()) {
-            phiManifest.setSomaticConsentTumorResponse(convertBooleanToYesNo(consentAnswer.get()));
+            phiManifest.setSomaticConsentTumorResponse(convertBooleanToYesNo((Boolean)consentAnswer.get()));
```
denniscunningham, line 164: I don't think this cast provides any value.

denniscunningham (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
zyme: Alas the compiler says otherwise 😞

PR #2782, opened 2024-01-29
PEPPER-526 Consent Part Deux

zyme:
Continuation of the [original PR](https://github.com/broadinstitute/ddp-study-server/pull/2777).

The PHI manifest was returning "No" instead of blank when there is no value given for CONSENT_ADDENDUM.SOMATIC_CONSENT_ADDENDUM_TUMOR (LMS) or CONSENT_ADDENDUM.SOMATIC_CONSENT_TUMOR (OS).  For pediatrics, these values are not set, so the manifest must render them as blank instead of "No" when they are missing (which is the case when the participant is a pediatric case).

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
@@ -72,7 +72,7 @@ public class ElasticSearchParticipantDto {
      * and question do not exist.
      */
     @VisibleForTesting
-    public boolean changeQuestionAnswer(String activityCode, String questionStableId, String value) {
+    public boolean changeQuestionAnswer(String activityCode, String questionStableId, Object value) {
         for (Activities activity : getActivities()) {
             if (activity.getActivityCode().equals(activityCode)) {
                 for (Map<String, Object> questionAnswer : activity.getQuestionsAnswers()) {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
@@ -68,22 +68,23 @@ public class ElasticSearchParticipantDto {
     /**
      * Changes the value for the given question's answer.
      * Does not make any modification to underlying elastic data.
+     * @return true if the value was changed, false if the activity
+     * and question do not exist.
      */
     @VisibleForTesting
-    public void changeQuestionAnswer(String activityCode, String questionStableId, String value) {
+    public boolean changeQuestionAnswer(String activityCode, String questionStableId, String value) {
         for (Activities activity : getActivities()) {
             if (activity.getActivityCode().equals(activityCode)) {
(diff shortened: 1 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/search/ElasticSearchParticipantDto.java
```diff
@@ -72,7 +72,7 @@ protected ElasticSearchParticipantDto() {  }
      * and question do not exist.
      */
     @VisibleForTesting
-    public boolean changeQuestionAnswer(String activityCode, String questionStableId, String value) {
+    public boolean changeQuestionAnswer(String activityCode, String questionStableId, Object value) {
```
zyme, line 75: I expanded this to `Object` because the underlying ES data will have these as booleans, and I wanted more realistic data here for testing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/phimanifest/PhiManifestService.java
```diff
@@ -204,11 +204,11 @@ private boolean hasParticipantConsentedToTumor(ElasticSearchParticipantDto parti
      * If the participant has answered the question, it will be returned
      * as a boolean.  If they have not, an empty optional will be returned.
      */
-    private static Optional<Object> getAdultParticipantConsentedToTumorAnswer(ElasticSearchParticipantDto participant, String studyGuid) {
+    public static Optional<Object> getAdultParticipantConsentedToTumorAnswer(ElasticSearchParticipantDto participant, String studyGuid) {
         if (DBConstants.LMS_STUDY_GUID.equals(studyGuid)) {
-            return participant.getParticipantAnswerInSurvey(CONSENT_ADDENDUM_PEDIATRICS_ACTIVITY_CODE, LMS_QUESTION_SOMATIC_CONSENT_ADDENDUM_TUMOR_STABLE_ID);
```
zyme, line 209: Here was the bug: I put in the wrong activity code for adult cases.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/phimanifest/PhiManifestTest.java
```diff
@@ -34,7 +35,7 @@
 public class PhiManifestTest extends DbAndElasticBaseTest {
     private static final DDPInstanceDao ddpInstanceDao = new DDPInstanceDao();
     private static final String instanceName = "phi_report_instance";
-    private static final String lmsStudyGuid = "cmi-lms";
+    private static final String lmsStudyGuid = DBConstants.LMS_STUDY_GUID;
```
zyme, line 38: Little bit of D-R-Y

pepper-apis/dsm-core/src/test/resources/elastic/lmsActivitiesSharedLearningEligible.json
```diff
@@ -44,6 +44,12 @@
     "SOMATIC_CONSENT_ADDENDUM_TUMOR" : true,
     "questionType" : "BOOLEAN"
     },
+    {
```
zyme, line 47: Because there was no SOMATIC_CONSENT_TUMOR value in the json, the test could not reset the answer.  Adding the block here means that the test can manipulate the value as needed.

denniscunningham (approved):

PR #2786, opened 2024-01-30
PEPPER-59 Java17 for docker build/test setup

zyme:
Bringing over some relevant java17 test bits from the isolated instance work.  The changes here give us java17 for builds and tests, resulting in a hybrid where java11 is still the app engine runtime, but the jars are built with java17 (still using 11 as a target).

Note that with sonar disabled on mainline, there will be merge conflicts as this branch will need to keep sonar enabled in order to verify that our java17 setup eliminates the java11 errors with sonar.


pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -29,9 +29,9 @@ network:
   name: managed
   instance_tag: study-server # necessary for firewall rules so that DSM will accept https requests from pepper
 {{if and (or (eq $environment "prod") (eq $environment "staging") (eq $environment "dev") (eq $environment "test")) (eq $gae "true") }}
-entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=pepper-backend,-cprof_cpu_use_per_thread_timers=true,-logtostderr -Xmx1640m -jar DataDonationPlatform.jar
+entrypoint: java -javaagent:tcell/tcellagent.jar -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_enable_heap_sampling=true,-cprof_service=pepper-backend,-cprof_cpu_use_per_thread_timers=true,-logtostderr -Xmx1640m -Dio.grpc.netty.shaded.io.netty.transport.noNative=false -jar DataDonationPlatform.jar
```
zyme, line 32: I discovered in last year's isolated CMI instance that a netty upgrade (which may or may not be required along with java17) broke things in really cryptic ways that took a few days to troubleshoot.  This may also be needed for cloud functions.

pepper-apis/configure.rb
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/ruby -w
 # Docker Image used to call out to generate config files
-$dsde_toolbox_image_name = "broadinstitute/study-server-build:java-2022-06-01A"
+
+$dsde_toolbox_image_name = "broadinstitute/study-server-build:java-2023-03-28C"
```
zyme, line 4: We may want to get even more recent versions of deps in dockerfiles.  The readme describes how to label and push these builds so that whenever we're running things on this branch, we can keep the images in sync without cluttering mainline.

denniscunningham (commented):
Sorry, I realize this is not up for review yet, but do we need to also update the POM files that specify a `maven-compiler-plugin`?

zyme:
@denniscunningham that's a great question.  I was thinking that this PR would just cover the docker/circle based builds, and that subsequent PRs could cover maven builds and java versions that devs will need to change locally, but I'm up for bundling this all together too.

denniscunningham:
@zyme As far as I know, that is the only other global/PR level change needed, so perhaps it fits here? Since we use Maven as part of the CI process, making the Docker changes without the Maven changes might put things out of whack?

denniscunningham (approved):

PR #2787, opened 2024-02-01
[PEPPER-1169] DSM sprint 18 RC bug fixes

denniscunningham:
## Context
[PEPPER-1169]
[PEPPER-843]

There are two distinct bug fixes in the PR. If that proves confusing I can split them into two.

Bug #1: This manifest as a bad OS1 ES export on test environment. This is because OS1 instance names are different on dev and test (thus tests on dev passed). DSM uses the instance name to identify instances when exporting, so this can create problems. For OS1 the instance name is 'osteo' on dev, but 'Osteo' on test and prod, so case-insensitive comparisons fix the problem. Obviously it would be better to have consistent instance names across environments, and ideally DSM would use stable instance IDs instead of instance names for identification. Each of these fixes will take more time.

Bug #2: This manifest as ES participantData entities for registration status being out of sync with corresponding ParticipantData in the DSM DB: the DB was correct, but the data was not reflected in ES. Registration status changes, processed by `WorkflowStatusUpdate`, never included an ES update, which is a bug. This bug was masked by repeated processing of AT default data for each registration update. This was also a bug, since that code added new genome IDs for each registration status change, but was fixed for the current RC candidate. However, updating ES after creating the erroneous genome IDs had the affect of updating the registration status. The fix was to add an ES update after updating registration status. 

## Checklist

## FUD Score

## Testing

## Release

[PEPPER-1169]: https://broadworkbench.atlassian.net/browse/PEPPER-1169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PEPPER-843]: https://broadworkbench.atlassian.net/browse/PEPPER-843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -88,46 +90,56 @@ public static void updateCustomWorkflow(Map<String, String> attributesMap, Strin
...
-                basicDefaultDataMaker.generateDefaults(studyGuid, ddpParticipantId);
-            }
-        } catch (Exception e) {
-            throw new DsmInternalError("Could not add AT default values", e);
-        }
+        exportWorkflowToES(workflow, status, ddpParticipantId, instance, setting, participantDataList);
+        updateEsParticipantData(ddpParticipantId, instance);
     }
 
     private static boolean isATStudy(String studyGuid) {
```
zyme, line 148: Glad to see we're already doing case-insensitive match on the guid here too.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/ATDefaultValues.java
```diff
@@ -71,12 +68,7 @@ protected boolean setDefaultData() {
                 return false;
             }
 
-            ObjectTransformer objectTransformer = new ObjectTransformer(instance.getName());
```
zyme, line 74: Looks like this moved to updateEsParticipantData, with minimal logging changes and no exception wrapping, since that's already done here, correct?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/AdditionalMedicalRecordsRetriever.java
```diff
@@ -17,10 +17,10 @@ class AdditionalMedicalRecordsRetriever extends AdditionalRecordsRetriever<Medic
     }
 
     static Optional<AdditionalMedicalRecordsRetriever> fromRealm(String realm) {
-        if  (OLD_OSTEO_INSTANCE_NAME.equals(realm)) {
+        if  (OLD_OSTEO_INSTANCE_NAME.equalsIgnoreCase(realm)) {
             return Optional.of(new AdditionalMedicalRecordsRetriever(NEW_OSTEO_INSTANCE_NAME));
         }
-        if (NEW_OSTEO_INSTANCE_NAME.equals(realm)) {
```
zyme, line 23: Since these comparisons are scattered about, we might benefit from a separate class that just does something like isOsteo1() and isOsteoPECGS().

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -88,46 +90,56 @@ public static void updateCustomWorkflow(Map<String, String> attributesMap, Strin
...
+                fieldSettingsDao.getFieldSettingByColumnNameAndInstanceId(instance.getDdpInstanceIdAsInt(), workflow);
         if (fieldSetting.isEmpty()) {
-            throw new DsmInternalError(String.format("Invalid workflow name %s for instance %s", workflow, instance.getName()));
+            throw new DsmInternalError(String.format("Invalid workflow name %s for instance %s", workflow, instanceName));
         }
 
         FieldSettingsDto setting = fieldSetting.get();
         String fieldType = setting.getFieldType();
-        boolean isOldParticipant = participantDatas.stream().anyMatch(
+        boolean isOldParticipant = participantDataList.stream().anyMatch(
```
zyme: What does "old participant" mean here?  If it's just a pre-existing participant vs. a soon-to-be-created one, maybe rename to `doesParticipantExist` because oldness could also mean something like "participant is of a certain age" or "this participant was first enrolled in datstat" (or something else).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -88,46 +90,56 @@ public static void updateCustomWorkflow(Map<String, String> attributesMap, Strin
...
-            log.info("Creating participant {} data for participant {} in study {} via workflow {} with status {}",
-                    fieldType, ddpParticipantId, studyGuid, workflow, status);
+            log.info("Creating participantData {} for participant {} in study {} via workflow {} with status {}",
+                    fieldType, ddpParticipantId, instanceName, workflow, status);
             createParticipantData(workflow, status, ddpParticipantId, instance.getDdpInstanceIdAsInt(), fieldType);
         }
-        exportWorkflowToES(workflow, status, ddpParticipantId, instance, setting, participantDatas);
 
-        try {
-            if (isATStudy(studyGuid)) {
```
zyme, line 122: This is bug #2 that masked bug #1, correct?

zyme: Also I think we both are learning something about markdown pound-number issue link macros today 🙄

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -206,11 +218,11 @@ protected static int createParticipantData(String workflow, String status, Strin
         return participantDataId;
     }
 
-    protected static void updateProbandStatus(String workflow, String status, ParticipantData participantData,
-                                              String fieldSettingsFieldType) {
+    protected static boolean updateProbandStatus(String workflow, String status, ParticipantData participantData,
```
zyme: Quick javadoc for why boolean is returned?  Also, the test does not check the return value.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/defaultvalues/ReferralSourceServiceTest.java
```diff
@@ -285,7 +284,7 @@ private void applyFoundOutAnswer(List<Activities> activities, List<String> answe
         refSourceQA.put(DDPActivityConstants.ACTIVITY_QUESTION_ANSWER, answer);
     }
 
-    protected static List<Activities> getActivities() {
+    public static List<Activities> getActivities() {
```
zyme, line 287: Candidate for `@VisibleForTest`?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/pubsub/UpdateWorkflowStatusTest.java
```diff
@@ -69,79 +98,168 @@ private static ParticipantData createParticipantData(String participantIdSeed) {
...
-        if (participantDataId >= 0) {
-            participantDataDao.delete(participantDataId);
-        }
-    }
-
-    private static void createFieldSettings() {
+    private static void createMemberTypeFieldSetting() {
         FieldSettingsDto fieldSettingsDto = new FieldSettingsDto.Builder(ddpInstanceDto.getDdpInstanceId())
-                .withFieldType(fieldTypeId)
+                .withFieldType(memberTypeFieldTypeId)
```
zyme, line 107: Given the recent discovery with osteo name differences across environments, are we sure we've got consistently named field types and column names across environments?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/pubsub/UpdateWorkflowStatusTest.java
```diff
@@ -69,79 +98,168 @@ private static ParticipantData createParticipantData(String participantIdSeed) {
...
-            JsonObject dataJsonObject = gson.fromJson(data, JsonObject.class);
-            Assert.assertEquals(status, dataJsonObject.get(workflow).getAsString());
-        } finally {
-            deleteParticipantData(participantDataId);
-        }
+
+        String workflow = "REGISTRATION_STATUS";
+        String status = "ENROLLED";
+        String fieldType = participantData.getFieldTypeId().orElseThrow();
+        WorkflowStatusUpdate.updateProbandStatus(workflow, status, participantData, fieldType);
```
zyme: Should we be checking the return value here too?

zyme (approved):
Thanks for putting these together--PR overview plus tickets saved me a bunch of time getting up to speed.  A few small things, but otherwise ✅

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/defaultvalues/ATDefaultValues.java
denniscunningham: Correct. This code is now in WorkflowStatusUpdate, at least temporarily. No need to add a redundant layer of exception catching for its current usage.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/AdditionalMedicalRecordsRetriever.java
denniscunningham: I'd rather compare by instance ID (which requires restructuring), or even better, remove this code when we make a better fix for OS1/OS2 handling. I consider all this a hack that needs to go.
Does that sound ok?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
denniscunningham: I can rename. This is legacy code I've been working on bit by bit for months. I try not to change too much, like variable names, to preserve some of the original code, but at this point I have rewritten this method a few times. (See the git history.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
denniscunningham: This is actually just an unfortunate git diff. Mostly what I did here was break out a hunk of code in `updateWorkflowData` to `updateWorkflowData`. Overdue since it helps readability and testing. (As I mentioned, I've been refactoring this spaghetti code for a while, but been conservative until I can pair changes with tests.)
I did not intend to use #1 and #2 as markdown. The two bugs have nothing to do with each other.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
denniscunningham: Yes, I partially refactored our comparison of study GUIDs in a prior PR, but there is still more work to do (you can see TODOs in the code).

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/defaultvalues/ReferralSourceServiceTest.java
denniscunningham: This is already a test, so unnecessary.
At some point I will reorganize this and similar into test utilities.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/pubsub/UpdateWorkflowStatusTest.java
denniscunningham: I'm not sure we do, since DSM users can change or delete these whenever. That is a huge flaw in this "feature". 
Like the many flaws in DSM, we can find a way to bandaid this, or re-implement correctly. 
I don't think a bandaid is appropriate in this PR. What do you think?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/defaultvalues/ReferralSourceServiceTest.java
zyme: doh, sorry I totally lost that this was already in a test!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/pubsub/UpdateWorkflowStatusTest.java
zyme: As long as we know at the moment that the types and column names we're looking for here are consistent, I think we don't need a change in this PR.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/AdditionalMedicalRecordsRetriever.java
zyme: Given the larger plan to separate OS1/2 indexes, I'm okay with leaving this as-is.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/pubsub/UpdateWorkflowStatusTest.java
denniscunningham: Actually, I take that back. I do not think these particular field settings are editable by users, but I have to relearn which are and which are not, and where we apply those restrictions. 
Also, *some* field settings are in Liquibase, but not all. In short, it is a mess, but I forget the details of the mess at this point.

PR #2789, opened 2024-02-01
PEPPER-526 failed lookups for tissue_id 0

zyme:
Bug in PEPPER-526.  rs.getInt() returns 0 when the value is null, and that was causing exceptions when trying to lookup nonexistent tissue with id 0.  PhiManifestService:L109 was expecting nulls and will filter them out properly already, but because getInt() uses 0 to represent null, that filter did not have the intended effect of removing the null tissue since 0 is not null.

PR to merge this into the current RC is: https://github.com/broadinstitute/ddp-study-server/pull/2790

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
@@ -317,7 +317,8 @@ public class MercuryOrderDao implements Dao<MercuryOrderDto> {
                                 .withStatusDate(rs.getLong(DBConstants.MERCURY_STATUS_DATE))
                                 .withOrderStatus(rs.getString(DBConstants.MERCURY_ORDER_STATUS))
                                 .withDdpParticipantId(rs.getString(DBConstants.DDP_PARTICIPANT_ID))
-                                .withTissueId(rs.getInt(DBConstants.TISSUE_ID))
+                                .withTissueId(rs.getObject(DBConstants.TISSUE_ID) != null
+                                        ? rs.getInt(DBConstants.TISSUE_ID) : null)
                                 .withDsmKitRequestId(rs.getInt(DBConstants.DSM_KIT_REQUEST_ID)).build();
                         ordersWithOrderId.add(mercurySequencingDto);
                     }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
@@ -99,7 +99,7 @@ public class MercuryOrderDto {
             return this;
         }
 
-        public Builder withTissueId(int tissueId) {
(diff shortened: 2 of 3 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
```diff
@@ -317,7 +317,8 @@ public List<MercuryOrderDto> getByOrderId(String orderId) {
                                 .withStatusDate(rs.getLong(DBConstants.MERCURY_STATUS_DATE))
                                 .withOrderStatus(rs.getString(DBConstants.MERCURY_ORDER_STATUS))
                                 .withDdpParticipantId(rs.getString(DBConstants.DDP_PARTICIPANT_ID))
-                                .withTissueId(rs.getInt(DBConstants.TISSUE_ID))
+                                .withTissueId(rs.getObject(DBConstants.TISSUE_ID) != null
```
zyme, line 320: Here's where sql null turns into 0, leading to downstream query failures when looking for a nonexistent tissue with id 0.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
```diff
@@ -99,7 +99,7 @@ public Builder withCreatedByUserId(String createdByUserId) {
             return this;
         }
 
-        public Builder withTissueId(int tissueId) {
+        public Builder withTissueId(Integer tissueId) {
```
zyme, line 102: Let the null through so caller checks on nullity will work properly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/mercury/MercuryOrderDao.java
denniscunningham: Looks like this is new-ish code, only used in PHI manifest.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
denniscunningham: Looks like `MercuryOrderDto` is heavily used.
Do we have a sense that no code is relying on the `0` value?

denniscunningham (commented):
Just one question.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/mercury/MercuryOrderDto.java
zyme: None that I could find.  Although I did find other code that is expecting getTissueId() to be null sometimes, so those paths might be triggered.

zyme: Although if the objects are created with getInt(), the behavior will remain the same.

denniscunningham (approved):

PR #2791, opened 2024-02-03
[PEPPER-1311] Implement ParticipantDataFixupService for AT genomic IDs

denniscunningham:
## Context
PEPPER-1311

This PR uses the AdminOperation facility to enable programmatic deletion of unneeded AT ParticipantData records. In this case the target is additional genomic ID and exit status records created for AT participants, which are causing confusion for AT study staff.

This particular service, `ParticipantDataFixupService`, can be extended in the future to handle other types of ParticipantData fixups. 

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
```diff
@@ -0,0 +1,214 @@
...
+
+    protected enum UpdateStatus {
+        UPDATED,
+        NOT_UPDATED,
+        ERROR,
+        NO_PARTICIPANT_DATA
+    }
+
+    private static final Gson gson = new Gson();
+    protected List<String> validRealms = List.of("atcp");
```
ssettipalli, line 39: Just a thought.
If the service is specific to only ATCP, maybe the name should reflect it like ATParticipantDataFixupService.java
OR have ATParticipantDataFixupService.java as separate class extending ParticipantDataFixupService.java supporting only ATCP and have realm check in there.

ssettipalli (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
denniscunningham: Yup, at this point it's hard to say what the next ParticipantData fixup would look like, and we may have no other cases, or we could have like 5.

What I tried to do here is strike a balance: The service is parameterized (at the HTTP request level) so we can use that to indicate the type of fixup, but of course there is only one now. I think if the next fixup is a lot like this one (uses much of the same code) then I would probably just keep the structure as is and put conditionals in the code. If the next case is different enough to warrant a separate class, then likely we would implement a class hierarchy as you describe, or perhaps just go with two separate (but renamed) classes.

I am working under the "you're not gonna need it (YAGNI)" principal, and I don't think this implementation has ruled out any future options.

Does that make sense? What do you think?

denniscunningham (commented):
Thanks for your review @ssettipalli!
I left a response to you comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
ssettipalli: sounds ok to me.

PR #2797, opened 2024-02-14
PEPPER-1264 . update somatic pex for lms

ssettipalli:
fix pex so that DSS angular validation doesnt fail


aweng98 (approved):
let's try this in dev today.

PR #2805, opened 2024-02-25
[PEPPER-1353] Add ParticipantInitService to update missing RGP default data

denniscunningham:
## Context
[PEPPER-1353]

With PEPPER-1161 we changed the way RGP participant data gets created (see https://docs.google.com/document/d/1KMP-5oWFRNKktmVbzgaAbeZBMmOOp9TfEa39yNz3vzE/edit). That ensured there will be no new RGP participants without family IDs, but it left a gap for the existing RGP participants who have not yet completed enrollment: they will be missing a family ID until they complete enrollment.

This PR adds an AdminOperation, ParticipantInitService, to add family IDs and default data for those participants. It also handles a few edge cases -- essentially bad data -- by either fixing the data, or noting the problem in the operation log. ParticipantInitService has a dry run mode, since the operation considers all RGP participants, not just a list of participants. Dry run mode can also be used to simply identify bad/inconsistent data issues, so it might be something we expand in the future.

I ran ParticipantInitService locally against Dev data, to ensure it worked properly.

RgpParticipantDataTestUtil contains a set of test utility methods, culled from RgpParticipantDataServiceTest. These methods are now needed by more than one test, and the class adds the benefit of automatic resource management. 

UpdateStatus was moved from individual AdminOperation implementing classes to UpdateLog, where it can be shared.

The PR includes a handful of code cleanups:

1. Renamed and refactored ParticipantDataDao.getParticipantData. The prior name was inaccurate. Unfortunately this had a ripple effect into several files.
2. Removed unused imports and fixed checkstyle issues from a handful of files I recently edited. (IOW, I overlooked the issues after making changes.)
3. Cleaned up ElasticSearchUtil. getDDPParticipantsFromES. The log messages were ambiguous, since the same log message is used for more than one method, and there was unnecessary code. There is a lot of work to do in this file, and I chip away as methods are brought under test.

## Checklist

## FUD Score

## Testing

## Release

[PEPPER-1353]: https://broadworkbench.atlassian.net/browse/PEPPER-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
@@ -28,6 +28,16 @@ import org.broadinstitute.dsm.service.participantdata.RgpFamilyIdProvider;
 import org.broadinstitute.dsm.service.participantdata.RgpParticipantDataService;
 import org.broadinstitute.dsm.util.ElasticSearchUtil;
 
+/**
+ * Service to initialize RGP participant data for participants that have not yet been initialized.
+ * The service also detects errors in participant data, fixes those that have a direct fix,
+ * and logs errors for those that require investigation. The 'logging' is part of the operation result which is
+ * stored in the DB. See AdminOperationRecord for details.
+ * </p>
+ * This service has a dry-run mode that logs changes that would be made without actually making them,
+ * including logging errors. Dry-run mode should always be run first to ensure the operation will yield the expected
+ * results.
+ */
 @Slf4j
 public class ParticipantInitService implements AdminOperation {
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
@@ -8,6 +8,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import com.google.gson.Gson;
 import lombok.extern.slf4j.Slf4j;
@@ -103,11 +104,19 @@ public class ParticipantInitService implements AdminOperation {
     protected static UpdateLog initParticipant(String ddpParticipantId, ElasticSearchParticipantDto esParticipant,
                                                DDPInstance ddpInstance, FamilyIdProvider familyIdProvider,
                                                boolean isDryRun) {
+        ParticipantData probandData = null;
         List<ParticipantData> ptpData = RgpParticipantDataService.getRgpParticipantData(ddpParticipantId);
+        if (!ptpData.isEmpty()) {
+            UpdateLog updateLog = new UpdateLog(ddpParticipantId, UpdateLog.UpdateStatus.ERROR.name());
+            probandData = getProbandRgpData(ptpData, updateLog);
+            if (probandData == null) {
+                return updateLog;
+            }
+        }
 
         Optional<Dsm> dsm = esParticipant.getDsm();
         if (dsm.isPresent() && StringUtils.isNotBlank(dsm.get().getFamilyId())) {
-            if (ptpData.isEmpty()) {
+            if (probandData == null) {
                 return new UpdateLog(ddpParticipantId, UpdateLog.UpdateStatus.ERROR.name(),
                         "Participant has family ID in ES but no RGP data");
             }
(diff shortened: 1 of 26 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
```diff
@@ -0,0 +1,178 @@
...
+            throw new DsmInternalError("Invalid realm: " + realm);
+        }
+
+        String esIndex = ddpInstance.getParticipantIndexES();
+        if (StringUtils.isEmpty(esIndex)) {
+            throw new DsmInternalError("No ES participant index for study " + realm);
+        }
+
+        // require dryRun since the user cannot specify a participant list
+        String dryRun = attributes.get("dryRun");
```
ssettipalli, line 74: If property doesnt exist and not true , maybe we can consider as NOT dryrun by default but enforcing like here is ok too

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
```diff
@@ -0,0 +1,201 @@
...
+import org.broadinstitute.dsm.model.elastic.search.ElasticSearchParticipantDto;
+import org.broadinstitute.dsm.service.admin.AdminOperation;
+import org.broadinstitute.dsm.service.admin.AdminOperationRecord;
+import org.broadinstitute.dsm.service.participantdata.FamilyIdProvider;
+import org.broadinstitute.dsm.service.participantdata.RgpFamilyIdProvider;
+import org.broadinstitute.dsm.service.participantdata.RgpParticipantDataService;
+import org.broadinstitute.dsm.util.ElasticSearchUtil;
+
+@Slf4j
+public class ParticipantInitService implements AdminOperation {
```
ssettipalli, line 42: would help to add a line like where the data is initialized. Is it DSM DB ? elastic ?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
```diff
@@ -0,0 +1,201 @@
...
+        if (StringUtils.isBlank(familyId)) {
+            updateLog.setError("Participant has proband RGP data that does not contain a family ID");
+        }
+        return participantData;
+    }
+
+    protected static UpdateLog createDefaultData(String ddpParticipantId, ElasticSearchParticipantDto esParticipant,
+                                                 DDPInstance ddpInstance, FamilyIdProvider familyIdProvider,
+                                                 boolean isDryRun) {
+        if (isDryRun) {
```
ssettipalli, line 185: Is dryrun just doing some validations and not actually doing any updates in DB and then rolling back ?
I know we cannot do any updates to elastic. 
Trying to understand dryrun purpose

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
denniscunningham: Per the PR comments, I wanted the user to be specific about what they want to do, since this operation affects potentially all participants in a study. IOW, the user should definitely do a dry run first, so defaulting to no dry run is not what we want here.
We could also default to dry run = true, but at that point the user needs to know about the dry run flag anyway, so just requiring this seems more straightforward.
Does that make sense?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
denniscunningham: Dry run just looks the data in ES and the DB, and logs what action to expect in when not doing a dry run. For instance, it will say UPDATED or NOT_UPDATED or ERROR, etc. This provides a full log of what to expect when this is run for real.
In non-dry-run mode (real changes), this will update the DB and/or ES.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
denniscunningham: Not sure I understand.
Do you mean a comment about what potentially gets changed?
I realize now that the instructions for this are largely in the PR comments, so it makes sense to describe usage somewhere in the code. I can add that.

denniscunningham (commented):
Thanks for your review @ssettipalli!
If I understood your comments correctly, it seems like there is at least one update I can make (instructions)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
ssettipalli: I meant java doc like some text for developers to understand what we are updating, not a  big deal any way

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantInitService.java
ssettipalli: I meant by default dryrun = false but like I said I am ok with enforcing

ssettipalli:
> Thanks for your review @ssettipalli! If I understood your comments correctly, it seems like there is at least one update I can make (instructions)?

ssettipalli (approved):
looks ok to me

PR #2813, opened 2024-03-04
[PEPPER-1267] Implement ParticipantData legacy ID analysis

denniscunningham:
## Context
[PEPPER-1267]

The primary purpose of this PR is to provide an AdminOperation to analyze ParticipantData that has a legacy participant ID for it's ddp_participant_id. The specific issue is some participants have a set of ParticipantData records with both legacy PIDs and standard PIDs. In the case observed in PEPPER-1267, the ParticipantData records have this pattern, and also have overlapping data. (For instance two AT_GROUP_GENOME_STUDY records with conflicting information.)

Since these cases are not easy to detect with a DB query (some of the required info is only in ES), this `ParticipantDataFixupService` addition is only intended to enumerate and analyze cases (dryRun mode). Once the extent of the issue is understood, the fix will either be coded into `ParticipantDataFixupService`, or addressed on a case by case basis with DB updates. In either case, ParticipantData with a legacy PID will be converted to use a standard ID (once record conflicts are resolved).

One issue identified in PEPPER-1267 is that only the ParticipantData with legacy PIDs is found in ES. That 'bug' is in `BaseMigrator`, since when resolving ParticipantData with legacy PIDs it overwrites ParticipantData for the same participant. That bug should not be fixed since it will result in conflicting duplicate ParticipantData records.

This PR includes a handful of conversions from GSON to Jackson serialization in the code the PR touches, since GSON has become problematic with the upgrade to Java 17. The full conversion will take a while -- these are just opportunistic changes. Specifically:

1. Introduced Jackson based serialization for response objects: `JsonResponseTransformer`.
2. Changed ParticipantData data map serialization to Jackson.
3. Added a non-null Jackson ObjectMapper, since this is required for ES dates.
4. Now using Jackson serialization for AdminOperations.

This PR also includes the following refactoring and cleanups:

1. Started ElasticSearchService, which is overdue. The idea is to add ES capabilities as they come under test. The relationship between `ElasticSearch` and `ElasticSearchUtil` is messy, and ``ElasticSearch` is not really structured correctly, but as I continue to cleanup `ElasticSearchUtil` I will move methods from there,
2. Added abstract class `ParticipantAdminOperationService` to consolidate common AdminOperationService patterns.
3. Converted `ATDefaultValues` to `ATParticipantDataService`. I had largely rewritten this code, but the include patterns for callers had become awkward.
4. Added `ParticipantDataService`, again overdue. This currently includes support code for `ParticipantDataFixupService`, and code moved from `WorkflowStatusUpdate` .
5. Added `ATParticipantDataTestUtil.java` to consolidate and expand AT test support.

## Checklist

## FUD Score

- [] They are user-visible in dev as a regular user journey and require no additional instructions.

## Testing

## Release

[PEPPER-1267]: https://broadworkbench.atlassian.net/browse/PEPPER-1267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantData.java
```diff
@@ -136,10 +133,7 @@ public Map<String, String> getDataMap() {
         if (StringUtils.isBlank(data)) {
             return null;
         }
-        Type type = new TypeToken<HashMap<String, String>>() {
-        }.getType();
-        Gson gson = new Gson();
-        cachedDataMap = gson.fromJson(data, type);
+        cachedDataMap = ObjectMapperSingleton.readValue(data, new TypeReference<Map<String, String>>() {});
```
zyme: I think this can be simplified to `new TypeReference<>()`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -67,22 +66,25 @@ private boolean hasReachedToBatchLimit(BulkExportFacade bulkExportFacade) {
...
 
-    private Map<String, Object> replaceLegacyAltPidKeysWithGuids(Map<String, Object> participantRecords) {
-        participantRecords = new ConcurrentHashMap<>(participantRecords);
-        List<String> legacyAltPids =
-                participantRecords.keySet().stream().filter(ParticipantUtil::isLegacyAltPid).collect(Collectors.toList());
-        Map<String, String> guidsByLegacyAltPids = elasticSearch.getGuidsByLegacyAltPids(index, legacyAltPids);
-        for (Map.Entry<String, String> entry : guidsByLegacyAltPids.entrySet()) {
-            String legacyAltPid = entry.getKey();
-            if (!participantRecords.containsKey(legacyAltPid)) {
+    private Map<String, Object> replaceLegacyPIDs(Map<String, Object> participantRecords) {
```
zyme, line 72: How about `replaceLegacyPIDsWithGuids` so it's clear what the pid is being replaced with?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Return a list of all participant guids in a participant ES index
+     */
+    public List<String> getAllParticipantGuids(String esParticipantsIndex) {
+        ElasticSearch es = elasticSearch.getAllParticipantsDataByInstanceIndex(esParticipantsIndex);
+        List<String> participantIds = new ArrayList<>();
+        for (ElasticSearchParticipantDto participantDto : es.getEsParticipants()) {
+            if (participantDto.getProfile().isPresent()) {
+                Profile profile = participantDto.getProfile().get();
+                if (StringUtils.isNotBlank(profile.getGuid())) {
```
zyme, line 48: Should we be checking for guid uniqueness and removing/logging duplicate participant guids?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/util/JsonResponseTransformer.java
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsm.route.util;
+
+import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
+import spark.ResponseTransformer;
+
+public class JsonResponseTransformer implements ResponseTransformer {
```
zyme: We have `JsonTransformer` and now `JsonResponseTransformer`, both of which implement `ResponseTransformer`.  How about if we change `JsonTransfomer` to `GsonResponseTransformer` and/or this one to `JacksonResponseTransformer`?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
```diff
@@ -3,35 +3,63 @@
...
+            this.label = label;
+        }
+
+        public static FixupType valueOfLabel(String label) {
+            for (FixupType fixupType : values()) {
+                if (fixupType.label.equals(label)) {
+                    return fixupType;
+                }
+            }
+            return null;
```
zyme, line 68: Maybe throw `DsmInternalError` here instead, if there's no valid use case for a fixup with no fixupType.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
```diff
@@ -3,35 +3,63 @@
...
+import org.broadinstitute.dsm.service.participantdata.ParticipantDataService;
+import org.broadinstitute.dsm.util.ParticipantUtil;
+import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
 
 /**
- * AdminOperation service to asynchronously fix various ParticipantData issues
+ * AdminOperation service to asynchronously fix various ParticipantData issues.
+ * </p>
+ * The operation requires a fixupType attribute. The supported fixup types are:
+ * - atcpGenomicId: remove duplicate genomic IDs and participant exit statuses
```
zyme, line 30: Since these strings are in `FixupType`, maybe move these comments to the enum and then reference the enum here in the class javadoc.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        List<ElasticSearchParticipantDto> participants = elasticSearch.parseSourceMaps(response.getHits().getHits());
+        log.info("Got {} participants from ES index {} (getLegacyPidsByGuid)", participants.size(), esIndex);
+
+        Map<String, String> guidToPid = new HashMap<>();
+        participants.forEach(participant -> {
+            if (participant.getProfile().isPresent()) {
+                Profile profile = participant.getProfile().get();
+                if (StringUtils.isNotBlank(profile.getGuid()) && StringUtils.isNotBlank(profile.getLegacyAltPid())) {
+                    guidToPid.put(profile.getGuid(), profile.getLegacyAltPid());
```
zyme, line 88: Is it possible for us to have multiple legacy pids for the same guid?  Should we be checking that here?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/TestParticipantUtil.java
```diff
@@ -79,4 +95,17 @@ public static ParticipantDto createIneligibleSharedLearningParticipant(String pa
...
                 "elastic/lmsActivitiesSharedLearningIneligible.json");
     }
+
+    /**
+     * Create a profile with a unique HRUID (GUID assignment left to caller), and email constructed from first and
+     * last name.
+     */
+    public static Profile createProfile(String firstName, String lastName, int uniqueDigits) {
+        Profile profile = new Profile();
+        profile.setEmail(String.format("%s%s@broad.org", firstName, lastName));
```
zyme: If there's any chance of emails being sent somewhere as a result of this, let's use `@broad.dev` since `@broad.org` goes somewhere else.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/proxy/jackson/ObjectMapperSingleton.java
```diff
@@ -4,28 +4,38 @@
...
 
 public class ObjectMapperSingleton {
 
+    private static final ObjectMapper objectMapperInstance = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .setSerializationInclusion(JsonInclude.Include.NON_NULL);
+
+    private static final ObjectMapper nonNullObjectMapper = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
zyme, line 21: I think we might want to use `ALWAYS`, since there are a number of `serializeNulls` calls in DSM code that is using gson.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
denniscunningham: Erf. Hit a nerve, or peeve, or something.

I'm not a fan of overly descriptive method names, which plague DSM, for a few reasons:

1. They are non-standard in the Java world. Exhibit A: any popular Java library.
2. They hurt readability. This is a matter of opinion, but they unnecessarily lengthen code lines, and they often need to be read as a clause although Java clauses are typically a group of tokens.
3. They give coders a false sense that the method name accurately encompasses the full behavior of the method. (Perhaps there are cases where they do, but there is no guarantee that they do.

Instead Java shops typically use a couple of conventions: 

1. Java doc for all but obvious public methods. If we think this important, there is a *lot* of remedial work to do in DSM.
2. For private/protected methods, reading and understanding the class code before making changes.

This case falls into the latter case, but I will add Java doc just to be clear.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantData.java
denniscunningham: I am definitely not an expert, but it does not look I can. But perhaps I am missing your point?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/util/JsonResponseTransformer.java
denniscunningham: Sure. 
I was reluctant to make a sweeping change to the existing method name. There are typically so many minor improvements I can make that I don't want bloat PRs with them.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
denniscunningham: These are here for a specific reason. 
For an AdminOperation user, this is the closest we have to instructions on how to create a request, since there is no UI.
I suppose I could make a file that is just instructions, but that might be overkill.
And/or I could make a copy of this comment for the enum.
In any case, I beefed up the class doc since as instructions it was not complete.

What do you think?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
denniscunningham: This is an enum that is already a bit heavy.
It's a bit unusual for enums to throw exceptions, at least in my experience. That makes them more like a class than a type.
However, I cleaned-up the calling code, since it had an odd way of validating the values. I think the validation and error handling are clearer now.
What do you think?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
denniscunningham: This is a method I moved and cleaned-up since it was under test and eligible for this class. (I am not the original author, and this method is currently only used for testing.)
However, I can improve the method if needed. I suppose it is possible that the GUIDs in the profile do not match the ES document, and it is an easy enough change to make.
In non-test environments profiles are created by DSS. Have we had instances where this was done incorrectly?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/proxy/jackson/ObjectMapperSingleton.java
denniscunningham: Not sure of what you mean, but I definitely have a typo in `objectMapperInstance`... `.setSerializationInclusion(JsonInclude.Include.NON_NULL)` should not be there.
Is that what you meant or something else?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
denniscunningham: I can check duplicate GUIDs here and throw an error.
But based on your prior comment it sounds like you are concerned about the integrity of participant GUID in Profiles?
If so, a better check would be to assure that the profile GUID matches the doc ID.  That's a more fundamental check for the situation you are concerned about here. I can wire that up if it seems important, but it feels like that's a pretty severe error we would have heard about already?

denniscunningham (commented):
Thank you @zyme for your review!
I included you on the reviewer list since you mentioned this last week, but thx for the thorough review since I realize you have other priorities.
I committed changes due to the review, but I also have some questions about whether more work is needed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
zyme: One thing that I think would help reviewers is a PR comment about what methods are being moved, on both sides of the refactor (the original class with the method, and the new method's home).  Without this information, the reviewer must do their own analysis to determine whether the "new" method is actually new, or just different, and that's often time-consuming.  When it looks like it _might_ be different, I end up making suggestions like this, which are probably a distraction at best.  I haven't seen cases where we have duplicate guids, so I'm okay just leaving this as-is until we have a signal that shows otherwise.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
zyme: Fantastic topic for process and standards!  A while back we had a working agreement that we would do javadoc for all but the most obvious methods.  I don't think we've stuck to it, so let's discuss the best way to make things more readable.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantData.java
zyme: Interesting...my intellij suggests it is safe.  Just looking to eliminate redundant explicit typing if it's truly optional.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/util/JsonResponseTransformer.java
zyme: How about updating this class to reflect its jackson-ness, but leaving the gson transformer class as-is?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
zyme: Looks good to me now!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/adminoperation/ParticipantDataFixupService.java
zyme: Very nice!  Thank you.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/proxy/jackson/ObjectMapperSingleton.java
zyme: Ah, now I get it.  I misunderstood how `includeNullValues` and was looking too broadly at existing route transforms in gson, which aren't relevant to these changes.  Sorry!

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantData.java
denniscunningham: Ah, I get it now. You were not suggesting getting rid of the empty implementation, `{}`.

I can make this change, although this style is prevalent in the code and provides some level of documentation.
Objectively, the second parameter to `readValue` is pretty clumsy, so perhaps just a placeholder is sufficient (rather than attempting to describe the object)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
denniscunningham: Totally agree. My bad for not doing that here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/util/JsonResponseTransformer.java
denniscunningham: Erf. I already made the change.
Happy to revert if that makes more sense.

zyme (approved):

PR #2814, opened 2024-03-04
Pepper-1334

ssettipalli:
Data Exporter changes to update second Osteo2 index


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -638,10 +649,26 @@ private void convertInfoToJSONAndExportToES(Handle handle,
...
         });
 
-        BulkResponse bulkResponse = esClient.bulk(bulkRequest, RequestOptions.DEFAULT);
+        if (finalBulkRequestOS2 != null) {
+            log.info("exporting {} participant records to index participants_structured.cmi.cmi-osteo2", participantRecords.size());
+            BulkResponse bulkResponse2 = esClient.bulk(finalBulkRequestOS2, RequestOptions.DEFAULT);
+            if (bulkResponse2.hasFailures()) {
+                log.error(bulkResponse2.buildFailureMessage());
+            }
+        }
```
aweng98: Q:

1. Is it intentional to update OS2 index before OS1 index?
2. If OS2 update fails, should run OS1 update always or stops?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -638,10 +649,26 @@ private void convertInfoToJSONAndExportToES(Handle handle,
                     .doc(value, XContentType.JSON)
                     .docAsUpsert(true);
             bulkRequest.add(updateRequest);
+
+            if (isOsteoParticipantStructuredIndex) {
```
aweng98, line 654: Variables `isOsteoParticipantStructuredIndex`, `osteoParticipantStructuredIndex` and `finalBulkRequestOS2` seems to be redundant to me because I can do `if (bulkRequestOS2 != null)`. Is it simpler?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -638,10 +649,26 @@ private void convertInfoToJSONAndExportToES(Handle handle,
...
+                        .id(key)
+                        .doc(value, XContentType.JSON)
+                        .docAsUpsert(true);
+                finalBulkRequestOS2.add(updateRequestOS2);
+            }
         });
 
-        BulkResponse bulkResponse = esClient.bulk(bulkRequest, RequestOptions.DEFAULT);
+        if (finalBulkRequestOS2 != null) {
+            log.info("exporting {} participant records to index participants_structured.cmi.cmi-osteo2", participantRecords.size());
```
aweng98: Cleaner if 
`log.info("exporting {} participant records to index {}", participantRecords.size(), CMI_OSTEO2_INDEX);`

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
ssettipalli: we are just logging and continuing. 
DSM wanted to use OS2 , new index as new osteo2 index so just made the call to do it first.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
ssettipalli: variables in lambda expressions forced me to make them effectively final, hence those extra variable assignments. I will take another look.

ssettipalli: ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.12.1:compile (default-compile) on project dss-core: Compilation failure
[ERROR] /Users/sampath/IdeaProjects/ddp-study-server/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java:[661,17] local variables referenced from a lambda expression must be final or effectively final

aweng98 (approved):
Thank you!

PR #2822, opened 2024-03-18
PEPPER-1372 allow null address2 for unvalidated mailing addresses.

zyme:
## Context

PEPPER-1372 fixes a bug that was preventing addresses from being created when `street2` was null.  At first, this seemed simple: just remove `@NonNull` from [EasyPostUtil.createAddressWithoutValidation()](https://github.com/broadinstitute/ddp-study-server/pull/2822/files#r1528475914), and add a [test](https://github.com/broadinstitute/ddp-study-server/pull/2822/files#r1528478924) or [two](https://github.com/broadinstitute/ddp-study-server/pull/2822/files#r1528480757).

Along the way, however, I realized that `EasyPostUtil` does not properly manage apiKeys.  If you look at the constructors, it seems like you can create many `EasyPostUtil` objects that each have different apiKeys.  In reality, however, they are just setting the singleton `EasyPost.apiKey` field, which will lead to race conditions in DSM where you don't really know which apiKey you are using.  I have added javadoc comments to warn us about this.  

There is a bit of refactoring going on here as well, EasyPostUtil had two constructors, one with a single string, and one with two strings.  Callers that used the two-string version always set the first string to null, so I created a few `from...()`-style factory methods to clarify usage.

Fixing this is a bigger deal that I'm leaving for another time.  Our version of easypost (3.4.0) is way out of date, and the [latest 7.1.0](https://github.com/EasyPost/easypost-java) appears to bring many benefits, including the ability to better manage multiple apikeys concurrently.  It's also unclear to me at this point whether the use of different apiKeys really matters at all.  They all roll up to the same credit card, and there are not differences in permission levels for apiKeys--they are all or nothing.  If we did have to separate costs by apiKey use in logs, we would likely have the wrong numbers.  But we don't, so I think we're okay for now.

## FUD Score

- [x ] :sweat_smile: There might be some issues here

## Testing

- [ x] I have added regression test labels to the jira ticket

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -38,8 +40,12 @@ public DeliveryAddress(String street1, String street2, String city, String state
         this.zip = zip;
         this.country = country;
 
-        empty = (StringUtils.isBlank(street1)) && (StringUtils.isBlank(street2)) && (StringUtils.isBlank(city))
-                && (StringUtils.isBlank(state)) && (StringUtils.isBlank(zip)) && (StringUtils.isBlank(country));
+        empty = (StringUtils.isBlank(this.street1))
```
zyme: Code above this is already setting state, so if we're going to set state in a different way from the params, we should be checking the state, not the args.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -49,30 +55,7 @@ public DeliveryAddress(String street1, String street2, String city, String state
         this.phone = phone;
     }
 
-    private DeliveryAddress(@NonNull Address address) {
```
zyme, line 52: Never used, deleted.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -49,30 +55,7 @@ public DeliveryAddress(String street1, String street2, String city, String state
...
-            Address address = Address.retrieve(id);
-            deliveryAddress = new DeliveryAddress(address);
-        } catch (Exception ex) {
-            logger.error(LOG_PREFIX + "An error occurred during address retrieval.", ex);
-        }
-        return deliveryAddress;
-    }
-
-    public void validate() {
+    public boolean validate() {
```
zyme, line 63: Returned the `valid` value for convenience.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -228,7 +264,7 @@ public Address createAddress(DDPParticipant ddpParticipant, String phone) throws
         return Address.create(toAddressMap);
     }
 
-    public Address createAddressWithoutValidation(@NonNull String name, @NonNull String street1, @NonNull String street2,
```
zyme, line 231: Main change here: `street2` can be null.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/lddp/util/DeliveryAddressTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class DeliveryAddressTest {
+
+    @BeforeClass
+    public static void setUp() {
+        EasyPostUtil.initializeEasyPostFromTypeSafeConfig();
+    }
+
+    @Test
+    public void testNullAddress2() {
```
zyme, line 16: Sanity check that I built along the way to understand the behavior.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/EasyPostUtilTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.broadinstitute.ddp.util.ConfigManager;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class EasyPostUtilTest {
+
+    private EasyPostUtil easyPostUtil =  EasyPostUtil.initializeEasyPostFromTypeSafeConfig();
+
+    @Test
+    public void testNullAddress2() {
```
zyme, line 13: Basic test that creates the address without verifying it, and then verifies it just to make sure there's nothing wrong with it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
denniscunningham: Not sure I understand how this is different, since all of the params are assigned to instance variables. Or is this more of a semantic change than a functional change?
But it is also unclear why a class is needed for this something that could be a method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -71,7 +71,7 @@ public KitResponse createNonPepperKit(JuniperKitRequest juniperKitRequest, Strin
      * @param kitTypeName the specific name of the request kit type
      *
      * @return an instance of KitResponse, either as an error or KitStatus
-     * **/
+     **/
```
denniscunningham: Nit. This should be `*/`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -30,6 +32,15 @@
...
 
+/**
+ * Although various constructors that use apiKeys implies
+ * that each new object will use a particular apiKey, that is not
+ * the case.  In reality, apiKey is used
+ * as a static singleton via {@link EasyPost#apiKey}.
+ * It's likely that DSM is
+ * using different apiKeys under load, when setting up
+ * kits for different realms/projects/clients.
+ */
```
denniscunningham, line 43: Should we open a ticket to fix this? I see from you PR comments we are unsure of the consequences.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
```diff
@@ -76,20 +87,45 @@ public class EasyPostUtil {
...
+     * an unexpected api key.
+     */
+    public static EasyPostUtil fromInstanceName(@NonNull String instanceName) {
+        String apiKey = DSMServer.getDDPEasypostApiKey(instanceName);
         if (StringUtils.isNotBlank(apiKey)) {
-            EasyPost.apiKey = apiKey;
-            logger.info("Setup of EasyPost api key");
+            return new EasyPostUtil(apiKey);
         } else {
             throw new RuntimeException("No EasyPost api key was found");
```
denniscunningham: This should be a `DSMInternalError`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -95,8 +78,17 @@ public void validate() {
...
+                AddressVerification deliveryVerification = address.getVerifications().get("delivery");
+                if (!deliveryVerification.getSuccess()) {
+                    List<Error> errors = deliveryVerification.getErrors();
+                    StringBuilder errorMessage = new StringBuilder();
+                    if (errors != null && !errors.isEmpty()) {
+                        for (Error error : errors) {
+                            errorMessage.append("validation error: " + error.getMessage() + "\n");
+                        }
+                    }
+                    logger.info(LOG_PREFIX + "Address verification failed for "
```
denniscunningham, line 95: Maybe a TODO to expose this to the eventual caller?
Or do we have a more extensive rewrite planned?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -110,10 +102,11 @@ public void validate() {
                     phone = address.getPhone();
                 }
             } catch (Exception ex) {
-                logger.error(LOG_PREFIX + "An error occurred during address verification.", ex);
+                logger.error(LOG_PREFIX + "An error occurred during address verification of " + address.prettyPrint(), ex);
```
denniscunningham, line 112: Ideally this error is (re)-thrown and we sort out how to expose it at the highest level possible, but I suspect we've got different error handling techniques peppered throughout this code. (I know I found a few discrepancies when I was last working in the area.)
Maybe just a TODO for now?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/lddp/util/DeliveryAddressTest.java
```diff
@@ -0,0 +1,21 @@
...
+        var address = new DeliveryAddress("75 Ames St.", null, "Cambridge", "MA", "02141", "US");
```
denniscunningham, line 17: Nit: We don't use `var` in Pepper.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
denniscunningham: If you now understand what this method/class does it would be great to add a comment. IOW, what does the EasyPost validation do? Are we assured of having a valid shipping address, or is this just checking that the proper fields are included?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/EasyPostUtilTest.java
```diff
@@ -0,0 +1,30 @@
...
+        if (StringUtils.isBlank(EasyPost.apiKey)) {
+            EasyPost.apiKey = ConfigManager.getInstance().getConfig().getString("easyPostApiKey");
+        }
```
denniscunningham: It feels like we are avoiding rather than testing a basic service capability, although perhaps I am not understanding this correctly. 
Perhaps the `initializeEasyPostFromTypeSafeConfig` call should go here, and if it fails the test fails? This code seems to be circumventing the behavior of the module under test, rather than testing this aspect of its behavior.

denniscunningham (commented):
Do we have any idea if allowing null `street2` fields through to EasyPost causes an error. I am confused about why we are so strict in `createAddressWithoutValidation`, but apparently not elsewhere. (In `DeliveryAddress`, for example.)
IOW, why is `phone` required to be non-null here?

How was this working before, since `street2` is often not part of a valid address?

zyme:
I have no information about why we required non-null `street2` in the past and suspect that marking it as such was itself a bug.  The test does perform EasyPost validation of an address with null `street2`.  Since EasyPost marks our test address without it as valid, I feel confident that EasyPost won't object to it.  Apparently `street2` was non-null during parsing of upload sheets.  The other place where we use non-null street2 is in our own return address labels, where `kit_return_information.return_address_street2` is always set, since we include additional information about which building (320) the kits should be returned to.

Initially, I had logic that did a "if street2 is null, set it to empty string" and when the tests passed with that change, I then removed the change to see if the tests would pass with `null`.  I was delighted and surprised to see that they did pass.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/EasyPostUtilTest.java
zyme: Actually I meant to remove L17-19.  Will do that.  I had factored out the common code to find a usable `apiKey`.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/lddp/util/DeliveryAddressTest.java
zyme: On 2023-10-03, we decided we could use `var` in the [process and standards meeting](https://docs.google.com/document/d/1EOXXHBJ44A0FZpf1HFesXDlV9UCptncioaxuUQJJd4M/edit).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
zyme: Will do--and I think if we hit an exception, `valid` should be reset to `false`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
zyme: The caller can know this by either checking the return value or calling `isValid()`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
zyme: Moving out to a separate method.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/EasyPostUtil.java
zyme: I don't think it's worth it at this point.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -37,9 +38,15 @@ public DeliveryAddress(String street1, String street2, String city, String state
         this.state = state;
         this.zip = zip;
         this.country = country;
+    }
 
-        empty = (StringUtils.isBlank(street1)) && (StringUtils.isBlank(street2)) && (StringUtils.isBlank(city))
```
zyme, line 41: Moved to `isEmpty()`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
```diff
@@ -49,34 +56,15 @@ public DeliveryAddress(String street1, String street2, String city, String state
...
-        this.street1 = address.getStreet1();
-        this.street2 = address.getStreet2();
-        this.city = address.getCity();
-        this.state = address.getState();
-        this.zip = address.getZip();
-        this.country = address.getCountry();
-        this.phone = address.getPhone();
-    }
-
-    public static DeliveryAddress populateFromEasyPost(@NonNull String id) {
```
zyme, line 64: `populateFromEasyPost` was never used, so I removed it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
denniscunningham: Much better. As a class with setters, the prior 'empty' check was wrong.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/lddp/util/DeliveryAddress.java
denniscunningham: Sorry, I see my comment was ambiguous. 
It seems like the logged information here would be good to include in a message to the requester/caller. So I was thinking a comment/todo to that effect. But maybe we just need to refactor all this stuff.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/lddp/util/DeliveryAddressTest.java
denniscunningham: I would like to revisit that after my time on Juniper.
Now that I see that practice in use I have an opinion. :-)

denniscunningham (approved):

PR #2830, opened 2024-03-25
PEPPER-849 osteo1 user enrollment API

ssettipalli:
API to create and enroll Osteo1 user (non prod scenario and mainly for test support)

Create user, Auth0 user, create and complete consent-v1 and release-v1 versions for self and pediatric. If all good publishes participant_registered event to DSM and pushes new user to elastic
consent, parental_consent, consent_assent, release_self, release_minor activities

Auth
Needs a valid DSM token

Config driven:
Vault conf has property "allowOsteo1UserCreation" which will drive if this API can be invoked or not 

endpoint for the API
{{api_base_url}}/pepper/v1/dsm/studies/cmi-osteo/ddp/osteo1user

Sample payload
{
    "email": "name+os1.self.mar25.dev.3@broad.dev",
	"firstName": "firstName self OS1",
	"lastName": "OS1LN",
	"birthDate": "1980-10-10"
}


pepper-apis/config/application.conf.ctmpl
```diff
@@ -23,6 +23,7 @@
                 {{if $i}},{{end}}"{{ $email }}"
             {{end}}
     ],
+    "allowOsteo1UserCreation": {{$conf.Data.allowOsteo1UserCreation}},
```
ssettipalli, line 26: Drives if API can be invoked or not in a particular env.
Will be disabled in prod and staging

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -177,8 +177,10 @@ public Object handle(Request request, Response response, UserRegistrationPayload
                 var pair = signUpNewOperator(response, handle, study, auth0UserId.get(),
                         payload, clientConfig, mgmtClient);
                 operatorUser = pair.getOperatorUser();
-                triggerUserRegisteredEvents(handle, study, operatorUser, pair.getParticipantUser());
-                publishRegisteredPubSubMessage(studyGuid, pair.getParticipantUser().getGuid());
+                if (!payload.skipTriggerEvents()) {
```
ssettipalli, line 180: This is the only core DSS change and shouldn't impact normal prod scenarios. 
will checkout in dev for sure.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -530,6 +531,7 @@ private static void start() throws MalformedURLException {
 
         post(API.DSM_NOTIFICATION, new ReceiveDsmNotificationRoute(), jsonSerializer);
         post(API.DSM_TERMINATE_USER, new DsmExitUserRoute(), responseSerializer);
+        post(API.OSTEO1_USERS, new Osteo1UserCreationRoute(new TaskPubSubPublisher()), responseSerializer);
```
zyme: Would it be safer to only register this endpoint only when `cfg.getBoolean(ConfigFile.ALLOW_OS1_USER_CREATION)` instead of doing the check inside the route?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
```diff
@@ -139,6 +139,9 @@ public class ConfigFile {
     public static final String AUTH0_IP_ALLOW_LIST = "auth0.ipAllowList";
     public static final String PREFERRED_SOURCE_IP_HEADER = "preferredSourceIPHeader";
     public static final String EMAIL_PATTERN_DENY_LIST = "emailPatternDenyList";
+    public static final String ALLOW_OS1_USER_CREATION = "allowOsteo1UserCreation";
+    public static final String OS1_USER_DEFAULT_PWD = "osteo1UserDefaultPwd";
+    public static final String OSTEO_AUTH0_CLIENT_ID = "osteoClientId";
```
zyme, line 144: Do we have a new client just for this route?  If we're re-using an existing client, I think we can join `umbrella_study -> auth0_tenant -> client` and hit the db directly, assuming the study_guid is stable across non-prod.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/UserRegistrationPayload.java
```diff
@@ -59,11 +59,20 @@ public class UserRegistrationPayload {
     @SerializedName("timeZone")
     private String timeZone;
 
+    //only used for Osteo1 user creation
+    private boolean skipTriggerEvents = false;
```
zyme, line 63: Makes sense to me to put this here, since this route is locked down IP-wise so that only Auth0 can call it (and in any event, the route isn't visible on prod).

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/json/users/requests/UserCreationPayload.java
```diff
@@ -22,7 +22,6 @@ public class UserCreationPayload {
     @SerializedName("email")
     private String email;
 
-    @NotEmpty
```
zyme, line 25: Given this change, I think we want someting in `UserCreationRoute` around L52 so that we throw an error when this isn't present.  Am I right that it's no longer required just for OS1 user creation route?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -177,8 +177,10 @@ public Object handle(Request request, Response response, UserRegistrationPayload
...
+                    triggerUserRegisteredEvents(handle, study, operatorUser, pair.getParticipantUser());
+                    publishRegisteredPubSubMessage(studyGuid, pair.getParticipantUser().getGuid());
+                }
```
zyme, line 185: Might be worth a log.warn in an else block, just so we know when these events are skipped, as a backup in case something gets to prod and it shouldn't.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/Osteo1UserCreationRoute.java
```diff
@@ -0,0 +1,671 @@
...
+    }
+
+    @Override
+    protected int getValidationErrorStatus() {
+        return HttpStatus.SC_UNPROCESSABLE_ENTITY;
+    }
+
+    @Override
+    public Object handle(Request request, Response response, UserCreationPayload payload) throws Exception {
+        String studyGuid = request.params(RouteConstants.PathParam.STUDY_GUID);
```
zyme, line 52: Can you break this out into some helper classes and create some tests for those helper classes?  And then see if we can get this route down to 20-30 lines?  With all the logic wrapped up in a route, testing is going to be hard.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/constants/ConfigFile.java
ssettipalli: unless I am missing something, tenant is same for cmi studies and I needed osteo angular client to register as osteo user.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/Osteo1UserCreationRoute.java
ssettipalli: refactored

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/service/participants/Osteo1ParticipantCreator.java
```diff
@@ -0,0 +1,557 @@
...
+                null,
+                null,
+                null
+        ));
+
+        jdbiActivityInstanceStatus.insert(activityDto.getId(), InstanceStatusType.COMPLETE, ddpCreatedAt + 2000,
+                activityDto.getParticipantId());
+    }
+
+    private UserRegistrationResponse registerGovernedUser(Handle handle, GovernedUserRegistrationPayload payload, LocalDate birthDate,
```
ssettipalli, line 488: Thought of moving this to utils too but didnt want to risk someone invoking it witout any Authorization check in here

zyme (approved):

PR #2841, opened 2024-04-01
[PEPPER-1353] Add temporary logging for missing ES data

denniscunningham:
## Context
PEPPER-1353

This PR contains temporary logging to help debug a problem we observed on Prod. Specifically, we are seeing DSM fail to get profile information from participant ES documents. Based in the logging timestamps the profile should be in place, so it is unclear if DSM is just failing to read the ES document, or if the timestamps are misleading.

Included in this PR:

1. Added logging to retrieve the ptp document in cases where DSM cannot get the ptp profile (RgpParticipantDataService).
2. Added logging to indicate when we are retrying the PARTICIPANT_REGISTERED handling (we already have this for UPDATE_CUSTOM_WORKFLOW handling). 
3. Moved `getParticipantDocument` methods from ElasticTestUtil to ElasticSearchService. Get a document by ID is a fundamental service that we have not provided for non-test cases.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
@@ -25,6 +26,8 @@ import org.broadinstitute.dsm.util.proxy.jackson.ObjectMapperSingleton;
 import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.action.delete.DeleteRequest;
 import org.elasticsearch.action.delete.DeleteResponse;
+import org.elasticsearch.action.get.GetRequest;
+import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.action.support.WriteRequest;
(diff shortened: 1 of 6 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participantdata/RgpParticipantDataService.java
```diff
@@ -240,4 +245,15 @@ private static String getPhoneFromActivities(List<Activities> activities) {
             return phoneAnswer.map(answer -> answer.get(DDPActivityConstants.ACTIVITY_QUESTION_ANSWER)).orElse("");
         }).orElse("");
     }
+
+    // TODO: remove this method after debugging is complete. -DC
+    private static void logParticipantDocument(String ddpParticipantId, String esIndex) {
+        Optional<String> ptpDoc =
+                ElasticSearchService.getParticipantDocumentAsString(ddpParticipantId, esIndex);
```
zyme, line 252: Any concerns about the size of this string, or whether the inability to hit ES to get the data may itself cause an exception that prevents existing log messages (like in L164)?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
```diff
@@ -176,6 +179,38 @@ public Dsm getRequiredDsmData(String ddpParticipantId, String index) {
                         ddpParticipantId, index)));
     }
 
+    /**
```
zyme, line 182: All moved from ElasticTestUtil with same signatures, exception for the `Optional<String>` coming back from getParticpantDocumentasString?

denniscunningham: Pretty much. `_getParticipantDocument` is pretty much the same with a small change.
The other two did not check the response value in ElasticTestUtil (assuming the document was always there).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participantdata/RgpParticipantDataService.java
denniscunningham: Not really. It's possible to IOExceptions from ES, but that would likely be true of the ES call that DSM just made to get the profile data. The case of a missing document is handled.

The string will easily fit in memory.
I've used a version of this is testing for a while without problems.

Worse case is this code throws an exception, which is handled but ends the retry sequence. Our ace is that we now have an AdminOperation to fix cases that go wrong, which we will use to resolve the existing Prod error cases.

Does that make sense?

denniscunningham (commented):
Thanks for your review @zyme.
Responded to your questions.

zyme (approved):
LGTM!  Thank you.

PR #2842, opened 2024-04-01
Pepper 1276 lms  notification fix

ssettipalli:
Fix LMS notification fix
self should show self message and NOT child notiifcation
Made a generic change so that we can run the generic patch to update only notification templates for any study.
Main issue was a bug in the patch where there are multiple events that fall under same category 
ex: activity-complete-notification


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/UpdateTemplatesInPlace.java
```diff
@@ -196,20 +224,33 @@ private void traverseEventConfigurations(Handle handle, long studyId) {
         }
     }
 
+
     // Best-effort attempt at making a unique identifier for event configuration.
     // watch out for scenarios where diff events might end up with same unique identifier
     // ex:- trigger and order are same but diff pre/cancel expressions !
     private String hashEvent(Config eventCfg) {
-        return String.format("%s-%d",
+        String key = String.format("%s-%d",
```
ssettipalli, line 228: Add additional parameters so that events can be made unique and not override others

pegahtah (approved):

PR #2843, opened 2024-04-02
[PEPPER-1141] Fixing SampleQueue missing in the download

pegahtah:
## Context

This PR is fixing the `SampleQueue` column (or Status column as it is called in the UI) not appearing in the download. After investigating the frontend, I have found out that this column is dynamically generated based on the kit's data and is not a real field in DB or ES. Therefore, in this PR the code is fixed to inject this value in the download and calculate it on the flight for each kit. 

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -56,19 +49,18 @@ public Object processRequest(Request request, Response response, String userId)
             response.status(403);
             return UserErrorMessages.NO_RIGHTS;
         }
-        Instant startInstant = Instant.now();
```
pegahtah, line 59: Not used

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -94,28 +86,4 @@ protected void setResponseHeaders(Response response, String filename) {
...
-            if (filteredSubset.getTotalCount() < currentTo) {
-                break;
-            }
-            currentFrom = currentTo;
-            currentTo += MAX_RESULT_SIZE;
-        }
-
-        return allResults;
-    }
-
```
pegahtah, line 120: Moved this to the `DownloadParticipantListService` so that it is easier to test and cleaner

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -301,4 +323,21 @@ public static void removeDsmField(String esIndex, String ddpParticipantId, Strin
...
+            Dsm dsm = gson.fromJson(dsmJson, Dsm.class);
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, Object> valueMap = mapper.convertValue(dsm, Map.class);
+            Map<String, Object> dsmMap = Map.of("dsm", valueMap);
+            ElasticSearchUtil.updateRequest(ddpParticipantId, esIndex, dsmMap);
+        } catch (Exception e) {
+            e.printStackTrace();
+            throw new RuntimeException(e);
+//            Assert.fail("Unexpected exception creating profile for participant " + ddpParticipantId);
+        }
```
pegahtah: Creates a participant with the specified data in the profile, so we can use filters to get exactly this participant in the tests.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1708,6 +1709,28 @@ public static String getCollaboratorSampleId(int kitTypeId, String participantCo
         }
     }
 
+    public static Object getSampleQueueValue(Map<String, Object> kit) {
```
denniscunningham: These methods probably belong in `DownloadParticipantListService` since they are only used there, and it seems unlikely they would be used by any kit shipping code. IOW, they really have to do with the download service -- the fact that the subject is a kit is unimportant.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1708,6 +1709,28 @@ public static String getCollaboratorSampleId(int kitTypeId, String participantCo
         }
     }
```
denniscunningham: This method needs JavaDoc since its purpose is unclear.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1708,6 +1709,28 @@ public static String getCollaboratorSampleId(int kitTypeId, String participantCo
...
 
+    public static Object getSampleQueueValue(Map<String, Object> kit) {
+        return isStringValuePresent(kit.get("externalOrderStatus")) ? kit.get("externalOrderStatus") :
+                isStringValuePresent(kit.get("deactivatedDate")) ? SampleQueue.DEACTIVATED.uiText :
```
denniscunningham: This code is hard to read and should be refactored (see SonarLint). In any case, there should be a comment explaining what's going on since that is unclear from reading the code.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/at/SampleQueue.java
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.dsm.model.at;
+
+public enum SampleQueue {
```
denniscunningham, line 7: A JavaDoc here would be helpful.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/at/SampleQueue.java
```diff
@@ -0,0 +1,16 @@
...
+
+    DEACTIVATED("Deactivated"),
```
denniscunningham, line 9: Do these values match the values in the UI?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
denniscunningham: Now that you have `DownloadParticipantListService` it seems odd to leave some of the logic here. 
Perhaps create another `DownloadParticipantListService` to encapsulate that logic?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,53 @@
...
+        int currentFrom = DEFAULT_FROM;
+        int currentTo = MAX_RESULT_SIZE;
+        while (true) {
+            // For each batch of results, add the DTOs to the allResults list
+            filter.setFrom(currentFrom);
+            filter.setTo(currentTo);
+            ParticipantWrapperResult filteredSubset = (ParticipantWrapperResult) filter.filter(queryParamsMap, new UnparsedDeserializer());
+            filteredSubset.getParticipants().forEach(participantWrapperDto -> {
+                        UnparsedESParticipantDto participantDto = (UnparsedESParticipantDto) participantWrapperDto.getEsData();
+                        HashMap<String, List> dsmData =  (HashMap<String, List>) participantDto.getDataAsMap().get(DSM);
```
denniscunningham: This should be `Map` instead of `HashMap`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,53 @@
...
+import org.broadinstitute.dsm.model.elastic.search.UnparsedDeserializer;
+import org.broadinstitute.dsm.model.elastic.search.UnparsedESParticipantDto;
+import org.broadinstitute.dsm.model.filter.Filterable;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperDto;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperResult;
+import spark.QueryParamsMap;
+
+public class DownloadParticipantListService {
+    /**
+     * Fetches participant information from ElasticSearch in batches of MAX_RESULT_SIZE
```
denniscunningham: This JavaDoc seems a bit off. 
As a public method, it seems to be explaining *how* it works, rather than explaining how to use it.
In general public methods are a bit opaque, so unless they have important side affects it is more important to explain the inputs and output.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,53 @@
...
+        List<ParticipantWrapperDto> allResults = new ArrayList<>();
+        int currentFrom = DEFAULT_FROM;
+        int currentTo = MAX_RESULT_SIZE;
+        while (true) {
+            // For each batch of results, add the DTOs to the allResults list
+            filter.setFrom(currentFrom);
+            filter.setTo(currentTo);
+            ParticipantWrapperResult filteredSubset = (ParticipantWrapperResult) filter.filter(queryParamsMap, new UnparsedDeserializer());
+            filteredSubset.getParticipants().forEach(participantWrapperDto -> {
+                        UnparsedESParticipantDto participantDto = (UnparsedESParticipantDto) participantWrapperDto.getEsData();
```
denniscunningham: Nit, but this code fragment is probably clearer as a separate method since it is doing special case work.
It also makes the loop easier to comprehend.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+            } else if (kitRequestShippingMap.get("ddpLabel").equals("newKit")) {
+                assertEquals("Waiting on GP", kitRequestShippingMap.get("sampleQueue"));
+            }else if (kitRequestShippingMap.get("ddpLabel").equals("shippedKit")) {
+                assertEquals("Shipped", kitRequestShippingMap.get("sampleQueue"));
+            }
+        });
+    }
+
+    private QueryParamsMap buildMockQueryParams(boolean onlyMostRecent, boolean humanReadable, String fileFormat) {
+        QueryParamsMap mainQueryParamsMap = mock(QueryParamsMap.class);
```
denniscunningham: Curious why you chose to mock rather than just create a `QueryParamsMap` directly?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+        when(mainQueryParamsMap.get("filters")).thenReturn(nestedFileFormatQueryParamsMap);
+        when(nestedFileFormatQueryParamsMap.value()).thenReturn("[]");
+
+        return mainQueryParamsMap;
+    }
+
+    private ManualFilterParticipantList getFilterFromFile(String fileName){
+        try {
+            String filterJson = TestUtil.readFile(fileName);
+            ManualFilterParticipantList filterable = new Gson().fromJson(filterJson, ManualFilterParticipantList.class);
```
denniscunningham: We are trying to use Jackson instead of Gson, so if that works here please replace.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -205,6 +205,28 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static Profile addParticipantProfile(String esIndex, String ddpParticipantId, String shortId, String firstName,
+                                               String lastName, String email, String fileName) {
+        Gson gson = new Gson();
+        try {
+            String profileJson = TestUtil.readFile(fileName);
+            profileJson.replace("<shortId>", shortId);
```
denniscunningham: Does this actually work? It seems like the result of `replace` is not used.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -205,6 +205,28 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static Profile addParticipantProfile(String esIndex, String ddpParticipantId, String shortId, String firstName,
```
denniscunningham: I suggest not passing the file since it is a template, but rather hardcoding it here.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -301,4 +323,21 @@ public static void removeDsmField(String esIndex, String ddpParticipantId, Strin
...
+        try {
+            String dsmJson = TestUtil.readFile(fileName);
+            Dsm dsm = gson.fromJson(dsmJson, Dsm.class);
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, Object> valueMap = mapper.convertValue(dsm, Map.class);
+            Map<String, Object> dsmMap = Map.of("dsm", valueMap);
+            ElasticSearchUtil.updateRequest(ddpParticipantId, esIndex, dsmMap);
+        } catch (Exception e) {
+            e.printStackTrace();
+            throw new RuntimeException(e);
```
denniscunningham: Avoid RuntimeException.

pepper-apis/dsm-core/src/test/resources/elastic/dsmWithKitRequestShipping.json
```diff
@@ -0,0 +1,74 @@
+{"kitRequestShipping" : [
```
denniscunningham: Nit re: filename.... This is not `dsmWith` but simply `kitRequestShipping`, right?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+                assertEquals("Deactivated", kitRequestShippingMap.get("sampleQueue"));
+            } else if (kitRequestShippingMap.get("ddpLabel").equals("receivedKit")) {
+                assertEquals("Received", kitRequestShippingMap.get("sampleQueue"));
+            } else if (kitRequestShippingMap.get("ddpLabel").equals("errorKit")) {
+                assertEquals("GP manual Label", kitRequestShippingMap.get("sampleQueue"));
+            } else if (kitRequestShippingMap.get("ddpLabel").equals("newKit")) {
+                assertEquals("Waiting on GP", kitRequestShippingMap.get("sampleQueue"));
+            }else if (kitRequestShippingMap.get("ddpLabel").equals("shippedKit")) {
+                assertEquals("Shipped", kitRequestShippingMap.get("sampleQueue"));
+            }
```
denniscunningham: Perhaps a final `else` to assert failure?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+        assertNotNull(downloadList);
+        assertEquals(1, downloadList.size());
+        List<Object> kitRequestShippings = ((List<Object>) ((Map<String, Object>)
+                ((Map<String, Object>) downloadList.get(0).getEsDataAsMap().get("dataAsMap")).get("dsm")).get("kitRequestShipping"));
+        assertNotNull(kitRequestShippings);
+        assertEquals(5, kitRequestShippings.size());
+        assertCorrectSampleQueueValue(kitRequestShippings);
+    }
+
+    private void assertCorrectSampleQueueValue(List<Object> kitRequestShippings) {
```
denniscunningham: Nit, the term `correct` here adds no value.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+    private static DDPInstanceDto ddpInstanceDto;
+    private static String guid = "PT_SAMPLE_QUEUE_TEST";
+    private static ParticipantDto participantDto = null;
+    private DownloadParticipantListService downloadParticipantListService = new DownloadParticipantListService();
+    static UserAdminTestUtil userAdminTestUtil = new UserAdminTestUtil();
+
+    @BeforeClass
+    public static void doFirst() {
+        esIndex = ElasticTestUtil.createIndex(instanceName, "elastic/lmsMappings.json", null);
+        userAdminTestUtil.createRealmAndStudyGroup(instanceName, instanceName, "downloadTest", groupName, esIndex);
```
denniscunningham: I don't think you need `UserAdminTestUtil`. Something like this should be sufficient:
`        ddpInstanceDto = DdpInstanceGroupTestUtil.createTestDdpInstance(instanceName, esIndex);`

And then in the teardown:
`        DdpInstanceGroupTestUtil.deleteInstance(ddpInstanceDto);`

denniscunningham (commented):
I left some comments.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/at/SampleQueue.java
pegahtah: Yes, screen shot from dev
Screenshot 2024-04-04 at 2 31 00 PM

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
pegahtah: great idea

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: GREAT CATCH!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: what's the appropriate exception we use in test?

pepper-apis/dsm-core/src/test/resources/elastic/dsmWithKitRequestShipping.json
pegahtah: This might be confusing now that I think about it, but since the json structure of this file has `kitRequestShipping[ ]` in another object, so this file should be uploaded as the participant's dsm, hence the dsm in the fileName

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
pegahtah: Jackson did not work here, it could not understand the abstract class hierarchies in the filters. (This could also be me having very limited experience with Jackson)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
pegahtah: The constructor had a protected access so I chose to mock this. If there is a better way please let me know

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/at/SampleQueue.java
denniscunningham: The comment above helped!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.broadinstitute.dsm.util.TestParticipantUtil;
+import org.broadinstitute.dsm.util.TestUtil;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import spark.QueryParamsMap;
+
+@Slf4j
+public class SampleQueueTest extends DbAndElasticBaseTest {
```
denniscunningham: This should probably be `DownloadParticipantListServiceTest`?

pepper-apis/dsm-core/src/test/resources/elastic/dsmParticipantProfile.txt
```diff
@@ -0,0 +1,13 @@
+{
```
denniscunningham: Regarding the file title....
`dsm` does not apply since `profile` is not part of the DSM object, and the term `template` probably best distinguishes this file. Also, this is a JSON file. So something like: `participantProfileTemplate.json` would work and fit with the other resources files in this directory.

pepper-apis/dsm-core/src/test/resources/elastic/dsmWithKitRequestShipping.json
denniscunningham: Perhaps get rid of `with` then, since there is no DSM object in the file, so `with` either adds no value or is confusing.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+import spark.Response;
+
+@Slf4j
+
+public class DownloadParticipantListService {
+    /**
+     * Fetches participant data from Elasticsearch based on the given filter and query parameters.
+     * returns a list of ParticipantWrapperDto objects
+     */
+    @VisibleForTesting
```
denniscunningham, line 40: Once you fix the test file name, this can be `protected`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+     *
+     * @param value The object to check for validity, which can be any type.
+     * @return {@code true} if the object is not null and not empty (for strings);
+     * {@code false} otherwise.
+     */
+    private static boolean isStringValuePresent(Object value) {
+        return value != null && !value.toString().trim().isEmpty();
+    }
+
+    public Object createParticipantDownloadZip(Filterable filterable, DownloadParticipantListParams params, TabularParticipantParser parser,
```
denniscunningham: This could use a JavaDoc, since it is public, and the role of `Reponse`, at least, is unclear.
JavaDoc on the private methods is less important.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+import org.broadinstitute.dsm.model.filter.FilterFactory;
+import org.broadinstitute.dsm.model.filter.Filterable;
+import org.broadinstitute.dsm.model.participant.DownloadParticipantListParams;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperDto;
+import org.broadinstitute.dsm.model.participant.ParticipantWrapperResult;
+import spark.QueryParamsMap;
+import spark.Response;
+
+@Slf4j
+
```
denniscunningham: Unnecessary blank line.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+                break;
+            }
+            currentFrom = currentTo;
+            currentTo += MAX_RESULT_SIZE;
+        }
+
+        return allResults;
+    }
+
+    private void setSampleQueue(ParticipantWrapperDto participantWrapperDto){
```
denniscunningham: All of the methods in this class can be static. 
It is fine if the public methods are not static (design choice), but the private methods should be static to illustrate they are not depending on class state.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import com.netflix.servo.util.VisibleForTesting;
+import lombok.extern.slf4j.Slf4j;
+import org.broadinstitute.dsm.db.KitRequestShipping;
```
denniscunningham: This include should be no longer necessary.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+    public List<ParticipantWrapperDto> fetchParticipantEsData(Filterable filter, QueryParamsMap queryParamsMap) {
+        List<ParticipantWrapperDto> allResults = new ArrayList<>();
+        int currentFrom = DEFAULT_FROM;
+        int currentTo = MAX_RESULT_SIZE;
+        while (true) {
+            // For each batch of results, add the DTOs to the allResults list
+            filter.setFrom(currentFrom);
+            filter.setTo(currentTo);
+            ParticipantWrapperResult filteredSubset = (ParticipantWrapperResult) filter.filter(queryParamsMap, new UnparsedDeserializer());
+            filteredSubset.getParticipants().forEach(participantWrapperDto -> setSampleQueue(participantWrapperDto));
```
denniscunningham, line 50: You can probably use a lambda expression here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
```diff
@@ -0,0 +1,151 @@
...
+            currentFrom = currentTo;
+            currentTo += MAX_RESULT_SIZE;
+        }
+
+        return allResults;
+    }
+
+    private void setSampleQueue(ParticipantWrapperDto participantWrapperDto){
+        UnparsedESParticipantDto participantDto = (UnparsedESParticipantDto) participantWrapperDto.getEsData();
+        Map<String, List> dsmData =  (HashMap<String, List>) participantDto.getDataAsMap().get(DSM);
```
denniscunningham: `HashMap` here is incorrect.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -205,6 +205,28 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static Profile addParticipantProfile(String esIndex, String ddpParticipantId, String shortId, String firstName,
+                                               String lastName, String email) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            String profileJson = TestUtil.readFile("elastic/participantProfile.json");
```
denniscunningham: It feels like this still cannot be working, since this is not the correct file.
I'm concerned that this profile is either unnecessary, or that your tests are not correct, if this error and the prior error can go unnoticed.

Perhaps you can get rid of this method and instead use `TestParticipantUtil.createParticipantWithEsProfile`?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
denniscunningham: If you look at the rest of the methods in the file, we assert failure.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -2,44 +2,26 @@
...
-import java.util.stream.Collectors;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipOutputStream;
-
-import static org.broadinstitute.dsm.util.ElasticSearchUtil.DEFAULT_FROM;
-import static org.broadinstitute.dsm.util.ElasticSearchUtil.MAX_RESULT_SIZE;
-
 public class DownloadParticipantListRoute extends RequestHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(DownloadParticipantListRoute.class);
```
denniscunningham: Logger is no longer used.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -301,4 +323,21 @@ public static void removeDsmField(String esIndex, String ddpParticipantId, Strin
         UpsertPainless upsert = new UpsertPainless(esIndex, QueryBuilders.termsQuery("_id", ddpParticipantId));
         upsert.export(script, Map.of(), field);
     }
+
+    public static void addParticipantDsmFromFile(String esIndex, String fileName, String ddpParticipantId) {
```
denniscunningham: Looks like a comment from my prior review got lost, but...
This should be combined with `addDsmEntityFromFile`, rather than creating a new method that does a similar thing differently.
If it is unclear how to do that, I can suggest a way.

denniscunningham (commented):
A few more comments.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/SampleQueueTest.java
pegahtah: Well... this is not really testing the download's correctness, only the sampleQueue field, but I think you are right, the class could be called `DownloadParticipantListServiceTest` and the unit test reflect that the test is specifically checking sampleQueue value.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: Thank you for finding the problem with this file, I needed to investigate a little more to find out why it was still working as I wanted and realized my filter file was missing filters, so I was getting all in the ES and not the participant I wanted. I have made the changes for this to work as intended

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: I changed it but I think when we assert failure, we lose some part of the logs of what caused the exception.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
pegahtah: No, I tried it but because setSampleQueue is void and is not returning anything, java didn't like using lambdas

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
denniscunningham: Not sure why you resolved this?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
denniscunningham: The stack trace above provides that.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
denniscunningham: Unclear why this is now resolved.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -205,6 +205,26 @@ public static Profile addParticipantProfileFromFile(String esIndex, String fileN
         }
     }
 
+    public static void addParticipantProfileFromTemplate(String esIndex, String ddpParticipantId, String shortId, String firstName,
```
denniscunningham, line 213: Given my prior comments, I am not sure why you kept this method.
Can you explain?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/download/DownloadParticipantListServiceTest.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private ManualFilterParticipantList getFilterFromFile(String fileName){
+        try {
+            String filterJson = TestUtil.readFile(fileName);
+            filterJson = filterJson.replace(":shortId", shortId);
+            ManualFilterParticipantList filterable = new Gson().fromJson(filterJson, ManualFilterParticipantList.class);
+            return filterable;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
denniscunningham: Assert.fail or a more specific exception with something above catching it would work better.

denniscunningham (commented):
A few more questions and suggestions.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: Of course!
I understand your concerns regarding the method in question. My interpretation was that the primary issue was with the initial implementation of the method and its inability to perform as intended. With your input, I've been able to revise the method to address these concerns.

I believe this method remains valuable for dynamically creating profiles with specific values for testing, which would mean having a more flexible and maintainable approach than creating multiple JSON files for each test scenario. This method allows us to easily adjust test parameters and conditions without the overhead of managing numerous files.

However, I'm open to further discussions or suggestions if you believe there's a better approach or if my understanding of the method's utility differs from its intended purpose. Your insights are greatly appreciated.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
pegahtah: This was resolved by mistake and my reply to it is lost! 
I meant to write that this can not be protected because the test class needs access to it and they are not in the same package. Please let me know if I am missing something, I would appreciate it

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: This method is removed and I am using `ElasticTestUtil.addDsmEntityFromFile(esIndex, "elastic/dsmKitRequestShipping.json", ddpParticipantId, "1990-10-10", null);` I think this addresses your concern but please let me know if I resolved it by mistake

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/download/DownloadParticipantListService.java
denniscunningham: I see.
Why are they not in the same package?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
denniscunningham: Yup, I understand.
My concern was that the flexibility this provides was not actually in use. In general we try not to create things that might be useful in the future but are not actually in use now (YAGNI).

denniscunningham (approved):
Looks good!
Just one change to consider re: package names.

PR #2844, opened 2024-04-03
[PEPPER-1335] Filter instances in MedicalRecordInstanceProvider

denniscunningham:
## Context
PEPPER-1335

This PR fixes a bug where DSM considers too many instances when requesting institutions from DSS. See `MedicalRecordInstanceProvider`. The other changes are just logging improvements.

Some context... 

Much of the code in this area is new, adapted from prior code, intended to include support creation of OS1 participants. The prior code was untestable, thus the new code. Having said that, while most of the new code is under test, there is a small portion that is difficult to test with JUnit (not impossible, just a lot of setup that would more or less contrive a situation that code would handle anyway). This change falls within that small untested area.

Note that we already fixed a bug due to the lack of uniformity in `ddp_instance_role` in each environment, so even if this code was correct, there are still factors that cause bugs in practice. Which is a long way of saying we need to verify this behavior manually/visually in Test to feel good about releasing to Prod.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/medicalrecord/MedicalRecordInstanceProvider.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/medicalrecord/MedicalRecordInstanceProvider.java
@@ -19,7 +19,8 @@ public class MedicalRecordInstanceProvider implements DDPInstanceProvider {
      */
     @Override
     public List<DDPInstance> getApplicableInstances() {
-        return DDPInstance.getDDPInstanceListWithRole(DBConstants.HAS_MEDICAL_RECORD_ENDPOINTS);
+        return DDPInstance.getDDPInstanceListWithRole(DBConstants.HAS_MEDICAL_RECORD_ENDPOINTS).stream()
+                .filter(DDPInstance::isHasRole).toList();
     }
 
     /**
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -335,6 +335,11 @@ public class DDPInstance {
         return (String) resultsGroup.resultValue;
     }
 
+    /**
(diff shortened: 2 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/medicalrecord/MedicalRecordInstanceProvider.java
```diff
@@ -19,7 +19,8 @@ public class MedicalRecordInstanceProvider implements DDPInstanceProvider {
      */
     @Override
     public List<DDPInstance> getApplicableInstances() {
-        return DDPInstance.getDDPInstanceListWithRole(DBConstants.HAS_MEDICAL_RECORD_ENDPOINTS);
+        return DDPInstance.getDDPInstanceListWithRole(DBConstants.HAS_MEDICAL_RECORD_ENDPOINTS).stream()
```
zyme, line 22: I think we need a method comment on getDDPInstanceListWithRole, and or a javadoc comment on getApplicableInstances().  I read getDDPInstanceListWithRole and think "this returns ddp instances that have the given role".  But that appears to not be the case, since this additional filter must be applied.

zyme: It looks like other uses of getDDPInstanceListWithRole() (outside of test code) do this addition hasRole() on the client side, so presumably they are doing the "right" thing.

zyme (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/medicalrecord/MedicalRecordInstanceProvider.java
denniscunningham: Added a comment.
Yes, I agree. There is no good reason this method should not return a filtered list.

denniscunningham (commented):
Thanks for you review!
Added JavaDoc for getDDPInstanceListWithRole.

zyme (approved):

PR #2846, opened 2024-04-06
[PEPPER-1390] Improve handling of missing ES participant document

denniscunningham:
## Context
PEPPER-1390

The  essential issue for the problems described in PEPPER-1390 is elsewhere, but this PR fixes an issue with how WorkflowStatusUpdate was handling ESMissingParticipantDataException. Along the way I made the following improvements:

1.  Added ElasticSearchService.getParticipantDocumentById (and related) to replace use of ElasticSearchUtil.getParticipantESDataByParticipantId. The latter did not provide a way for callers to detect a missing participant document, and was overly complex. Also added a corresponding test in ElasticSearchServiceTest.
2. Refactored BasicDefaultDataMaker.SetDefaultData. Since this method is used differently now it could be simplified. Also now checking for a valid ES participant document here.
3. Added logging for the retry exhausted case in DMtasksSubscription.generateStudyDefaultValues.
4. Removed temporary logging in RgpParticipantDataService since we now know what the underlying problem is.
5. Rewrote and simplified RgpParticipantDataService.insertEsFamilyId, since it was not handling missing participant documents correctly. Updated the corresponding test, and added support in ElasticSearchService.
6. Stopped handling participants with legacy PIDs in ParticipantInitService. These participants do not need initialization, and the current code does not handle them correctly. Added a test  for that. 
7. The rest of the changes are converting from ElasticSearchUtil.getParticipantESDataByParticipantId to ElasticSearchService.getParticipantDocumentById (and related) in files I recently created. 

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
@@ -79,8 +79,8 @@ public class ElasticSearchService {
     /**
      * Get a single participant document based on participant GUID. Throw an exception if document is not found.
      */
-    public ElasticSearchParticipantDto getRequiredParticipantDocumentById(String ddpParticipantId, String index) {
-        return getParticipantDocumentById(ddpParticipantId, index)
+    public ElasticSearchParticipantDto getRequiredParticipantDocument(String ddpParticipantId, String index) {
+        return getParticipantDocument(ddpParticipantId, index)
                 .orElseThrow(() -> new ESMissingParticipantDataException("Participant document %s not found in index %s"
                         .formatted(ddpParticipantId, index)));
     }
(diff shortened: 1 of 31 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
```diff
@@ -74,6 +76,78 @@ public List<String> getAllParticipantGuids(String esParticipantsIndex) {
         return participantIds;
     }
 
+    /**
+     * Get a single participant document based on participant GUID. Throw an exception if document is not found.
+     */
+    public ElasticSearchParticipantDto getRequiredParticipantDocumentById(String ddpParticipantId, String index) {
```
pegahtah: Just a suggestion: Do we need both of the words `get` and `required` in the name? If the only difference between this and `getParticipantDocumentById` is the failure when the participant doesn't exist, can we name it something like `getParticipantDocumentByIdOrThrow` or `requireParticipantDocumentById` maybe?

pegahtah (approved):
Thx

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/elastic/ElasticSearchService.java
denniscunningham: Ah, good catch. I added `byId` to disambiguate with another method in this class, but the `byId` methods are now much more widely used. The `byId` adds no value, so I changed the conflicting method name, and got rid of the `byId` part. (The method I wanted to disambiguate is not really an overload of these, so I was reluctant to use the same name for both.)

As for your suggestions, `get` is important and expected here, so I cannot get rid of that. For `orThrow` vs `required`... I've been using `required` for methods like this and phasing out `orThrow`, since the former better describes the intent. That is, I expect the value to be there, so raise and exception if not. The `orThrow`  term is ambiguous (throw for what conditions?), and has a particular and somewhat different meaning when used with `Optional`, so that can be confusing.

Given that, it seems like the term `required` did not resonate with you. Does a better term come to mind?

denniscunningham (commented):
@pegahtah Thanks for your review!
I left one question.

PR #2847, opened 2024-04-09
PEPPER-1420 removing openapi docs

zyme:
PEPPER-1420 Since nobody is using the [pepper openAPI docs](https://pepper.datadonationplatform.org/docs), so we are going to stop generating them in our builds and stop hosting them in GAE.

The changes are limited to our circleci config.  Once these changes get up to `develop`, we can verify circleCI behavior and then remove the `pepper-api-spec` appengine service in each environment as the RC takes shape and moves to dev, test, staging, and prod.

We'll leave most of the code in place to generate the api docs so if we change our mind, it'll be easy to re-enable the docs.  These changes just eliminate the build from generating and deploying the docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ parameters:
     default: false
   project:
     type: enum
-    enum: ["dss", "dsm", "nothing"]
+    enum: ["dss", "dsm","api-docs", "nothing"]
     default: "nothing"
   action:
     type: enum
@@ -28,7 +28,7 @@ workflows:
     jobs:
       - continuation/continue:
           configuration_path: .circleci/main-config.yml
-          parameters: '{"action":"build-test-deploy","dss": true}'
+          parameters: '{"action":"build-test-deploy","dss": true,"api-docs": false}'
 
   build-test-workflow:
     description: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -79,7 +81,7 @@ workflows:
       # Only DSS projects and API docs affected!
       - continuation/continue:
           configuration_path: .circleci/main-config.yml
-          parameters: '{"action":"build-test-store","dss": true,"api-docs": true}'
```
zyme, line 82: Primary change: set `api-docs` to `false` everywhere.

.circleci/main-config.yml
```diff
@@ -1180,7 +1180,8 @@ workflows:
       - dss-parallel-compile-and-test-job:
           parallelism: << pipeline.parameters.test_parallelism >> # the caller of this workflow will tell us the parallelism
       - dsm-compile-and-run-tests-job
-      - build-api-docs-job
```
zyme, line 1183: In addition, remove the call to run the docs job, but leave it commented out so it's straightforward to turn it back on if we so desire.

denniscunningham (approved):

PR #2848, opened 2024-04-10
PEPPER-1379 remove tcell clutter

zyme:
Although we disabled tcell in sprint 20 as part of PEPPER-1379, there's a bunch of clutter in deployment configuration that mentions tcell.  This PR removes the clutter.


pepper-apis/config/manifest.rb
```diff
@@ -15,5 +15,4 @@
 copy_secret_from_path "secret/pepper/#{$env}/pepper-cert.crt"
 render "ddp.conf.ctmpl"
 copy_file "post_deploy_smoketest.sh"
-copy_file "waf.conf"
```
zyme, line 18: There are probably other things in `manifest.rb` that are a distraction, but this PR is just focused on tcell.

pepper-apis/config/tcell_agent.config.ctmpl
```diff
@@ -1,26 +0,0 @@
-
-{{with $environment := env "ENVIRONMENT"}}
```
zyme, line 2: Don't need a config file for tcell since we aren't using tcell.

pepper-apis/config/DataDonationPlatform.yaml.ctmpl
```diff
@@ -36,7 +36,5 @@ entrypoint: java -Xmx1640m -Dio.grpc.netty.shaded.io.netty.transport.noNative=fa
 # GAE does not like the "=" character in entrypoint and does not support "." in environment variables
 env_variables:
   config_file: application.conf
-  TCELL_AGENT_HOST_IDENTIFIER: study-server
```
zyme, line 39: Don't need these env vars to drive tcell since we aren't using tcell.

PR #2851, opened 2024-04-16
[DSM] PEPPER-1244

pegahtah:
## Context

This PR fixes the Clinical  Orders download by changing the export config checks. The participant column class in the frontend for Clinical Orders is a special case, because it is a DSM field but has an `object` field. (it is `dsm.object`). My first attempt was to change this through frontend to have the column match other fields but it broke searching and the impact was big, so I switched my changes to the backend and added a small check

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
```diff
@@ -40,13 +40,13 @@ public class TabularParticipantParser {
     private final List<String> nestedArrayObjects = Arrays.asList(ESObjectConstants.KIT_TEST_RESULT);
     private Map<String, Map<String, Object>> activityDefs;
 
-    public TabularParticipantParser(List<Filter> filters, DDPInstance ddpInstance, boolean humanReadable, boolean onlyMostRecent,
```
pegahtah, line 43: I think it's nice if we start moving away from DDPInstance and just use the Dto

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -92,7 +91,8 @@ public static Alias of(SortBy sortBy) {
 
     public static Alias of(ParticipantColumn column) {
         Alias esAlias;
-        if (Objects.nonNull(column.getObject()) && Alias.ofOrNull(column.getObject()) != null) {
+        if (Objects.nonNull(column.getObject()) && Alias.ofOrNull(column.getObject()) != null
+                && !Alias.CL.name().equalsIgnoreCase(column.getTableAlias())) {
```
pegahtah: Clinical Order is an oddly shaped column, it has an "object" field but it is not a DSS field. I tried changing the column but it broke the search, so I thought this is the fix with the smallest impact

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -106,14 +106,8 @@ public static Alias of(String alias) {
         return Enums.getIfPresent(Alias.class, alias.toUpperCase()).or(ACTIVITIES);
     }
 
-    public static Alias aliasByValue(String value) {
-        return Arrays.stream(Alias.values())
-                .filter(alias -> alias.value.equals(value))
-                .findFirst()
-                .orElse(ACTIVITIES);
-    }
-
```
pegahtah, line 115: not used and not likely to get used

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/download/DownloadParticipantListServiceTest.java
```diff
@@ -30,24 +38,24 @@
...
-    private static final String shortId = "PT_SHORT";
 
     @BeforeClass
     public static void doFirst() {
         esIndex = ElasticTestUtil.createIndex(instanceName, "elastic/lmsMappings.json", null);
         ddpInstanceDto = DdpInstanceGroupTestUtil.createTestDdpInstance(instanceName, esIndex);
-        String ddpParticipantId = TestParticipantUtil.genDDPParticipantId(guid);
-        participantDto = TestParticipantUtil.createParticipant(ddpParticipantId, ddpInstanceDto.getDdpInstanceId());
+        guid = TestParticipantUtil.genDDPParticipantId(guid);
+        participantDto = TestParticipantUtil.createParticipant(guid, ddpInstanceDto.getDdpInstanceId());
```
denniscunningham: This seems more confusing than the code it replaces, particularly since `ddpParticipantId` is more descriptive, is used widely in the code, and is the name of the corresponding parameters in `ElasticTestUtil`.
IOW, it feels like there is a one line change here that would achieve your purpose?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -261,6 +261,7 @@ public static Dsm addDsmEntityFromFile(String esIndex, String fileName, String d
             } else {
                 json = json.replace("\"dateOfMajority\" : \"<dateOfMajority>\",", "");
             }
+            json = json.replace(":guid", ddpParticipantId);
```
denniscunningham: Good change, but perhaps a better one would be to refactor this method to be more like similar methods in this file (addParticipantAddressFromFile, addParticipantProfileFromFile)?
In those cases we first deserialize the file contents, then do the assignment. Feels cleaner and clearer than this approach.
(BTW, Git says I am the author of this method, but I did not add the replacement technique.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/tabular/TabularParticipantParser.java
denniscunningham: I understand why you made this change, since `DownloadParticipantListServiceTest ` uses DDPInstanceDto. But I suggest not making further arbitrary changes to replace DDPInstance with DDPInstanceDto, as your comment suggests.

DDPInstanceDto and DDPInstance proliferate the code, and are used interchangeably. I find it mildly frustrating, but have come to peace with things as is, which is often the case with existing Pepper code.
I don't think making an effort to eliminate one or the other makes much difference in code quality or readability. I feel like our efforts are better spent on changes required for code quality improvements, particularly to enable testing.

IOW, not that this particular instance is such a big deal, but making arbitrary changes to remedy historical anomalies in the code, however well intentioned, just muddies code reviews, etc. without much benefit, particularly since there is *so much* we could do. Just an opinion -- happy to discuss.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
denniscunningham: Not sure I understand, but since 'object' is such a generic term, a code comment here explaining what's going on would be helpful, particularly since this seems to be the essence of the fix.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/download/DownloadParticipantListServiceTest.java
```diff
@@ -58,45 +66,84 @@ public static void cleanUp() {
...
         queryParamsMap = buildMockQueryParams(false, false, "tsv");
         downloadList = DownloadParticipantListService.fetchParticipantEsData(filterable, queryParamsMap);
-        assertDownloadList(downloadList);
+        assertKitRequestShipping(downloadList);
+    }
+
+    @Test
+    public void testClinicalOrdersDownloadFromElastic() {
+        ManualFilterParticipantList filterable = getFilterFromFile("elastic/filtersWithClinicalOrderColumns.json");
+        List<Filter> columnNames = getColumnNames("elastic/filtersWithClinicalOrderColumns.json");
```
denniscunningham, line 109: This reads the same file twice and gets different information?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/download/DownloadParticipantListServiceTest.java
```diff
@@ -58,45 +66,84 @@ public static void cleanUp() {
...
+        assertNotNull(participantExporter);
+        assertNotNull(participantExporter.getExportFilename());
+        assertParticipantExporterMap(participantExporter);
+
+    }
+
+    private void assertParticipantExporterMap(TabularParticipantExporter participantExporter) {
+        assertEquals(1, participantExporter.participantValueMaps.size());
+        Map<String, String> participantValues = participantExporter.participantValueMaps.get(0);
+        assertNotNull(participantValues);
```
denniscunningham, line 136: Perhaps add a comment that these values match those in the test fixture file?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/DownloadParticipantListRoute.java
```diff
@@ -33,7 +37,10 @@ public Object processRequest(Request request, Response response, String userId)
         }
         setResponseHeaders(response, realm + "_export.zip");
         Filterable filterable = FilterFactory.of(request);
-        return DownloadParticipantListService.createParticipantDownloadZip(filterable, params, realm, request.queryMap(), payload, response);
+        DDPInstanceDto ddpInstanceDto = ddpInstanceDao.getDDPInstanceByInstanceName(realm)
```
denniscunningham: You might want to do this before the call to `checkUserAccess`, since it uses realm in a query. This is standard input validation, so should happen right at the beginning of the method.

denniscunningham (commented):
Left some comments.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/download/DownloadParticipantListServiceTest.java
pegahtah: Yes because one extracts the filters and one the display columns and in the main code both of them come from the request (which is in the same shape of the contents of that file)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: I agree, I made the changes, and it seems like that change is not necessary, there is no ddpParticipantId in the DSM class

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
pegahtah: This method checks the display columns to determine the correct location of the data required for downloads. Each participant column contains fields that specify where the data is stored:

tableAlias: Indicates which DSM database table holds the data.
object: Points to the specific object in the JSON document where this data is located (e.g., profile, files, or dsm).
For DSS-related fields, such as date of birth, emails, and survey answers, the table alias is typically data, and the object specifies which part of the JSON document contains this data.

An exception is the Clinical Order columns. Although these also have an object field, their table alias `cl` is the identifier that should be used. The reasons behind this design choice are not fully explained, but any changes to this structure will break the filtering processes and can also have unforeseen impacts.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
```diff
@@ -110,4 +109,18 @@ private static Alias ofOrNull(String alias) {
         return Enums.getIfPresent(Alias.class, alias.toUpperCase()).orNull();
     }
 
+
+    /**
+     * checks if the column has an object field that can be used to determine the alias, which is true for all the columns with an
```
denniscunningham, line 114: Nice!

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
```diff
@@ -247,22 +248,21 @@ public static Address addParticipantAddressFromFile(String esIndex, String fileN
     /**
      * Add a DSM entity to the participant doc
      *
-     * @param dob date of birth to replace in DSM entity
+     * @param dob            date of birth to replace in DSM entity
```
denniscunningham, line 251: Does something in you IDE make these formatting changes, or is it something you do by hand?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/sort/Alias.java
denniscunningham: Thx for adding the comment and breaking out the method. Very helpful.
This ^^ explanation even better!
You should feel free to add longer explanations like this to the code. Someone will likely benefit from it.

denniscunningham (approved):

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/util/ElasticTestUtil.java
pegahtah: This is all checkstyle

denniscunningham: Checkstyle is doing the reformatting automatically, or complaining about the indenting?
I don't see that in my IDE, and usually it shows checkstyle violations.

image

PR #2854, opened 2024-04-18
PEPPER-1392 and PEPPER-1412 app engine ah start and GAE instance logging

zyme:
This started with just PEPPER-1392, getting DSS, DSM, and Housekeeping responding to `_ah/start` routes quickly on boot, but I also added in more GAE instance details in the logs as per PEPPER-1412.

Start with [SparkBootUtil](pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java), a class that centralizes much of the shared, copy-pasted spark bootup process.  While making this class, I found a number of 100% copy-pasted classes in the dsm and dss directories, and moved them into ddp-common, so you'll see what looks like file deletes but are actually just moves.  I've called them out inline.

Now DSS, DSM, and Housekeeping use a lot more of the same code directly when it comes to starting up Spark.  Left for later is how each app will respond to stop routes (PEPPER-1409).

Things were way more complicated than expected not because of the `_ah/start` route, but because of our circleCI test infrastructure.  Particularly time consuming was debugging flakiness resulting from not waiting properly for IntegrationTestUtil to finish starting.  See [here](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1571285695) and [here](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1571285356) for how this has been addressed.  

Prior to these changes, all apps waited to boot spark routes until after everything (liquibase, caching, etc.) had been started, so using `nc` to ping a port worked reasonably well (although there were some "just for good measure" sleeps thrown in that masked yet more interesting timing issues).  While testing these changes, it eventually became clear that  parallel tests in circleCI were starting prematurely because `IntegrationTestSuite` was not being given enough time to do more time-intensive operations like running liquibase.  It dawned on me that this was the root cause of the instability, but it sure took a while.

## FUD Score

## Release


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/logging/LogUtil.java
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.ddp.logging;
+
+import org.slf4j.MDC;
+
+public class LogUtil {
```
zyme, line 5: Moved from here.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/logging/LogUtil.java
```diff
@@ -1,5 +0,0 @@
-package org.broadinstitute.ddp.logging;
-
```
zyme, line 2: Moved to [here](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1570786921) so that DSM and DSS can use a single class for this.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/jetty/JettyConfig.java
```diff
@@ -1,153 +0,0 @@
...
-import spark.embeddedserver.EmbeddedServerFactory;
-import spark.embeddedserver.EmbeddedServers;
-import spark.embeddedserver.jetty.EmbeddedJettyServer;
-import spark.embeddedserver.jetty.HttpRequestWrapper;
-import spark.embeddedserver.jetty.JettyServerFactory;
-import spark.http.matching.MatcherFilter;
-import spark.route.Routes;
-import spark.staticfiles.StaticFilesConfiguration;
-
-public class JettyConfig {
```
zyme, line 27: Moved to ddp-common so to eliminate copy-pasted source.  Now both DSS and DSM can use the same class.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/logging/LogUtil.java
zyme, line 5: Moved from [dss-core](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1570801631) so that both DSM and DSS can use the same class.  Added values from PEPPER-1412.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/logging/LogUtil.java
```diff
@@ -1,5 +0,0 @@
...
-public class LogUtil {
```
zyme, line 3: Moved to [ddp-common](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1570800530) so both DSM and DSS can use the same class.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/ConfigManager.java
```diff
@@ -173,7 +173,7 @@ public Config getConfig() {
         for (Map.Entry<String, String> override : overrides.entrySet()) {
             ConfigValue overrideValue = ConfigValueFactory.fromAnyRef(override.getValue(), "Overriding via " + this.getClass().getName());
             cfgWithOverride = cfgWithOverride.withValue(override.getKey(), overrideValue);
-            log.info("Overriding config value for key {}", override.getKey());
+            log.info("Overriding config value for key {} to {}", override.getKey(), override.getValue());
```
zyme, line 176: Including the value proved helpful while reproducing circleCI issues locally, so I'd like to leave this in for everyone.

.circleci/main-config.yml
```diff
@@ -700,11 +701,11 @@ commands:
           command: |
             for i in `seq 1 << parameters.timeout >>`;
             do
-              nc -z << parameters.server_address >> << parameters.port >> && echo Success && exit 0
+              nc -z << parameters.server_address >> << parameters.port >> && echo Success connecting to << parameters.server_address >> << parameters.port >> after  [$i] cycles && exit 0
```
zyme: Including the value proved helpful while reproducing circleCI issues locally, so I'd like to leave this in for everyone.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -226,10 +222,6 @@ public class DataDonationPlatform {
     private static final Map<String, String> pathToClass = new HashMap<>();
     private static Scheduler scheduler = null;
 
-    private static final AtomicBoolean isReady = new AtomicBoolean(false);
```
zyme, line 229: No longer doing the complicated block/wait/timeout approach for boots.  We just `start()` and it does all the work.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -254,16 +255,10 @@ public static void main(String[] args) {
                 }
 
                 new DSMConfig(cfg);
-
```
zyme, line 257: Moved much of this to central SparkBootUtil

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java
```diff
@@ -260,65 +252,50 @@ public static void shutdown(int millisecs) {
...
         Config cfg = ConfigManager.getInstance().getConfig();
         int maxConnections = cfg.getInt(ConfigFile.NUM_POOLED_CONNECTIONS);
 
-        int requestThreadTimeout = cfg.getInt(ConfigFile.THREAD_TIMEOUT);
         String healthcheckPassword = cfg.getString(ConfigFile.HEALTHCHECK_PASSWORD);
         String sendGridEventsVerificationKey = cfg.hasPath(EVENTS_VERIFICATION_KEY)
                 ? cfg.getString(EVENTS_VERIFICATION_KEY) : null;
         String auth0LogEventsToken = cfg.hasPath(AUTH0_LOG_EVENTS_TOKEN) ? cfg.getString(AUTH0_LOG_EVENTS_TOKEN) : null;
 
-        // app engine's port env var wins
```
zyme, line 284: Moved much of this to SparkBootUtil

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -284,11 +286,6 @@ public static void start(String[] args, SendGridSupplier sendGridSupplier) {
         heartbeatMonitor = new StackdriverMetricsTracker(StackdriverCustomMetric.HOUSEKEEPING_CYCLES,
                 PointsReducerFactory.buildMaxPointReducer());
 
-        String envPort = System.getenv(ENV_PORT);
```
zyme, line 287: Moved much of this to SparkBootUtil

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -273,66 +272,32 @@ public static void tearDown() {
...
-        boolean spawnProcess = cfg.getBoolean(ConfigFile.BOOT_TEST_APP_IN_SEPARATE_PROCESS);
-
         Map<String, String> serverArgs = new HashMap<>();
         serverArgs.put("config.file", RouteTestUtil.getConfigFilePath());
         serverArgs.put("cachingDisabled", isCacheDisabled + "");
 
-        int bootTimeoutSeconds = 30;
         log.info("Starting app on port {}", port);
 
-        if (spawnProcess) {
```
zyme, line 293: Seems like nobody is using the separate process stuff so I'm removing it to simplify the code.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
```diff
@@ -114,7 +115,8 @@ private TransactionWrapper(int maxConnections, String dbUrl, DB db) {
      * Should only be called during testing.
      */
     public static synchronized void reset() {
-        log.warn("reset() should only be called in the context of testing");
+        log.warn("reset() should only be called in the context of testing.  Call stack follows");
+        log.warn(Arrays.toString(Thread.currentThread().getStackTrace()).replace(',', '\n'));
```
zyme: I added a stack trace here when trying to reproduce some circleCI issues locally.  Thoughts on keeping it around?

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -273,66 +272,32 @@ public static void tearDown() {
...
-            } catch (IOException e) {
-                log.error("App starter failed.", e);
-            }
-        } else {
-            runDdpServer(isCacheDisabled);
-        }
-
-    }
-
-    private static boolean isDebugEnabled() {
```
zyme, line 312: Was only used by `bootTestAppInSeparateProcess`-related code, so I've removed it.

.circleci/main-config.yml
```diff
@@ -699,12 +700,13 @@ commands:
           name: Waiting for server to start
           command: |
             for i in `seq 1 << parameters.timeout >>`;
+            # todo arz is this responding prematurely, leading to tests to start before server is truly ready?
             do
-              nc -z << parameters.server_address >> << parameters.port >> && echo Success && exit 0
+              nc -z -v << parameters.server_address >> << parameters.port >> && echo Success connecting to << parameters.server_address >> << parameters.port >> after  [$i] cycles && exit 0
```
zyme, line 703: Added a bit more verbose logging for each attempt to connect

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Called when the {@link org.broadinstitute.ddp.constants.RouteConstants.GAE#STOP_ENDPOINT}
+     * is called by the app engine dispatcher
+     */
+    @FunctionalInterface
+    public interface StopRouteCallback {
+
+        void onStop();
```
zyme, line 86: Leaving a stub here for now.  There are other tickets for handling shutdown requests, which need different things across DSM, DSS, and Housekeeping.

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -396,11 +362,47 @@ private static boolean isTestServerRunning() {
...
+    }
+
+    /**
+     * Server socket that allows one connection
+     * and then closes
+     */
+    public static class SingleUseServerSocket {
+
+        public SingleUseServerSocket(int port) {
+            ServerSocket serverSocket = null;
```
zyme, line 393: This appears a bit clunky at first--why not just use another spark port?  The answer is that making multiple spark services on different ports in the same VM requires more extensive changes, and also requires all future new routes to reference a particular spark service instead of just `get(...), post(...)` etc.  It would also require changes to our circleCI env to check curl return codes and result bodies.  Since circleCI is already doing simple TCP connections via `nc`, I decided to stick with that pattern and make a simple port available here for circleCI.  [See yaml file change here](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1571285695).

.circleci/main-config.yml
```diff
@@ -715,7 +715,7 @@ commands:
           additional_options: '-Dexec.mainClass=org.broadinstitute.ddp.route.IntegrationTestSuite -Dexec.classpathScope=test'
           background: true
       - wait_for_server:
-          port: 5556
+          port: 5999 # we're waiting for the post-liquibase port to open up, not the main spark port
```
zyme, line 718: See [corresponding java code change](https://github.com/broadinstitute/ddp-study-server/pull/2854/files#r1571285356).

pepper-apis/dss-core/src/main/resources/logback.xml
```diff
@@ -4,7 +4,7 @@
         <!-- this filter is here to help prevent sensitive data like PHI from bleeding into stdout -->
         <filter class="org.broadinstitute.ddp.logging.NonSecureFilter" />
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} %X{LogBreadCrumb} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg D:%X{GAE_DEPLOYMENT_ID} I:%X{GAE_INSTANCE}%n</pattern>
```
zyme: Sneaking in PEPPER-1413 here.  Adding placeholders for GAE instance and deployments.  I put them to the right of the log line since the left side was getting pretty unwieldy already.

pepper-apis/dss-core/src/main/resources/logback.xml
```diff
@@ -4,7 +4,7 @@
         <!-- this filter is here to help prevent sensitive data like PHI from bleeding into stdout -->
         <filter class="org.broadinstitute.ddp.logging.NonSecureFilter" />
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} %X{LogBreadCrumb} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %X{LogBreadCrumb} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg D:%X{GAE_DEPLOYMENT_ID} I:%X{GAE_INSTANCE}%n</pattern>
```
zyme: Adding in support for PEPPER-1412 as well.  Logback line was getting pretty busy on the left side, so I added these GAE values on the right side of the log message.

pepper-apis/ddp-common/pom.xml
```diff
@@ -130,5 +130,18 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
```
zyme, line 134: Necessary for SparkBootUtil in ddp-common.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -0,0 +1,82 @@
...
+     *                          when the _ah/stop route is called by GAE
+     */
+    public static void startSparkServer(StopRouteCallback stopRouteCallback) {
+        Config cfg = ConfigManager.getInstance().getConfig();
+        String preferredSourceIPHeader = null;
+        if (cfg.hasPath(ConfigFile.PREFERRED_SOURCE_IP_HEADER)) {
+            preferredSourceIPHeader = cfg.getString(ConfigFile.PREFERRED_SOURCE_IP_HEADER);
+        }
+        int configFilePort = cfg.getInt(ConfigFile.PORT);
+        String appEnginePort = System.getenv(APPENGINE_PORT_ENV_VAR); // the port defined by app engine "wins"
```
denniscunningham: Style nit... end of line comments are passe, and this comment seems to be about the following code.
The code below might be clearer with an in-line conditional since it avoids the triple assignment.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -0,0 +1,82 @@
...
+        int port = -1;
+        if (appEnginePort != null) {
+            port = Integer.parseInt(appEnginePort);
+        } else {
+            port = configFilePort;
+        }
+
+        int requestThreadTimeout = 60000;
```
denniscunningham: Is this milliseconds?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -0,0 +1,82 @@
...
+        if (cfg.hasPath(ConfigFile.THREAD_TIMEOUT)) {
+            requestThreadTimeout = cfg.getInt(ConfigFile.THREAD_TIMEOUT);
+        }
+        JettyConfig.setupJetty(preferredSourceIPHeader);
+        threadPool(-1, -1, requestThreadTimeout);
```
denniscunningham: Looks like this threadPool is associated with Spark, but perhaps a comment would help explain why we need it, and why the min and max threads are -1?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -0,0 +1,82 @@
...
+
+        log.info("Starting spark on port {}", port);
+        Spark.port(port);
+        Spark.get(RouteConstants.GAE.START_ENDPOINT, (request, response) -> {
```
denniscunningham, line 61: It feels like your PR description has great info on why we want to return from these immediately, and how that does not cause problems on startup. 
Since that description will quickly fade, perhaps some of that info could go here?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
denniscunningham: Stack traces are primal alerts for developers and this will splat out on every DB test setup, etc.
Also, I wonder how the GCP log trawler will handle this?
Any reason to not have this at debug level? If there is, perhaps move it down from warn, since it is not clear a warning is warranted here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -231,6 +230,8 @@ public class DSMServer {
     private static Auth0Util auth0Util;
 
     public static void main(String[] args) {
+        LogUtil.addAppEngineEnvVarsToMDC();
+        SparkBootUtil.startSparkServer(null);  // respond GAE dispatcher endpoints as soon as possible
```
denniscunningham: Nit: Can this comment be moved above this line to better match our coding norms?

pepper-apis/dsm-core/src/main/resources/logback.xml
```diff
@@ -2,7 +2,7 @@
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} %X{LogBreadCrumb} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %X{LogBreadCrumb} %X{X-Forwarded-For} C:%X{ClientId} S:%X{UserId} [%thread] %-5level %logger{36} %msg D:%X{GAE_DEPLOYMENT_ID} I:%X{GAE_INSTANCE}%n</pattern>
```
denniscunningham: Dumb question, but which log messages does this apply to?
Seems like this info is only useful for service restarts?

pepper-apis/dss-core/src/test/java/org/broadinstitute/ddp/route/IntegrationTestSuite.java
```diff
@@ -396,11 +362,47 @@ private static boolean isTestServerRunning() {
...
+    /**
+     * Called by circleci when creating a dss backend
+     * to run tests against.  Starting a test instance
+     * from scratch can take 30s or more, during which
+     * time most routes will not work properly.  To check
+     * for readiness, ping port 5999 with a utility
+     * like netcat.  When you can connect to port 5999,
+     * the instance is ready for traffic.  Note that
+     * this administrative port is only available for one
+     * connection.  After that, the port closes.
```
denniscunningham, line 374: Nice!

.circleci/main-config.yml
denniscunningham: Thx for that comment!

denniscunningham (approved):
Some comments and questions but looks good!

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
zyme: Yes, will change name to make it clear.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
zyme: I'll do some commit history digging to see if I can discover why we set it this way.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/db/TransactionWrapper.java
zyme: I'm going to ditch this and revert to the original, simpler warn.

pepper-apis/dsm-core/src/main/resources/logback.xml
zyme: It will apply to every single log line.  Since the logs from all instances end up in the same stream, my thought was that just having this information on boot won't tell us which instance is doing what, since there may be moments of overlap during startup/shutdown.

Perhaps that's too much log bloat, though, since the immediate issue is understanding the timing of startup and shutdown and the overlap in instances during restarts, and we should instead just log this information along with start/stop requests.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
zyme: Poking around some ancient commits and old jira tickets, and I found [this ticket](https://broadinstitute.atlassian.net/browse/DDP-602), which I'll drop in as a comment.

PR #2862, opened 2024-04-30
[DSM] PEPPER-1439 Label Creation Job improvement

pegahtah:
## Context

This PR focuses on adding better exception handling for label creation job, as well as making sure the job can resets itself and to restart

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -172,96 +173,116 @@ public static void createLabel(List<KitRequestCreateLabel> kitsLabelTriggered, E
...
+                        kitToCreateLabelFor.getDdpParticipantId(), kitToCreateLabelFor.getInstanceName(), kitToCreateLabelFor.getDsmKitId(),
+                        e);
             }
         }
-
         DBUtil.updateBookmark(0, BOOKMARK_LABEL_CREATION_RUNNING);
     }
 
-    protected static DDPParticipant getDDPParticipant(String ddpParticipantId, DDPInstance ddpInstance) {
+    private static void createCollaboratorPartiticpantId(KitRequestCreateLabel kitToCreateLabelFor, DDPInstanceDto ddpInstanceDto) {
```
zyme: New method--can you put in some javadoc since there's a lot going on in this one?  Also might be worth breaking out a separate method to generate the id from updating the value in the database/dtos.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1303,13 +1312,14 @@ public static String createRandom(int len) {
...
-        if (addressId == null && participant == null) { //if both are set to null then it is return label!
-            toAddress =
-                    easyPostUtil.createAddressWithoutValidation(kitRequestSettings.getReturnName(), kitRequestSettings.getReturnStreet1(),
-                            kitRequestSettings.getReturnStreet2(), kitRequestSettings.getReturnCity(), kitRequestSettings.getReturnZip(),
-                            kitRequestSettings.getReturnState(), kitRequestSettings.getReturnCountry(), kitRequestSettings.getPhone());
+        if (isReturnAddress(addressId, participant)) {
+            try {
+                toAddress = createReturnAddress(kitRequestSettings, easyPostUtil);
+            } catch (EasyPostException e) {
+                logger.error("Couldn't generate return address for a kit for ddpInstance " + kitRequestSettings.getDdpInstanceId(), e);
```
zyme, line 1326: In this situation, what happens when the callers receive a `null` address?  Is that something that we should look at for improving robustness?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1325,6 +1335,18 @@ public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequ
...
     }
 
+    public static Address createReturnAddress(KitRequestSettings kitRequestSettings, EasyPostUtil easyPostUtil) throws EasyPostException {
+        return easyPostUtil.createAddressWithoutValidation(kitRequestSettings.getReturnName(), kitRequestSettings.getReturnStreet1(),
+                kitRequestSettings.getReturnStreet2(), kitRequestSettings.getReturnCity(), kitRequestSettings.getReturnZip(),
+                kitRequestSettings.getReturnState(), kitRequestSettings.getReturnCountry(), kitRequestSettings.getPhone());
+
+    }
+
+    private static boolean isReturnAddress(String addressId, DDPParticipant participant) {
```
zyme, line 1365: Is there a more reliable way of knowing the return address?  Passing all nulls into a method smells like a bug.  Is there metadata we could add in easypost to help us keep track of our return addresses?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/jobs/LabelCreationJob.java
```diff
@@ -24,11 +24,13 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
                 if (!kitsLabelTriggered.isEmpty()) {
                     KitUtil.createLabel(kitsLabelTriggered, null);
                 }
-            } else if (labelCreationStarted < System.currentTimeMillis() - (SystemUtil.MILLIS_PER_HOUR * 2)) {
-                logger.error("Label creation job is running for over 2 hours now...");
+            } else if (labelCreationStarted < System.currentTimeMillis() - (SystemUtil.MILLIS_PER_HOUR)) {
```
zyme: Check out `Instant`--it has really readable ways to do date comparisons with different units, stuff like `Instant.now().minus(1, ChronoUnit.HOURS)`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: This method is called in `getToAddressId`, which generates the EasyPost address IDs necessary for creating a shipping label. A shipping label requires two key pieces of information: the destination address (where the kit is being sent) and the sender's address (for returning the kit if delivery fails). Typically, in DSM, each kit is already associated with an address ID that was generated when the kit was registered, or it uses the participant's address as the shipping address.

However, in the same process of buying a label, we need to determine the return address for a specific lab at Broad without having its pre-existing address IDs or being able to use the participant data. In this scenario of these two being null, `getToAddressId` knows to retrieve the return address directly from the KitRequestSettings, ensuring that the address used is appropriate for the specified lab's requirements. So in here `isReturnAddress` is just telling `getToAddressId` where to look for address

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
pegahtah: Great catch!! Thank you!! I added a check for it before returning `toAddress`

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1327,23 +1332,37 @@ public static Address getToAddressId(@NonNull EasyPostUtil easyPostUtil, KitRequ
...
                     toAddress = easyPostUtil.createAddress(participant, kitRequestSettings.getPhone());
-                    return toAddress;
+                } else {
+                    toAddress = getAddressForStudiesWithCareOfField(easyPostUtil, kitRequestSettings, participant, ddpInstanceDto);
                 }
-                toAddress = getAddressForStudiesWithCareOfField(easyPostUtil, kitRequestSettings, participant, ddpInstanceDto);
             }
         }
+        if (toAddress == null) {
+            throw new DsmInternalError("Couldn't get address for kit request");
```
zyme, line 1341: Add a bit more detail about the kit request in question--anything you can include in the exception message about method params?

zyme (approved):

PR #2870, opened 2024-05-09
PEPPER-332 It's okay to delete a somatic upload with no bytes

zyme:
This PR is a dependency of this [frontend PR](https://github.com/broadinstitute/ddp-angular/pull/2376) that cleans up properly when a somatic upload fails.  Previously, an attempt to delete an upload would result in an error if the file blob was empty.  Because we have found that some uploads fail and don't write any bytes, we need the delete to go through even if there are no bytes in the blob.  This PR removes the exception that was thrown if the byte stream was empty.

For good measure, I added a test to verify that the delete works properly.

## Testing

## Release

### DIFF
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -214,7 +214,7 @@
       {{end}}
     }
     "somatic": {
-        "maxFileSize": 3000000,
+        "maxFileSize": 30000000,
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -214,7 +214,7 @@
       {{end}}
     }
     "somatic": {
-        "maxFileSize": 1000000000,
+        "maxFileSize": 3000000,
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -215,7 +215,7 @@
     }
     "somatic": {
         "maxFileSize": 30000000,
-        "mediaTypes": "application/pdf,application/json"
+        "mediaTypes": "application/pdf,application/json",
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -217,7 +217,7 @@
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
-            {"realm": "somatic-upload-test", "uploadBucket":"cmi-lms-uploaded-files"}
+            {"realm": "somatic-upload-test", "uploadBucket":"cmi-lms-uploaded-files", "maxFileSize": 100000000L}
         ]
     }
 }
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -214,7 +214,6 @@
       {{end}}
     }
     "somatic": {
-        "maxFileSize": 30000000,
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -214,10 +214,11 @@
       {{end}}
     }
     "somatic": {
+        "maxFileSize": 1000000000,
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
-            {"realm": "somatic-upload-test", "uploadBucket":"cmi-lms-uploaded-files", "maxFileSize": 100000000L}
+            {"realm": "somatic-upload-test", "uploadBucket":"cmi-lms-uploaded-files"}
         ]
     }
 }
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -215,6 +215,7 @@
     }
     "somatic": {
         "maxFileSize": 30000000,
+        "mediaTypes": "application/pdf"
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -215,7 +215,7 @@
     }
     "somatic": {
         "maxFileSize": 30000000,
-        "mediaTypes": "application/pdf"
+        "mediaTypes": "application/pdf,application/json"
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
             {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -216,7 +216,8 @@
     "somatic": {
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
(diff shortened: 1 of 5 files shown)


pepper-apis/config/testing-inmemorydb.conf.ctmpl
```diff
@@ -216,7 +216,8 @@
     "somatic": {
         "realmToBucketMappings": [
             {"realm": "osteo2", "uploadBucket":"cmi-osteo2-uploaded-files"},
-            {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"}
+            {"realm": "cmi-lms", "uploadBucket":"cmi-lms-uploaded-files"},
+            {"realm": "somatic-upload-test", "uploadBucket":"cmi-lms-uploaded-files"}
```
zyme, line 220: Without this, the test fails because it can't find the upload bucket for the new test realm.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -256,6 +260,9 @@ private static SomaticResultUpload getSingleSomaticResultUpload(long documentId,
         if (documents.isEmpty()) {
             throw new DSMBadRequestException("Bad request for document with id " + documentId);
         }
+        if (documents.get(0) == null) {
```
zyme, line 263: Added for good measure, just in case something weird happens, to avoid returning null.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -124,6 +126,8 @@ public class SomaticResultUpload implements HasDdpInstanceId {
             + "LEFT JOIN ddp_instance as ddp ON (ddp.ddp_instance_id = p.ddp_instance_id) "
             + "WHERE p.ddp_participant_id = ? AND ddp.instance_name= ?)";
 
+    private static final String HARD_DELETE_SOMATIC_DOCUMENT = "delete from somatic_documents where id = ?";
```
denniscunningham: It looks like the pattern here (and elsewhere in DSM) is to prefix the SQL strings with `SQL_`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SomaticResultUpload.java
```diff
@@ -312,6 +319,23 @@ private static SomaticResultUpload getResult(ResultSet rs) throws SQLException {
             rs.getLong(DBConstants.CREATED_DATE));
     }
 
+    /**
+     * Only use this for testing.  Hard deletes the row.  For production code,
+     * we only soft delete, so please use other delete methods for normal operations.
+     */
+    @VisibleForTesting
+    public static void hardDeleteSomaticDocumentById(Handle handle, long id) {
```
denniscunningham, line 327: Sad that the currency in the code is long, since the ID is actually an int in the DB.
Not worth fixing at this point.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
+        esIndex = ElasticTestUtil.createIndex(instanceName, "elastic/lmsMappings.json", null);
+        ddpInstanceDto = DdpInstanceGroupTestUtil.createTestDdpInstance(instanceName, esIndex, instanceName);
         Config loadedConfig = ConfigManager.getInstance().getConfig();
         somaticResultUploadSerivce = SomaticResultUploadService.fromConfig(loadedConfig);
     }
 
+    @AfterClass
+    public static void tearDown() {
+        try {
+            Assert.assertNotNull(testParticipant);
```
denniscunningham, line 49: I suggest deleting `testParticipant` in an after method class, since it is setup by a particular test.
(This will matter when another test is added.)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
+            TestParticipantUtil.deleteParticipant(testParticipant.getRequiredParticipantId());
+            Assert.assertNotNull(ddpInstanceDto);
```
denniscunningham, line 51: I suggest using `DdpInstanceGroupTestUtil.deleteInstance(ddpInstanceDto)` to provide symmetry with the setup, and to take advantage of whatever that class provides now or in the future.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
+            ddpInstanceDao.delete(ddpInstanceDto.getDdpInstanceId());
+        } catch (Exception e) {
+            e.printStackTrace();
+            Assert.fail("Unexpected exception: " + e);
+        }
+        ElasticTestUtil.deleteIndex(esIndex);
+    }
+
+    static SomaticResultUploadService somaticResultUploadSerivce;
```
denniscunningham: Feels like this should be marked `private` and moved alongside the other class instance variables?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
     }
+
+    @Test
+    public void testDelete() {
+        String ddpParticipantId = TestParticipantUtil.genDDPParticipantId("somaticupload");
+        testParticipant = TestParticipantUtil.createParticipant(ddpParticipantId, ddpInstanceDto.getDdpInstanceId());
+        SomaticResultMetaData uploadData = new SomaticResultMetaData("testFile.pdf", "application/pdf", 1L);
+        SomaticResultUploadService.AuthorizeResult uploadAuth = somaticResultUploadSerivce.authorizeUpload(
+                instanceName, Integer.toString(testParticipant.getRequiredParticipantId()),
+                testParticipant.getRequiredDdpParticipantId(), uploadData);
```
denniscunningham: We already have a `ddpParticipantId` variable. Not a big deal, but definitely helps readability.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
+        Assert.assertEquals(SomaticResultUploadService.AuthorizeResultType.OK, uploadAuth.getAuthorizeResultType());
+        Assert.assertNotNull(uploadAuth.getSomaticResultUpload());
+        SomaticResultUpload deleteResult = somaticResultUploadSerivce.deleteUpload(
+                testParticipant.getRequiredParticipantId(), uploadAuth.getSomaticResultUpload().getSomaticDocumentId(),
+                instanceName);
+        long secondsSinceDeletion = Duration.between(Instant.ofEpochSecond(deleteResult.getDeletedAt()), Instant.now()).getSeconds();
+        assertNotNull(deleteResult);
```
denniscunningham, line 76: Perhaps assert this above, since you use it in the prior statement.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
+        Assert.assertTrue(secondsSinceDeletion < 30);
```
denniscunningham, line 80: Perhaps a comment about what this test verifies?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
     public void test_authorizeValidRequest() {
         assertNotNull(somaticResultUploadSerivce);
     }
+
+    @Test
+    public void testDelete() {
+        String ddpParticipantId = TestParticipantUtil.genDDPParticipantId("somaticupload");
+        testParticipant = TestParticipantUtil.createParticipant(ddpParticipantId, ddpInstanceDto.getDdpInstanceId());
+        SomaticResultMetaData uploadData = new SomaticResultMetaData("testFile.pdf", "application/pdf", 1L);
+        SomaticResultUploadService.AuthorizeResult uploadAuth = somaticResultUploadSerivce.authorizeUpload(
```
denniscunningham, line 68: This depends on DSM being able to contact GCP to get a signed URL.
Is it clear we can always do that, regardless of where this test runs? (Locally, Circle CI, etc.)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
```diff
@@ -3,22 +3,83 @@
...
 
-public class SomaticResultUploadServiceTest {
+import java.time.Duration;
+import java.time.Instant;
 
-    static SomaticResultUploadService somaticResultUploadSerivce;
+public class SomaticResultUploadServiceTest extends DbAndElasticBaseTest {
+
+    private static final DDPInstanceDao ddpInstanceDao = new DDPInstanceDao();
+    private static final String instanceName = "somatic-upload-test";
```
denniscunningham, line 29: Perhaps note that this name matches the configuration and thus care must be taken to change it.

denniscunningham (commented):
Looks good. 
Left some comments.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
zyme: Oooh good catch!

zyme: Will fix this by creating the test participant once in the beforeClass.  There are already a few tests in this class, but only one makes use of the generated test participant, so when additional tests are added, they can (hopefully) use the same participant.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/SomaticResultUploadServiceTest.java
zyme: It does pass for me locally, and should always work since it relies on an SA that we use in the test config.  No IP address restrictions.

denniscunningham (approved):

PR #2873, opened 2024-05-22
Cleaning up logging during shutdown and during liquibase error handling.

zyme:
## PEPPER-1413 cleanups

Post deployment, I noticed a few things that needed some tidying up in order to keep our logs in better shape and avoid false positive errors.  

One issue is that when a liquibase exception occurs during a timeout, the underlying issue is hard to see because of an NPE.  Simple fix to the error logging.

The other issue was that during shutdowns, it appears that the database pool shuts down before quartz jobs, despite the fact that code-wise, the quartz job is shutdown before the connection pool.  It's probably the case that quartz just takes a while to clean itself up async, and by the time it does so, the connection pool has already closed, resulting in a raft of db connectivity issues during shutdown.  The fix here was to track a bit of state about whether the app is shutting down, and decide whether to log an error or an info depending on whether the app is shutting down.

### DIFF
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
@@ -26,6 +26,9 @@ public class SparkBootUtil {
 
     private static int numShutdownAttempts = 0;
 
+    // set when the shutdown process begins
+    private static boolean isShuttingDown;
+
     /**
      * Reads configuration and environment settings common to
      * DSM, DSS, and Housekeeping and starts a spark server on the appropriate port.
@@ -69,6 +72,7 @@ public class SparkBootUtil {
         });
 
         Spark.get(RouteConstants.GAE.STOP_ENDPOINT, (request, response) -> {
+            isShuttingDown = true;
             log.info("Received GAE stop request [{}] for instance {} deployment {}", request.url(),
                     System.getenv(LogUtil.GAE_INSTANCE), System.getenv(LogUtil.GAE_DEPLOYMENT_ID));
             if (stopRouteCallback != null) {
@@ -89,8 +93,15 @@ public class SparkBootUtil {
         });
 
         if (stopRouteCallback != null) {
-            Runtime.getRuntime().addShutdownHook(new Thread(() -> stopRouteCallback.onTerminate()));
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                isShuttingDown = true;
+                stopRouteCallback.onTerminate();
+            }));
         }
+    }
+
+    public static boolean isShuttingDown() {
+        return isShuttingDown;
     }
 
     /**
--- a/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/LiquibaseUtil.java
+++ b/pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/LiquibaseUtil.java
@@ -205,7 +205,7 @@ public class LiquibaseUtil implements AutoCloseable {
                 liquibase.update(new Contexts(context));
             } catch (LiquibaseException originalError) {
                 log.error("LiquibaseException: {}", originalError.getMessage());
-                if (tag != null) {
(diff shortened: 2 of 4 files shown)


pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/util/LiquibaseUtil.java
```diff
@@ -205,7 +205,7 @@ private void runMigrations(String changelogFile, String context) throws Liquibas
                 liquibase.update(new Contexts(context));
             } catch (LiquibaseException originalError) {
                 log.error("LiquibaseException: {}", originalError.getMessage());
-                if (tag != null) {
+                if (tag != null && originalError.getCause() != null) {
```
zyme, line 208: While testing the error handling for liquibase locks, I discovered this NPE.  The fix is to only log this addition message if `getCause()` is not null.

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
```diff
@@ -89,10 +93,17 @@ public static void startSparkServer(AppEngineShutdown stopRouteCallback, Config
         });
 
         if (stopRouteCallback != null) {
-            Runtime.getRuntime().addShutdownHook(new Thread(() -> stopRouteCallback.onTerminate()));
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                isShuttingDown = true;
```
pegahtah, line 97: question: Is there a case where this needs to go back to false, for example in case of an error?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
```diff
@@ -502,7 +511,11 @@ private static void shutdownPubSub() {
             try {
                 pubSubTaskConnectionService.destroy();
             } catch (PubSubTaskException e) {
-                log.error("Failed to shutdown PubSubTask API", e);
+                if (!SparkBootUtil.isShuttingDown()) {
+                    log.error("Failed to shutdown PubSubTask API", e);
+                } else {
+                    log.info("Exception during shutdown", e);
```
pegahtah, line 517: I am not very familiar with housekeeping, but shouldn't this also be at an ERROR level?

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/event/publish/pubsub/PubSubPublisherInitializer.java
```diff
@@ -58,7 +59,11 @@ public static void shutdownPublishers() {
             try {
                 publisher.getValue().shutdown();
             } catch (Exception e) {
-                log.error("Error shutting down publisher {}", publisher.getKey(), e);
+                if (!SparkBootUtil.isShuttingDown()) {
+                    log.error("Error shutting down publisher {}", publisher.getKey(), e);
+                } else {
+                    log.info("Exception during shutdown", e);
```
pegahtah, line 65: same here, I am just wondering if logging an exception at info level will make it easy to miss?

pepper-apis/ddp-common/src/main/java/org/broadinstitute/ddp/appengine/spark/SparkBootUtil.java
zyme: I'm thinking that this is really a one-way thing: once the signal to shutdown is received, the shutdown is going to happen, regardless of errors.  Check out [this doc](https://cloud.google.com/appengine/docs/standard/how-instances-are-managed#shutdown) that describes the shutdown process.

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/Housekeeping.java
zyme: My thinking here is that during shutdown, things are going to be messy, and while a best attempt is made to shut down cleanly, if something blows up during VM termination, it's interesting but not something that we want to consider as a production alert.

pegahtah (approved):

pegahtah:
Approved, thank you for working on this!

PR #2880, opened 2024-05-30
PEPPER-1407 Exception mapper for handling unique constraints

zyme:
## PEPPER-1407

Added exception mapper to help handle attempts to insertdata that will break unique constraints or is otherwise considered duplicative.  Along the way, changed a few things to simplify how we check for uniqueness of filters.

## FUD Score

* Make a new filter that has the same name as an existing filter.  When saving the filter with the same name, you should see a better error message in the red popup at the bottom of the screen.

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -44,6 +44,7 @@ import org.broadinstitute.dsm.exception.AuthenticationException;
 import org.broadinstitute.dsm.exception.AuthorizationException;
 import org.broadinstitute.dsm.exception.DSMBadRequestException;
 import org.broadinstitute.dsm.exception.DsmInternalError;
+import org.broadinstitute.dsm.exception.DuplicateEntityException;
 import org.broadinstitute.dsm.exception.UnsafeDeleteError;
 import org.broadinstitute.dsm.jobs.DDPEventJob;
 import org.broadinstitute.dsm.jobs.DDPRequestJob;
@@ -1067,6 +1068,13 @@ public class DSMServer {
             logger.warn("Authentication error while processing request: {}: {}", request.url(), exception.toString());
             response.status(401);
             response.body(exception.getMessage());
+        });
+        exception(DuplicateEntityException.class, (e, request, response) -> {
+            // Tell the user to retry with a different name.  Not logged as an error,
+            // since this can happen during normal operations.
+            response.status(400);
+            response.body(String.format("%s %s is already taken.  Please retry with a different value.",
+                    e.getEntityName(), e.getEntityValue()));
         });
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -1072,7 +1072,7 @@ public class DSMServer {
         exception(DuplicateEntityException.class, (e, request, response) -> {
             // Tell the user to retry with a different name.  Not logged as an error,
             // since this can happen during normal operations.
-            response.status(400);
+            response.status(422);
             response.body(String.format("%s %s is already taken.  Please retry with a different value.",
                     e.getEntityName(), e.getEntityValue()));
         });
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -1073,8 +1073,8 @@ public class DSMServer {
             // Tell the user to retry with a different name.  Not logged as an error,
             // since this can happen during normal operations.
             response.status(422);
-            response.body(String.format("%s %s is already taken.  Please retry with a different value.",
-                    e.getEntityName(), e.getEntityValue()));
+            response.body(String.format("%s %s is already taken.  Please retry with a different %s.",
+                    e.getEntityName(), e.getEntityValue(), e.getEntityName()));
         });
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
@@ -108,7 +108,7 @@ public class ViewFilter {
                       Integer[] realmId) {
         this.userId = userId;
         this.fDeleted = fDeleted;
-        this.filterName = filterName;
+        setFilterName(filterName);
         this.columns = columns;
         this.shared = shared;
         this.id = id;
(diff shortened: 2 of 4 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -120,42 +121,30 @@ public ViewFilter(String filterName, Map<String, List<Object>> columns, String i
         this.realmId = realmId;
     }
 
-    private static SimpleResult checkUniqueFilterName(String suggestedName) {
```
zyme, line 123: I reworked this a little bit to clarify what the method does, and returning a boolean instead of the more opaque `SimpleResult`.  Since there's only one caller and the method is used to check for uniqueness, there was an opportunity to simplify this a bit.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -120,42 +121,30 @@ public ViewFilter(String filterName, Map<String, List<Object>> columns, String i
...
+                throw new DsmInternalError("Error checking uniqueness of filter " + filterName, e);
             }
-            return dbVals;
+            return null;
         });
-        return results;
+        return filterExists.get();
     }
 
-    public static Object saveFilter(@NonNull String filterViewToSave, String userId, @NonNull Map<String, DBElement> columnNameMap,
```
zyme, line 144: Instead of an opaque string input that is a blob of anything, I changed it to `ViewFilter` so that the method can trust its inputs, and leave conversion from string json into the object to callers.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -120,42 +121,30 @@ public ViewFilter(String filterName, Map<String, List<Object>> columns, String i
...
+    public static Object saveFilter(@NonNull ViewFilter viewFilter, String userId, @NonNull Map<String, DBElement> columnNameMap,
                                     @NonNull String ddpGroupId) {
-        ViewFilter viewFilter = new Gson().fromJson(filterViewToSave, ViewFilter.class);
-        if (viewFilter == null) {
-            throw new RuntimeException("The request for saving filter doesn't have a ViewFilter");
-        }
         String query;
-        String suggestedName = viewFilter.getFilterName();
-        SimpleResult results = checkUniqueFilterName(suggestedName);
```
zyme, line 152: Check for uniqueness has been moved outside, to callers, so this method just does the save.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ViewFilterRoute.java
```diff
@@ -67,7 +69,14 @@ public Object processRequest(Request request, Response response, String userId)
                 }
             } else if (request.url().contains(RoutePath.SAVE_FILTER)) {
                 String ddpGroupId = DDPInstance.getDDPGroupId(realm);
-                return ViewFilter.saveFilter(json, userIdRequest, patchUtil.getColumnNameMap(), ddpGroupId);
+                ViewFilter viewFilter = new Gson().fromJson(json, ViewFilter.class);
+
+                String filterName = viewFilter.getFilterName();
```
denniscunningham, line 75: It might make sense to trim `filterName` and check for empty.
Not sure of the request body origin, but we've definitely seen problems with extra whitespace in input strings.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -120,42 +121,30 @@ public ViewFilter(String filterName, Map<String, List<Object>> columns, String i
...
     }
 
-    private static SimpleResult checkUniqueFilterName(String suggestedName) {
-        SimpleResult results = inTransaction((conn) -> {
-            SimpleResult dbVals = new SimpleResult();
+    /**
+     * Returns true if a filter with the given name exists in the database,
+     * false if it does not exist.
+     */
+    public static boolean doesFilterExist(String filterName) {
```
denniscunningham, line 135: Does it make sense to add a test for this method, or is manual testing sufficient?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ViewFilterRoute.java
zyme: Agreed, will apply this inside of `FilterView` state so that it's trimmed when constructed.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
zyme: `saved-views-custom-columns.spec.ts` and `saved-views-custom-columns-filter-condition.spec.ts` do cover this via PW, so we'll catch any issues there.  I think if we did add another test, it would be most helpful to test `processRequest` here.  Since we're time boxing to a day, I'd lean towards merging this as-is, seeing what comes out of PW, and if there are subsequent issues, I can put in a more detailed junit test.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -1068,6 +1069,13 @@ private void setupRouteGenericErrorHandlers() {
             response.status(401);
             response.body(exception.getMessage());
         });
+        exception(DuplicateEntityException.class, (e, request, response) -> {
+            // Tell the user to retry with a different name.  Not logged as an error,
+            // since this can happen during normal operations.
+            response.status(422);
+            response.body(String.format("%s %s is already taken.  Please retry with a different value.",
```
pegahtah: does it make sense to change the last word from "value" to `e.getEntityName()`, 
"Please try with a different filter name" for example

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ViewFilterRoute.java
```diff
@@ -67,7 +70,14 @@ public Object processRequest(Request request, Response response, String userId)
...
+                if (ViewFilter.doesFilterExist(filterName)) {
+                    // tell the user to retry with a different name for the filter if the name is already taken
+                    throw new DuplicateEntityException("filter", filterName);
```
pegahtah: I think the error is easier to read if the entityName is "filter name"

pegahtah (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
denniscunningham: Do we have a way to test `processRequest`?
I've tried to deal with the Spark Java classes on a couple of occasions, including a few hours today, and have found it very frustrating.

denniscunningham (commented):
Follow-up question.

PR #2883, opened 2024-05-31
[PEPPER-1429] Refactor ParticipantExitRoute and add tests

denniscunningham:
PEPPER-1429

The fix here was easy (trim input fields), but building appropriate tests was hard, due to Spark Java, which is a poor, outdated framework, and due to the existing structure of the code which was also poor and not conducive to testing.

The main work was in rewriting and testing ParticipantExitRoute -- it was poorly structured, did not handle a handful of cases properly, and mixed route and service code (which makes building tests difficult.) I broke out the exit logic into ParticipantExitService, which enabled testing the route code, and provides a clean break should service tests be added in the future. (I did not add those tests here since that would have taken an additional day, at least, and required additional restructuring downstream. For now we will continue to rely on our PW system test for coverage.)

Since we do not have JUnit tests for ParticipantExitService, I largely kept the code as it was in ParticipantExitRoute, but did some refactoring of `_exitParticipant` since it had a couple obvious deficiencies. I also added some notes where the code intention was unclear.

Other things:

- Renamed `ParticipantNotExist` exception to `EntityNotFound`, and added handling for it in DSMServer, since we did not have a general not found response exception.
- Added a non-static method to UserUtil to allow for mocking. UserUtil is really a service, but at this point there is limited value in setting it up properly and refactoring adjacent code.

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -44,6 +44,7 @@ import org.broadinstitute.dsm.exception.AuthenticationException;
 import org.broadinstitute.dsm.exception.AuthorizationException;
 import org.broadinstitute.dsm.exception.DSMBadRequestException;
 import org.broadinstitute.dsm.exception.DsmInternalError;
+import org.broadinstitute.dsm.exception.EntityNotFound;
 import org.broadinstitute.dsm.exception.UnsafeDeleteError;
 import org.broadinstitute.dsm.jobs.DDPEventJob;
 import org.broadinstitute.dsm.jobs.DDPRequestJob;
@@ -1066,6 +1067,11 @@ public class DSMServer {
             // this is a fallback exception, log it warn level to see why it is happening
             logger.warn("Authentication error while processing request: {}: {}", request.url(), exception.toString());
             response.status(401);
+            response.body(exception.getMessage());
+        });
+        exception(EntityNotFound.class, (exception, request, response) -> {
+            logger.info("Entity not found while processing request: {}: {}", request.url(), exception.toString());
+            response.status(404);
             response.body(exception.getMessage());
         });
     }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ParticipantExitRoute.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ParticipantExitRoute.java
@@ -1,115 +1,75 @@
 package org.broadinstitute.dsm.route;
 
-import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
-import java.util.Optional;
 
-import com.google.api.client.http.HttpStatusCodes;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
-import lombok.NonNull;
+import com.google.gson.Gson;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
-import org.broadinstitute.dsm.DSMServer;
-import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.KitDiscard;
-import org.broadinstitute.dsm.db.KitRequestShipping;
 import org.broadinstitute.dsm.db.ParticipantExit;
-import org.broadinstitute.dsm.db.dao.ddp.instance.DDPInstanceDao;
-import org.broadinstitute.dsm.db.dto.ddp.instance.DDPInstanceDto;
-import org.broadinstitute.dsm.exception.ParticipantNotExist;
+import org.broadinstitute.dsm.exception.AuthorizationException;
+import org.broadinstitute.dsm.exception.DSMBadRequestException;
+import org.broadinstitute.dsm.route.request.ParticipantExitRequest;
 import org.broadinstitute.dsm.security.RequestHandler;
-import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.service.participant.ParticipantExitService;
 import org.broadinstitute.dsm.statics.RequestParameter;
 import org.broadinstitute.dsm.statics.RoutePath;
-import org.broadinstitute.dsm.statics.UserErrorMessages;
-import org.broadinstitute.dsm.util.DDPRequestUtil;
 import org.broadinstitute.dsm.util.UserUtil;
-import org.broadinstitute.lddp.handlers.util.Result;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
 
+@Slf4j
 public class ParticipantExitRoute extends RequestHandler {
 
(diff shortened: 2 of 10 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
```diff
@@ -0,0 +1,93 @@
...
+public class ParticipantExitService {
+
+    public Map<String, ParticipantExit> getExitedParticipants(String realm) {
+        return ParticipantExit.getExitedParticipants(realm, true);
+    }
+
+    public List<KitDiscard> exitParticipant(String realm, String ddpParticipantId, String userId, boolean inDDP) {
+        _exitParticipant(realm, ddpParticipantId, userId, inDDP);
+
+        // NOTE: The following code was lifted as-is from ParticipantExitRoute.java. I did not attempt to
```
denniscunningham, line 44: As noted in the code, this method is pretty much the same as it was in `ParticipantExitRoute`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
```diff
@@ -0,0 +1,93 @@
...
+        return kitsNeedAction;
+    }
+
+    /**
+     * Exit a participant from DSM and DSS
+     * @param inDDP true if the participant is in DSS. Note: The role of this parameter is unclear. It is passed from
+     *             the front end and the pre-existing code recorded it in the DB, but used
+     *             different criteria to determine if the participant is in DSS. -DC
+     */
+    protected void _exitParticipant(String realm, String ddpParticipantId, String userId, boolean inDDP) {
```
denniscunningham, line 70: This code is refactored a bit so should be compared to the former `exitParticipant` method in `ExitParticipantRoute`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -255,7 +255,7 @@ public void onAhStop() {
             }
 
             public void onTerminate() {
-                logger.info("Terminating DSM instance", LogUtil.getAppEngineInstance());
```
zyme, line 258: Thank you

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -254,6 +254,10 @@ private static void getList(Connection conn, String query, String role, String u
         }
     }
 
+    public boolean userHasRole(String realm, String userId, String role, String userIdRequest) {
```
zyme, line 257: Why not go full bore and just rename the underlying method?  Happy to put a thumb on that as a separate PR.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ParticipantExitRoute.java
```diff
@@ -1,115 +1,75 @@
...
-        if (StringUtils.isNotBlank(realm)) {
-            if (UserUtil.checkUserAccess(realm, userId, ROLE, null)) {
-                return ParticipantExit.getExitedParticipants(realm).values();
-            } else {
-                response.status(500);
-                logger.warn("User with id={} could not access realm={} with role={}", userId, realm, ROLE);
-                return new Result(500, UserErrorMessages.NO_RIGHTS);
-            }
+    public Object processRequest(Request request, Response response, String userId) {
+        String requestMethod = request.requestMethod();
```
zyme, line 29: Just to make sure I get it: instead of switching behavior based on the existence of the `realm` param, now we switch based on get/post?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
```diff
@@ -0,0 +1,93 @@
...
+            } else {
+                //refund label of kits which are not sent yet
+                KitRequestShipping.refundKit(kit.getDsmKitRequestId(), DSMServer.getDDPEasypostApiKey(realm),
+                        maybeInstance.orElse(null));
+            }
+        }
+        return kitsNeedAction;
+    }
+
+    /**
```
zyme, line 64: Should this javadoc be on the public method instead of the protected one?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
```diff
@@ -0,0 +1,93 @@
...
+     * Exit a participant from DSM and DSS
+     * @param inDDP true if the participant is in DSS. Note: The role of this parameter is unclear. It is passed from
```
zyme, line 66: `inDDP` looks like it is always set to true in the frontend, as `!ParticipantExitComponent.deletedFromDDP`.  The ability to change this value in the UI has been commented out, and the majority of rows in the table are set to true.

zyme (approved):
Good stuff!  Nice balance of factoring out code and making it more testable (and tested).

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/ParticipantExitRoute.java
denniscunningham: Yes. The former code was relying on an implementation byproduct rather than testing for the request method directly.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
denniscunningham: I mostly wanted to draw attention to `inDDP`, since it's use was odd.
I'll take a crack at the public method, but was hoping to avoid explaining why it returns a list of `KitDiscard` since I do not think it is used in the UI. Although I can hazard a guess that it is kits left dangling by the withdrawal.
Do you know what that's for?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/UserUtil.java
denniscunningham: 'checkUserAccess` is used in many places, so did not want to expand the number of file in the PR.
Although I think `userHasRole` may be more accurate, the method does a bit more checking on input. 
The motivation for changing was to avoid a collision with the existing method name, not to provide a better name, so we would need to resolve that issue to make the change you suggest.

denniscunningham (commented):
Thanks for your review @zyme!
A couple questions back to you before I make an update.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/participant/ParticipantExitService.java
zyme: I don't, but Pegah probably knows.  Either way, l'm up for holding off on this for another time.

PR #2884, opened 2024-06-05
[DSM] PEPPER-1314

pegahtah:
## Context

This is not a direct fix for the problem in PEPPER-1314, but it is a fix for only prod error I found that could be related to this error. 
I have changed the frontend to pass in null values intead of a blank string and changed the backend to be able to handle a null value. 
However this error may not be related to this ticket.

## Checklist

## FUD Score

## Testing

## Release


pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/elastic/export/generate/MappingGeneratorTest.java
```diff
@@ -95,26 +95,32 @@ public void generateMapping() {
...
-            public Map<String, String> extract() {
-                return Map.of();
-            }
-        });
-        Map<String, Object> parseJson = mappingGenerator.parseJson();;
-        Map<String, Object> additionalValuesJson = (Map) parseJson.get("dynamicFields");
-        Assert.assertNotNull(additionalValuesJson);
-        Assert.assertEquals(TypeParser.TEXT_KEYWORD_MAPPING,
-                ((Map) additionalValuesJson.get(PROPERTIES)).get(CamelCaseConverter.of("DDP_INSTANCE").convert()));
+        try {
```
pegahtah, line 98: Just added a try-catch here

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/PatchFactory.java
```diff
@@ -3,7 +3,6 @@
 import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
-import org.broadinstitute.dsm.db.OncHistoryDetail;
```
pegahtah, line 6: unused import

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/BasePatch.java
```diff
@@ -235,6 +235,10 @@ protected boolean isNameValuePairs() {
 
     protected List<NameValue> setWorkflowRelatedFields(@NonNull Patch patch) {
         List<NameValue> nameValues = new ArrayList<>();
+        if (patch.getNameValue().getValue() == null) {
+            // if the date that was sent in is null, we don't want to set the workflow fields
+            return nameValues;
+        }
```
pegahtah, line 241: If the value that was sent was null, it means we should not set the request to sent

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/generate/BaseGenerator.java
```diff
@@ -76,13 +77,13 @@ public Object collect() {
         Object sourceToUpsert;
         try {
             sourceToUpsert = parseJson();
-        } catch (JsonParseException jpe) {
+        } catch (JsonParseException | DsmNullDateException e) {
```
pegahtah: In case if `DsmNullDateException` we also need to parse the nameValue as single element

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/generate/MappingGenerator.java
```diff
@@ -65,10 +66,13 @@ public Map<String, Object> getCompleteMap(Object propertyMap) {
...
 
     @Override
-    protected Map<String, Object> parseJson() {
+    protected Map<String, Object> parseJson() throws DsmNullDateException {
         Map<String, Object> resultMap = new HashMap<>();
         Map<String, Object> fieldsByValues = parseJsonToMapFromValue();
-
+        if (fieldsByValues == null) {
+            //This happens when the value is null, and here dsm throws an exception so that the null value is processed as single values
+            throw new DsmNullDateException("fieldsByValues is null");
```
pegahtah: See this [comment](https://github.com/broadinstitute/ddp-study-server/pull/2884/files#r1628545574)

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/patch/OncHistoryDetailPatchTest.java
```diff
@@ -131,12 +135,75 @@ public void doPatchTest() {
...
+        patch.setRealm(instanceName);
+        patch.setId(String.valueOf(oncHistoryDetailId));
+
+        NotificationUtil mockNotificationUtil = mock(NotificationUtil.class);
+        BasePatch existingOncHistoryDetailPatch = ExistingRecordPatchFactory.produce(patch, mockNotificationUtil);
+        existingOncHistoryDetailPatch.setElasticSearchExportable(true);
+        existingOncHistoryDetailPatch.doPatch();
+
+        try {
+            TimeUnit.SECONDS.sleep(1);
```
zyme, line 162: What are we waiting for here?  ES sync?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/generate/MappingGenerator.java
zyme: Why do we need a separate exception class here?  Based on a null `fieldsByValue`, how do we know that what's missing was a date vs. something else?  The fact this exception is caught and handled in the exact same way as `JsonParseException` in BaseGenerator makes me think that the separate exception class does not lead to different exception handling logic.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/model/patch/OncHistoryDetailPatchTest.java
pegahtah: Yes, found this in the other tests in this file too, and if I remove it sometimes ES does not catch up in time

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/export/generate/MappingGenerator.java
pegahtah: Excellent point. Here’s how `fieldsByValue` is calculated: we either return a Map or throw one of two different exceptions.

The only way the Map could be null is if the value itself was null. Until today, we handled null values by converting them to blank strings in the frontend. However, now that we need to handle null values specifically for dates, I thought this should be addressed by a separate exception.

What we actually want is the same outcome as when the value couldn't be parsed into a map. Here’s the relevant code:

```protected Map<String, Object> parseJsonToMapFromValue() {
        try {
            return ObjectMapperSingleton.instance().readValue(String.valueOf(getNameValue().getValue()), Map.class);
        } catch (com.fasterxml.jackson.core.JsonParseException | JsonMappingException je) {
            throw new JsonParseException(je.getMessage());
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    }
```

I think one way to maybe help with your concerns could be to only throw this exception if the  `getNameValue().getName()` is one of the date fields for fax sents, because those are the only fields we could be sending nulls for

zyme (approved):

PR #2894, opened 2024-06-12
[DSM] PEPPER-1397 moved the retry variables to config

pegahtah:
## Context

I moved the variables to config so we can test different scenarios for this situation

## Checklist

## FUD Score

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
@@ -15,8 +15,10 @@ import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.dao.SkippedParticipantEventDao;
 import org.broadinstitute.dsm.db.dao.queue.EventDao;
 import org.broadinstitute.dsm.model.KitDDPNotification;
+import org.broadinstitute.dsm.statics.ApplicationConfigConstants;
 import org.broadinstitute.dsm.statics.RoutePath;
 import org.broadinstitute.dsm.util.DDPRequestUtil;
+import org.broadinstitute.dsm.util.DSMConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -96,8 +98,11 @@ public class EventService {
     @VisibleForTesting
     protected static boolean triggerDssWithEvent(@NonNull String eventType, DDPInstance ddpInstance, long eventDate,
                                         @NotNull String ddpParticipantId, @NotNull String eventInfo, KitReasonType reason) {
-        final long initialInterval = 500; // base delay in milliseconds
-        final double multiplier = 2.0; // exponential backoff multiplier
+
+        final long initialInterval = DSMConfig.getIntFromConfig(ApplicationConfigConstants.EVENT_RETRY_INTERVAL_MS); // base delay in
+        // milliseconds
+        final double multiplier = DSMConfig.getDoubleFromConfig(ApplicationConfigConstants.EVENT_RETRY_MULTIPLIER); // exponential backoff
+        // multiplier (1.5, 2, etc.)
 
         IntervalFunction intervalFn = IntervalFunction.ofExponentialBackoff(initialInterval, multiplier);
         RetryConfig retryConfig = RetryConfig.custom()
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsm.service;
 
 import java.util.Collection;
-import java.util.Optional;
 import java.util.concurrent.Callable;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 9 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
```diff
@@ -96,8 +98,11 @@ protected static boolean triggerDssByKitEvent(KitDDPNotification kitDDPNotificat
     @VisibleForTesting
     protected static boolean triggerDssWithEvent(@NonNull String eventType, DDPInstance ddpInstance, long eventDate,
                                         @NotNull String ddpParticipantId, @NotNull String eventInfo, KitReasonType reason) {
-        final long initialInterval = 500; // base delay in milliseconds
-        final double multiplier = 2.0; // exponential backoff multiplier
+
+        final long initialInterval = DSMConfig.getIntFromConfig(ApplicationConfigConstants.EVENT_RETRY_INTERVAL_MS); // base delay in
```
ssettipalli: looks like you might get null if not in conf file. 
May be default to some valid int value if not in path? 
same for multiplier

pegahtah: oh great catch! will fix, Thx

zyme: I think it would be better to pass these values in as args to `triggerDssWithEvent()`, instead of reading the config values inside the method.  Reading global config values inside a method like this tends to make testing harder.  They could also become instance variables in `EventService`, and then the methods that need them could move to instance methods instead of static methods.  Alternatively, they can become static final variables at the class level.

zyme: I think it's safe to fail if it's not in the config.  With this change, we expect the value to be in the config, so it's okay to leave as is IMO.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/DSMConfig.java
```diff
@@ -23,6 +24,28 @@ public static String getSqlFromConfig(@NonNull String queryName) {
         return config.getString(queryName);
     }
 
+    public static Integer getIntFromConfig(@NonNull String name) {
```
zyme: Take a look at ConfigManager.getInstance().  I think these methods could be replaced with calls to `ConfigManager.getInstance().getConfig().getInt()` and `getDouble()`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/DSMConfig.java
```diff
@@ -23,6 +24,28 @@ public static String getSqlFromConfig(@NonNull String queryName) {
...
+        try {
+            if (hasConfigPath(name)) {
+                return config.getInt(name);
+            }
+            return null;
+        } catch (Exception e) {
+            throw new DsmInternalError("Error getting int from config for " + name, e);
```
zyme: I think the existing exceptions that typesafe throws when it can't find a key convey the same information to the caller.  Does throwing a wrapped DsmInternalError buy us some different exception handling here?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitTestUtil.java
```diff
@@ -513,5 +516,16 @@ public void setEsIndex(String esIndex) {
         new DDPInstanceDao().updateEsParticipantIndex(ddpInstanceId, esIndex);
     }
 
+
+    /**
+     * reads the test config and sets it as what dsm uses as config.
+     * This is needed for the tests that use values from config/secret to run
+     * */
+    public static void setDsmConfig() {
```
zyme: It's a little weird to have a setter with no args.  If in all cases we always want to use the configManager, then the guts of this method should probably go into the KitTestUtil constructor so that they are always called.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitTestUtil.java
```diff
@@ -513,5 +516,16 @@ public void setEsIndex(String esIndex) {
...
+        ConfigManager configManager = ConfigManager.getInstance();
+        Config cfg = configManager.getConfig();
+        new DSMConfig(cfg);
```
zyme: Because internally newing up DSMConfig actually changes static state, it's possible that this may run into race conditions during tests, where some tests attempt to set DSMConfig.config differently between local envs vs circle.  If in EventConfig we set things like so:

``` 
long initialInterval = ConfigManager.getInstance().getConfig().getInt(ApplicationConfigConstants.EVENT_RETRY_INTERVAL_MS); // base delay in
    double multiplier = ConfigManager.getInstance().getConfig().getDouble(ApplicationConfigConstants.EVENT_RETRY_MULTIPLIER); // exponential backoff
```
then I think we can avoid this risk.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/DSMConfig.java
pegahtah: Not really but this is how most of the other DSM exceptions are wrapped, and the server knows how to handle and communicate that to the frontend if needed.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitTestUtil.java
pegahtah: That solved the problem! 

The problem here was that the config for DSM tests does not come from the dsm secret, it comes from the testing-inmemory-db while DSM is set to use DSMConfig that is created upon server start. `ConfigManager.getInstance()` was a great way to handle this, thanks

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/kits/KitTestUtil.java
pegahtah: You'd be happy to know that I got rid of this completely

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
pegahtah: It was cleaner to make them static final variables in the class level so I moved them there

zyme (approved):

PR #2896, opened 2024-06-13
PEPPER-397 Redux

zyme:
## PEPPER-1397

These changes largely restore[ the earlier commits from this PR](https://github.com/broadinstitute/ddp-study-server/pull/2894).  I tried a few approaches to pass in the retry timeouts as instance params, but they all quickly pulled on the thread to far, so I returned to what was working earlier and made a few improvements to `DSMConfig` along the way.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -263,7 +263,7 @@ public class DSMServer {
         synchronized (isReady) {
             try {
                 logger.info("Starting up DSM");
-                new DSMConfig(cfg);
+                DSMConfig.setConfig(cfg);
                 DSMServer server = new DSMServer();
                 server.configureServer(cfg);
                 isReady.set(true);
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
@@ -15,6 +15,7 @@ import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.dao.SkippedParticipantEventDao;
 import org.broadinstitute.dsm.db.dao.queue.EventDao;
 import org.broadinstitute.dsm.model.KitDDPNotification;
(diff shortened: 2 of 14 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -263,7 +263,7 @@ public void onTerminate() {
         synchronized (isReady) {
             try {
                 logger.info("Starting up DSM");
-                new DSMConfig(cfg);
+                DSMConfig.setConfig(cfg);
```
zyme, line 266: I was puzzled by why we had a number of these new-ing up of DSMConfig without assignment.  I realized what's really going on is setting the config and not newing up a new instance, so I changed the constructor to a setter, since that more accurately reflects what's going on.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/EventService.java
```diff
@@ -27,11 +27,6 @@ public class EventService {
     protected static final int MAX_TRIES = 5;
     private static final EventDao eventDao = new EventDao();
     private static final SkippedParticipantEventDao skippedParticipantEventDao = new SkippedParticipantEventDao();
-    private static final int INITIAL_RETRY_INTERVAL = ConfigManager.getInstance().getConfig()
```
zyme, line 30: The static initializer approach (that I had suggested earlier) turned out to be unworkable in this case, so I moved these lookups back to where they were originally, inside the method that does the retries.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/util/DSMConfig.java
```diff
@@ -3,35 +3,47 @@
...
         if (config == null) {
             throw new RuntimeException("Config is null ");
         }
+    }
 
-        if (!config.hasPath(queryName)) {
-            throw new RuntimeException("Conf is missing query named " + queryName);
-        }
+    public static int getIntFromConfig(String path) {
+        checkNullConfig();
```
zyme, line 32: Small refactor to write a checkNullConfig method instead of copy pasting the check.

denniscunningham (approved):

PR #2908, opened 2024-07-09
PEPPER-749 . Include sampleType while export to ES on kit deactivate or reactivate

ssettipalli:
PEPPER-749 
Include sampleType while export to ES on kit deactivate or reactivate
Change to handle scenario where kit was reactivated but not yet scanned, make sure sampleType(kitType) is included in ES upsert.
Similarly on deactivate, query and include sampleType(kitType) in ES upsert.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
@@ -182,8 +182,8 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                     + "LEFT JOIN kit_type ty ON (req.kit_type_id = ty.kit_type_id) WHERE req.ddp_participant_id = ? "
                     + "AND realm.instance_name = ?";
     public static final String SQL_SELECT_KIT =
-            "SELECT * FROM (SELECT realm.instance_name, re.dsm_kit_request_id FROM ddp_kit_request re, ddp_instance realm "
-                    + "WHERE realm.ddp_instance_id = re.ddp_instance_id) AS request "
+            "SELECT * FROM (SELECT realm.instance_name, re.dsm_kit_request_id, kt.kit_type_name FROM ddp_kit_request re, ddp_instance realm, kit_type kt "
+                    + "WHERE realm.ddp_instance_id = re.ddp_instance_id and kt.kit_type_id = re.kit_type_id) AS request "
                     + "LEFT JOIN (SELECT * FROM (SELECT k.easypost_to_id, k.easypost_return_id, "
                     + "k.deactivated_date, k.deactivation_reason, k.dsm_kit_request_id, k.easypost_address_id_to, k.dsm_kit_id "
                     + "FROM ddp_kit k INNER JOIN( "
@@ -396,6 +396,13 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
     public KitRequestShipping(Integer dsmKitRequestId, Long dsmKitId, String easypostToId, String easypostAddressId, Boolean error,
                               String message) {
         this(null, null, null, null, null, null, dsmKitRequestId, dsmKitId, null, null, null, null, null, null, null, error, message, null,
+                easypostAddressId, null, null, null, null, easypostToId, null, null, null, null, null, null, null, null, null, null, null,
+                null, null, null, null, null, null, null, null, null, null, null);
+    }
+
+    public KitRequestShipping(Integer dsmKitRequestId, Long dsmKitId, String easypostToId, String easypostAddressId, Boolean error,
+                              String message, String kitTypeName) {
+        this(null, null, null, null, null, kitTypeName, dsmKitRequestId, dsmKitId, null, null, null, null, null, null, null, error, message, null,
                 easypostAddressId, null, null, null, null, easypostToId, null, null, null, null, null, null, null, null, null, null, null,
                 null, null, null, null, null, null, null, null, null, null, null);
     }
@@ -1676,7 +1683,7 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                         dbVals.resultValue =
                                 new KitRequestShipping(rs.getInt(DBConstants.DSM_KIT_REQUEST_ID), rs.getLong(DBConstants.DSM_KIT_ID),
                                         rs.getString(DBConstants.EASYPOST_TO_ID), rs.getString(DBConstants.EASYPOST_ADDRESS_ID_TO), false,
-                                        null);
+                                        null, rs.getString(DBConstants.KIT_TYPE_NAME));
                     }
                 }
             } catch (Exception ex) {
@@ -1693,7 +1700,6 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
 
             long dsmKitId = KitRequestShipping.writeNewKit(dsmKitRequestId, kitRequestShipping.getEasypostAddressId(), message, false);
             kitRequestShipping.setDsmKitId(dsmKitId);
-            kitRequestShipping.setKitTypeName(kitRequestDao.getKitTypeByKitRequestId(dsmKitRequestId));
             try {
                 UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto, ESObjectConstants.DSM_KIT_ID,
                         ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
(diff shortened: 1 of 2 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -854,8 +854,9 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
             kitRequestShipping.setDsmKitRequestId(dsmKitRequestId);
             kitRequestShipping.setDeactivationReason(deactivationReason);
             kitRequestShipping.setDeactivatedDate(deactivatedDate);
-
+            kitRequestShipping.setKitTypeName(kitRequestDao.getKitTypeByKitRequestId(dsmKitRequestId));
             try {
+                //todo: if multiple kits exist for same kit-request(dsmKitRequestId), we are updating all elements? edge case though
```
ssettipalli, line 866: just added todo with what was noticed. edge case though

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/kitrequest/KitRequestDao.java
```diff
@@ -134,6 +139,31 @@ public String getKitLabelFromDsmKitRequestId(long dsmKitRequestId) {
...
+                    }
+                }
+            } catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Couldn't get kit type for kit request: " + kitRequestId, results.resultException);
```
denniscunningham: Pls use DsmInternalError or DSMBadRequestException or similar in place of RuntimeException, which is too general.

Also, this pattern can be simplified. You are expecting to return a value, otherwise it is an error. 

Here is an example. (It's not identical, but an illustration of what simpler looks like.)

```    
public static AdminOperationResponse.AdminOperationResult getOperationRecord(int operationId) {
        return inTransaction(conn -> {
            AdminOperationResponse.AdminOperationResult result = null;
            try (PreparedStatement stmt = conn.prepareStatement(SQL_SELECT_OPERATION_BY_ID)) {
                stmt.setInt(1, operationId);
                    if (rs.next()) {
                        result = getResult(rs);
                    }
                }
            } catch (SQLException e) {
                throw new DsmInternalError("Error getting admin operation result for operation " + operationId, e);
            }
            return result;
        });
    }
```

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1692,7 +1693,7 @@ public static void reactivateKitRequest(Integer dsmKitRequestId, String message,
 
             long dsmKitId = KitRequestShipping.writeNewKit(dsmKitRequestId, kitRequestShipping.getEasypostAddressId(), message, false);
             kitRequestShipping.setDsmKitId(dsmKitId);
-
+            kitRequestShipping.setKitTypeName(kitRequestDao.getKitTypeByKitRequestId(dsmKitRequestId));
```
denniscunningham: There is not a way to get this in the query above (SQL_SELECT_KIT...)?

ssettipalli: SQL was used elsewhere too and we needed sampleType is re-activation code too, hence felt its cleaner this way.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/kitrequest/KitRequestDao.java
denniscunningham: It looks like you chose to not simplify this method?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
denniscunningham: But can you add those fields to the query and just ignore the additional fields in other usages?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/kitrequest/KitRequestDao.java
ssettipalli: overlooked and just edited the Exception. refactored it now.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: refactored

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -182,8 +182,8 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                     + "LEFT JOIN kit_type ty ON (req.kit_type_id = ty.kit_type_id) WHERE req.ddp_participant_id = ? "
                     + "AND realm.instance_name = ?";
     public static final String SQL_SELECT_KIT =
-            "SELECT * FROM (SELECT realm.instance_name, re.dsm_kit_request_id FROM ddp_kit_request re, ddp_instance realm "
-                    + "WHERE realm.ddp_instance_id = re.ddp_instance_id) AS request "
+            "SELECT * FROM (SELECT realm.instance_name, re.dsm_kit_request_id, kt.kit_type_name FROM ddp_kit_request re, ddp_instance realm, kit_type kt "
+                    + "WHERE realm.ddp_instance_id = re.ddp_instance_id and kt.kit_type_id = re.kit_type_id) AS request "
```
denniscunningham, line 186: Will this new restriction (where predicate) work for all users of SQL_SELECT_KIT? Is this necessary because you added columns to the select statement?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -854,8 +861,9 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
             kitRequestShipping.setDsmKitRequestId(dsmKitRequestId);
             kitRequestShipping.setDeactivationReason(deactivationReason);
             kitRequestShipping.setDeactivatedDate(deactivatedDate);
-
+            kitRequestShipping.setKitTypeName(kitRequestDao.getKitTypeByKitRequestId(dsmKitRequestId));
```
denniscunningham, line 864: Sorry, my question may have been unclear.
Is this still necessary now that you get the value via the select?

denniscunningham (commented):
A couple more questions.
I may be confused about your recent changes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: new column I needed to query is in another table and hence needed to join to other table. 
All the kit-requests should have a kit-type I believe.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: Yes Dennis. This part of code is not using that SELECT statement , actually not doing a DB query, hence need to query and get it. 
line 860: KitRequestShipping kitRequestShipping = new KitRequestShipping();

denniscunningham: I am confused then.
What was the purpose of the changes to this file in the most recent commit?

ssettipalli: Its used in other place reactivateKitRequest method and avoided extra separate query to DB. 
I thought , thats what you recommended in previous comment

denniscunningham: No, I was commenting about `kitRequestDao.getKitTypeByKitRequestId`, since it is making a second query while processing the results of the first query.
I thought perhaps you could just add the fields you needed to the first query to avoid the second query.

I'm not sure I see the other place you mentioned, but will go back through the commits to try to figure that out.

denniscunningham (approved):

PR #2909, opened 2024-07-15
PEPPER-1497 . skip elastic update if non pepper studies for kit deactivate/reactivate scenarios

ssettipalli:
PEPPER-1497
skip elastic update if non pepper studies  like eating-disorders for kit deactivate/reactivate scenarios

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
@@ -877,11 +877,11 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
             }
         } else {
             //todo check out when we need to request refund and probably remove else once we confirm its never invoked.
-            if (easypostApiKey != null) {
-                KitRequestShipping.refundKit(dsmKitRequestId, easypostApiKey, ddpInstanceDto);
-                logger.info("Requested refund for kitRequest: {} ", dsmKitRequestId);
-            }
             logger.error("DDPInstanceDto null for deactivation of kitRequest: {}", dsmKitRequestId);
+            if (easypostApiKey != null) {
+                logger.info("Requesting refund for kitRequest: {} ", dsmKitRequestId);
+                KitRequestShipping.refundKit(dsmKitRequestId, easypostApiKey, ddpInstanceDto);
+            }
         }
     }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
@@ -870,8 +870,8 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                             ESObjectConstants.DSM_KIT_REQUEST_ID, ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
                             new PutToNestedScriptBuilder()).export();
                 } catch (Exception e) {
-                    logger.error(String.format("Error updating kit request shipping deactivate reason with dsm kit request id: %s in "
-                            + "ElasticSearch. Exception: %s", dsmKitRequestId, e.getMessage()));
+                    logger.error("Error updating kit request shipping deactivate reason with dsm kit request id: {} "
+                            + " in ElasticSearch. Exception: {}", dsmKitRequestId, e.getMessage());
                     e.printStackTrace();
                 }
             }
@@ -1712,8 +1712,7 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                             ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
                             new PutToNestedScriptBuilder()).export();
                 } catch (Exception e) {
-                    logger.error(String.format("Error inserting reactivated kit request shipping with id: %s in ElasticSearch. Exception: %s ",
-                            dsmKitRequestId, e.getMessage()));
+                    logger.error("Error inserting reactivated kit request shipping with id: {} in ElasticSearch. Exception: {} ", dsmKitRequestId, e.getMessage());
                     e.printStackTrace();
                 }
             }
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
@@ -857,22 +857,24 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                     results.resultException);
         }
         if (Objects.nonNull(ddpInstanceDto)) {
-            KitRequestShipping kitRequestShipping = new KitRequestShipping();
-            kitRequestShipping.setDsmKitRequestId(dsmKitRequestId);
-            kitRequestShipping.setDeactivationReason(deactivationReason);
-            kitRequestShipping.setDeactivatedDate(deactivatedDate);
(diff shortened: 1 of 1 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -857,22 +857,24 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
...
-            try {
-                //todo: if multiple kits exist for same kit-request(dsmKitRequestId), we are updating all elements? edge case though
-                UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto,
-                        ESObjectConstants.DSM_KIT_REQUEST_ID, ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
-                        new PutToNestedScriptBuilder()).export();
-            } catch (Exception e) {
-                logger.error(String.format("Error updating kit request shipping deactivate reason with dsm kit request id: %s in "
-                        + "ElasticSearch", dsmKitRequestId));
-                e.printStackTrace();
+            if (ddpInstanceDto.isESUpdatePossible()) {
```
ssettipalli, line 860: Thought of adding this if condition to previous line if (Objects.nonNull(ddpInstanceDto)) { line# 859} but to avoid impact to the else condition , ended up adding another if.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -857,22 +857,24 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
...
+                            ESObjectConstants.DSM_KIT_REQUEST_ID, ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
+                            new PutToNestedScriptBuilder()).export();
+                } catch (Exception e) {
+                    logger.error(String.format("Error updating kit request shipping deactivate reason with dsm kit request id: %s in "
+                            + "ElasticSearch", dsmKitRequestId));
+                    e.printStackTrace();
+                }
             }
-
         } else {
```
ssettipalli, line 878: wondered if we ever get to this else ? 
when would ddpInstanceDto be null? All realms in DSM should have a ddpInstanceDto !

denniscunningham: I agree.
Looking at the callers, there is only one possibility of ddpInstanceDto being null, and that seems impossible/unlikely.
I suggest putting a 'logger.error' statement here, along with a comment that says we should remove the code in the future if the error never reports.
That at least provides an opinion about the code and protection in the unlikely case that the code path is used.

Does that make sense?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -857,22 +857,24 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
...
+                kitRequestShipping.setKitTypeName(kitRequestDao.getKitTypeByKitRequestId(dsmKitRequestId));
+                try {
+                    //todo: if multiple kits exist for same kit-request(dsmKitRequestId), we are updating all elements? edge case though
+                    UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto,
+                            ESObjectConstants.DSM_KIT_REQUEST_ID, ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
+                            new PutToNestedScriptBuilder()).export();
+                } catch (Exception e) {
+                    logger.error(String.format("Error updating kit request shipping deactivate reason with dsm kit request id: %s in "
+                            + "ElasticSearch", dsmKitRequestId));
+                    e.printStackTrace();
```
denniscunningham, line 875: The exception message should be included in the logged error.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1700,14 +1702,17 @@ public static void reactivateKitRequest(Integer dsmKitRequestId, String message,
...
-                        dsmKitRequestId));
-                e.printStackTrace();
+            if (ddpInstanceDto.isESUpdatePossible()) {
+                //skip if no elastic index. ex: juniper studies/eating-disorder
+                try {
+                    UpsertPainlessFacade.of(DBConstants.DDP_KIT_REQUEST_ALIAS, kitRequestShipping, ddpInstanceDto, ESObjectConstants.DSM_KIT_ID,
+                            ESObjectConstants.DSM_KIT_REQUEST_ID, dsmKitRequestId,
+                            new PutToNestedScriptBuilder()).export();
+                } catch (Exception e) {
+                    logger.error(String.format("Error inserting reactivated kit request shipping with id: %s in ElasticSearch",
```
denniscunningham: The exception message should be included in the logged error.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: Adding error or some log message will tell us if it ever gets there but I was more of wondering when we should try to do refund (below else code) . As it stands its probably never invoked and probably we're not requesting refunds. Wanted to check with Whitney, if there is a refund scenario at all in deactivate. If so it probably should be invoked for all deactivate and has a label assuming we get charged once label is created.
if (easypostApiKey != null) {
       KitRequestShipping.refundKit(dsmKitRequestId, easypostApiKey, ddpInstanceDto);
}

I will add a error log though

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: Its existing code and we have a e.printStackTrace() right below but I agree its helps to have exception message in alert. 
Added to the error log so that we get the message as part of the alert

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
ssettipalli: Its existing code and updated it.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
denniscunningham: BTW, you do not need `String.format` in log messages since '{}' works and makes the code easier to read.

denniscunningham (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -857,26 +857,31 @@ public static void deactivateKitRequest(int dsmKitRequestId, @NonNull String dea
...
+                }
             }
-
         } else {
+            //todo check out when we need to request refund and probably remove else once we confirm its never invoked.
             if (easypostApiKey != null) {
                 KitRequestShipping.refundKit(dsmKitRequestId, easypostApiKey, ddpInstanceDto);
+                logger.info("Requested refund for kitRequest: {} ", dsmKitRequestId);
             }
+            logger.error("DDPInstanceDto null for deactivation of kitRequest: {}", dsmKitRequestId);
```
denniscunningham: This should go at the top of this clause so that it always gets logged (in case there is an error in the intervening code).

PR #2924, opened 2024-08-08
PEPPER-1509 . sequencing-order collection date fix

ssettipalli:
PEPPER-1509

Some scenarios on collection date addition in sequencing order tab, backend is throwing 500 error.
In scenarios where a kit-request was de-activated and re-activated  new row is written into DDP_KIT table for same ddp-kit-request-id. . WHen patch is updating ddp-kit with ddp-kit-request-id, its updating multiple rows and throwing error post successful DB update.
Transaction is committed any way.
Solution: Updated to not throw error but log as a warning incase we want to track these warnings.

QA
Tested locally and looks good

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
@@ -160,7 +160,8 @@ public class Patch {
                 if (result == 1) {
                     logger.info("Updated {} value in table {}, record ID={}", nameValue.getName(), dbElement.getTableName(), id);
                 } else {
-                    logger.warn("Updated {} values in table {}, record ID={} , rows updated: {}", nameValue.getName(), dbElement.getTableName(), id, result);
+                    //logging as warn in case we want to investigate or track the scenarios where we updated more than one row.
+                    logger.warn("Updated more than one {} values in table {}, record ID={} , rows updated: {}", nameValue.getName(), dbElement.getTableName(), id, result);
                 }
             } catch (SQLIntegrityConstraintViolationException ex) {
                 throw new DsmInternalError(toErrorMessage(dbElement.getTableName(), dbElement.getColumnName(),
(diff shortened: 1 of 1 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -160,8 +160,7 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
                 if (result == 1) {
                     logger.info("Updated {} value in table {}, record ID={}", nameValue.getName(), dbElement.getTableName(), id);
                 } else {
-                    throw new DsmInternalError(toErrorMessage(dbElement.getTableName(), dbElement.getColumnName(),
```
ssettipalli, line 163: In scenarios (edge cases) where a kit-request was de-activated and re-activated, new row is written into DDP_KIT table for same ddp-kit-request-id. These multiple rows are updated when collection date is added to ddp-kit-request.
When multiple rows are updated, no point in throwing error because transaction is committed any way. Logging as warning incase we want to track these and move on.

ssettipalli:
Screenshot 2024-08-08 at 3 09 34 PM
Tested locally and looks good

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/patch/Patch.java
```diff
@@ -160,8 +160,7 @@ public static boolean patch(@NonNull String id, @NonNull String user, @NonNull N
...
-                            nameValue.getValue(), dbElement.getPrimaryKey(), id) + ": " + result + " rows updated");
+                    logger.warn("Updated {} values in table {}, record ID={} , rows updated: {}", nameValue.getName(), dbElement.getTableName(), id, result);
```
denniscunningham: It would be helpful for log readers to understand why this is a warning.
Perhaps starting the message like this would provide that information: "Updated more than one {} values..."

denniscunningham (approved):

PR #2925, opened 2024-08-09
PEPPER-1365 . Special handling of lms release v2 pdfs config with consent v1

ssettipalli:
PEPPER-1365
Handle scenarios where lms user completed release v2 but had old consent v1.

Tried the approach to update release v2 pdf config to remove reference to consent v1 . Approach was clean with one issue for DSM. Since MEDICAL_RELEASE activity was same for all self/parental and consent-assent, DataExporter was updating elastic with ALL 3 release pdfs and DSM is displaying ALL 3 pdf config names . 
Basically DSM displays some additional invalid release configs as download options. 
To avoid that issue and confusion ended with this hardcoded special handling similar to what we did for pancan. 

Other option is add special handling to remove other pdf config names which is equally hardcoded/ugly.
With this way, no pdfconfig changes (studybuilder patch) needed atleast and less invasive

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
@@ -981,7 +981,6 @@ public class DataExporter {
         if (!studyConfigs.isEmpty() && studyConfigs.get(0).getStudyGuid().equals(LMS_GUID)) {
             PdfConfigInfo lmsReleasePdfConfigV2 = getLmsReleaseV2PdfConfigIfNeeded(studyConfigs, userActivityVersions);
             if (lmsReleasePdfConfigV2 != null && !userPdfConfigs.contains(lmsReleasePdfConfigV2)) {
-                log.info("-----------------------------------> Adding LMS Release V2 PDF Config");
                 userPdfConfigs.add(lmsReleasePdfConfigV2);
             }
         }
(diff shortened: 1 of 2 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -971,6 +977,13 @@ private List<PdfConfigInfo> findPdfConfigsForStudyUser(List<PdfConfigInfo> study
             }
         }
 
+        //lms special case to handle scenarios where lms user consented-v1 but had release-v2
+        if (!studyConfigs.isEmpty() && studyConfigs.get(0).getStudyGuid().equals(LMS_GUID)) {
```
ssettipalli, line 981: special handling of pdf config with users consented - v1 and completed newer release - v2

pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/export/DataExporter.java
```diff
@@ -994,6 +1007,25 @@ private PdfConfigInfo getPancanReleaseV2PdfConfigIfNeeded(
         return releasePdfConfig;
     }
 
+    private PdfConfigInfo getLmsReleaseV2PdfConfigIfNeeded(
```
ssettipalli, line 1010: looked into other pancan special handling code added recently to avoid code duplication but activity names, pdf config names being different didn't make it feasible

zyme (approved):
LGTM

PR #2930, opened 2024-08-23
PEPPER-1364 . Additional PHI Manifest changes

ssettipalli:
PEPPER-1364
PHI manifest changes
Add new columns, replace a column and change order


zyme (approved):
LGTM!

PR #2935, opened 2024-09-16
PEPPER-1527 in person kits from juniper

zyme:
## Context

PEPPER-1527 adds a few new fields to the endpoint that juniper uses to make a kit request in order to support "in-person" kits, whereby participants complete a kit in-person, obviating the need to ship the kit to the participant. The new fields are `returnTrackingId`, `returnOnly`, and `kitLabel`.  More information about what these fields do is [here](https://github.com/broadinstitute/ddp-study-server/pull/2935/files#r1761772405).

When `returnOnly` is set, a kit is created at the same time the kit request is made, the kit is given a scan date of today, and the kit is marked as complete.  This ensures that the kit requests do not appear in DSM's "kits without label" queue.  The kit can still be queried in DSM, but since the kit does not need to be shipped, it's paramount that the kit does not appear in a worklist of kits to be mailed.

Beyond changing the json coming over from juniper, most of the changes flow from passing through these additional params through methods in `NonPepperKitCreationService`.

Along the way, I noticed that some aspects of kit request creation went across transactions, so I did some clean up to pass around `conn` so that the endpoint is a little bit more atomic.

## FUD Score

Hop over to juniper and use the admin pages to scan in a kit for a participant.  Then come back to DSM and you should be able to query the kit, but it should not be available in the kits without label view.

## Testing

## Release

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
@@ -209,7 +209,7 @@ public class KitRequestShipping extends KitRequest implements HasDdpInstanceId {
                     + "ON kit.dsm_kit_request_id = groupedKit.dsm_kit_request_id AND kit.dsm_kit_id = groupedKit.kit_id "
                     + "SET deactivated_date = ?,deactivation_reason = ?, deactivated_by = ? WHERE kit.dsm_kit_request_id = ?";
     private static final String INSERT_KIT =
-            "INSERT INTO ddp_kit (dsm_kit_request_id, easypost_address_id_to,  error, message, needs_approval) VALUES (?,?,?,?,?)";
+            "INSERT INTO ddp_kit (dsm_kit_request_id, easypost_address_id_to,  error, message, needs_approval, kit_complete, tracking_return_id, kit_label) VALUES (?,?,?,?,?,?,?,?)";
     private static final String UPDATE_KIT =
             "UPDATE ddp_kit SET label_url_to = ?, label_url_return = ?, easypost_to_id = ?, easypost_return_id = ?, tracking_to_id = ?, "
                     + "tracking_return_id = ?, easypost_tracking_to_url = ?, easypost_tracking_return_url = ?, error = ?, message = ?, "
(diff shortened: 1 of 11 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/JuniperKitRequest.java
```diff
@@ -21,4 +21,25 @@ public class JuniperKitRequest extends KitRequest {
     private String juniperParticipantID;
     private String juniperStudyID;
     private String easypostAddressId;
+
```
zyme, line 24: New fields for in-person kits.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -962,60 +963,56 @@ public static void addKitRequests(@NonNull String instanceId, @NonNull String ki
      *      4. BSP and Mercury dummy kit routes in non-prod
      * @return dsm_kit_request_id of the new kit
      * */
-    public static String writeRequest(@NonNull String instanceId, @NonNull String ddpKitRequestId, int kitTypeId,
+    public static String writeRequest(Connection conn, @NonNull String instanceId, @NonNull String ddpKitRequestId, int kitTypeId,
```
zyme, line 966: Added a few more params here so we can capture more information about the kit as part of the same transaction that writes the kit request.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -945,9 +946,9 @@ public static void addKitRequests(@NonNull String instanceId, @NonNull String ki
                     errorMessage += "bspCollaboratorSampleId was too long ";
                 }
             }
-            writeRequest(instanceId, kitRequestId, kitTypeId, participantId, collaboratorParticipantId, collaboratorSampleId, "SYSTEM",
-                    null, errorMessage, externalOrderNumber, needsApproval, uploadReason, ddpInstance, bspCollaboratorSampleType,
-                    subkitsDdpLabel);
+            writeRequest(conn, instanceId, kitRequestId, kitTypeId, participantId, collaboratorParticipantId, collaboratorSampleId,
```
ssettipalli, line 949: good that writeRequest is brought into the main transaction boundary

ssettipalli (approved):
Its good that we didn't need any schema changes and didn't need save the kits as "returnOnly" in DB minimizing the impact of changes to Non Pepper kits . 
However wondered that by not saving the kits as returnOnly in DB, in future if we have a need to find All in-person kits, do we have a easy way to distinguish them.  
Assuming Juniper is tracking it and maybe we don't need to track in DSM.

PR #2940, opened 2024-10-02
PEPPER-1389 auth0 action updates

ssettipalli:
PEPPER-1389
New changes to handle token claim updates and user AppMetadata updates post pepper/register API call
Main change is user different npm package axios and handke async/await for the registration API response.
Change works for ALL tenants similar to rule. 
Reverted some jave code changes made to UserRegistrationRoute that are NO more needed.

CMI rule has additional check to restrict users trying to login to other cmi studies using study_guid which is not needed for other non-cmi studies, hence separated the cmi rule.

Auth0 changes applied and tested in DEV. looks good.
Prion enrollment/logins work as expected.
CMI tenant restricts users from logging to different cmi studies. (Current prod behavior is retained)

Once merged will update current RC and will redeploy to `test` env.

### DIFF
--- a/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
+++ b/pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
@@ -16,7 +16,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
-import org.broadinstitute.ddp.client.ApiResult;
 import org.broadinstitute.ddp.client.Auth0ManagementClient;
 import org.broadinstitute.ddp.constants.ErrorCodes;
 import org.broadinstitute.ddp.db.TransactionWrapper;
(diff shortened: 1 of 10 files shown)


pepper-apis/dss-core/src/main/java/org/broadinstitute/ddp/route/UserRegistrationRoute.java
```diff
@@ -592,30 +591,13 @@ private void initializeProfile(Handle handle, User user, UserRegistrationPayload
             log.info("User {} has auth0 account, proceeding to sync user_metadata", user.getGuid());
             Map<String, Object> metadata = new HashMap<>();
             metadata.put(User.METADATA_LANGUAGE, languageDto.getIsoCode());
-            metadata.put("user_guid", user.getGuid());
```
ssettipalli, line 595: reverted to current prod code base.

pepper-apis/studybuilder-cli/tenants/register-cmi-user-in-pepper-action.js
```diff
@@ -221,82 +220,57 @@ exports.onExecutePostLogin = async (event, api) => {
...
-      if (pepper_params.mode && pepper_params.mode === 'login' && event.user.app_metadata.user_guid) {
-        console.log('setting userGUID claim from user.app_metadata');
-        api.idToken.setCustomClaim(pepperUserGuidClaim, event.user.app_metadata.user_guid);
-      }
-
       var pepperUrl = event.secrets.pepperBaseUrl;
       if (event.client.metadata.backendUrl) {
         pepperUrl = event.client.metadata.backendUrl;
       }
-      var request = require('request');
```
ssettipalli, line 246: removed dependency on old require and added axios for API calls

pepper-apis/studybuilder-cli/tenants/register-cmi-user-in-pepper-action.js
```diff
@@ -221,82 +220,57 @@ exports.onExecutePostLogin = async (event, api) => {
...
-          console.log('Error during registration: ' + err.message);
-          return api.access.deny(JSON.stringify(error));
-        } else if (response && response.statusCode !== 200) {
-          console.log('Response: ' + JSON.stringify(response));
+
+      var axios = require('axios');
+      console.log('Invoking DDP registration : ' + pepperUrl);
+
+      try {
+        const response = await axios.post(`${pepperUrl}/pepper/v1/register`, pepper_params, {timeout: 15000});
```
ssettipalli, line 235: used axios and async/await is what was missing before.

zyme (commented):
On L100 in pepper-apis/studybuilder-cli/tenants/register-cmi-user-in-pepper-action.js, should we deny access if `study_guid` is empty since we are expecting `study_guid` to be passed in on the query params?  It seems like this value is required going forward.  In `UserRegistrationRoute`, it looks like this might be caught when the study is queried since the return value of the lookup in the database will return null.

zyme (approved):

PR #2942, opened 2024-10-10
PEPPER-123. LMS consent_assent pdf template update.

ssettipalli:
PEPPER-123
Update Template and update new template(pdf bytes) in DB.
I added generic changes to support  updating existing pdf versioned config template bytes in DB.
Added new functionality in PdfBuilder.java

QA
Tested locally and looked good


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/PdfBuilder.java
```diff
@@ -76,6 +76,70 @@ public void updatePdfs(Handle handle) {
         }
     }
 
+    /**
+     * Update existing pdf templates (pdf bytes only) in the pdf config.
+     * Does not update the pdf config like sources/substitutions or new insertions
+     */
+    public void updatePdfTemplates(Handle handle, long studyId, Config pdfCfg, String pdfFilePath) {
```
ssettipalli, line 83: Can be used for any future pdf byte updates into DB.
Supports new versioned pdf configuration.
Can update ALL pdf templates in a config OR selected pdf file only in the config.

pepper-apis/studybuilder-cli/studies/lms/patches/consent_assent-pdf-v2.conf
```diff
@@ -0,0 +1,134 @@
+{
```
ssettipalli, line 1: just copied from existing pediatric pdf config, nothing new. 
copied consent_assent pdf config since other file contained both assent and parental pdf config.

zyme (approved):

PR #2947, opened 2024-10-31
PEPPER-1438 Request Status filed_settings value update

ssettipalli:
PEPPER-1438 
Add new status value "receivedIncomplete" to onc_history filed_settings possible values


denniscunningham (commented):
This will change lms and osteo2, but it looks like the ticket wants a change to Pancan too?

ssettipalli:
> This will change lms and osteo2, but it looks like the ticket wants a change to Pancan too?

@denniscunningham 
onc_history upload and template download features are for pe-cgs only and these changes are to support those features
However for pancan new status will be available to add and search, those changes are handled in dsm angular.
Thanks

denniscunningham (approved):

PR #2957, opened 2024-12-18
PEPPER-1562 . PHI manifest update

ssettipalli:
PEPPER-1562 
PHI manifest update to make collaborator participant Id is included in the scenario where there is tumor only orders.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
@@ -190,28 +190,19 @@ public class KitDao {
         return booleanCheckFoundAsName(ddpLabel, SQL_IS_BLOOD_KIT_QUERY);
     }
 
-    public String getCollaboratorParticipantId(String ddpParticipantId) {
-        if (StringUtils.isBlank(ddpParticipantId)) {
-            return null;
-        }
-
-        SimpleResult results = inTransaction(conn -> {
-            SimpleResult dbVals = new SimpleResult("");
+    public String getCollaboratorParticipantId(String ddpParticipantId) throws Exception {
+        return inTransaction(conn -> {
             try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_PARTICIPANT_ID)) {
                 stmt.setString(1, ddpParticipantId);
                 try (ResultSet rs = stmt.executeQuery()) {
                     if (rs.next()) {
-                        dbVals.resultValue = rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
+                        logger.debug("Found collaborator participant Id {} for ddpParticipant {} ", ddpParticipantId);
+                        return rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
                     }
                 }
-            } catch (SQLException ex) {
-                dbVals.resultException = new DsmInternalError("Error getting collaborator participant Id for participant: " + ddpParticipantId, dbVals.resultException);
             }
-            logger.debug("Found collaborator participant Id {} for ddpParticipant {} ", dbVals.resultValue, ddpParticipantId);
-            return dbVals;
+            return null;
         });
-
-        return results.resultValue.toString();
     }
 
     public Boolean hasTrackingScan(String kitLabel) {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
@@ -201,9 +201,6 @@ public class KitDao {
                         return rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
                     }
                 }
-            } catch (SQLException ex) {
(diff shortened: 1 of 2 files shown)


denniscunningham (commented):
Is there a way to test this via PhiManifestTest?

ssettipalli:
> Is there a way to test this via PhiManifestTest?

tested it locally running DSM and having execution go through new code. 
Considering a very edge case scenario, we should be good. prod has 2 cases that the code execution will go through new code to get it.
Test seem to be focusing on values from elastic and this field is being retrieved from DB

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
```diff
@@ -186,6 +190,30 @@ public Boolean isBloodKit(String ddpLabel) {
...
+            SimpleResult dbVals = new SimpleResult("");
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_PARTICIPANT_ID)) {
+                stmt.setString(1, ddpParticipantId);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
+                    }
+                }
+            } catch (SQLException ex) {
+                dbVals.resultException = new DsmInternalError("Error getting collaborator participant Id ", dbVals.resultException);
```
denniscunningham: Including the ddpParticipantId in the message will help debugging.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
```diff
@@ -186,6 +190,30 @@ public Boolean isBloodKit(String ddpLabel) {
         return booleanCheckFoundAsName(ddpLabel, SQL_IS_BLOOD_KIT_QUERY);
     }
 
+    public String getCollaboratorParticipantId(String ddpParticipantId) {
+        if (StringUtils.isBlank(ddpParticipantId)) {
+            return null;
```
denniscunningham: Is null a useful return value? I see the caller is immediately assigning this value.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
```diff
@@ -186,6 +190,30 @@ public Boolean isBloodKit(String ddpLabel) {
...
+            SimpleResult dbVals = new SimpleResult("");
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_PARTICIPANT_ID)) {
+                stmt.setString(1, ddpParticipantId);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
+                    }
+                }
+            } catch (SQLException ex) {
+                dbVals.resultException = new DsmInternalError("Error getting collaborator participant Id for participant: " + ddpParticipantId, dbVals.resultException);
```
denniscunningham: This code effectively swallows this exception.
I suggest either throwing an exception, or logging a ERROR and returning a value the caller can use, perhaps NULL as above.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
```diff
@@ -186,6 +190,30 @@ public Boolean isBloodKit(String ddpLabel) {
...
+            return null;
+        }
+
+        SimpleResult results = inTransaction(conn -> {
+            SimpleResult dbVals = new SimpleResult("");
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_PARTICIPANT_ID)) {
+                stmt.setString(1, ddpParticipantId);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = rs.getString(DBConstants.COLLABORATOR_PARTICIPANT_ID);
```
denniscunningham: This code would be clearer if you removed the indirection through `SimpleResults`, and just returned a value here. (You can also add the logging from below here.) That way anything that falls through the bottom of these conditionals can be handled uniformly. 
Likewise you can either catch and rethrow exceptions below, or not handle exceptions at this level, depending on your preferred error handling approach.

denniscunningham (changes requested):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
ssettipalli: not really needed. cleaned up

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
ssettipalli: refactored and cleaned up

denniscunningham (approved):

PR #2961, opened 2025-01-09
PEPPER-1088 . restrict onc history upload if not consented to tissue sample

ssettipalli:
PEPPER-1088

onc history upload change to restrict onc history upload if not consented to tissue sample

Added unit test and tested the code by running DSM locally


ssettipalli:
Local testing screenshots
Screenshot 2025-01-09 at 11 59 59 AM
Screenshot 2025-01-09 at 12 01 00 PM

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -174,6 +181,12 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
                     + "Please remove onc histories for these withdrawn participants from the file and upload it again: " + exitedParticipants);
         }
 
+        if (!nonTissueConsentedParticipants.isEmpty()) {
+            log.error("Found {} participants: {} in Onc History Upload who did not consent to tissue sample", nonTissueConsentedParticipants.size(), nonTissueConsentedParticipants);
```
denniscunningham: You do not want to log at error level: this is a user error, not a system error, and this will generate an alert in GCP. Likewise for the exit code you added previously.
Also, I'm not sure why there are two separate messages here since the user should have the same information as the log. I suggest creating a single message a pass it to logging and the exception. (Likewise for the exit code.)

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -174,6 +181,12 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
...
+            throw new OncHistoryValidationException("One or more of the uploaded onc histories is associated with a participant who did not consent to tissue sample. "
+                    + "Please remove onc histories for these participants from the file and upload it again: " + nonTissueConsentedParticipants);
+        }
```
denniscunningham, line 199: What does this message look like? Does it make sense to format the list?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -351,6 +360,11 @@ private static Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> re
                 // expecting an exit record and exception
                 Assert.assertTrue(e.getMessage().contains("One or more of the uploaded onc histories is associated with a withdrawn participant"));
                 return null;
+            } else if (records.get(0).getParticipantTextId().startsWith("abc-no-tissue-consent")) {
+                e.printStackTrace();
```
denniscunningham: Is this stack trace for debugging?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -351,6 +360,11 @@ private static Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> re
                 // expecting an exit record and exception
                 Assert.assertTrue(e.getMessage().contains("One or more of the uploaded onc histories is associated with a withdrawn participant"));
                 return null;
+            } else if (records.get(0).getParticipantTextId().startsWith("abc-no-tissue-consent")) {
```
denniscunningham: This code does not check the case when the exception is not thrown, so does not effectively demonstrate that the feature works properly. Likewise for the exit case.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -188,6 +193,10 @@ private void writeToDb(String realm, String testFile) {
             // expecting an exit record and exception which was checked in getParticipantIds
             return;
         }
+        if (rows.get(0).getParticipantTextId().startsWith("abc-no-tissue-consent")) {
```
denniscunningham: I'm not sure I understand what this code does.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -174,6 +181,12 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
                     + "Please remove onc histories for these withdrawn participants from the file and upload it again: " + exitedParticipants);
         }
 
+        if (!nonTissueConsentedParticipants.isEmpty()) {
```
denniscunningham: This code is going to give inconsistent results depending on the order of records in the file. It also has the possibility of requiring multiple passes to detect all the validation errors. 
I suggest checking all records for both exited ptps and those ptps that did not consent to tissue.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -149,7 +151,12 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
...
-            Long ptpId = ptpData.getDsm().get().getParticipant().get().getParticipantId();
+            Dsm dsmData = ptpData.getDsm().get();
+            if (!dsmData.isHasConsentedToTissueSample()) {
+                nonTissueConsentedParticipants.add(rec.getParticipantTextId());
+                continue;
+            }
+            Long ptpId = dsmData.getParticipant().get().getParticipantId();
             int participantId;
             try {
                 participantId = ptpId.intValue();
```
denniscunningham, line 153: This code seems like it should run before the check for a no tissue ptp, since this is effectively a system error that will be overlooked as is.

denniscunningham (changes requested):

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
ssettipalli: I am not sure I understand this comment correctly. we have a test case where onc histories will be loaded and verified. That should cover positive scenario and functionality, right ?

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
ssettipalli: Its the same code used for all unit tests and hence checking for no-tissue-consent record and skipping further check for actual onc history records creation in DB. Does that explain?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
ssettipalli: Format looks good. PR has one screenshot with how error is displayed to users. 
For multiple shortIds it looks like this
One or more of the uploaded onc histories is associated with a participant who did not consent to tissue sample. Please remove onc histories for these participants from the file and upload it again: [abc-no-tissue-consent, HRUID-2]

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -163,15 +161,34 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
...
         }
 
         if (!exitedParticipants.isEmpty()) {
-            log.error("Found {} exited participants: {} in Onc History Upload", exitedParticipants.size(), exitedParticipants);
-            throw new OncHistoryValidationException("One or more of the uploaded onc histories is associated with a withdrawn participant. "
-                    + "Please remove onc histories for these withdrawn participants from the file and upload it again: " + exitedParticipants);
+            String message = String.format("One or more of the uploaded onc histories is associated with a withdrawn participant. "
+                    + "Please remove onc histories for these withdrawn participants from the file and upload it again: %s", exitedParticipants);
+            log.warn(message);
+            throw new OncHistoryValidationException(message);
```
denniscunningham: This code is now better in detecting both exited ptps and ptps who did not consent, but the reporting is still separate. I suggest collecting information for both types of validation failures and reporting them together. Perhaps something like "The following errors were detected with the upload:\n One or more withdraw.... \n One or more participants who did not...".

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
denniscunningham: I think a separate test for exited and non-consent would make more sense. I'm not sure why the exited case was added here.
My point is that if the code under test fails to detect an exited ptp, for instance, this test will still pass.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadService.java
```diff
@@ -163,15 +161,43 @@ protected Map<Integer, Integer> getParticipantIds(List<OncHistoryRecord> oncHist
...
                     this.realm, updateElastic);
             medIds.put(participantId, medId);
         }
 
-        if (!exitedParticipants.isEmpty()) {
-            log.error("Found {} exited participants: {} in Onc History Upload", exitedParticipants.size(), exitedParticipants);
-            throw new OncHistoryValidationException("One or more of the uploaded onc histories is associated with a withdrawn participant. "
-                    + "Please remove onc histories for these withdrawn participants from the file and upload it again: " + exitedParticipants);
+        if (!exitedParticipants.isEmpty() || !nonTissueConsentedParticipants.isEmpty()) {
+            StringBuilder message = new StringBuilder("One or more of the uploaded onc histories is associated with ");
```
denniscunningham, line 180: I was thinking of something simpler, with a header and a list of problems (1 or 2 problems), but his works.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -160,6 +168,36 @@ public void testGetParticipantIdForTextId() {
...
+
+        try {
+            uploadService.getParticipantIds(rows, new TestParticipantIdProvider(shortIdToId), false);
+        } catch (Exception e) {
+            // expecting an exit record / not tissue consented record and exception
+            Assert.assertTrue(e.getMessage().contains(expectedMessage));
+            return;
+        }
+
+        Assert.fail("Expected exception");
```
denniscunningham: You can make the logic more evident by moving this final assert to line 192. That way you can also remove the return statement at line 195.

pepper-apis/dsm-core/src/test/java/org/broadinstitute/dsm/service/onchistory/OncHistoryUploadServiceTest.java
```diff
@@ -100,13 +100,21 @@ public void testWriteToDb() {
     }
 
     @Test
-    public void testLmsWriteToDb() {
```
denniscunningham: Did we lose a test here: testLmsWriteToDb?

denniscunningham (commented):
Looks good!
Just want to be sure we did not lose a test before approving.

denniscunningham (approved):

PR #2972, opened 2025-05-16
PEPPER-1613 LMS adult consent version 3

ssettipalli:
PEPPER-1613
Update LMS consent text, revision to v3


pepper-apis/studybuilder-cli/src/main/java/org/broadinstitute/ddp/studybuilder/task/RevisionStudyActivityVariablesSupport.java
```diff
@@ -0,0 +1,165 @@
+package org.broadinstitute.ddp.studybuilder.task;
```
ssettipalli, line 1: Generic class which can be extended by other activity version classes. 
Currently supports only versioning existing template variables

Rinchan20 (approved):

PR #2999, opened 2025-09-10
PEPPER-1642 sample counter offsets

zyme:
## Context

Overall motivation and design is in [a google doc here](https://docs.google.com/document/d/1Zwxc1h2iEP6cb5c-dYzsxwgSyKd5tlhGdOPJ1tSI7fU/edit?tab=t.0#heading=h.r7imonaoa15y).

The new  [`sample_counter_offset`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#r2337532697) is the way DSM sets its sample counters and legacy collaborator participant ids for kits that were tracked outside of DSM.  When there are values in this table for a `ddp_instance`, DSM uses these values to determine the sample id count suffix.

The main changes are in the [`generateBspSampleID()`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#r2337536498) and [`getCollaboratorParticipantId()`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#r2337538049) methods of `KitRequestShipping`.  These methods consult `sample_counter_offset` to figure how how to set kit counters and collaborator sample id.  If these values aren't present, DSM uses its previous logic, which involves looking up previous kits in kit request tables, and generating a collaborator participant id based on the ddp_instance's prefix value.

[`SampleCounterOffsets`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#diff-11aa855590561cf112f920a092bc84ef72f1f62999f4300ac463a764c594c298), [`SampleCounterOffset`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#diff-578e10eb2dab982b4e2788f7c5f53ed5054b875d43b39ea9d6654e8be2ced66c), and [`SampleCounterOffsetDao`](https://github.com/broadinstitute/ddp-study-server/pull/2999/files#diff-8107427ae4af3ef52174051491cb69fc4d6bc5d430765a0db812726749adf993) do the supporting work here, and there are a few new automated tests focused on sample id and collab participant id generation in a few different scenarios.  Further integration testing is needed and will be done manually as we prepare for the A-T migration.

I had to change`EventServiceTest` to use explicit matchers due to keep the tests passing.  Without these changes, I had to resort to polluting `DDPInstance` with yet more builder-specific code.  I chose to add complexity to `EventServiceTest` rather than to add it to DDPInstance and associated builders.

After A-T is migrated to Juniper, we will seed the `sample_counter_offset` with an extract of Juniper's new short ids, the legacy collaborator_participant_ids, and per-kit-type counts (mostly 1s, perhaps a few 2s) by query DSM's genomic sample id custom fields and reviewing these with Alex.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -13,6 +13,7 @@ import java.util.List;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.Getter;
@@ -31,6 +32,7 @@ import org.slf4j.LoggerFactory;
 
 @Data
 @Builder(toBuilder = true, setterPrefix = "with")
+@AllArgsConstructor()
 public class DDPInstance {
 
     public static final String SQL_SELECT_ALL_ACTIVE_REALMS =
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -108,7 +108,7 @@ public class DDPInstance {
                        int daysMrAttentionNeeded, int daysTissueAttentionNeeded, boolean hasAuth0Token, List<String> notificationRecipient,
                        boolean migratedDDP, String billingReference, String participantIndexES, String activityDefinitionIndexES,
                        String usersIndexES, String researchProject, String displayName, String mercuryOrderCreator,
-                       InstanceSettings instanceSettings, String studyGuid, LegacyKits legacyKits) {
+                       InstanceSettings instanceSettings, String studyGuid) {
         this.ddpInstanceId = ddpInstanceId;
         this.name = name;
         this.baseUrl = baseUrl;
@@ -128,11 +128,6 @@ public class DDPInstance {
         this.displayName = displayName;
         this.instanceSettings = instanceSettings;
         this.studyGuid = studyGuid;
-        if (legacyKits == null) {
-            this.legacyKits = new LegacyKits(Collections.emptyList());
-        } else {
-            this.legacyKits = legacyKits;
-        }
     }
 
 
@@ -430,7 +425,7 @@ public class DDPInstance {
                 rs.getString(DBConstants.BILLING_REFERENCE), rs.getString(DBConstants.ES_PARTICIPANT_INDEX),
                 rs.getString(DBConstants.ES_ACTIVITY_DEFINITION_INDEX), rs.getString(DBConstants.ES_USERS_INDEX),
                 rs.getString(DBConstants.RESEARCH_PROJECT), rs.getString(DBConstants.DISPLAY_NAME),
-                rs.getString(DBConstants.MERCURY_ORDER_CREATOR), null, rs.getString(DBConstants.STUDY_GUID), null);
+                rs.getString(DBConstants.MERCURY_ORDER_CREATOR), null, rs.getString(DBConstants.STUDY_GUID));
     }
 
     private static DDPInstance getDDPInstanceFormResultSet(@NonNull ResultSet rs) throws SQLException {
@@ -447,7 +442,7 @@ public class DDPInstance {
                 rs.getString(DBConstants.BILLING_REFERENCE), rs.getString(DBConstants.ES_PARTICIPANT_INDEX),
                 rs.getString(DBConstants.ES_ACTIVITY_DEFINITION_INDEX), rs.getString(DBConstants.ES_USERS_INDEX),
                 rs.getString(DBConstants.RESEARCH_PROJECT), rs.getString(DBConstants.DISPLAY_NAME),
-                rs.getString(DBConstants.MERCURY_ORDER_CREATOR), null, rs.getString(DBConstants.STUDY_GUID), null);
+                rs.getString(DBConstants.MERCURY_ORDER_CREATOR), null, rs.getString(DBConstants.STUDY_GUID));
     }
 
     //assumption: base url of pepper studies will always end like: dsm/studies/<STUDYNAME>
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -113,7 +113,7 @@ public class DDPInstance {
      */
     @Getter(value = AccessLevel.NONE)
     @Setter(value = AccessLevel.NONE)
-    private SampleCounterOffsets sampleCounterOffsets = new SampleCounterOffsets(new ArrayList<>());
+    private final SampleCounterOffsets sampleCounterOffsets = new SampleCounterOffsets(new ArrayList<>());
 
     public DDPInstance(String ddpInstanceId, String name, String baseUrl, String collaboratorIdPrefix, boolean hasRole,
                        int daysMrAttentionNeeded, int daysTissueAttentionNeeded, boolean hasAuth0Token, List<String> notificationRecipient,
@@ -573,7 +573,7 @@ public class DDPInstance {
      */
     @VisibleForTesting
     public void setSampleCounterOffsets(SampleCounterOffsets sampleCounterOffsets) {
-        this.sampleCounterOffsets = sampleCounterOffsets;
+        this.sampleCounterOffsets.initFrom(sampleCounterOffsets);
     }
 
     // todo arz move out of this class
@@ -585,14 +585,27 @@ public class DDPInstance {
      */
     public static class SampleCounterOffsets {
 
-        private Map<String, SampleCounterOffset> sampleCounterOffsetsByShortId = new HashMap<>();
+        private final Map<String, SampleCounterOffset> sampleCounterOffsetsByShortId = new HashMap<>();
 
-        private Map<String, SampleCounterOffset> sampleCounterOffsetsByCollabParticipantId = new HashMap<>();
+        private final Map<String, SampleCounterOffset> sampleCounterOffsetsByCollabParticipantId = new HashMap<>();
 
         public SampleCounterOffsets(List<SampleCounterOffset> sampleCounterOffsets) {
             for (SampleCounterOffset sampleCounterOffset : sampleCounterOffsets) {
                 sampleCounterOffsetsByCollabParticipantId.put(sampleCounterOffset.getLegacyCollaboratorParticipantId(), sampleCounterOffset);
                 sampleCounterOffsetsByShortId.put(sampleCounterOffset.getShortId(), sampleCounterOffset);
+            }
+        }
+
+        public void clear() {
+            this.sampleCounterOffsetsByCollabParticipantId.clear();
+            this.sampleCounterOffsetsByShortId.clear();
+        }
+
+        public void initFrom(SampleCounterOffsets copyFrom) {
+            clear();
+            if (copyFrom != null) {
+                sampleCounterOffsetsByShortId.putAll(copyFrom.sampleCounterOffsetsByShortId);
+                sampleCounterOffsetsByCollabParticipantId.putAll(copyFrom.sampleCounterOffsetsByCollabParticipantId);
             }
         }
 
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -102,7 +102,8 @@ public class DDPInstance {
     private final String displayName;
     private InstanceSettings instanceSettings;
     private final String studyGuid;
-    private LegacyKits legacyKits = new LegacyKits(Collections.emptyList());
+    // marked final to avoid state management complexity with lombok Builder
+    private final LegacyKits legacyKits = new LegacyKits(Collections.emptyList());
 
     public DDPInstance(String ddpInstanceId, String name, String baseUrl, String collaboratorIdPrefix, boolean hasRole,
                        int daysMrAttentionNeeded, int daysTissueAttentionNeeded, boolean hasAuth0Token, List<String> notificationRecipient,
@@ -559,7 +560,8 @@ public class DDPInstance {
      */
     @VisibleForTesting
     public void setLegacyKits(LegacyKits legacyKits) {
-        this.legacyKits = legacyKits;
+        // legacyKits is final to avoid state management complexity from this class's use of lombok builder
+        this.legacyKits.initFrom(legacyKits);
     }
 
     public static class LegacyKits {
@@ -581,6 +583,19 @@ public class DDPInstance {
 
         public LegacyKitSummary getKitSummaryByCollaboratorParticipantId(String collabParticipantId) {
             return legacyKitsByCollabParticipantId.get(collabParticipantId);
+        }
+
+        public void clear() {
+            this.legacyKitsByCollabParticipantId.clear();
+            this.legacyKitsByShortId.clear();
+        }
+
+        public void initFrom(LegacyKits copyFrom) {
+            clear();
+            if (copyFrom != null) {
+                legacyKitsByShortId.putAll(copyFrom.legacyKitsByShortId);
+                legacyKitsByCollabParticipantId.putAll(copyFrom.legacyKitsByCollabParticipantId);
+            }
         }
 
         public static class LegacyKitSummary {
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
@@ -105,6 +105,12 @@ public class DDPInstance {
     private InstanceSettings instanceSettings;
     private final String studyGuid;
 
+    /**
+     * Sample id offsets control the number at which to start
+     * the sample id suffix when generating sample ids
+     * for kits.  In the absence of this value, the suffix will
+     * start at 0.
+     */
     @Getter(value = AccessLevel.NONE)
     @Setter(value = AccessLevel.NONE)
     private final LegacyKits legacyKits = new LegacyKits(new ArrayList<>());
@@ -548,6 +554,10 @@ public class DDPInstance {
      * consistent participant id naming for sequencing data that was generated prior to juniper, we use the legacy participant id.
      */
     public LegacyKits getLegacyKits() { // todo arz rename to indicate this study was on pepper and now on juniper vs. was on datstat and now on pepper
+        return legacyKits;  // getLegacyKitsForJuniper kitCounterOffsets
+    }
+
+    public void loadLegacyKitsFromDb() {
         // todo arz query via new legacy_kits table
 
         /*
@@ -555,7 +565,6 @@ public class DDPInstance {
             legacyKits = queryLegacyKitsFromDb();
         }
         */
-        return legacyKits;
     }
 
     /**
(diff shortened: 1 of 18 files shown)


pepper-apis/dsm-core/src/main/resources/liquibase/PEPPER-1642_sample_counter_offsets.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="andrew" id="PEPPER-1642_sample_counter_offsets">
```
zyme, line 8: New table here.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1316,7 +1316,21 @@ public static String generateBspSampleID(@NonNull Connection conn, String collab
                 collaboratorSampleId += "_" + type;
             }
 
-            int counter = getKitCounter(conn, collaboratorSampleId, kitTypeId);
+            int counter = 0;
```
zyme, line 1319: Changes for generateBspSampleID()

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1838,6 +1852,18 @@ public static String getCollaboratorParticipantId(DDPInstance ddpInstance, Strin
             return handleGen2MigratedKit(instanceId, ddpParticipantId, shortId, collaboratorIdPrefix,
                     collaboratorParticipantLengthOverwrite);
         }
+
```
zyme, line 1855: Changes for getCollaboratorParticipantId()

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1838,6 +1852,18 @@ public static String getCollaboratorParticipantId(DDPInstance ddpInstance, Strin
...
+        SampleCounterOffsets sampleCounterOffsets = ddpInstance.getSampleCounterOffsets();
+        if (sampleCounterOffsets != null) {
+            // if this study is using sample kit offsets, use them if you have them for the participant
+            SampleCounterOffset legacyKitsSummary = sampleCounterOffsets.getKitSummaryByShortId(shortId);
```
denniscunningham: The instance name here is oddly combined with the type, which does not help readability.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
```diff
@@ -0,0 +1,70 @@
...
+                    offsetForShortCode.setSampleCounterOffsetForKitTypeId(kitTypeId, offset);
+                }
+            }
+        }
+        return offsets;
+    }
+
+    /**
+     * Return the counter offset for the given shortcode in the given
+     * list of shortcodes
```
denniscunningham: It looks like this method takes a list of offsets, not shortcodes.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class SampleCounterOffsetDao {
+
+    private static final String QUERY_ALL_OFFSETS_FOR_INSTANCE =
+            "select o.participant_short_code, o.legacy_collaborator_participant_id, o.kit_type_id, o.offset "
+            + "from sample_counter_offset o, ddp_instance i, kit_type kt "
+            + "where o.ddp_instance_id = i.ddp_instance_id and kt.kit_type_id = o.kit_type_id "
+            + "and i.ddp_instance_id = ?";
+
+    public static List<SampleCounterOffset> querySampleCounterOffsetsForInstance(Connection conn, int ddpInstanceId) throws SQLException {
```
denniscunningham: Nit, but typically this method would be named `get....`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+        return offsets;
+    }
+
+    /**
+     * Return the counter offset for the given shortcode in the given
+     * list of shortcodes
+     */
+    public static SampleCounterOffset getOffsetForShortCode(String participantShortCode, List<SampleCounterOffset> offsets) {
```
denniscunningham: Is this public for testing?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
```diff
@@ -0,0 +1,70 @@
...
+        try (PreparedStatement stmt = conn.prepareStatement(QUERY_ALL_OFFSETS_FOR_INSTANCE)) {
+            stmt.setInt(1, ddpInstanceId);
+            try (ResultSet rs = stmt.executeQuery()) {
+                while (rs.next()) {
+                    String participantShortCode = rs.getString(PARTICIPANT_SHORT_CODE);
+                    String collaboratorParticipantId = rs.getString(LEGACY_COLLABORATOR_PARTICIPANT_ID);
+                    Integer kitTypeId = rs.getInt(DBConstants.KIT_TYPE_ID);
+                    Integer offset = rs.getInt(KIT_COUNTER_OFFSET);
+
+                    SampleCounterOffset offsetForShortCode = getOffsetForShortCode(participantShortCode, offsets);
```
denniscunningham, line 40: This deserves a comment since it is not obvious what's going on. 
When an `offsetForShortCode` is found, it is not added to the result?
Is this something that could not be added to the query?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
```diff
@@ -200,53 +200,33 @@ private void addJuniperKitRequest(Connection conn, String kitTypeName, KitReques
             throw new DsmInternalError("EasyPost addressId could not be received for " + kit.getJuniperKitId(), e);
         }
 
-        if (StringUtils.isNotBlank(kitRequestSettings.getExternalShipper())) {
```
denniscunningham, line 203: Was this case not valid?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -28,6 +32,7 @@
 
 @Data
 @Builder(toBuilder = true, setterPrefix = "with")
+@AllArgsConstructor
```
denniscunningham, line 35: Is this constructor used somewhere?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -1316,7 +1316,21 @@ public static String generateBspSampleID(@NonNull Connection conn, String collab
                 collaboratorSampleId += "_" + type;
             }
 
-            int counter = getKitCounter(conn, collaboratorSampleId, kitTypeId);
+            int counter = 0;
+
+            SampleCounterOffsets sampleCounterOffsets = ddpInstance.getSampleCounterOffsets();
+            SampleCounterOffset sampleCounterOffset = null;
```
denniscunningham, line 1322: This logic can be refactored for to improve readability.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
```diff
@@ -0,0 +1,46 @@
+package org.broadinstitute.dsm.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Each kit type can have its own offset for a participant.  In addition,
```
denniscunningham: Perhaps explaining what an offset is/is used for would help this comment.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
```diff
@@ -0,0 +1,46 @@
...
+ * the offset can declare a collaboratorParticipantId, which can be used
+ * instead of the prefix + hruid based naming for participants whose
+ * kits predate DSM.
+ */
+public class SampleCounterOffset {
+
+    private String legacyCollaboratorParticipantId;
```
denniscunningham: Are collaborator participant and participant short id not mapped elsewhere in the DB?
Does the term `legacy` add any value here?
Would changing the `shortId` to `participantShortId` help readability?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private String shortId;
+
+    // key is the kit type id, value is the sample counter offset for that kit type
```
denniscunningham, line 24: Using the term offset here instead of num would help readability, and jive with `setSampleCounterOffsetForKitTypeId`.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
```diff
@@ -0,0 +1,46 @@
+package org.broadinstitute.dsm.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Each kit type can have its own offset for a participant.  In addition,
+ * the offset can declare a collaboratorParticipantId, which can be used
```
denniscunningham: It says 'can declare' but `collaboratorParticipantId` is required.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
```diff
@@ -0,0 +1,70 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.broadinstitute.dsm.statics.DBConstants.KIT_COUNTER_OFFSET;
```
denniscunningham: These are only used here. Is there a reason to have them defined globally?

denniscunningham (commented):
I have some questions so I can better understand the approach.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -532,4 +546,25 @@ public static DDPInstance from(DDPInstanceDto ddpInstanceDto) {
...
+     * used a one-off participant id generator specific to A-T.  To ensure
+     * consistent participant id and sample id naming
+     * for sequencing data that was generated prior to juniper,
+     * use the sample counter offsets.
+     */
+    public SampleCounterOffsets getSampleCounterOffsets() {
+        if (sampleCounterOffsets == null) {
+            sampleCounterOffsets = TransactionWrapper.inTransaction(conn -> {
+                try {
+                    return new SampleCounterOffsets(SampleCounterOffsetDao.querySampleCounterOffsetsForInstance(conn, getDdpInstanceIdAsInt()));
```
denniscunningham: Somewhere we setup sample counter offsets for all DDP instances?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -532,4 +546,25 @@ public static DDPInstance from(DDPInstanceDto ddpInstanceDto) {
...
     }
+
+    /**
+     * For older studies (A-T, for example), kits created prior to juniper
+     * used a one-off participant id generator specific to A-T.  To ensure
+     * consistent participant id and sample id naming
+     * for sequencing data that was generated prior to juniper,
+     * use the sample counter offsets.
+     */
+    public SampleCounterOffsets getSampleCounterOffsets() {
```
denniscunningham, line 557: So this is effectively a cache of sample counter offsets?
Did you explore just getting this data as needed, since I assume that would simplify the code?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
zyme: Yes, [in existing autogenerated mock testing code here](https://app.circleci.com/pipelines/github/broadinstitute/ddp-study-server/22684/workflows/37cce263-6304-4da0-8847-6022aef98e3e/jobs/45263).  Without this annotation, there compile-time errors related to

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
zyme: I did--and current use is limited to `JuniperShipKitRoute`.  I decided against putting special setter logic there so that any future use of this capability would not require changes to set this data.  Given the proliferation of constructors and the number of additional files and tests that would have to change to add this to constructor params to further future-proof the code, I opted for this approach.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
zyme: At the moment we set this in a one-off sql script based on a few queries from DSM and Juniper, just for A-T.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
zyme: I'm open to suggestions--do you have specifics in mind?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
zyme: It's the collab participant id that combines the the instance's given collab participant prefix with the kit type.  I'm not sure I follow what you mean by the instance name being combined with the type.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
zyme: It's marked as nullable in the schema--did I miss some code that implies otherwise?

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/SampleCounterOffset.java
zyme: The values live in kit request and kit tables, but as strings related to kits and kit requests, not as properties of a participant object.  Will change to remove `legacy`, as that was a holdover from a previous iteration before I realized that `legacy` is indeed not helpful and might cause more confusion here.  For `shortId`, the pattern I've seen elsewhere in code makes me think that `shortId` is only ever used as a short id for the participant--nothing else seems to use this term.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
zyme: I've been trying to follow what appears to be a general pattern of putting column names into `DBConstants`.  Not sure if that's a good pattern to follow at the moment, though.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
zyme: Ah yes!  I had moved this around during some refactors and agree that since it's in a dao, `get` is more appropriate.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
zyme: Will add a comment to explain that there is only ever at most one offset per participant + kit type in the context of a ddp_instance, so they don't sum.  They do get added to the dynamic count of previous kits in pepper-NonPepperKitCreationService.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
zyme: In a previous version, yes, but now this should be private since the testing is happening up a few levels.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/nonpepperkit/NonPepperKitCreationService.java
zyme: Ah thank you for asking about this--I forgot to explain it in previous PR comments.  The "external shipper" business was only used for a previous (now completed) study that used a different carrier from easypost.  The external carrier was a company called GBF, and we won't be working with them via DSM again.  This looks like it was copy-pasted into this class when the class was created.  Given the way that GP handles these sorts of kits now, it is extremely unlikely that Juniper in particular would ever make use of DSM's "non-easypost" (aka "external") shipping mechanism, so I decided to remove this to avoid distraction.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
denniscunningham: I can't think of a case where these constants would be useful outside of this DAO.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
denniscunningham: Typically you would expect an instance name like `sampleCounterOffset`. By using a completely different name I assume you want to convey additional information, but that information was not clear to me.

denniscunningham (approved):

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
zyme: Will change the name as suggested--I missed this cleanup from an earlier refactor.

pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/SampleCounterOffsetDao.java
zyme: Agreed, will move these to the dao.

PR #3003, opened 2025-10-24
Upgrading ES to v8

zyme:
Elastic v7 is EOL at the end of the year.  This PR includes changes to include special "preserve v7 behavior" headers when running against v8 ([elastic doc here](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/7.17/migrate-hlrc.html)) and updates the docker image in integration tests to use v8.  

I _think_ I found all the places in code where this flag needs to be enabled.  Experiments on test and staging environments are looking promising, and I'm grateful for all the existing tests that hit ES--thanks Dennis!  All these tests pass, which has been hugely reassuring.

Along the way, I merged up the changes that had been on test/staging/prod branches to downgrade our pepper API backends from two instances to one, since we are seeing such low load overall.

Dev/staging/and test ES instances are now running v8.  The migration process has been pretty smooth, so the rollout plan to prod would be perhaps 30m of downtime while ES is upgraded, and then deploying DSS and DSM with these changes.

### DIFF
--- a/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/mapping/FieldTypeExtractor.java
+++ b/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/mapping/FieldTypeExtractor.java
@@ -7,6 +7,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import org.broadinstitute.dsm.exception.DsmInternalError;
 import org.broadinstitute.dsm.model.elastic.export.generate.MappingGenerator;
 import org.broadinstitute.dsm.util.ElasticSearchUtil;
 import org.elasticsearch.client.RequestOptions;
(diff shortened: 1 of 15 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/model/elastic/mapping/FieldTypeExtractor.java
```diff
@@ -25,11 +28,15 @@ public class FieldTypeExtractor implements TypeExtractor<Map<String, String>> {
...
-            for (Map.Entry<String, GetFieldMappingsResponse.FieldMappingMetadata> entry : mapping.entrySet()) {
-                fieldTypeMapping.put(getRightMostFieldName(entry.getKey()), extractType(entry.getKey(), entry.getValue()));
+            if (mapping != null) {
+                Map<String, String> fieldTypeMapping = new HashMap<>();
+                for (Map.Entry<String, GetFieldMappingsResponse.FieldMappingMetadata> entry : mapping.entrySet()) {
+                    fieldTypeMapping.put(getRightMostFieldName(entry.getKey()), extractType(entry.getKey(), entry.getValue()));
+                }
+                cachedFieldTypes.putAll(fieldTypeMapping);
+            } else {
+                logger.warn("No mapping found for index {}", index);
```
zyme, line 38: This has been helpful as part of diagnosing v7 -> v8 issues.

PR #3004, opened 2025-11-17
PEPPER-1646 using two easypost api keys

zyme:
As described in PEPPER-1646, because we aren't migrated historic shipment information between the old and new easypost account, we need a way for DSM to fallback to using the "old" easypost account api key for shipments that were created using the previous account.  Without this ability, DSM can't keep track of updates on older kits.

Along the way, I realized that DSM is running queries every hour to look up shipment information for some very old kits--going back to 2018.  We tend to give up on kits that are more than 12 months old, both because of historical return rates and because return labels _do_ expire.  I've set the cutoff at 18 months so that DSM now only checks for status of kits that haven't been received _and_ are at most 18 months old.

### DIFF
--- a/pepper-apis/config/testing-inmemorydb.conf.ctmpl
+++ b/pepper-apis/config/testing-inmemorydb.conf.ctmpl
@@ -85,6 +85,7 @@
     "elasticsearchBatchSize": 100,
     "healthcheckPassword": "{{$testingConf.Data.healthcheckPassword}}",
     "easyPostApiKey": "{{$testingConf.Data.easyPostApiKey}}",
+    "legacyEsyPostApiKey": "{{$testingConf.Data.legacyEasyPostApiKey}}",
     # by default, boot in-processs for testing.  change rendered .conf file manually for local development
     "bootTestAppInSeparateProcess": false,
     "apiBaseUrl": "http://localhost:5555",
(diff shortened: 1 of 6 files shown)


pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/db/dao/kit/KitDao.java
```diff
@@ -38,7 +38,7 @@ public class KitDao {
                     + "req.bsp_collaborator_participant_id, req.bsp_collaborator_sample_id, req.ddp_participant_id, req.ddp_label, "
                     + "req.dsm_kit_request_id, req.ddp_kit_request_id,"
                     + "req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, "
-                    + "req.external_response, kt.no_return, req.created_by FROM kit_type kt, ddp_kit_request req, ddp_instance ddp_site "
+                    + "req.external_response, kt.no_return, req.created_by, from_unixtime(req.created_date/1000) as request_created_on FROM kit_type kt, ddp_kit_request req, ddp_instance ddp_site "
```
zyme, line 41: Adding kit creation time here so it can be filtered in KitUtil below.

pepper-apis/config/testing-inmemorydb.conf.ctmpl
```diff
@@ -85,6 +85,7 @@
     "elasticsearchBatchSize": 100,
     "healthcheckPassword": "{{$testingConf.Data.healthcheckPassword}}",
     "easyPostApiKey": "{{$testingConf.Data.easyPostApiKey}}",
+    "legacyEasyPostApiKey": "{{$testingConf.Data.legacyEasyPostApiKey}}",
```
zyme, line 88: Vault has been updated with this new (old) value.

